PR #880, opened 2015-01-14
Adding managed preferences class

khawkins:
Starting with the wrapper class for managed preferences.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.h
@@ -1,0 +1,62 @@
+/*
+ Copyright (c) 2015, salesforce.com, inc. All rights reserved.
+ 
+ Redistribution and use of this software in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this list of conditions
+ and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice, this list of
+ conditions and the following disclaimer in the documentation and/or other materials provided
+ with the distribution.
+ * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.h
```diff
@@ -0,0 +1,62 @@
...
+ The associated labels for the provided login hosts.
+ */
+@property (nonatomic, readonly) NSArray *loginHostLabels;
+
+/**
+ The raw NSDictionary of managed preferences.
+ */
+@property (nonatomic, strong, readonly) NSDictionary *rawPreferences;
+
+@end
```
wmathurin: On Android, we also have remote access consumer key and oauth redirect url (to allow runtime replacement for the values in bootconfig.json). See https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/unstable/libs/SalesforceSDK/src/com/salesforce/androidsdk/config/RuntimeConfig.java

wmathurin:
Looks fine otherwise.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.h
khawkins: @akhileshgupta had told me to leave those settings out.  Akhilesh, should we not have parity across platforms for these?

khawkins: @wmathurin I went ahead and added those in.  I also added a notification listener for changes to the settings that can be pushed in at runtime.

khawkins: (https://github.com/khawkins/SalesforceMobileSDK-iOS/commit/5a9ff6f1f80a2437074f1ae0cc38c61be1dd67f9)

bhariharan:
:+1:

PR #885, opened 2015-01-15
Consume managed settings

khawkins:
- Changed advanced auth to be a tri-state check—managed settings will short-circuit the process to go into advanced authentication.

- Consuming any OAuth Consumer Key and OAuth Redirect URI values sent by MDM.


wmathurin:
:+1:

PR #889, opened 2015-01-15
Include FMDB as a CocoaPod dependency

jeffbailey:
Currently FMDB code is included as part of the Salesforce SDK.  This causes problems (duplicate symbols) when trying to integrate Salesforce and another Cocoapod that depends on FMDB.  I updated the podspec to pull in FMDB as a dependency.  Also, in order to resolve other build issues I had to add a subspec for SaleforceSDKCommon.


khawkins:
We can't remove FMDB from the SDK, as CocoaPods is not the sole source of package management for the Mobile SDK, and it's not going to be.  We can export the included FMDB symbols as a means of "namespacing" them to remove the conflicts.

khawkins:
The addition of SalesforceSDKCommon was made today, so that update should not be necessary if you get latest on unstable.

jeffbailey:
Understood about not being able to remove FMDB. I didn't consider the non-CocoaPods case.  Can you clarify what "We can export the included FMDB symbols" means?  Do you plan to rename the FMDB classes included in the SDK?

khawkins:
Rather than rename the classes themselves, just set the build flags to export different symbols for those artifacts to consumers.  Since FMDB is only used internally in SalesforceSDKCore, this becomes simply a project configuration step.

Let me put together a PR for that.  Shouldn't take but a few minutes.

jeffbailey:
Thanks Kevin.  Appreciate the fast resolution.  You rock!

PR #891, opened 2015-01-15
Rename exported FMDB symbols so as not to clash with other versions of FMDB

khawkins:
@jeffbailey This should do the trick.  You should be able to include the FMDB podspec side by side with the Mobile SDK podspec now.


bhariharan:
:+1:

PR #892, opened 2015-01-16
Clearing clipboard on background, based on managed config

khawkins:


wmathurin:
:+1: 

We haven't done that on Android (yet).

bhariharan:
:+1: 

IIRC, there's no clean way to do this on ```Android```. Maybe writing an empty string using ```CliboardManager``` could work, but I can't seem to find any clean API for this.

wmathurin:
I couldn't find a way either. If you long-tap, you can get to previously copied entries, so setting an empty string would not get rid of those. 

Maybe instead of clearing the clipboard, we could disable copy/paste from the app. So whatever you had copied before would not get lost.
However the airwatch folks mentioned being able to disable copy/paste using app restrictions, no code was required in the target app according to them.

PR #893, opened 2015-01-16
Implement login hosts

khawkins:
- Allow managed login host properties to consume a backing string or array of strings.
- Use the first managed login host, if present, as the default host when the app is launched for the first time.


wmathurin:
:+1:

bhariharan:
:+1:

PR #896, opened 2015-01-22
Moving SFReachability to SalesforceSDKCommon

khawkins:
- Moved `SFReachability` out of SalesforceHybridSDK and into SalesforceSDKCommon, as it's more appropriately available in a low-level library that any other libraries and apps can reference/share.
- Renamed to `SFSDKReachability`, to avoid any potential clashes with other code.
- Took the opportunity to grab the latest sample version of this class from Apple.
- Additionally namespaced the `NetworkStatus` enum and all of its members, as well as extern constants and methods not encapsulated in the class definition (reachability callback, print flags).  Basically, anything that could possibly show up as a symbol clash with someone else's version of the class.


khawkins:
Ping.  Any comments on this?

bhariharan:
Looks fine to me.

PR #898, opened 2015-01-26
Changing podspec to refer to tag

khawkins:
Looking through the [Podspec Syntax Reference](http://guides.cocoapods.org/syntax/podspec.html), I notice that there's no specification of `branch` in there.  Looks like you only override that location in your [Podfile definition](http://guides.cocoapods.org/using/the-podfile.html).  So I'm switching to the more common `tag` specification for the pod spec.  We were also pointing to the wrong branch anyway. ;-)


bhariharan:
:+1:

PR #903, opened 2015-01-26
Removing decimal portion to avoid error when running install.sh

wmathurin:
If your xcode_ver_num has a decimal portion (mine is 61.1) then you would get the following error:
./install.sh: line 37: [[: 61.1: syntax error: invalid arithmetic operator (error token is ".1")

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -32,9 +32,9 @@ then
     echo "Could not determine Xcode version.  Is xcodebuild on your path?"
     exit 3
 fi
-xcode_ver_num=`echo $xcode_ver | sed 's/^Xcode \([0-9][0-9]*\)\.\([0-9][0-9]*\)/\1\2/'`
+xcode_ver_num=`echo $xcode_ver | sed 's/^Xcode \([0-9][0-9]*\)\.\([0-9][0-9]*\).*$/\1\2/'`
 xcode_ver_str=`echo $xcode_ver | sed 's/^Xcode //'`
-if [[ ${xcode_ver_num%.*} -lt $XCODE_MIN_VERSION ]]
+if [[ $xcode_ver_num -lt $XCODE_MIN_VERSION ]]
 then
     echo "Current configured Xcode version ($xcode_ver_str) is less than the minimum required version ($XCODE_MIN_VERSION_STR)."
     exit 4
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -34,7 +34,7 @@ then
 fi
 xcode_ver_num=`echo $xcode_ver | sed 's/^Xcode \([0-9][0-9]*\)\.\([0-9][0-9]*\)/\1\2/'`
 xcode_ver_str=`echo $xcode_ver | sed 's/^Xcode //'`
-if [[ $xcode_ver_num -lt $XCODE_MIN_VERSION ]]
```
wmathurin: Using a bash expression to remove the decimal portion
From the bash man page:
${parameter%word}
${parameter%%word}
         Remove matching suffix pattern.  The word is expanded to produce a pattern just as in pathname expansion.   If
         the pattern matches a trailing portion of the expanded value of parameter, then the result of the expansion is
         the expanded value of parameter with the shortest matching pattern (the ``%'' case) or  the  longest  matching
         pattern  (the  ``%%'' case) deleted.  If parameter is @ or *, the pattern removal operation is applied to each
         positional parameter in turn, and the expansion is the resultant list.  If parameter is an array variable sub-
         scripted  with  @  or *, the pattern removal operation is applied to each member of the array in turn, and the
         expansion is the resultant list.

khawkins: I'd rather fix the sed substitution, which is not working right in this case.  My expectation was that line 35 would lop off / ignore anything beyond the first two numerics in an x.y.z.p.d.q representation, but apparently not.  Taking a look now.

khawkins: I forget how regexs differ between platform implementations.  The regex in line 35 should be:

        sed 's/^Xcode \([0-9][0-9]*\)\.\([0-9][0-9]*\).*$/\1\2/'

khawkins: Similar change for line 18.

wmathurin:
thanks Kevin, I put in your sed expression https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/59649bb444545a21fcf7162b88aba92510f35b69

khawkins:
Could you update line 18 as well?  It just needs a `.*$` at the end of the regex.  May as well fix that one too for when someone has iOS 7.1.5 installed. :)

wmathurin:
@khawkins Done.

khawkins:
Thanks! :+1:

PR #904, opened 2015-01-26
Fixing errors reported by 'pod spec lint'

khawkins:
Overlooked that our MKNetworkKit has a dependency on SalesforceCommonUtils.


wmathurin:
:+1:

PR #921, opened 2015-01-29
Xcode 6.1 seems to have issues with code signing

khawkins:
From teh Googles, it looks like setting `CODE_SIGN_RESOURCE_RULES_PATH` will fix it.


bhariharan:
I trust the Googles! :+1:

PR #930, opened 2015-02-10
Implementation of merge mode LEAVE_IF_CHANGED for sync up

bhariharan:
What's left:

* Native tests.

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -631,16 +631,18 @@ static NSMutableDictionary *syncMgrList = nil;
 }
 
 - (BOOL)isNewerThanServer:(NSString *)objectType objectId:(NSString *)objectId lastModifiedDate:(long long)lastModifiedDate {
-    __block BOOL done = NO;
     __block BOOL isNewer = NO;
     __block long long serverLastModified = -1;
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
     SFSmartSyncSoqlBuilder *soqlBuilder = [SFSmartSyncSoqlBuilder withFields:kSyncManagerLastModifiedDate];
     [soqlBuilder from:objectType];
     [soqlBuilder where:[NSString stringWithFormat:@"Id = '%@'", objectId]];
     NSString *query = [soqlBuilder build];
     SFRestRequest *request = [[SFRestAPI sharedInstance] requestForQuery:query];
-    [self sendRequestWithSmartSyncUserAgent:request failBlock:^(NSError *error) {
-            done = YES;
+    [self sendRequestWithSmartSyncUserAgent:request
+        failBlock:^(NSError *error) {
+            [self log:SFLogLevelError format:@"REST request failed with error: %@", error];
+            dispatch_semaphore_signal(semaphore);
         }
         completeBlock:^(NSDictionary* d) {
             if (nil != d) {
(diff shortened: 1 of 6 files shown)


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -544,6 +545,21 @@ - (void) syncUpOneEntry:(SFSyncState*)sync recordIds:(NSArray*)recordIds index:(
...
+     * on the server AFTER the client's last sync down. If the merge mode
+     * passed in tells us to leave the record alone under these
+     * circumstances, we will do nothing and return here.
+     */
+    if (mergeMode == SFSyncStateMergeModeLeaveIfChanged &&
+        (action == kSyncManagerActionUpdate || action == kSyncManagerActionDelete) &&
+        ![self isNewerThanServer:objectType objectId:objectId lastModifiedDate:lastModifiedDate]) {
+        // Next
+        [self syncUpOneEntry:sync recordIds:recordIds index:i+1 updateSync:updateSync failRest:failRest];
+    }
```
wmathurin: You need to return out of the function otherwise the rest will execute

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -614,6 +630,40 @@ - (void)sendRequestWithSmartSyncUserAgent:(SFRestRequest *)request failBlock:(SF
...
+                    }
+                }
+            }
+            if (serverLastModified <= lastModifiedDate) {
+                isNewer = YES;
+            }
+            dispatch_semaphore_signal(semaphore);
+        }
+     ];
+    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
```
wmathurin: Should we really block the thread?

khawkins: Yeah, I'd change this function to return an async result, and the consuming logic accordingly.

khawkins: But then, if it's already running in an async thread, perhaps this is just as good?

wmathurin: In that case, we could rewrite the whole sync up to be more readable (and a simple loop when calling syncUpOneEntry)

bhariharan:
Native tests [here](https://github.com/bhariharan/SalesforceMobileSDK-iOS/commit/7cfff15abbc091cc4324b67ae5d654fb354cb2a1).

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
bhariharan: Good catch! I was blindly following some previously written code a few lines above for the same use case. I'll add return statements to both.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
bhariharan: Completion blocks are probably the way to go. I'll change the pattern here.

bhariharan:
I've made the changes suggested. Should be good to go now.

wmathurin:
Looks reasonable

khawkins:
Nice. :+1:

PR #936, opened 2015-02-13
Refactoring of SFSyncTarget (first part)

wmathurin:
SFSyncTarget is now an abstract class with three concrete subclasses SFMruSyncTarget, SFSoqlSyncTarget and SFSoslSyncTarget
Not done yet: the code fetching data from the server during syncDown is still in SFSmartSyncSyncManager instead of being in the SF*SyncTarget class
All smartsync tests (hybrid and native) passing


bhariharan:
Looks good.

PR #937, opened 2015-02-18
Refactoring of SFSyncTarget (second part)

wmathurin:
Data fetching now handled by SFSyncTarget sub classes
Next: allowing custom target

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Util/SFSyncTarget.m
+++ b/libs/SmartSync/SmartSync/Classes/Util/SFSyncTarget.m
@@ -26,31 +26,34 @@
 #import "SFMruSyncTarget.h"
 #import "SFSoqlSyncTarget.h"
 #import "SFSoslSyncTarget.h"
-
-#define ABSTRACT_METHOD {\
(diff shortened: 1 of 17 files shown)


libs/SmartSync/SmartSync/Classes/Util/SFSyncTarget.m
```diff
@@ -27,6 +27,11 @@
 #import "SFSoqlSyncTarget.h"
 #import "SFSoslSyncTarget.h"
 
+#define ABSTRACT_METHOD {\
```
khawkins: Is this a concept we should generalize to the greater SDK (e.g. `SalesforceSDKConstants.h`)?

khawkins:
Overall, looks great.

libs/SmartSync/SmartSync/Classes/Util/SFSyncTarget.m
wmathurin: Will do.

bhariharan:
:+1:

PR #950, opened 2015-03-04
If upsert fails on syncDown, that should be fatal

khawkins:
Call the "sync fail" block if local data save fails on sync down.


wmathurin:
:+1:

PR #952, opened 2015-03-10
Create mycopy

clejeune:


khawkins:
I presume this was a mistake....

PR #956, opened 2015-03-12
SFSyncServerTarget

khawkins:
Functionality for breaking the sync up logic out into sync server targets, for more granular control of the sync up process.

Left to do: More tests around the new functionality.  That's next on my list.  All current tests are passing.


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -479,76 +497,37 @@ - (void)resumeSyncUpOneEntry:(SFSyncState*)sync recordIds:(NSArray*)recordIds in
...
-    // Send request
-    [self sendRequestWithSmartSyncUserAgent:request failBlock:failRest completeBlock:completeBlock];
-}
-
-- (void)sendRequestWithSmartSyncUserAgent:(SFRestRequest *)request failBlock:(SFRestFailBlock)failBlock completeBlock:(id)completeBlock {
-    //[request setHeaderValue:[SFRestAPI userAgentString:kSmartSync] forHeaderName:kUserAgent];
-    [self.restClient sendRESTRequest:request failBlock:failBlock completeBlock:completeBlock];
-}
-
-- (void)isNewerThanServer:(SFSyncState*)sync recordIds:(NSArray*)recordIds index:(NSUInteger)i record:(NSMutableDictionary*)record action:(SFSyncManagerAction)action updateSync:(SyncUpdateBlock)updateSync failRest:(SFRestFailBlock)failRest {
```
bhariharan: Did this become ```fetchRecordModificationDates```?

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -90,30 +76,46 @@ @implementation SFSmartSyncSyncManager
...
++ (instancetype)sharedInstance:(SFUserAccount *)user {
+    if (user) {
+        SFSmartStore *store = [SFSmartStore sharedStoreWithName:kDefaultSmartStoreName user:user];
+        return [self sharedInstanceForStore:store];
+    } else {
+        return nil;
+    }
+}
+
++ (instancetype)sharedInstanceForStore:(SFSmartStore *)store {
```
bhariharan: Can we have ```sharedInstance``` and ```sharedInstanceForStore``` pass through a single method underneath? So, this method would call that one with a ```nil``` user, and the previous one would call it with a ```nil``` store, or something along those lines. I'm wondering if there's a possibility that someone would want an instance of a user with a store they pass in.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -386,15 +388,19 @@ - (void) syncUp:(SFSyncState*)sync updateSync:(SyncUpdateBlock)updateSync failSy
...
-    [self syncUpOneEntry:sync recordIds:dirtyRecordIds index:0 updateSync:updateSync failRest:failRest];
+    [self syncUpOneEntry:sync recordIds:dirtyRecordIds index:0 updateSync:updateSync failBlock:failBlock];
 }
 
-- (void) syncUpOneEntry:(SFSyncState*)sync recordIds:(NSArray*)recordIds index:(NSUInteger)i updateSync:(SyncUpdateBlock)updateSync failRest:(SFRestFailBlock)failRest {
+- (void)syncUpOneEntry:(SFSyncState*)sync
+             recordIds:(NSArray*)recordIds
+                 index:(NSUInteger)i
+            updateSync:(SyncUpdateBlock)updateSync
+             failBlock:(void (^)(NSError *))failBlock {
```
wmathurin: Why not have a typdef for the failblock?

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -431,44 +437,56 @@ - (void) syncUpOneEntry:(SFSyncState*)sync recordIds:(NSArray*)recordIds index:(
...
-                if (record[fieldName] != nil)
-                    fields[fieldName] = record[fieldName];
-            }
-        }
-    }
-
+    
     // Delete handler
-    SFRestDictionaryResponseBlock completeBlockDelete = ^(NSDictionary *d) {
+    void (^completeBlockDelete)(NSDictionary *) = ^(NSDictionary *d) {
```
wmathurin: Typedef for the block signature?

wmathurin:
Looks great overall. Have you tried the sync manager tests?

wmathurin:
Never mind I just read your summary comment about running the tests.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
khawkins: Yes.  The `fetch` method decouples the comparison process, simply returning the relevant dates to the sync manager, which then implements the check.  That way we can support other kinds of checks in the future, without modifying the targets.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
khawkins: So you're thinking of an additional convenience method like `sharedInstanceForUser:(SFUserAccount *)user storeName:(NSString *)storeName` ?  We could add that in.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
bhariharan: :+1:

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
bhariharan: Yes, I think that would be nice. It'll also unify the code paths.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
khawkins: We could do that.  I had initially focused on decoupling the sync manager from the SFRest functionality, but we can add local typedefs to keep parity with `SyncUpdateBlock`.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
khawkins: I can do that here as well.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
khawkins: Okay, I'll add that in.

khawkins:
@wmathurin @bhariharan The last two commits have the changes you requested: some new singleton methods for sync manager, and block typedefs for the sync up functionality.  Working on more tests now.

bhariharan:
Otherwise, looks pretty good.

libs/SmartSync/SmartSync/Classes/Util/SFSyncServerTarget.m
```diff
@@ -0,0 +1,176 @@
...
+
+#pragma mark - Serialization and factory methods
+
++ (instancetype)newFromDict:(NSDictionary*)dict {
+    NSString *implClassName;
+    NSString *targetTypeString = (dict[kSFSyncTargetTypeKey] == nil ? kSFSyncServerTargetTypeRestStandard : dict[kSFSyncTargetTypeKey]);
+    switch ([self targetTypeFromString:targetTypeString]) {
+        case SFSyncServerTargetTypeCustom:
+            implClassName = dict[kSFSyncTargetiOSImplKey];
+            return [NSClassFromString(implClassName) newFromDict:dict];
```
khawkins: I basically copied this implementation from `SFSyncTarget`, but it seems like an anti-pattern.  We're breaking de facto inheritance here.  Shouldn't this be calling into a common initializer, like `initWithDict`?

wmathurin: We could change it around. I don't feel strongly one way or the other.

wmathurin:
For sync down, the sync target defines two things: 
- what records to bring down and 
- how to bring them down (i.e. what end points to go against). 
For sync up, the sync target currently doesn't define the "target" i.e. set of records to send up, it only defines how to send them up (i.e. what end points to go against).

It would be nice the sync down and sync up targets were more similar and would both define the what and the how. 
It is conceivable that one would want to sync up a specific subset of dirty records (or even records that were modified but didn't have the dirty flags) and that type of use cases would be covered if we moved the selection of records to send up to a method on the sync server target.

Thoughts?

wmathurin:
Also should we rename sync target as sync down target and sync server target as sync up target?
After all they both deal with the server side in their implementation.

wmathurin:
@khawkins if you want, you can merge this change as is, and I can proceed with the suggested changes in a followup pull request.

khawkins:
@wmathurin :

- Re: defining the target records in sync up, that seems like a good idea.  It would be nice for sync up targets and sync down targets to have more parity.
- Re: renaming the targets, I only avoided it because of potential backward compatibility issues.  Otherwise, yes, it would be better if the naming conventions were more consistent.

For either/both of these, I'd prefer to proceed with merging in, and we can iterate from there.  I have some tests coming in behind this too...figured I'd leave them out of the main request, to keep it easier to review.

wmathurin:
That's fine, go ahead and merge.
I'm currently working on the Android implementation. We can align both implementations afterwards.

khawkins:
Sounds good.

PR #963, opened 2015-03-18
Aligning sync up target between Android and iOS

wmathurin:
- Renamed SFSyncTarget as SFSyncDownTarget
- Renamed SFSyncServerTarget as SFSyncUpTarget
- Changed methods exposed by SFSyncUpTarget to be more like Android's
- Added method to SFSyncUpTarget to actually specify targeted records for the sync up

- Native and hybrid tests passing
- Sample apps (native and hybrid) running


bhariharan:
:+1:

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -515,22 +516,41 @@ - (void)resumeSyncUpOneEntry:(SFSyncState*)sync
...
+    
+    SFSyncUpTarget *target = (SFSyncUpTarget*) sync.target;
+    NSArray *fieldList = sync.options.fieldlist;
+    
+    // Getting type and id
+    NSString* objectType = [SFJsonUtils projectIntoJson:record path:kObjectTypeField];
+    NSString* objectId = record[kId];
+    
+    // Fields to save (in the case of create or update)
+    NSMutableDictionary* fields = [NSMutableDictionary dictionary];
```
khawkins: The decision to move this into the sync up target was deliberate, to take the decision making about the shape of the data out of the hands of the sync server.  What's the impetus for the decision to move it back?

libs/SmartSync/SmartSync/Classes/Util/SFSyncState.h
```diff
@@ -25,7 +25,8 @@
 #import <Foundation/Foundation.h>
 
 @class SFSyncTarget;
```
khawkins: This can go now, yes?

khawkins: Ah, didn't see that SFSyncTarget becomes the base class.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
wmathurin: Agreed it would be more flexible but given that the sync up is parameterised via the fieldlist (passed through the sync options), won't we simply end up with the same code in all custom targets (looping through the fieldlist, and projecting out the chosen fields)? Also, if you really need the whole record, you could simply pass [""] for the field list and you end up with {"":record} in fields?

wmathurin: Part of me feel that sync options and sync target could be merged. With possibly the merge mode staying out.
Anyway for your app, are the methods current signature too restrictive?

khawkins: *&gt; won't we simply end up with the same code in all custom targets (looping through the fieldlist, and projecting out the chosen fields)?*

Well sure, but that's the idea: it ends up in the *target*, which (in my thinking anyway) should have ownership over what defines the structure of the data it's sending.

Not sure about the "" part—isn't `[record objectForKey@""]` going to return `nil`?

Anyway, I don't know that the switch in this model will be too restrictive for our use cases.  I do think that merging targets and options could make sense, which would be another path into that flexibility.  But this is fine at this point...we can see how it works in our consumption model, and ping you guys with our results.

khawkins:
The rest looks fine to me.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
wmathurin: obj[""] will return nil, but I thought we were using [SFJsonUtils projectIntoJson:record path:fieldName]; -- which we are not.
If I switch it over to using "project", then it will allow people to get the whole record with "", or nested field values with paths like "a.b.c" and behave like before for regular attributes.

khawkins: :+1: for projecting the field name.

PR #965, opened 2015-03-19
A fix for swift

wmathurin:


khawkins:
Oh cool, that fix is easier than I thought it would be. :+1:

PR #966, opened 2015-03-20
Adding an overload to append re-syncing filter based on generic mod date field

khawkins:


bhariharan:
Ooh - fancy! :+1:

wmathurin:
Nice - we need to add that on the hybrid and Android side (and Windows)

khawkins:
I can do that.

khawkins:
So wait, are we proposing that we add net-new configuration to the hybrid side?  Right now there is no configuration for specifying a modification date field (of course).  Would this be in the config section of the options of the sync?

wmathurin:
Actually going back to your change.
How is that going to work? Currently SFSmartSyncSyncManager captures the max time stamp during a sync down in getMaxTimeStamps and always read from the standard field, instead it should read from the other field. And maybe that alternate field could be part of the sync down target definition.

khawkins:
*sigh* Yeah, you're right.  Let me look at another solution.

PR #972, opened 2015-03-28
Removing store removal plugin methods

bhariharan:


bhariharan:
Based on our offline discussion.

wmathurin:
:+1:

khawkins:
I presume the hybrid change never made it?

bhariharan:
@khawkins Yes, you are correct. Nothing to change there.

PR #974, opened 2015-04-01
Making sure all public headers are referenced in each lib

khawkins:
I'll add the script that updates the headers, and its inclusion in the post-build steps, in another PR to unstable.  This is just the result of running the script to true up the public header files.


wmathurin:
:+1:

bhariharan:
:+1:

khawkins:
I updated the tag in the spec to (eventually) point to `pod_v#{s.version}`.  I'll push the new tag up once this is merged.

PR #980, opened 2015-04-07
Fixing compilation issues

wmathurin:


bhariharan:
:+1:

hybrid/SampleApps/NoteSync/NoteSync/SFContentSoqlSyncDownTarget.m
```diff
@@ -198,7 +199,7 @@ - (id)initWithQueryLocator:(NSString*)queryLocator
 
 - (void)prepareRequestForSend
 {
-    NSString *sessionId = [CSFNetwork currentNetwork].account.credentials.accessToken;
+    NSString *sessionId = [SFAuthenticationManager sharedManager].coordinator.credentials.accessToken;
```
khawkins: What was not working here?  If we're not getting the token from the account of the current network, we should at least be getting the token from the current user.  Credentials being owned by `SFAuthenticationManager` is a legacy—the true owner of credentials at this point should be `SFUserAccount` instances.

PR #982, opened 2015-04-07
Adding xcconfig files for missing projects

bhariharan:
Added xcconfig files for:

* ```SalesforceHybridSDK```
* ```SalesforceRestAPI```
* ```SmartSync```


wmathurin:
:+1:

libs/SalesforceHybridSDK/SalesforceHybridSDK-Common.xcconfig
```diff
@@ -0,0 +1,15 @@
...
+//  Created by Bharath Hariharan on 4/7/15.
+//  Copyright (c) 2015 salesforce.com. All rights reserved.
+//
+
+GCC_PRECOMPILE_PREFIX_HEADER = YES
+GCC_PREFIX_HEADER = $(PROJECT_NAME)/$(PROJECT_NAME)-Prefix.pch
+
+PRODUCT_NAME = SalesforceHybridSDK
+
+HEADER_SEARCH_PATHS = "$(BUILT_PRODUCTS_DIR)/Headers" "$(TARGET_BUILD_DIR)/Headers" "$(PROJECT_DIR)/../../external/cordova-ios/cordova-ios/CordovaLib/Classes" "$(PROJECT_DIR)/../../external/ThirdPartyDependencies/SalesforceCommonUtils/Headers" $(inherited)
```
khawkins: The CordovaLib/Classes may need a `/**` at the end, if it's supposed to be recursive (which I think it is).

khawkins:
Looks fine otherwise.  Do we want to remove the old config values, like `HEADER_SEARCH_PATHS`, from the project files?

libs/SalesforceHybridSDK/SalesforceHybridSDK-Common.xcconfig
bhariharan: You're right. I'll add that in.

bhariharan:
Should be good to go now.

khawkins:
Cool, looks good.

PR #983, opened 2015-04-08
Avoiding SFRootViewManager to present/dismiss inspector

wmathurin:
Dev can either call present (which now takes a UIViewController) and dismiss or get the sharedInstance and present it themselves
SFRootViewManager was getting stuck in waitForPresentCompletion on iOS8

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreInspectorViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreInspectorViewController.m
@@ -88,14 +88,28 @@ static NSUInteger   const kLabelTag              = 99;
 
 #pragma mark - Present / dimiss
 
-+ (void) present
++ (void) present:(UIViewController*)currentViewController
 {
-    [[SFRootViewManager sharedManager] pushViewController:[SFSmartStoreInspectorViewController sharedInstance]];
+    if (![NSThread isMainThread]) {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [SFSmartStoreInspectorViewController present:currentViewController];
+        });
+        return;
+    }
+
+    [currentViewController presentViewController:[SFSmartStoreInspectorViewController sharedInstance] animated:NO completion:nil];
 }
 
 + (void) dismiss
 {
-    [[SFRootViewManager sharedManager] popViewController:[SFSmartStoreInspectorViewController sharedInstance]];
+    if (![NSThread isMainThread]) {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [SFSmartStoreInspectorViewController dismiss];
+        });
+        return;
+    }
+
+    [[[SFSmartStoreInspectorViewController sharedInstance] presentingViewController] dismissViewControllerAnimated:NO completion:nil];
 }
 
 #pragma mark - Results setter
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreInspectorViewController.m
```diff
@@ -88,14 +88,28 @@ + (SFSmartStoreInspectorViewController *) sharedInstance
...
 {
-    [[SFRootViewManager sharedManager] pushViewController:[SFSmartStoreInspectorViewController sharedInstance]];
+    if (![NSThread isMainThread]) {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            [SFSmartStoreInspectorViewController present:currentViewController];
+        });
+        return;
+    }
+
+    [currentViewController presentViewController:[SFSmartStoreInspectorViewController sharedInstance] animated:NO completion:nil];
```
khawkins: Any reason to keep it as a singleton?  Any reason there couldn't potentially be *n* instances of `SFSmartStoreInspectorViewController` scattered through different portions of the view hierarchy?

wmathurin: The reason is that if you have typed a complex query then went back into the app and then back to the inspector it would still be there. 
We could also leave it to the devs to keep the instance around if they need that behavior.
So present/dismiss could become instance methods instead also, the constructor could take a store to be able to inspect the global stores or other stores.

bhariharan:
Looks good to me.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreInspectorViewController.m
khawkins: Ah, that makes sense.  As far as the `present` and `dismiss` methods go, it seems like we're writing convenience wrappers around the one-liners `presentViewController:animated:completion` and `dismissViewControllerAnimated:completion`.  Are we adding value there to developers?

wmathurin: It's more than a one liner if you are not running on the main thread.
Also we have already documented present/dismiss so I didn't want to drop them entirely (even though I changed the signature of present).
But anyway, they don't have to use those methods, they can get a controller and present it themselves.

khawkins: I see your main thread, and raise you:

        dispatch_async(dispatch_get_main_queue(), ^{ [self.viewController presentViewController:myViewController animated:YES completion:NULL });

:)  I'm okay keeping the pattern for the reasons you stated.  My only point is that iOS developers learn this pattern from day one, and use it every day.  I don't feel like we should be getting into the practice of refactoring or otherwise changing development patterns that are already commonplace in the development environments we support.

wmathurin:
Merging for now. But might do another pull request to allow the store to be passed in the constructor.
And maybe then I'll get rid of the sharedInstance / present and dismiss methods?

khawkins:
I think this is fine.  We can look at different approaches going forward, wherever necessary.

PR #985, opened 2015-04-08
Updates to data management

khawkins:
No real functionality changes here.  Just changing a couple of the `SFSmartStoreDatabaseManager` methods from instance level to class level, to allow the encryption/decryption process to be isolated from SmartStore itself.  `encryptOrUnencryptDb:name:path:oldKey:newKey:error:` is a new internal method signature, but it's encapsulating the functionality that was already there.

Tests pass, SmartStore sample apps all seem good.


wmathurin:
Looks reasonable.

bhariharan:
:+1:

PR #986, opened 2015-04-08
where is a reserved word in Swift

wmathurin:
No longer using "where" as parameter or variable name

NB: Common utils still does in some of its classes (SFSoqlBuilder.h and SFSoslReturningBuilder.h)


bhariharan:
These classes are originally from there (part of the 'open source effort' of ```SalesforceCommonUtils```).

bhariharan:
:+1:

khawkins:
:+1:

PR #987, opened 2015-04-08
Replacing deprecated cordova calls

wmathurin:


libs/SalesforceHybridSDK/HybridPluginTestAppTests/ForcetkTestSuite.m
```diff
@@ -35,10 +35,6 @@ - (void)setUp
 }
 
 
-- (void)testGetInfo {
```
wmathurin: This test is in SDKInfoTestSuite.m
(bad copy/paste when creating ForcetkTestSuite.m)

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFAdditions/CDVPlugin+SFAdditions.m
```diff
@@ -31,39 +31,23 @@ @implementation CDVPlugin (SFAdditions)
 
 #pragma mark - Cordova plugin support
 
-- (void)writeSuccessResultToJsRealm:(CDVPluginResult*)result callbackId:(NSString*)callbackId
```
wmathurin: Those two functions were only called by the other methods below.

bhariharan:
:+1:

PR #990, opened 2015-04-11
Adding network queue support for unauthenticated users

bhariharan:


bhariharan:
@khawkins Let me know if this approach works for you.

khawkins:
In the new network SDK, authentication is meant to be encapsulated in `CSFAction` action objects; actions supply the context of a network interaction, while `CSFNetwork` is merely the broker.

In that sense, I'd like all authentication determinations moved out of `CSFNetwork` entirely.  Generally, this means starting the network queue unsuspended for *all* actions, then moving the auth checks of `observeValueForKeyPath:ofObject:change:context:` in `CSFAction` to `CSFSalesforceAction`, since the particular checks that are currently implemented, are really specific to that action type.

We can further leverage the `requiresAuthentication` property of `CSFAction

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.h
```diff
@@ -78,7 +78,7 @@
  *
  * @see networkSuspended
  */
-@property (atomic, readonly, getter=isRefreshingAccessToken) BOOL refreshingAccessToken;
+@property (atomic, readonly, getter = isRefreshingAccessToken) BOOL refreshingAccessToken;
```
khawkins: `refreshingAccessToken

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.h
```diff
@@ -95,6 +95,14 @@
  */
 + (instancetype)networkForUserAccount:(SFUserAccount*)account;
 
+/** Returns a reference to the shared instance of this class for an unauthenticated user.
+ 
+ If an instance doesn't yet exist for this user, one will be created.
+ 
+ @see currentNetwork
+ */
++ (instancetype)networkForUnauthenticatedUser;
```
NachoMan: Since so many subsystems depend on an actual SFUserAccount existing, I had planned to implement this by instead creating a special Guest SFUserAccount, like so:

```
    [CSFNetwork networkForUserAccount:[SFUserAccount guestAccount]]
```

In this way anything that is keyed off of a user account object won't have to perform special checks to see if it's an authenticated or unauthenticated queue.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
```diff
@@ -109,17 +110,30 @@ + (instancetype)networkForUserAccount:(SFUserAccount*)account {
...
         [self.queue addObserver:self forKeyPath:@"operationCount" options:NSKeyValueObservingOptionNew context:kObservingKey];
         _online = YES;
-        
-        // Start the queue suspended, so we can unsuspend it when the user account object is set
-        self.queue.suspended = YES;
-        _networkSuspended = YES;
-        
+
+        // Start the queue unsuspended, since the queue could be used by an unauthenticated user.
+        self.queue.suspended = NO;
```
NachoMan: The default value of -[NSOperationQueue suspended] is already NO, so you don't have to do that here.

NachoMan:
I feel that this change is too focused on CSFNetwork, and as @khawkins points out, doesn't focus enough on the CSFAction behavior.  If you create a guest user object and use that to initialize the network, the changes will be minimal at the network level.  It then allows the CSFAction object to decide for itself (a) whether or not it's capable of performing the request based on its `requiresAuthentication` method, and (b) whether or not to add any HTTP authentication headers from the user object's information.

bhariharan:
@khawkins I've made the changes we discussed offline. The guest user support piece will be added in a different pull request later.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -209,6 +209,16 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
     if (context == kObservingKey) {
         [self willChangeValueForKey:@"ready"];
         [self didChangeValueForKey:@"ready"];
+        if ([keyPath isEqualToString:@"communityId"]) {
+            self.enqueuedNetwork.defaultConnectCommunityId = self.enqueuedNetwork.account.communityId;
+        } else if (self.enqueuedNetwork.account.credentials.accessToken
```
khawkins: This needs to be subject to the value of the `requiresAuthentication` property as well.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction+Internal.h
```diff
@@ -65,6 +65,7 @@ CSF_EXTERN NSTimeInterval const CSFActionDefaultTimeOut;
 @property (nonatomic, strong) NSURLSessionDownloadTask *downloadTask;
 @property (nonatomic, strong) NSMutableData *responseData;
 @property (nonatomic, strong) CSFAuthRefresh *authRefreshInstance;
+@property (nonatomic) BOOL credentialsReady;
```
bhariharan: Moved this property to the action from the queue.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
```diff
@@ -157,7 +157,7 @@ - (instancetype)initWithResponseBlock:(CSFActionResponseBlock)responseBlock {
         _method = @"GET";
         _authRefreshClass = [CSFTokenRefresh class];
         _requiresAuthentication = YES;
-
+        self.credentialsReady = YES;
```
bhariharan: Default is ```YES```.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -35,6 +35,11 @@
 
 static void * kObservingKey = &kObservingKey;
 
+@interface CSFNetwork() <SFUserAccountManagerDelegate> {
```
bhariharan: I had to move this here as well (seemed like the right place anyway), for user switching to work properly.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -67,24 +98,6 @@ - (NSDictionary *)headersForAction {
     return httpHeaders;
 }
 
-- (void)setEnqueuedNetwork:(CSFNetwork *)enqueuedNetwork {
```
bhariharan: This is no longer required (the check is in ```isReady```).

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -209,6 +222,18 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
     if (context == kObservingKey) {
         [self willChangeValueForKey:@"ready"];
         [self didChangeValueForKey:@"ready"];
+        if ([self requiresAuthentication] && (self.enqueuedNetwork.account == object)) {
```
bhariharan: Pretty much the same as before, except it has a new home, and the extra ```[self requiresAuthentication]``` conditional.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -234,4 +259,21 @@ - (BOOL)isEqualToAction:(CSFAction *)action {
     return YES;
 }
 
+#pragma mark SFAuthenticationManagerDelegate
+
+- (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
```
bhariharan: Delegate for user switching.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
```diff
@@ -115,11 +115,7 @@ - (id)init {
         self.queue = [NSOperationQueue new];
         [self.queue addObserver:self forKeyPath:@"operationCount" options:NSKeyValueObservingOptionNew context:kObservingKey];
         _online = YES;
-        
-        // Start the queue suspended, so we can unsuspend it when the user account object is set
-        self.queue.suspended = YES;
```
bhariharan: Unsuspended by default now.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
```diff
@@ -175,37 +165,7 @@ - (void)dealloc {
 
 - (void)setAccount:(SFUserAccount *)account {
     if (_account != account) {
-        if (_account) {
```
bhariharan: Moved over to the action.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
```diff
@@ -349,24 +299,6 @@ - (void)URLSession:(NSURLSession *)session downloadTask:(NSURLSessionDownloadTas
     [action sessionDownloadTask:downloadTask didFinishDownloadingToURL:location];
 }
 
-#pragma mark SFAuthenticationManagerDelegate
-
-- (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
```
bhariharan: Moved over to the action.

bhariharan:
@khawkins I've verified ```REST``` operations using ```RestApiExplorer```, including user switching. Also ran the relevant tests.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
khawkins: I'm presuming you just wanted the `SFUserAccountManagerDelegate` part, right?  You don't want the `CSFNetwork` private interface here.  You can piggy back the delegate declaration in `CSFSalesforceAction_Internal.h`.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -44,10 +49,36 @@ - (instancetype)initWithResponseBlock:(CSFActionResponseBlock)responseBlock {
         _apiVersion = CSFSalesforceDefaultAPIVersion;
         _pathPrefix = CSFSalesforceActionDefaultPathPrefix;
         self.authRefreshClass = [CSFSalesforceOAuthRefresh class];
+        CSFNetwork *network = self.enqueuedNetwork;
+        [network addObserver:self forKeyPath:@"credentials.accessToken"
```
khawkins: The key path for these will need to be e.g. `account.credentials.accessToken`, yes?

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
bhariharan: True. I'll move that over.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
bhariharan: Ah yes. Looks like I forgot to commit.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -234,4 +259,21 @@ - (BOOL)isEqualToAction:(CSFAction *)action {
     return YES;
 }
 
+#pragma mark SFAuthenticationManagerDelegate
+
+- (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
+    SFUserAccountManager *accountManager = (SFUserAccountManager*)notification.object;
+    if ([accountManager isKindOfClass:[SFUserAccountManager class]]) {
+        if (accountManager.currentUserIdentity.userId != self.enqueuedNetwork.account.credentials.userId) {
```
khawkins: This is the check that should be changed to:

        if (![accountManager.currentUserIdentity isEqual:self.enqueuedNetwork.account.accountIdentity])

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
```diff
@@ -53,7 +53,7 @@
     return [NSString stringWithFormat:@"%@-%@-%@", user.credentials.organizationId, user.credentials.userId, user.communityId];
 }
 
-@interface CSFNetwork() <SFAuthenticationManagerDelegate, SFUserAccountManagerDelegate> {
+@interface CSFNetwork() <SFAuthenticationManagerDelegate> {
```
khawkins: `SFAuthenticationManagerDelegate` should go too.

khawkins:
The following areas need to be cleaned up in `CSFNetwork.m`:

- `#import "CSFOAuthTokenRefreshInput.h"`, `#import "CSFOAuthTokenRefreshOutput.h"`, and `#import "CSFAuthRefresh.h"` are no longer necessary.
- The calls to `[[SFAuthenticationManager sharedManager] addDelegate:self];` and `[[SFUserAccountManager sharedInstance] addDelegate:self];` need to be removed from `init`.
- `[notificationCenter addObserver:self selector:@selector(authManagerDidLogout:) name:kSFUserLogoutNotification object:nil];` needs to be removed from `init`.
- `[[SFAuthenticationManager sharedManager] removeDelegate:self];` needs to be removed from `dealloc`.
- `#pragma mark - SFAuthenticationManagerDelegate` can be removed
- All of the alert view logic—`_deviceUnauthorizedAlert`, `cleanupDeviceUnauthorizedAlert`, `applicationDidBecomeActive`, and `alertView:clickedButtonAtIndex:`—should be removed.

bhariharan:
@khawkins I've made those changes.

khawkins:
By jove, it's looking pretty good. :)

bhariharan:
Sweet!

PR #996, opened 2015-04-16
Fixing MKNetworkKit issue in podspec

khawkins:
I don't know why this only started showing up with Xcode 6.3, but MKNetworkKit.h's

        #import "Reachability/Reachability.h"

started causing issues.  The line I added here just piggy backs on other logic to do the same with other imports in MKNetworkKit.h.  I've verified that it passes `pod lib lint --allow-warnings` and installs and runs properly in a consuming app.

Not sure if you want/need to update the tag/version.  I'll leave that to you guys.


khawkins:
Oh, this is only an issue with CocoaPods.  Doesn't seem to impact anything else.

bhariharan:
:+1:

PR #1008, opened 2015-04-24
Allow re-sync for all states other than running

khawkins:
Non-running syncs should be allowed to be retried, even if they previously failed.


bhariharan:
:+1:

PR #1011, opened 2015-05-04
Push notification unregistration should be done before revoking refresh token

bhariharan:


khawkins:
I thought we had that positioning before, but it had some sort of issues there.  @wmathurin Do you remember?

wmathurin:
Looks reasonable.

PR #1017, opened 2015-05-10
Network should use apiUrl, to support Communities REST API calls

khawkins:


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
```diff
@@ -61,17 +61,17 @@ + (NSURL*)urlForAction:(CSFAction*)action error:(NSError**)error {
         return nil;
     }
     
-    NSString *host = [action.enqueuedNetwork.account.credentials.instanceUrl host];
+    NSString *baseUrlString = [action.enqueuedNetwork.account.credentials.apiUrl absoluteString];
     NSString *path = [NSMutableString stringWithFormat:@"%@%@", action.basePath, action.verb];
     
     // Make sure path is not empty
-    if (!host || host.length == 0) {
+    if (baseUrlString.length == 0) {
```
bhariharan: Will ```apiUrl``` be set for an unauthenticated user?

bhariharan:
Looks good to me otherwise.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
khawkins: Realistically, this is another section of code that probably needs to be reworked in the context of unauthenticated users, as even `instanceUrl` will not be available in that scenario.  `apiUrl` is a convenience property that will return the Community URL if it's configured, and the `instanceUrl` otherwise.  But I agree that, as we get into unauthenticated scenarios, there needs to be a reconsideration of this flow.  In fact `CSFAction`, as the base action, should probably default to not making any presumptions about the URL.  The existing considerations here should be moved into `CSFSalesforceAction`.

bhariharan: Makes sense.

bhariharan:
:+1:

wmathurin:
:+1: 

Are we patching master with that fix?

khawkins:
@wmathurin I'll leave that decision to you guys. :)  Happy to go in that direction if that's what the team decides to do.

PR #1018, opened 2015-05-14
Error creating the API URL

khawkins:
An update to the previous API URL PR.  I got bitten by [this issue](http://stackoverflow.com/questions/2579544/nsstrings-stringbyappendingpathcomponent-removes-a-in-http) in the original.  This should fix it up.


bhariharan:
:+1:

PR #1019, opened 2015-05-15
Show auth view in willBeginAuthenticationWithView to prevent delay after pressing login button.

wojteklembryk:
Current state: auth web view is presented after is loaded (`[SFOAuthCoordinator webViewDidFinishLoad]`). When user has slow internet connection loading process can take some time (or fail after a timeout). This is not the greatest user experience :wink: 

After change from this PR auth web view is presented when authentication process begins (`[SFOAuthCoordinator beginUserAgentFlow]`) so we can prevent possible delay during loading time.

Moreover user is presented with an option to cancel the loading (right now he'll have to wait until something loads, or even worse - he might get bored and leave that screen before oauth gets presented).

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
@@ -1055,6 +1055,16 @@ static Class InstanceClass = nil;
             [delegate authManagerWillBeginAuthWithView:self];
         }
     }];
+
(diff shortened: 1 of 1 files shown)


bhariharan:
@wojteklembryk Thanks for your contribution! Would you mind re-submitting this against ```unstable```?

wojteklembryk:
Sure! Here is new PR: https://github.com/Taptera/SalesforceMobileSDK-iOS/commit/41fb270a8d346185dd3467f9778db4aee9b32fdc

PR #1022, opened 2015-05-20
SmartStore full-text search support for iOS

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSmartStoreTests.m
```diff
@@ -40,27 +38,19 @@
...
 #import "SFSecurityLockout+Internal.h"
-#import "SFUserAccountManager.h"
 #import <SalesforceSecurity/SFKeyStoreManager.h>
 #import <SalesforceSecurity/SFEncryptionKey.h>
 #import <SalesforceCommonUtils/NSString+SFAdditions.h>
 #import <SalesforceCommonUtils/NSData+SFAdditions.h>
 
-NSString * const kTestSmartStoreName   = @"testSmartStore";
-NSString * const kTestSoupName   = @"testSoup";
+#define kTestSmartStoreName  @"testSmartStore"
```
bhariharan: Any particular reason to use ```#define``` instead of ```const```? Isn't it more optimal in ```iOS``` to use a ```const``` instead of a preprocessor macro, since it's limited to this file (I could be wrong in my assumption)? :-)

bhariharan:
Looks pretty good otherwise. Just the one question about ```#define``` for string constants.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSmartStoreTests.m
wmathurin: That way we don't need to have constant names that are unique across the entire compilation unit.

bhariharan:
:+1:

PR #1023, opened 2015-05-20
Smartsync bug fix

wmathurin:
- Not blocking re-runs of failed sync
- Not staying stuck at 0% in a sync up when there is nothing to sync up


khawkins:
:+1: Thanks for pulling in the failed sync re-run!

PR #1025, opened 2015-05-22
Setting the kdf_iter to its 2.x value because the 3.x is slow and als…

wmathurin:
…o to be able to open 2.x databases without any migration


bhariharan:
:+1:

PR #1026, opened 2015-05-22
Test for sync up without local changes

wmathurin:
(And tweak to the code to not fire more updates than previously when they are indeed local changes)


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -375,8 +375,8 @@ - (void) syncUp:(SFSyncState*)sync updateSync:(SyncUpdateBlock)updateSync failSy
     // Call smartstore
     NSArray* dirtyRecordIds = [target getIdsOfRecordsToSyncUp:self soupName:soupName];
     NSUInteger totalSize = [dirtyRecordIds count];
-    updateSync(nil, totalSize == 0 ? 100 : 0, totalSize, kSyncManagerUnchanged);
```
wmathurin: We introduced that line a couple of days ago - which meant one extra update than previously when there are changes.

bhariharan:
Looks good. ```master``` or ```unstable```?

wmathurin:
Master because I also changed the code -- i will also update the smartsync.a in the plugin repo.

bhariharan:
:+1:

Issue #1014, opened 2015-05-05
After upgrade to 3.2 logging out breaks all future requests


cgreening:
After upgrading to 3.2 when we logout of our application we can login to the application successfully, but all requests fail with the following error:

__Network action must have an instance hosts__

paweldudek:
@cgreening they will fail if you log in using the same account. Check out my PR for details.

akhileshgupta:
We are working on this issue and will keep you posted once the fix is released.

wmathurin:
We have released a patch yesterday (https://github.com/forcedotcom/SalesforceMobileSDK-iOS/tree/v3.2.1) which we believe fixed you issue.

jeffreyjackson:
Anyone know when the tags may get bumped for cocoapods to catch this?  I'd rather not run :head on our Podfile for local troubleshooting and support purposes.

bhariharan:
@jeffreyjackson Coming up soon.

jeffreyjackson:
cc @brad36

PR #1033, opened 2015-05-28
Fixed issues with network layer not being usable after signing

paweldudek:
This PR fixes https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1014 by resetting network shared instance for given account upon sign out. 

Also, minor code beautification :wink:


paweldudek:
Looks like this one should go to unstable as well...

PR #1035, opened 2015-05-29
Updating ID data now posts user did change notification

paweldudek:
With current implementation there's no way of knowing when `currentUser`s `idData` has been updated as it gets silently updated by `SFAuthenticationManager`. It's a bit contrary to behavior of the user account manager object - other changes trigger notifications. This makes working with `SFUserAccountManager` a bit hard. 

This PR changes this by adding logic that posts notification upon `currentUser`s `idData` update.


paweldudek:
And I guess this should go into unstable... :wink: :

PR #1036, opened 2015-05-29
Updating ID data now posts user did change notification

paweldudek:
With current implementation there's no way of knowing when `currentUser`s `idData` has been updated as it gets silently updated by `SFAuthenticationManager`. It's a bit contrary to behavior of the user account manager object - other changes trigger notifications. This makes working with `SFUserAccountManager` a bit hard. 

This PR changes this by adding logic that posts notification upon `currentUser`s `idData` update.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountConstants.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountConstants.h
@@ -80,6 +80,6 @@ typedef NS_OPTIONS(NSUInteger, SFUserAccountChange) {
 
     /** The ID data changed
      */
-    SFUserAccountChangeIdData = 1 << 5,
+    SFUserAccountChangeIdData = 1 << 6
 };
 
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountConstants.h
```diff
@@ -77,5 +77,9 @@ typedef NS_OPTIONS(NSUInteger, SFUserAccountChange) {
     /** The community ID changed
      */
     SFUserAccountChangeCommunityId  = 1 << 5,
+
+    /** The ID data changed
+     */
+    SFUserAccountChangeIdData = 1 << 5,
```
bhariharan: @paweldudek This should be ```1 << 6```, yeah?

paweldudek: Oh yes, this looks like a classic copy&paste failure. Fixed!

bhariharan:
Thanks!

PR #1040, opened 2015-06-01
FTS enhanced syntax

wmathurin:
- updated tests
- latest shared
- latest 3rd party dependencies (that has the libsqlcipher.a)


bhariharan:
:+1:

PR #1042, opened 2015-06-02
Simple touch id support in passcode screen

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAbstractPasscodeViewController.m
```diff
@@ -104,6 +110,25 @@ - (void)validatePasscodeFailed
     });
 }
 
+- (BOOL) canShowTouchId;
+{
+    LAContext *context = [[LAContext alloc] init];
+    return cachedPasscode != nil && [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:nil];
```
wmathurin: Can only use touch id if:
- it's available on the device
- we have cached the passcode in memory (i.e. the user has entered it once since app was launched)

bhariharan:
:+1:

kchitalia:
:+1:

PR #1044, opened 2015-06-08
Proper cleanup of CSFNetwork shared instance on logout

wmathurin:
* Reverted fix submitted by external contributor
* Cleaning up the CSFNetwork in SharedInstances on logout the same way we do with smarstore, sync manager, metadata manager etc

### DIFF
--- a/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
+++ b/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
@@ -49,7 +49,7 @@ NSString *CSFNetworkInstanceKey(SFUserAccount *user) {
     return [NSString stringWithFormat:@"%@-%@-%@", user.credentials.organizationId, user.credentials.userId, user.communityId];
 }
 
-@interface CSFNetwork() {
+@interface CSFNetwork() <SFAuthenticationManagerDelegate>{
     //Flag to ensure that we file CSFActionsRequiredByUICompletedNotification only once through out the application's life cycle
     NSString *_defaultConnectCommunityId;
 }
@@ -104,6 +104,16 @@ static NSMutableDictionary *SharedInstances = nil;
     return instance;
 }
 
++ (void)removeSharedInstance:(SFUserAccount*)userAccount {
+    @synchronized (SharedInstances) {
+        NSString *key = CSFNetworkInstanceKey(userAccount);
+        [SharedInstances removeObjectForKey:key];
+    }
+}
+
+#pragma mark -
+#pragma mark object lifecycle
+
 - (id)init {
     self = [super init];
     if (self) {
@@ -121,7 +131,7 @@ static NSMutableDictionary *SharedInstances = nil;
         self.actionQueue = dispatch_queue_create("com.salesforce.network.action", DISPATCH_QUEUE_SERIAL);
         
         NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
-        #ifdef SFPlatformiOS
+#ifdef SFPlatformiOS
 		[notificationCenter addObserver:self
 												 selector:@selector(applicationDidBecomeActive:)
 													 name:UIApplicationDidBecomeActiveNotification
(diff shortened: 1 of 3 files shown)


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
```diff
@@ -304,11 +302,11 @@ - (void)URLSession:(NSURLSession *)session downloadTask:(NSURLSessionDownloadTas
...
 }
 
-#pragma mark - Handling Login State
+#pragma mark - SFAuthenticationManagerDelegate
 
-- (void)userWillLogOutNotification:(NSNotification *)notification {
-    SFUserAccount *userAccount = notification.userInfo[kSFUserAccountKey];
-    [[self class] cleanupNetworkForUserAccount:userAccount];
+- (void)authManager:(SFAuthenticationManager *)manager willLogoutUser:(SFUserAccount *)user {
+    [[self class] removeSharedInstance:user];
```
khawkins: My only concern with this approach is that it gets into potentially tenuous territory.  In this implementation, it's probably fine.  If you had something like this, on the other hand:

```obj-c
- (void)authManager:(SFAuthenticationManager *)manager willLogoutUser:(SFUserAccount *)user {
    [[self class] removeSharedInstance:user];
    self.someOtherProperty = nil;
```

The app would crash, because after the `removeSharedInstance` call, the instance no longer exists, and yet the execution path is still in the instance's code base.

Ideally, it would be a class-level responsibility for consuming the notification and cleaning up.  In practice, this is difficult, because the class itself cannot adopt the delegate protocol.  You could implement a helper class to facilitate that process, but it gets slightly convoluted.

I think we just need to be aware of this fact, and tread lightly, at any rate.

khawkins: The other alternative to satisfy such a flow would be to consume the notification from `SFAuthenticationManager`, which *can* be done at the class level.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.h
```diff
@@ -176,11 +176,6 @@ extern NSString * const kSFUserLoggedInNotification;
 extern NSString * const kSFAuthenticationManagerFinishedNotification;
 
 /**
- Key for account object present in user info in notifications.
- */
-extern NSString * const kSFUserAccountKey;
```
khawkins: What benefit is there to taking this away?

wmathurin: I reverted the fix submitted by the external contributor (that line came with that fix)

khawkins: Ah yes, got it.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Queue/CSFNetwork.m
wmathurin: We have taken the exact same approach in SFSmartStore, SFSmartSyncCacheManager, SFSmartSyncMetadataManager, SFSmartSyncSyncManager.

Also, if you were to set some property to self, I don't think it would crash. Once the current instance is dropped from the shared instances dictionary, it doesn't get deallocated right away, my understanding is that it gets an autorelease message (inserted by ARC) and will only be deallocated when the autorelease pool is drained (probably when exiting the method).

khawkins: I dunno, I fixed a similar bug recently where where a class's delegate nil'ed out the class instance as part of its delegate method implementation, and subsequent lines in the class made the app crash, because the memory had been deallocated.  So it's certainly something that *can* come up.  Interestingly, that crash only happened on device, not the simulator.

Anyway, I just wanted to point it out.  It's probably fine.

bhariharan:
:+1:

PR #1048, opened 2015-06-10
Fix for bug affecting dependent actions + test

wmathurin:
Fix for the bug "Dependent action does not start after parent action completes if dependent action is added while parent action is in flight."
The solution was suggested in the bug.

The tricky part was to write a test where realistic mock actions can be stopped in flight to verify the issue first and the solution next.

### DIFF
--- a/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
+++ b/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
@@ -354,9 +354,9 @@ NSTimeInterval const CSFActionDefaultTimeOut = 3 * 60; // 3 minutes
         return;
     }
     
-    [self willChangeValueForKey:@"executing"];
+    [self willChangeValueForKey:@"isExecuting"];
     _executing = YES;
-    [self didChangeValueForKey:@"executing"];
+    [self didChangeValueForKey:@"isExecuting"];
 
     // If this is a duplicate action, the parent must have just completed, so we can safely process our response based
     // on the result of that parent.  By calling the completion handler, the response will automatically be sent,
@@ -517,13 +517,13 @@ NSTimeInterval const CSFActionDefaultTimeOut = 3 * 60; // 3 minutes
 }
 
 - (void)completeOperationWithError:(NSError *)error {
-    [self willChangeValueForKey:@"executing"];
-    [self willChangeValueForKey:@"finished"];
+    [self willChangeValueForKey:@"isExecuting"];
+    [self willChangeValueForKey:@"isFinished"];
     _executing = NO;
     _finished = YES;
     self.error = error;
-    [self didChangeValueForKey:@"executing"];
-    [self didChangeValueForKey:@"finished"];
+    [self didChangeValueForKey:@"isExecuting"];
+    [self didChangeValueForKey:@"isFinished"];
     
     self.responseData = nil;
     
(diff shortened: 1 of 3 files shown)


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
```diff
@@ -354,9 +354,9 @@ - (void)start {
         return;
     }
     
-    [self willChangeValueForKey:@"executing"];
```
wmathurin: See KVO-compliant properties paragraph in https://developer.apple.com/library/mac/documentation/Cocoa/Reference/NSOperation_class/

bhariharan:
:+1:

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetworkTests/MockNetworkTests.m
```diff
@@ -29,6 +29,52 @@
...
+@interface MockAction : CSFAction
+
+@property (nonatomic, strong) NSString *actionName;
+@property (nonatomic, strong) dispatch_semaphore_t semaphore;
+
+- (instancetype) initWithName:(NSString*)actionName;
+
+@end
+
+@implementation MockAction
```
wmathurin: Mock action that does an actual http call (to /services/data so it doesn't need an access token).
However the completeOperationWithError is overridden, it waits on a semaphore before calling its super.
The semaphore can be released by calling completeAction. That way we can stop the action in flight.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetworkTests/MockNetworkTests.m
```diff
@@ -29,6 +29,52 @@
...
+    if (result != 0 && error == nil) {
+        error = [NSError errorWithDomain:CSFNetworkErrorDomain
+                                    code:CSFNetworkInternalError
+                                userInfo:@{ NSLocalizedDescriptionKey: @"Semaphore wait timed out",
+                                            CSFNetworkErrorActionKey: self }];
+    }
+    [super completeOperationWithError:error];
+}
+
+- (BOOL)isEqualToAction:(CSFAction *)action {
```
wmathurin: We need the action to be seen as a duplicate if it has the same actionName

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
agosar: As per the "... Upon starting the operation, your start method should also update the execution state of the operation as reported by the executing property. You do this by sending out KVO notifications for the __executing__ key path, which lets interested clients know that the operation is now running. Your executing property must also provide the status in a thread-safe manner.", should you leave the keypath as __executing__ in the start method?

wmathurin: I think the properties are executing and finished and the KVO key paths are isExecuting and isFinished (see the next paragraph in the doc). Also with the executing key path, my test below does not succeed.

wmathurin:
Please see my last checkin https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/75f85b5c3b18f8bb6229679e755e8e48450fed0f
Looks like the actions were never deallocated before (because never fully finished). Once I fixed the KVO keypath (earlier checkin), I got a bunch of test failures (because in dealloc we were removing observers that had never been registered - they were added too soon in the constructor).

bhariharan:
:+1:

PR #1049, opened 2015-06-12
Fix for resync bugs

wmathurin:
At start up we mark any syncs that were running when the app quits as failed to allow resync


wmathurin:
Working on a better solution - the change above doesn't help when the app is backgrounded (instead of killed) during a sync

PR #1050, opened 2015-06-13
Fix for resync bug

wmathurin:
Sync manager only does one sync at a time, so we now capture the currently running sync id in a property.
Resync checks if that property value matches the one you want to re-sync -- and only in that case does it not run.
So if the original sync was interrupted by backgrounding or killing the app, resync should succeed.

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -128,7 +128,7 @@ static NSMutableDictionary *syncMgrList = nil;
         self.queue = dispatch_queue_create(kSyncManagerQueue,  NULL);
         [[SFAuthenticationManager sharedManager] addDelegate:self];
         [SFSyncState setupSyncsSoupIfNeeded:self.store];
-        [SFSyncState cleanupUnfinishedSyncs:self.store];
+        [SFSyncState cleanupUnfinishedSyncs:self.store pageSize:kSyncManagerPageSize];
     }
     return self;
 }
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -57,6 +57,7 @@ typedef void (^SyncFailBlock) (NSString* message, NSError* error);
 
 @property (nonatomic, strong) SFSmartStore *store;
 @property (nonatomic, strong) dispatch_queue_t queue;
+@property (atomic) NSInteger syncIdRunning;
 
 @end
 
@@ -128,7 +129,6 @@ static NSMutableDictionary *syncMgrList = nil;
         self.queue = dispatch_queue_create(kSyncManagerQueue,  NULL);
         [[SFAuthenticationManager sharedManager] addDelegate:self];
         [SFSyncState setupSyncsSoupIfNeeded:self.store];
-        [SFSyncState cleanupUnfinishedSyncs:self.store pageSize:kSyncManagerPageSize];
     }
     return self;
 }
@@ -165,6 +165,19 @@ static NSMutableDictionary *syncMgrList = nil;
         [sync save:weakSelf.store];
         
         [weakSelf log:SFLogLevelDebug format:@"Sync update:%@", sync];
+        
+        
+        switch (sync.status) {
+            case SFSyncStateStatusNew:
+                break; // should not happen
+            case SFSyncStateStatusRunning:
+                weakSelf.syncIdRunning = sync.syncId;
+                break;
+            case SFSyncStateStatusDone:
+            case SFSyncStateStatusFailed:
+                weakSelf.syncIdRunning = -1;
+                break;
+        }
         
         if (updateBlock)
             updateBlock(sync);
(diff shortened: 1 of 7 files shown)


libs/SmartSync/SmartSyncTests/SyncManagerTests.m
```diff
@@ -763,6 +807,41 @@ - (void) testAddFilterForResync
...
+- (void) testResyncRunningSync
+{
+    // Create test data
+    [self createTestData];
+    
+    // Ids clause
+    NSString* idsClause = [self buildInClause:[idToNames allKeys]];
+    
+    // Create sync
+    NSString* soql = [@[@"SELECT Id, Name, LastModifiedDate FROM Account WHERE Id IN ", idsClause] componentsJoinedByString:@""];
```
wmathurin: We run a slow sync down (a custom sync down that is basically a soql sync down that sleeps for a second before fetching from the server). I didn't use a semaphore here because the sync target gets serialized/deserialized and the target instance that the tests builds will not be the one used by the sync manager.

We start the sync, call re sync, expect nil, then we wait for sync and retry the resync and this time we expect it to run.

libs/SmartSync/SmartSyncTests/SyncManagerTests.m
```diff
@@ -976,7 +1055,7 @@ - (NSString*) createLocalId
 
 - (void) deleteSyncs
 {
-    [store clearSoup:ACCOUNTS_SOUP];
+    [store clearSoup:kSFSyncStateSyncsSoupName];
```
wmathurin: Syncs were not cleaned up between tests because the clear was going to the wrong soup.

wmathurin:
I should do the same thing on the Android side.
@akhileshgupta do you want that fix in 3.2.1 ?

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -209,6 +223,11 @@ - (SFSyncState*) syncDownWithTarget:(SFSyncDownTarget*)target options:(SFSyncOpt
 /** Resync
  */
 - (SFSyncState*) reSync:(NSNumber*)syncId updateBlock:(SFSyncSyncManagerUpdateBlock)updateBlock {
+    if (syncId.integerValue == self.syncIdRunning) {
```
khawkins: But this is imposing an artificial constraint on the sync manager that doesn't otherwise exist: only one sync can run at a time.  That's a pretty drastic limitation to apps that manage many different sync data sets.

khawkins: Seems like this would be a good base for implementing a reasonable `isEqual:` on `SFSyncState`, and instead of having a single `syncIdRunning` property for the sync manager, have a map of `SFSyncState` objects to their running sync IDs.  Then you could have a similar comparison, and keep the ability to run multiple syncs at a time.

bhariharan:
@wmathurin @akhileshgupta I'm going to hold off on patching the ```Cordova``` fix on ```Android``` until we make a decision on whether we should patch this on ```Android``` too.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
wmathurin: This change is not introducing this constraint. Currently all syncs are dispatched through a serial queue and therefore there is only one running at given time (for a given store).
If we relax that constraint, we will have to at least prevent two concurrent syncs to/from the same soup. And then keeping track of the collection of running syncs like you suggested will be needed.

khawkins: Are we sure about that?  I know that we use a serial queue to initiate the dispatch of a sync, but there are a number of asynchronous processes that happen after the initial dispatch—most notably an async network service that dispatches independently of the SmartSync dispatching.  The serial execution of a dispatch queue is only as serial as the synchronous code encapsulated within the dispatch.  I haven't had time to test it out, but my guess is that, once a network request has been queued for one sync, another sync will be freed to start.

wmathurin: You are right. On the android side, everything is synchronous (smartstore / network) and dispatching the sync to a thread pool with a single worker guarantees that all syncs are done serially. But on iOS, networking is asynchronous, and so concurrent syncs are possible. There means bad things could happen if the developers are not careful. I feel we should limit concurrency (either completely like the current Android's implementation) or at least by soup.
For the resync fix, we could continue "living dangerously" and use a collection of "running sync ids" instead of a single "running sync id".

wmathurin:
@khawkins check out https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/1778583bf1c22fb82e5aa348b33e4f3b878e79bd now capturing running sync ids in a set

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -124,8 +125,9 @@ + (void)removeSharedInstanceForStore:(SFSmartStore *)store {
 - (instancetype)initWithStore:(SFSmartStore *)store {
     self = [super init];
     if (self) {
+        self.runningSyncIds = [NSMutableSet new];
         self.store = store;
-        self.queue = dispatch_queue_create(kSyncManagerQueue,  NULL);
+        self.queue = dispatch_queue_create(kSyncManagerQueue,  DISPATCH_QUEUE_SERIAL);
         [[SFAuthenticationManager sharedManager] addDelegate:self];
```
wmathurin: No change here DISPATCH_QUEUE_SERIAL is #define'd as NULL.

bhariharan:
:+1:

khawkins:
Sorry, I hadn't had a chance to get back to this.  Those changes look good to me.  :+1:

PR #1052, opened 2015-06-15
Fix crash caused by SFJsonUtils passing invalid json object

bhariharan:


bhariharan:
From Apple's doc:

```"If obj will not produce valid JSON, an exception is thrown. This exception is thrown prior to parsing and represents a programming error, not an internal error. You should check whether the input will produce valid JSON before calling this method by using isValidJSONObject:."```

wmathurin:
:+1:

khawkins:
:+1:

PR #1053, opened 2015-06-15
Fix for another crash

bhariharan:


bhariharan:
Original pull request came from the ```iOS``` team. Simply pulling this upstream.

kchitalia:
:+1:

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFSmartStore/SFSmartStorePlugin.m
```diff
@@ -155,12 +155,12 @@ - (void)pgQuerySoup:(CDVInvokedUrlCommand *)command
...
-        NSError* error;
+        NSError* error = nil;
         SFStoreCursor* cursor = [self runQuery:querySpec error:&error argsDict:argsDict];
-        if (cursor) {
-            if ([self isGlobal:argsDict]) {
+        if (cursor.cursorId) {
+            if ([self isGlobal:argsDict] && self.globalCursorCache) {
                 (self.globalCursorCache)[cursor.cursorId] = cursor;
-            } else {
+            } else if (self.userCursorCache) {
```
wmathurin: When does the cursor cache end up being nil?
Should we check for it not being nil in the other places where we access it?

PR #1060, opened 2015-06-17
Community ID shouldn't be nil'ed out when user accounts are loaded

khawkins:
Not sure why the community ID was being nil'ed originally, but the end result is that, when the app is launched again, the community ID of the current user will not be persisted.


wmathurin:
:+1:

bhariharan:
:+1:

PR #1061, opened 2015-06-18
kdf_iter fix

wmathurin:
We can't blindly use 4000 iterations for kdf_iter: if the database was created with SDK 3.1 or SDK 3.2 and cocoa pods, it will be using sqlcipher 3.1 with a kdf_iter of 64000.


bhariharan:
:+1:

Issue #1062, opened 2015-06-19
SmartSync: App crashes if sync up an object that has been deleted on server side


tuyennguyen266:
Step 1: sync down all data
Step 2: delete an object on server side
Step 3: modify that object on client side
Step 4: sync up then app crashes.

May be the clue below can help:
When call method fetchRecordModificationDates... to compare serverTime and localTime, there is no record returned from server make the following code crash.
```
        if (nil != response) {
            NSDictionary *record = response[@"records"][0]; // CRASH HERE
            if (nil != record) {
                NSString *serverLastModifiedStr = record[self.modificationDateFieldName];
                if (nil != serverLastModifiedStr) {
                    NSDate *testServerModifiedDate = [SFSmartSyncObjectUtils getDateFromIsoDateString:serverLastModifiedStr];
                    if (testServerModifiedDate != nil) {
                        serverLastModifiedDate = testServerModifiedDate;
                    }
                }
            }
        }
```

wmathurin:
We are reproducing the issue here and we are working on a fix.

wmathurin:
FYI, We fixed the issue in the unstable branch. It will ship in the next patch / release.

PR #1065, opened 2015-06-19
Sync up should handle remotely deleted records

wmathurin:
Reported here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1062

Previous behavior:
If one updates a remotely deleted record, the sync up fails (with a crash if merge mode was leave if changed)

New behavior:
If merge mode is overwrite, then the record is recreated on the server, the local record gets the new server id
If merge mode is leave if changed, then the local record is left unchanged / dirty
In both cases the sync up should continue

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -491,9 +491,27 @@ static NSMutableDictionary *syncMgrList = nil;
                   updateSync:(SyncUpdateBlock)updateSync
                    failBlock:(SFSyncUpTargetErrorBlock)failBlock {
     
+    SFSyncStateMergeMode mergeMode = sync.mergeMode;
     SFSyncUpTarget *target = (SFSyncUpTarget *)sync.target;
     NSString* soupName = sync.soupName;
     NSNumber* soupEntryId = record[SOUP_ENTRY_ID];
+    NSArray *fieldList = sync.options.fieldlist;
+    
+    // Getting type and id
+    NSString* objectType = [SFJsonUtils projectIntoJson:record path:kObjectTypeField];
+    NSString* objectId = record[target.idFieldName];
+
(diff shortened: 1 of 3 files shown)


libs/SmartSync/SmartSync/Classes/Util/SFSyncUpTarget.m
```diff
@@ -128,7 +128,7 @@ - (void)fetchRecordModificationDates:(NSDictionary *)record
     };
     
     id completeBlock = ^(NSDictionary *response) {
-        if (nil != response) {
```
wmathurin: Blindly accessing array index 0 would cause the crash when the remote record was gone

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -526,31 +544,30 @@ - (void)resumeSyncUpOneEntry:(SFSyncState*)sync
...
-    // Fields to save (in the case of create or update)
-    NSMutableDictionary* fields = [NSMutableDictionary dictionary];
-    if (action == SFSyncUpTargetActionCreate || action == SFSyncUpTargetActionUpdate) {
-        for (NSString *fieldName in fieldList) {
-            if (![fieldName isEqualToString:target.idFieldName] && ![fieldName isEqualToString:target.modificationDateFieldName]) {
-                NSObject* fieldValue = [SFJsonUtils projectIntoJson:record path:fieldName];
-                if (fieldValue != nil)
-                    fields[fieldName] = fieldValue;
+    // Update failure handler
+    SFSyncUpTargetErrorBlock failBlockUpdate = ^ (NSError* err){
```
wmathurin: When we get a 404 during an update, we don't fail anymore
We either recreate the record (if merge mode is overwrite)
Or we leave the local record unchanged (if merge mode is leave if changed)

bhariharan:
:+1:

khawkins:
Looks good.

PR #1071, opened 2015-06-25
Advanced user tracking

wmathurin:
Adding two pieces of information to the user agent:
- a randomly generated id to distinguish users doing anonymous requests or sharing sfdc accounts
- a counter incremented every time the app is foregrounded 

@akhileshgupta

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
@@ -46,7 +46,7 @@ static NSString * const kSFMobileSDKHybridDesignator = @"Hybrid";
 static NSString * const kAppSettingsAccountLogout = @"account_logout_pref";
 
 // Random id to uniquely identify user
-static NSString* uid;
+static NSString* uid = nil;
 
 // Incremented every time the app is foregrounded
 static int countForegrounds = 1;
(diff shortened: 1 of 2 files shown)


bhariharan:
Looks good. :+1: 

What is the purpose of counting number of times the user foregrounds the app (we can talk about this offline)?

wmathurin:
From that we could derive whether they do go into their app often or not, and whether they stay in app for long periods of time or not.

wmathurin:
Capturing the same information on Android is going to be tru

bhariharan:
:+1: 

```Activity``` foregrounding is the closest on ```Android```, unless we do some smart computations. I'm not aware of any direct ```Application``` foregrounding API.

**Comment by wmathurin** [comment]
True but they could write their own activities.

**Comment by bhariharan** [comment]
Sure, but we recommend subclassing ```SalesforceActivity```. That's not a good way anyway, because it will be triggered even if another ```Activity``` is pushed on the stack and then popped. This will lead to a lot of false counts.

The closest way (and this can be painful) is to listen for package launch notifications in a ```Service```, and every time the package is launched, it counts as a foregrounding of the app.

File: libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m [hunk 1]
```diff
@@ -45,6 +45,13 @@
 // app when it is re-opened.
 static NSString * const kAppSettingsAccountLogout = @"account_logout_pref";
 
+// Random id to uniquely identify user
+static NSString* uid;
```
khawkins: It's always good to initialize static objects to `nil`, for some weird edge case stuff.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -52,6 +59,7 @@ + (instancetype)sharedManager
     static dispatch_once_t pred;
     static SalesforceSDKManager *sdkManager = nil;
     dispatch_once(&pred, ^{
+        uid = [SalesforceSDKManager randomStringWithLength:12];
```
khawkins: Seems fine.  Seems easier to just use an `NSUUID`, but either way works.

khawkins:
Overall, :+1: from me.

PR #1073, opened 2015-06-25
Long queries should be allowed

bhariharan:
The query length is not limited to ```1024``` anymore.


wmathurin:
A new test maybe (could be in SalesforceRestAPITests - a soql query with  a select ... in very long)?

bhariharan:
I'll try to come up with a long query.

bhariharan:
@wmathurin Added a test [here](https://github.com/bhariharan/SalesforceMobileSDK-iOS/commit/229153e229e5320f2e31f28e39c890691902f29e).

wmathurin:
:+1:

PR #1076, opened 2015-06-26
Fix for SmartStore not getting wiped using 'Forgot Passcode'

bhariharan:
The problem was that ```SmartStore``` wouldn't get the delegate call because it hadn't been initialized yet. ```SalesforceSDKManager``` is the only class that's guaranteed to be initialized when an app starts, so this is better suited there.

I have verified it works on both native and hybrid apps. Didn't add a test, because logout and logging back in is a pain.


wmathurin:
:+1:

PR #1080, opened 2015-06-27
Fix for push alert bug - showing push alert only after launch

wmathurin:
Related bug: Login screen disappears after accepting the push notification alert


bhariharan:
:+1:

khawkins:
:+1:

PR #1081, opened 2015-06-29
Fix for compilation error

bhariharan:
We changed the signature on this method a couple of days ago.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSmartStoreTests.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSmartStoreTests.m
@@ -690,7 +690,7 @@
 - (void)createDbDir:(NSString *)dbName withManager:(SFSmartStoreDatabaseManager *)dbMgr
 {
     NSError *createError = nil;
-    [dbMgr createStoreDir:dbName error:&createError];
+    [dbMgr createStoreDir:dbName];
     XCTAssertNil(createError, @"Error creating store dir: %@", [createError localizedDescription]);
 }
 


wmathurin:
Thanks. I wish xcode would build everything and not just the currently selected project...

kchitalia:
:+1:

bhariharan:
Yeah, it sucks that the test targets are different.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSmartStoreTests.m
```diff
@@ -690,7 +690,7 @@ - (void)removeStoreForManager:(SFSmartStoreDatabaseManager *)dbMgr withName:(NSS
 - (void)createDbDir:(NSString *)dbName withManager:(SFSmartStoreDatabaseManager *)dbMgr
 {
     NSError *createError = nil;
-    [dbMgr createStoreDir:dbName error:&createError];
+    [dbMgr createStoreDir:dbName];
     XCTAssertNil(createError, @"Error creating store dir: %@", [createError localizedDescription]);
```
briansimmonsSF: I think it would make more sense to check the return value (BOOL) of the method "createStoreDir" b/c createError is always going to be nil now (i.e. it's never modified).

wmathurin: You are right, we should XCTAssertTrue the return value of [dbMgr createStoreDir]

wmathurin: fixing here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1083

PR #1083, opened 2015-06-29
Fixing tests

wmathurin:
A few commits back openDatabase was changed to only return readable databases

See https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/f4074e132e1b467afcad9d156665bc830c9a8a30


bhariharan:
:+1:

PR #1090, opened 2015-07-01
Disabling ATS

wmathurin:
Now able to run apps/tests on iOS 9


bhariharan:
So we uncomment those pieces once the server side work is done?

:+1:

wmathurin:
Currently the server support TLS 1.2 but not forward secrecy. Once they do we won't need any of those changes.
The commented out block doesn't disable ATS entirely but only disable the requirement for forward secrecy. I couldn't get it to work. We should try it out again once Xcode 7 beta 3 comes out.

PR #1091, opened 2015-07-02
For smartstore to be available in the background parent directories must not use NSFileProtectionComplete

wmathurin:
To try it out go to https://github.com/wmathurin/BackgroundTester
In Podfile, uncomment one of the configs
```
# for Mobile SDK 3.3 with parent directory using NSFileProtectionComplete ==> doesn't work when phone locked
# pod 'SalesforceMobileSDK-iOS', :git => 'https://github.com/wmathurin/SalesforceMobileSDK-iOS.git', :commit => '63c2627', :submodules => 'true'

# for Mobile SDK 3.3 with parent directory using NSFileProtectionCompleteUntilFirstUserAuthentication ==> works when phone locked
# pod 'SalesforceMobileSDK-iOS', :git => 'https://github.com/wmathurin/SalesforceMobileSDK-iOS.git', :commit => 'e3a3e93', :submodules => 'true'

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.m
@@ -45,7 +45,7 @@ static NSString * const kDefaultCommunityName = @"internal";
     if (![manager fileExistsAtPath:directory]) {
         return [manager createDirectoryAtPath:directory
                   withIntermediateDirectories:YES
-                                   attributes:@{NSFileProtectionKey: NSFileProtectionComplete}
+                                   attributes:@{NSFileProtectionKey: NSFileProtectionCompleteUntilFirstUserAuthentication}
                                         error:error];
     } else {
         return YES;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.m [hunk 1]
```diff
@@ -45,7 +45,7 @@ + (BOOL)ensureDirectoryExists:(NSString*)directory error:(NSError**)error {
     if (![manager fileExistsAtPath:directory]) {
         return [manager createDirectoryAtPath:directory
                   withIntermediateDirectories:YES
-                                   attributes:@{NSFileProtectionKey: NSFileProtectionComplete}
+                                   attributes:@{NSFileProtectionKey: NSFileProtectionCompleteUntilFirstUserAuthentication}
```
khawkins: Is there any upgrade step, for apps where the directory has previously been created?

wmathurin: @khawkins indeed and upgrade step is needed and is missing here, I'll add that tomorrow.

wmathurin:
With https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/e3a3e93c317db2ca009713cbc3d9632e6a490836 taking care of upgrade scenarios.
protectStoreDirIfNeeded (which is called defensively when opening existing smartstore) will check/set the right filesystem protection for the file and its parent and ancestor directories.

khawkins:
Seems good. :+1:

bhariharan:
:+1:

PR #1095, opened 2015-07-04
Add Content-Length header to POST requests

khawkins:
Some server endpoints want the Content-Length header on their POSTs.


bhariharan:
:+1:

camflan:
This is still an issue on iOS 7, works on iOS 8 and iOS 9. Any ideas?

PR #1098, opened 2015-07-06
Removing long generated method name

rwhitley:
Fixing the standard problem with the iOS reference master TOC format. We need to add this manual step to the release instructions.


bhariharan:
:+1: Thanks @rwhitley.

PR #1103, opened 2015-07-07
Custom Request Parsing

Jugale:
##### Description
This fixes a case where a picklist request such as the following cannot be parsed:

```objc
SFRestRequest *req =[SFRestRequest requestWithMethod:SFRestMethodGET path:@"/<redacted>/picklistValues" queryParams:nil];
req.endpoint = @"/services/apexrest";
```

The issue is the JSON that the API returns for the above request is actually a string. With this change there will be no failure within the `Salesforce iOS SDK` and the user can then parse the result using `NSJSONSerialization` again.

The option `NSJSONReadingAllowFragments` shouldn't break any other other request parsing as it adds supports for "fragments" rather than making them required

##### Next steps
Maybe we want to add another check underneath line 540 with something like:
```objc
if (!error && [content isKindOfClass:[NSString class]]) {

    NSString *stringContent = [content copy];
    content = [NSJSONSerialization JSONObjectWithData:[stringContent dataUsingEncoding:NSUTF8StringEncoding] options:kNilOptions error:&jsonParseError];
}
```


bhariharan:
@Jugale Could you send us a pull request against our ```unstable``` branch please?

PR #1106, opened 2015-07-09
CSFAction should allow requests to any URL

bhariharan:


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
```diff
@@ -60,10 +60,9 @@ + (NSURL*)urlForAction:(CSFAction*)action error:(NSError**)error {
     if (!action) {
         return nil;
     }
-    
-    NSMutableString *baseUrlString = [NSMutableString stringWithString:[action.enqueuedNetwork.account.credentials.apiUrl absoluteString]];
+    NSMutableString *baseUrlString = [NSMutableString stringWithString:[action.baseURL absoluteString]];
```
bhariharan: The basic bug - this factory method shouldn't be married to ```Salesforce``` - it should rely on the ```action``` passed in for ```baseURL```.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -273,4 +273,9 @@ - (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
     }
 }
 
+- (NSURLRequest*)createURLRequest:(NSError**)error {
+    self.baseURL = self.enqueuedNetwork.account.credentials.apiUrl;
```
bhariharan: Setting ```baseURL``` in the subclass (this one's for Salesforce specific actions). Ideally, we would want this in the ```init``` method, but sadly that's not possible, because setting the ```baseURL``` requires ```enqueuedNetwork``` to be setup first, which only happens on ```start```, which in turn only happens when the action is enqueued on the network object (can't be done before the action's ```init``` happens, and it depends on the caller). Long story short - this is where it was being set before, except that it was being set in ```CSFAction```, and now it's set in the subclass.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
```diff
@@ -131,16 +129,19 @@ + (NSError *)errorInResponseDataForAction:(CSFAction*)action {
...
     NSString *baseString = nil;
     if (url.port) {
         baseString = [NSString stringWithFormat:@"%@://%@:%@", url.scheme, url.host, url.port];
     } else {
         baseString = [NSString stringWithFormat:@"%@://%@", url.scheme, url.host];
     }
-    
     action.baseURL = [NSURL URLWithString:baseString];
-    action.verb = url.path;
+    NSMutableString *relativePath = [NSMutableString stringWithString:url.path];
```
bhariharan: This bug is somewhat interesting. We read the ```path``` here from the URL, but completely ignore any query params that may be set on the URL itself. This snippet of code adds the query params to ```verb``` too, which in turn adds them as ```CSFParameterStorage```. That's the good news. The bad news is that ```action.params``` is ignored as well in ```urlForAction```. I attempted to fix that, but it's impossible without some serious changes to the way that stuff is stored/handled/architected. That's a battle for another day.

Moral of the story: This class wasn't really built for external URLs @khawkins.

bhariharan:
Any objections if I merge this in @khawkins?

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
khawkins: I was afraid of that.  Well, it's the best we can do for now.

khawkins:
It's g2g from me. :+1:

PR #1109, opened 2015-07-11
POST requests cannot be replayed after auth refresh

khawkins:
The issue is that POST bodies are represented as `NSInputStream` objects, which are used as the `NSURLRequest.HTTPBodyStream` in the initial request.  But once the input stream is read through once in the initial request, it's no longer valid.  This presents a problem if the request needs to be replayed after a 401 from the original request.

I changed the input stream inputs across the board, to a block that returns the `NSInputStream`.  That way, the stream can be regenerated from its source, with the initiation of each request.


bhariharan:
:+1:

PR #1110, opened 2015-07-14
Fixing bug is POST duplicate comparison

bhariharan:


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Support/CSFParameterStorage.m
```diff
@@ -63,6 +63,9 @@ - (BOOL)isEqual:(CSFParameterStorage*)object {
     if (![self.mimetypes isEqualToDictionary:object.mimetypes])
         return NO;
 
+    if (![self.bodyStreamBlock isEqual:object.bodyStreamBlock])
```
khawkins: Generally speaking, this `isEqual:` implementation doesn't cover the `nil == nil` case, since a `nil` parameter is always going to return `NO` for any comparison.

bhariharan:
@khawkins I added ```nil``` case comparison.

khawkins:
The whole method needs it. :)

bhariharan:
Lol....I'll add it in.

bhariharan:
Done!

khawkins:
Sweet!  Looks good to me otherwise. :+1:

PR #1111, opened 2015-07-16
Setting default user agent

bhariharan:
We weren't setting any user agent on ```CSFNetwork```. All outgoing requests through ```RestApiExplorer``` currently have ```User-Agent``` set to ```nil```.


kchitalia:
This is just for my understanding, we need to send the user agent in Network Requests for?

bhariharan:
@kchitalia I'll tell you offline :-)

kchitalia:
:+1:

PR #1112, opened 2015-07-17
Round of bug fixes

bhariharan:


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
```diff
@@ -541,8 +541,8 @@ - (id)contentFromData:(NSData*)data fromResponse:(NSHTTPURLResponse*)response er
         content = [NSJSONSerialization JSONObjectWithData:data options:NSJSONReadingAllowFragments error:&jsonParseError];
     }
     
-    // If it's an error here, it's a basic parsing error.
-    if (jsonParseError && error) {
+    // If it's an error here, it's a basic parsing error and error is not 500 internal server error.
+    if (jsonParseError && error && response.statusCode != 500) {
```
bhariharan: In the absence of this, ```response``` gets polluted with raw HTML.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -197,12 +197,19 @@ - (id)contentFromData:(NSData*)data fromResponse:(NSHTTPURLResponse*)response er
                 network.securityToken = securityToken;
             }
         }
+    } else if (!content && !responseError && self.statusCodeToRaiseErrorWithEmptyResponse.count > 0) {
```
bhariharan: This is for cases where an app might want to treat certain response codes as errors, such as ```500``` in the above case, or ```304``` even (when ```etag``` headers are used and the content hasn't changed).

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
khawkins: This seems...an odd approach.  Wouldn't it make more sense to put the status code check in `completeOperationWithResponse:`, before reading the content?  Because I'm reading this as: 500 errors should be treated independently of response content parsing.  If that's the case, this doesn't belong here.

khawkins: Oh I see, it's supposed to fall through.  It seems to me that a better approach would be to institute classes of error handling here, like `CSFSalesforceAction` has.  For instance, maybe all non-4xx errors fall through the default JSON parser, rather than being parsed.  The parsing is predicated on the idea that the REST APIs will send back error responses for bad requests, and that's probably not going to happen with a non-4xx response code.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
khawkins: This seems like it's working around the existing design pattern of having an inherited class that overrides `contentFromData:fromResponse:error:`, if you need custom error handling behavior.

khawkins: Though I could see merit in having a way for the consumer to get a "first shot" at overriding error handling behavior in `CSFAction`.  In such a model, I would propose creating a block property on `CSFAction` that takes the content and response as an input, and returns an `NSError *` as output.  If set, you could put that in front of `CSFAction`'s call to `contentFromData:fromResponse:error:`.  The block would be responsible for returning an error if the response didn't pass its custom error handling, and `nil` if it didn't catch its custom errors.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
bhariharan: The fix came from @qingqingliu. @qingqingliu What do you think?

lqingq:
Yes, based on the suggestion, using sub-class pattern should be the way to go for 304 error. On the other hand though, I am questioning the default behavior of trying to parse response as JSON when server side returns error code and return response as error when failed to do that. That to me seems to be a bug and we should not use subclass to work around it, I would suggest to fix it by at least not return error when JSON parse failed.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
lqingq: Agree with Kevin, we can make the change so that JSON is only parsed for 4xx error

lqingq:
I will make change to use sub class

PR #1113, opened 2015-07-20
Prevent overwriting umbrella header if it hasn't changed

bhariharan:


bhariharan:
Original pull request from @jeanbovet.

khawkins:
:+1:

PR #1114, opened 2015-07-20
Empty index columns should be cleared on update

bhariharan:


bhariharan:
@khawkins I applied your pull request here.

khawkins:
Cool thanks! :+1:

PR #1119, opened 2015-07-27
Bug fixes

bhariharan:


bhariharan:
Per apple documentation using ```[[NSFileManager alloc] init] is better than [NSFileManager defaultManager], since the latter returns the same instance and can lock up.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecurityLockout.m
```diff
@@ -353,6 +352,7 @@ + (void)setupTimer
                                       selector:@selector(timerExpired:)
                                  timerInterval:securityLockoutTime];
 	}
+    [SFInactivityTimerCenter updateActivityTimestamp];
```
bhariharan: Passcode bugfix.

bhariharan:
Has been reviewed earlier - pulling it upstream from soma.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -383,10 +383,10 @@ - (BOOL)logoutOnInvalidCredentials:(NSError *)error
 - (NSURL *)fullFileUrlForPage:(NSString *)page
 {
     NSString *fullPath = [[[[NSBundle mainBundle] resourcePath] stringByAppendingPathComponent:self.wwwFolderName] stringByAppendingPathComponent:page];
-    if (![[NSFileManager defaultManager] fileExistsAtPath:fullPath]) {
```
wmathurin: What's wrong with the defaultManager?

bhariharan: It doesn't make a difference in this particular class, but might make a difference in ```UserAccountManager``` or ```SmartStore```:

Per apple documentation using ```[[NSFileManager alloc] init]``` is better than ```[NSFileManager defaultManager]```, since the latter returns the same instance and can lock up.

wmathurin:
:+1:

PR #1120, opened 2015-07-27
Move to NSURLSession

wmathurin:


bhariharan:
Looks good to me. Should we take this opportunity to use our network stack (```CSFAction``` and ```CSFNetwork```), instead of using ```NSURLSession``` directly? ```CSFAction``` should now work without requiring authentication to complete ;-)

khawkins:
@bhariharan I think there would be a little more thinking/finagling required, since the Network SDK still lives conceptually "above" SalesforceSDKCore.  It could be refactored for some of the cases, but others like SalesforceOAuth would create circular dependencies.

bhariharan:
Oh right. Thanks for jogging my memory @khawkins! :+1:

PR #1125, opened 2015-07-29
Refactor and bug fixes around anonymous and temporary user states

bhariharan:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecurityLockout.m
```diff
@@ -404,10 +404,7 @@ + (void)unlock:(BOOL)success action:(SFSecurityLockoutAction)action passcodeConf
 + (void)timerExpired:(NSTimer*)theTimer
 {
     [self log:SFLogLevelInfo msg:@"Inactivity NSTimer expired."];
-    [SFSecurityLockout setLockScreenFailureCallbackBlock:^{
-        [[SFAuthenticationManager sharedManager] logout];
-    }];
-	[SFSecurityLockout lock];
+    [SFSecurityLockout lock];
```
bhariharan: This one is actually a duplicate callback post failure.

bhariharan:
Thanks @jeanbovet for doing the heavy-lifting on this. I have tweaked things to fit in our current version, but it's pretty much the same as the work you did.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -584,45 +688,100 @@ - (SFUserAccount *)loadUserAccountFromFile:(NSString *)filePath {
...
+
+    // Try to load the user account assuming it is encrypted
+    SFUserAccount *user = nil;
+    if ([self loadUserAccountFromFile:filePath encrypted:YES account:&user error:nil]) {
+        return user;
+    } else {
+
+        // Unable to retrieve the encrypted user, maybe it's still in the old plain text format?
+        NSError *error = nil;
+        if ([self loadUserAccountFromFile:filePath encrypted:NO account:&user error:&error]) {
```
wmathurin: My ios9 fix (around the fact that unarchiveObjectWithFile doesn't throw an exception but return nil) need to make its way into loadUserAccountFromFile

bhariharan: Good catch @wmathurin! I've added that check back in.

jeanbovet:
Looks good, thanks @bhariharan !

wmathurin:
:+1:

PR #1127, opened 2015-07-31
Using cocoapod in native template

wmathurin:
There are also changes in forceios.js (separate pull request since it lives in SalesforceMobileSDK-iOS-Package)
After these changes we should probably bring forceios.js into the SalesforceMobileSDK-iOS repo (just like we do on the Android side) and get rid of the SalesforceMobileSDK-iOS-Package repo


build/app_template_files/createApp.sh.base
```diff
@@ -76,14 +71,13 @@ function usage()
   local appName=`basename $0`
   echoColor $TERM_COLOR_CYAN "Usage:"
   echoColor $TERM_COLOR_MAGENTA "$appName"
-  echoColor $TERM_COLOR_MAGENTA "   -t <Application Type> (native, hybrid_remote, hybrid_local)"
+  echoColor $TERM_COLOR_MAGENTA "   -t <Application Type> (native)"
```
wmathurin: Keeping the type option even though there is only one right now, because we will soon introduce support for swift native app.
Or maybe the language (swift or objective-c) should be a different option?

kchitalia:
Looks good

build/app_template_files/__NativeTemplateAppFiles__/Podfile
```diff
@@ -0,0 +1,9 @@
+# Uncomment this line to define a global platform for your project
+# platform :ios, '6.0'
+
+target '__NativeTemplateAppName__' do
+
+pod 'SalesforceMobileSDK-iOS', '~> 3.3'
```
bhariharan: Can we move ```3.3``` to a variable here and add a comment, so that we remember to update it every time we release?

jeffreyjackson:
you can drop the  `, '~> 3.3'` and it will pick up the newest release automatically.

build/app_template_files/createApp.sh.base
```diff
@@ -173,7 +159,7 @@ function parseOpts()
   
   # Validate that we got the required command line args.
   if [[ "${OPT_APP_TYPE}" == "" ]]; then
-    echoColor $TERM_COLOR_RED "No option specified for Application Type.  Must be 'native' or 'hybrid'."
+    echoColor $TERM_COLOR_RED "No option specified for Application Type.  Must be 'native'."
```
bhariharan: Do we even need this option anymore, if the only option is ```native```?

bhariharan:
Looks good otherwise.

build/app_template_files/__NativeTemplateAppFiles__/Podfile
jeffreyjackson: @bhariharan you can drop the , '~> 3.3' and it will pick up the newest release automatically.

build/app_template_files/createApp.sh.base
wmathurin: There are more native times coming soon (native with swift, native with react).

PR #1129, opened 2015-08-03
Forceios now lives in SalesforceMobileSDK-iOS

wmathurin:
The npm package generation is really quick (under 5s) and the generated package is much smaller (about 5.8 MB)


wmathurin:
prepublish.js and postpublish.js are now identical to the SalesforceMobileSDK-Android's versions and will be moved to SalesforceMobileSDK-Shared in the next pull request

wmathurin:
We no longer need the SalesforceMobileSDK-iOS-Package and the SalesforceMobileSDK-iOS-Distribution repos

bhariharan:
We can remove the dependency on the ```Dependencies``` repo too (it only has ```libSqlcipher.a```), but that can be another pull request.

.npmignore
```diff
@@ -0,0 +1,16 @@
...
+external/fmdb
+external/ocmock
+external/ThirdPartyDependencies
+external/shared/samples/accounteditor/
+external/shared/samples/contactexplorer/
+external/shared/samples/fileexplorer/
+external/shared/samples/simplesync/
+external/shared/samples/smartstoreexplorer/
+external/shared/samples/userandgroupsearch/
+external/shared/samples/usersearch/
```
bhariharan: Should we add ```xcuserdata``` and ```xcshareddata``` to this too? Maybe even the ```workspace``` file/folder?

node/forceios.js
```diff
@@ -0,0 +1,314 @@
...
+        usage();
+        process.exit(4);
+    }
+
+    //
+    // FIXME
+    // 
+
+    /*
+    // Copy dependencies
```
bhariharan: Should this be removed?

bhariharan:
Looks good otherwise.

node/forceios.js
wmathurin: I think so - we don't have it on the android side.
We could fix it to do pod update or cordova update but maybe it's best they use pod / cordova directly

.npmignore
wmathurin: We could

PR #1131, opened 2015-08-04
Add tests to travis CI

kchitalia:
The build will fail once this PR is merged. To get the build to pass, we need to fix the failing tests. Not adding the tests for RestAPI as that freezes the build completely. Once we fix the tests, will add that back in.


.travis.yml
```diff
@@ -19,6 +20,11 @@ script:
...
-
-   #run all tests
+   # run all tests
    - xctool test -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceSDKCommon -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+   - xctool test -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceSecurity -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+   - xctool test -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceOAuthTest -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+   - xctool test -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceSDKCoreUnitTestApp -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+   - xctool test -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceNetworkiOS -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+   - xctool test -workspace SalesforceMobileSDK.xcworkspace -scheme SmartSync -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
+   - xctool test -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceHybridSDK -sdk iphonesimulator ONLY_ACTIVE_ARCH=NO
```
wmathurin: Two seems to be missing OCMockLibTests and SmartSyncTests

kchitalia: The SmartSync one is the second last line from bottom. Is that incorrect?

wmathurin: I stand corrected the two missing ones are SalesforceRestAPITests and OCMockLibTests

kchitalia: The RestAPITests freezes the build completely, so I have excluded that for now. I'll add the OCMockLibTests.

wmathurin:
I was looking at what the build failure looks like - it's not easy to tell what tests are failing (and the output is truncated). Is there a way to improve that - maybe run a grep on the raw output?

kchitalia:
Yes, definitely we can do some better storing of test results so that its easy to make out whats failing.

.travis.yml
bhariharan: We need to add ```HybridPluginTestApp```, ```SalesforceNetworkiOSTests``` and ```SalesforceRestAPIUnitTestApp``` to this list. I think ```HybridPluginTestApp``` is the one that runs the hybrid plugin tests.

kchitalia: Will add HybridPluginTestApp and SalesforceNetworkiOSTests. The SalesforceRestAPIUnitTestApp freezes the build completely. I was thinking we could add that in after the tests are fixed?

bhariharan: Okay.

kchitalia:
Pretty printing of logs coming in a separate pull request

PR #1134, opened 2015-08-04
Fixing MetadataManagerTests and refactoring them to use XCTestExpectation

bhariharan:


libs/SalesforceRestAPI/SalesforceRestAPITests/SalesforceRestAPITests.m
```diff
@@ -42,14 +42,8 @@
 
 @interface SalesforceRestAPITests ()
 {
```
bhariharan: Nothing new here. Just the merge from @wmathurin's pull request.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncMetadataManager.m
```diff
@@ -387,7 +387,7 @@ - (void)loadMRUObjects:(NSString *)objectTypeName limit:(NSInteger)limit cachePo
             }
             NSString *whereClause = nil;
             if (objectContainedLastViewedDate) {
-                [queryBuilder from:[NSString stringWithFormat:@"%@ using MRU", objectTypeName]];
+                [queryBuilder from:[NSString stringWithFormat:@"%@ using SCOPE MRU", objectTypeName]];
```
bhariharan: Post ```v32.0```, ```SCOPE``` is required too.

wmathurin:
Cool, I like expectations (aka future in java and promise in javascript ;-)

wmathurin:
We should be able to refactor SFSDKTestRequestListener to use expectations?

bhariharan:
Yes, at some point we should be able to. For now, I'm just cherry-picking the work done by S1, due to lack of time ;-)

wmathurin:
Hmm I should have used expectations instead of my semaphore in last checkin.

khawkins:
Expectations are great...and a relatively new find for me as well.  One thing I've noticed in my use of them to be careful with is that (as near as I can tell) you can only use one actively at a time.  I.e. you can't nest them.  But I'm in favor of replacing our old system of waits with them.

PR #1140, opened 2015-08-06
Fix for testUserDefinedEndpoint

wmathurin:


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
```diff
@@ -99,24 +99,26 @@ - (id)contentFromData:(NSData*)data fromResponse:(NSHTTPURLResponse*)response er
     NSError *responseError = nil;
     id content = [super contentFromData:data fromResponse:response error:&responseError];
 
-    if (content && !responseError) {
+    if (!responseError) {
```
wmathurin: In the case of a 404 (which is what the test does), we have a nil content and no responseError. If we don't enter the if, the code will get back a nil object and no error.

bhariharan:
:+1: 

That big merge is going to get a bit more messy now ;-)

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
wmathurin: I wonder some of the code in this method in CSFSalesforceAction belongs in the super class?

kchitalia:
:+1:

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFSalesforceAction.m
bhariharan: Some of it, like the ```errorMsg``` logic is specific to ```Salesforce API```, that's why I left it here.

PR #1141, opened 2015-08-06
Fix for splits/migration

bhariharan:
I don't know how/when this regressed, but ```instanceUrl``` can change during the refresh token flow.


bhariharan:
@khawkins You probably want this too, although I don't anticipate ```DF``` getting split.

bhariharan:
@qingqingliu @brichey You guys want this for ```Insights```.

khawkins:
This came out of S1.  You may want to ask them about why they implemented it.  They were having some issue with state getting overwritten when it shouldn't.

bhariharan:
@khawkins Frozen white screen on ```S1``` is what led me to this investigation ;-) I understand why they would do this for ```communityId``` and ```communityUrl```, because they have this ```accessToken``` sharing thingy within an org for multiple communities. However, that won't work for Mobile SDK apps, and is not the correct way to handle changes in communities. Either way, ```instanceUrl``` and ```idUrl``` should be outside the ```if``` statement.

wmathurin:
:+1:

khawkins:
:+1:

PR #1143, opened 2015-08-07
Removing submodule for ThirdPartyDependencies

bhariharan:


wmathurin:
We should exclude that folders in .npmignore so that it doesn't get bundled in the forceio*.tgz

PR #1145, opened 2015-08-14
React native support

wmathurin:
In
---
New library SalesforceReact
Bridges for smartstore and smartsync
New app type in createApp.sh / forceios to create react native app

Not in
--------
The bridges need testing (will be part of a separate pull request).
We need a bridge for Rest API (since we can't use forcetk as it is).
Once we have that, we need to modify the template app to pull data from the service (right now it shows canned data).


.gitmodules
```diff
@@ -4,12 +4,12 @@
...
-	url = https://github.com/forcedotcom/SalesforceMobileSDK-iOS-Dependencies.git
 [submodule "external/ocmock"]
 	path = external/ocmock
 	url = https://github.com/erikdoe/ocmock.git
 [submodule "external/fmdb"]
 	path = external/fmdb
 	url = https://github.com/forcedotcom/fmdb
+[submodule "external/react-native"]
+	path = external/react-native
+	url = https://github.com/wmathurin/react-native.git
```
kchitalia: Will we eventually point to a forcedotcom fork of react-native?

build/app_template_files/__ReactNativeTemplateAppName__/__ReactNativeTemplateAppName__/AppDelegate.m
```diff
@@ -0,0 +1,205 @@
+/*
+ Copyright (c) 2011-2014, salesforce.com, inc. All rights reserved.
```
kchitalia: Should this be 2015?

build/app_template_files/__ReactNativeTemplateAppName__/__ReactNativeTemplateAppName__/main.m
```diff
@@ -0,0 +1,33 @@
+/*
+ Copyright (c) 2014, salesforce.com, inc. All rights reserved.
```
kchitalia: 2015?

.gitmodules
wmathurin: Fixed with last checkin (also just sent the pull request for the react native fork)

wmathurin:
I think Travis build failed because i checked in the .gitmodule change before merging the react-native pull request

kchitalia:
You should be able to re-run the build

kchitalia:
After re running the build, Travis is happy :)

PR #1146, opened 2015-08-18
New react native module to allow network calls through our iOS network library

wmathurin:
Changed template react native app to leverage new module
Also a fix for iOS 9


shared/react/react.force.net.js
```diff
@@ -0,0 +1,243 @@
+'use strict';
```
kchitalia: Do we need copyright info here?

wmathurin: Done here https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/3cf7d921701e05443d5dbbc83d36e25c64aa9843

kchitalia:
:+1:

PR #1149, opened 2015-08-19
Misc fixes

wmathurin:
User agent changes for react native
Podfile pointing to git


libs/SalesforceReact/SalesforceReact/Classes/SFNetReactBridge.m
```diff
@@ -51,6 +55,7 @@ @implementation SFNetReactBridge
     
     SFRestRequest* request = [SFRestRequest requestWithMethod:method path:path queryParams:queryParams];
     [request setCustomHeaders:headerParams];
+    [request setHeaderValue:[SFRestAPI userAgentString:kReactNative] forHeaderName:kUserAgent];
```
wmathurin: User agent will say nativeReact only for calls through SFNetReactBridge.m
To have all calls from the app have the modified user agent (e.g. the smartsync ones), we need a different approach. Will probably revisit this in the next pull request.

kchitalia:
:+1:

PR #1151, opened 2015-08-20
New app type in SalesforceSDKManager for react native

wmathurin:
- user agent for react native will show the correct app type regardless of whether SFNetReactBridge is used or not
- smartsync calls from react native app should show: ReactNativeSmartSync


kchitalia:
:+1:

PR #1153, opened 2015-08-21
Better output for tests

wmathurin:
Unfortunately we have a lot of failures (a bunch created by the network sdk merge) but others also

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,24 +15,24 @@ before_script:
 
 script:
    #build all schemes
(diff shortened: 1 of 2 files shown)


wmathurin:
Also added a script to locally build/test everything (see https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/6a6fe39850cef2598d6c01ce90870093cce9c868).
For android we already had a similar script (tools/sdk.sh)

kchitalia:
:+1:

PR #1154, opened 2015-08-21
Script to build / test everything in the workspace (or specific schemes)

wmathurin:
Very similar to SalesforceMobileSDK-Android/tools/sdk.sh

Usage: ./build/sdk.sh [options]
    -b, --build scheme               Build given scheme (pass all to build all the schemes)
    -t, --test scheme                Test given scheme (pass all to test all the schemes)
    -m, --max-time timeout           Maximum time that build/test is allowed to run

- xcodebuild is run in a separate thread which is killed if it takes too long
- output for builds is a one liner: build succeeded or build failed, you will need to build from Xcode to get more details
- output for tests is the list of failed tests and nothing else


build/sdk.sh
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/ruby
```
kchitalia: Copyright info required?

kchitalia:
:+1:

PR #1163, opened 2015-08-27
Fix for crasher

bhariharan:
Adding an operation to an operation queue that is finished is always a crasher. This can occur in a quick enqueue and cancel, since enqueuing can be async. From the docs for addOperation:

> An operation object can be in at most one operation queue at a time and this method throws an NSInvalidArgumentException exception if the operation is already in another queue. Similarly, this method throws an NSInvalidArgumentException exception if the operation is currently executing or has already finished executing.


wmathurin:
:+1:

PR #1176, opened 2015-09-04
Enhanced sync explorer native

wmathurin:
Added menu button that brings up popover with following actions:
- logout
- switch user
- db inspector

NB: using 3rd party popover (WYPopoverController) since UIPopoverController only works on iPad


bhariharan:
:+1:

PR #1180, opened 2015-09-08
Rest API enhanced to allow manual testing of almost all api wrappers

wmathurin:
Added support for:
- search scope and order
- search result layout
- owned files list
- files in users groups
- files shared with user
- file details
- file shares
- batch details
- add file share
- delete file share

Only the ones that involves binary content are not supported:
- requestForFileRendition
- requestForFileContents
- requestForUploadFile

Still not working on iPhone:
- would need to use flexible layout
- would need to use an iphone friendly pop over controller


bhariharan:
:+1:

PR #1181, opened 2015-09-09
SalesforceReact was not compiling in xcode/command line

wmathurin:
Now it is, and it also works from generated app (cocoapod)

### DIFF
--- a/build/sdk.sh
+++ b/build/sdk.sh
@@ -34,7 +34,6 @@ SCHEMES_TO_BUILD = [
   "NoteSync",
   "OCMock",
   "OCMockLib",
-  "React",
   "RestAPIExplorer",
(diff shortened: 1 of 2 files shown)


kchitalia:
:+1:

build/sdk.sh
```diff
@@ -34,7 +34,6 @@ SCHEMES_TO_BUILD = [
   "NoteSync",
   "OCMock",
   "OCMockLib",
-  "React",
```
bhariharan: Don't we need ```SalesforceReact``` here?

wmathurin: It is there line 42.
The react scheme doesn't exist. (We don't compile the React project from the SDK workspace).

bhariharan:
:+1:

PR #1186, opened 2015-09-15
SimpleSync is now SmartSyncExplorerHybrid

wmathurin:
Also misc cleanup


.gitignore
```diff
@@ -1,5 +1,6 @@
 .DS_Store
 xcuserdata
+xcshareddata
```
bhariharan: Actually, this could have shared data common to all users of the workspace.

bhariharan:
Everything else looks good.

.gitignore
wmathurin: Ok will revert that part

PR #1190, opened 2015-09-23
Fixing test failure (iOS9 only) SFUserAccountManagerTests.m

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -755,10 +755,21 @@ - (BOOL)loadUserAccountFromFile:(NSString *)filePath encrypted:(BOOL)encrypted a
...
+                }
+                return YES;
+            } else {
+                if (error) {
+                    *error = [NSError errorWithDomain:SFUserAccountManagerErrorDomain
+                                                 code:SFUserAccountManagerCannotReadDecryptedArchive
+                                             userInfo:@{ NSLocalizedDescriptionKey : reason} ];
+                }
+                return NO;
             }
```
wmathurin: This code path is only exercised from the test apparently and also need the special iOS9 handling.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerTests.m
```diff
@@ -352,7 +352,15 @@ - (void)testPlaintextToEncryptedAccountUpgrade {
     XCTAssertTrue([fm fileExistsAtPath:userFilePath], @"User directory for User ID '%@' and Org ID '%@' should exist.", userId, orgId);
     
     // Verify that the user data is now encrypted on the filesystem.
-    XCTAssertThrows([NSKeyedUnarchiver unarchiveObjectWithFile:userFilePath], @"User account data for User ID '%@' and Org ID '%@' should now be encrypted.", userId, orgId);
+    @try {
```
wmathurin: Since unarchive returns nil instead of throwing an exception on iOS9, we need to tweak the test so that it passes both on iOS9 and on older versions

bhariharan:
:+1:

PR #1194, opened 2015-09-29
Fixes for iOS 7 bodyStream/Content-Length issues

camflan:
Even though v3.3 is working great ton iOS 8/9 devices, theres an issue on iOS 7. As soon as you use HTTPBodyStream, it will clear Content-Length header, even if you set it after the bodyStream. This is considered functioning as designed for iOS 7. 

These are the fixes I had to use to get our app working on iOS 7/8/9 properly.


bhariharan:
@camflan Thanks for the pull request. However, at this time, we are considering dropping support for ```iOS 7``` (since the ```%``` of users on ```iOS 7``` is less than ```10%```). Given that, it might be best to implement this fix on your fork if you need support for ```iOS 7```. #safeharbor

camflan:
@bhariharan Yep, already done and we are dropping iOS 7 with the release after the current one as well.

thanks

PR #1201, opened 2015-10-05
Updated SalesforceCommonUtils and SQLCipher to support bitcode

MikeNachbaur:
This also updates SalesforceCommonUtils to work when used within Cocoa
Touch Frameworks, including Swift-based projects.


bhariharan:
:+1: Thanks!

PR #1224, opened 2015-10-30
Removing pre-iOS 8 conditionals and fixing warnings

bhariharan:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
```diff
@@ -121,12 +116,8 @@ - (void)registerForRemoteNotifications
     
     // register with Apple for remote notifications
     [self log:SFLogLevelInfo msg:@"Registering with Apple for remote push notifications"];
-    
-    // [UIApplication registerUserNotificationSettings:] is how iOS 8 and above registers for notifications.
     if ([[UIApplication sharedApplication] respondsToSelector:@selector(registerUserNotificationSettings:)]) {
```
khawkins: I'd probably go a step further on this by taking out the conditional, and renaming `registerNotificationsForiOS8` to `registerNotifications`, since it should be the only method for registering now.

khawkins:
Otherwise, looks fine to me.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
bhariharan: Done.

PR #1233, opened 2015-11-09
SFSmartStoreDatabaseManager cleaning up on logout

wmathurin:
That was causing the react native app to not work properly after logout

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreDatabaseManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreDatabaseManager.m
@@ -28,6 +28,7 @@
 #import <SalesforceCommonUtils/NSString+SFAdditions.h>
 #import "SFSmartStoreUtils.h"
 #import "SFUserAccountManager.h"
+#import "SFAuthenticationManager.h"
 #import "SFUserAccount.h"
 #import "SFDirectoryManager.h"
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreDatabaseManager.m
```diff
@@ -53,6 +54,10 @@
 static NSInteger  const kSFSmartStoreVerifyReadDbErrorCode = 7;
 static NSString * const kSFSmartStoreVerifyReadDbErrorDesc = @"Could not read from database at path '%@', for verification: %@";
 
+@interface SFSmartStoreDatabaseManager () <SFAuthenticationManagerDelegate>
```
khawkins: Can we piggy-back this in SFSmartStoreDatabaseManager+Internal.h?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SmartStore/SFSmartStoreDatabaseManager.m
```diff
@@ -126,6 +134,12 @@ - (id)initGlobalManager
     return self;
 }
 
+- (void)dealloc {
+    if (self.user) {
```
khawkins: Shouldn't this be driven by the implementation of `authManager:willLogoutUser:` in `SalesforceSDKManager`?  That's the current chain of cleanup events on logout.

PR #1234, opened 2015-11-10
Sync Manager logout issue

wmathurin:
Revealed by smart sync react native sample app.

On logout: sync manager's removeSharedInstanceForUser was calling SFSmartStore's sharedStoreWithName in order to get the store path (which was the key in the map of sync managers).
As a result a SFSmartStore instance was recreated and would linger after the logout.

Now: sync manager's can figure out the key from the store name / user and doesn't actually need the store object.

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -90,11 +90,11 @@ static NSMutableDictionary *syncMgrList = nil;
     @synchronized ([SFSmartSyncSyncManager class]) {
         if (store == nil || store.storePath == nil) return nil;
         
-        NSString *storePath = store.storePath;
-        id syncMgr = [syncMgrList objectForKey:storePath];
+        NSString *key = [SFSmartSyncSyncManager keyForStore:store];
+        id syncMgr = [syncMgrList objectForKey:key];
         if (syncMgr == nil) {
             syncMgr = [[self alloc] initWithStore:store];
-            syncMgrList[storePath] = syncMgr;
+            syncMgrList[key] = syncMgr;
         }
         return syncMgr;
     }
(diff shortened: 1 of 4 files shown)


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -107,19 +107,32 @@ + (void)removeSharedInstance:(SFUserAccount*)user {
...
+        [syncMgrList removeObjectForKey:key];
     }
 }
 
++ (NSString*)keyForStore:(SFSmartStore*)store {
+    return [SFSmartSyncSyncManager keyForUser:store.user storeName:store.storeName];
+}
+
++ (NSString*)keyForUser:(SFUserAccount*)user storeName:(NSString*)storeName {
+    NSString* keyPrefix = user == nil ? SFKeyForUserAndScope(nil, SFUserAccountScopeGlobal) : SFKeyForUserAndScope(user, SFUserAccountScopeUser);
```
bhariharan: This should default to ```SFUserAccountScopeCommunity``` I believe. In the absence of community, it will use org + user. User alone is inadequate for uniqueness.

bhariharan: Similar to what we've done in ```MetadataManager``` and ```CacheManager```.

khawkins:
This approach seems fine to me.  Something @bhariharan and I had discussed was the idea that, because `SFSmartSyncManager` is really a consumer of `SFSmartStore`, it could also make sense to codify that dependency in the form of a notification.  So for instance when a store is about to be removed, it would send a notification along the lines of "storeWillBeRemoved", and its consumers could respond accordingly.

Right now, the fact that these two objects respond to user events independently feels artificial—`SFSmartSyncManager` only nominally cares about users.  It's mostly dependent on stores.

wmathurin:
@bhariharan mentioned that and it's probably something worth doing. For now, I went with a smaller fix. But we should consider removing any ties between sync manager and users in the future.

khawkins:
:+1:

PR #1241, opened 2015-11-16
forceios script updates for Cordova 3.9.2

bhariharan:


node/forceios.js
```diff
@@ -118,8 +118,8 @@ function createHybridApp(config) {
 
     shelljs.exec('cordova create "' + projectDir + '" ' + config.companyid + ' ' + config.appname);
     shelljs.pushd(projectDir);
-    shelljs.exec('cordova platform add ios');
-    shelljs.exec('cordova plugin add https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin');
+    shelljs.exec('cordova platform add ios@' + cordovaPlatformVersion);
+    shelljs.exec('cordova plugin add https://github.com/forcedotcom/SalesforceMobileSDK-CordovaPlugin#unstable');
```
kchitalia: Should we add a line to install the minimum supported CLI version by default?

bhariharan: Developers might have different versions of CLI installed. Installing CLI across different platforms is too painful to support. Besides, I wouldn't go that far to step on a developer's toes.

wmathurin:
:+1:

node/forceios.js
kchitalia: You are right. No point in forcing developers to use a particular CLI version.

PR #1243, opened 2015-11-17
Splitting out OAuth test flow and coordinator delegate

khawkins:
In anticipation of testing a session refresher, which will have its own `SFOAuthCoordinator` delegate implementation.


wmathurin:
:+1:

PR #1244, opened 2015-11-17
Instrumentations classes

wmathurin:
Can be used for instance to collect performance data from SDK classes without having to change them.

For instance, to get the execution time of all pg* methods calls on SFSmartStorePlugin simply do the following at app initialization time:
```objective-c
    [[SFInstrumentation
      instrumentationForClass:objc_getClass("SFSmartStorePlugin")]

     instrumentForTiming:^BOOL(SEL selector) {
         return [NSStringFromSelector(selector) hasPrefix:@"pg"];
     }
      afterBlock:^(NSInvocation *invocation, NSTimeInterval executionTime) {
         // publish execution time for invocation.target (class) invocation.selector (method)
        //  If you pass nil for the afterBlock, you end up with a log line saying something like: TIMING className.methodName: x.xxx ms
     }];
```


bhariharan:
Looks good to me. :+1: 

Will we be doing the equivalent on ```Android``` too?

wmathurin:
I guess we could.

PR #1247, opened 2015-11-17
Updating test to fix failure

khawkins:
There seems to be an issue in the legacy encryption/decryption process, for strings that are too short.  However, a) that's not ever going to be an issue with access and refresh tokens, and b) I don't want to muck with the legacy logic to make sure it can be fixed without breaking its previous functionality, because c) I doubt that functionality is hardly ever leveraged anymore (we're a long time past those legacy token encryption schemes).


kchitalia:
:+1:

bhariharan:
:+1:

PR #1255, opened 2015-11-25
Reverting the openStoreQueueWithName method to what it was before

wmathurin:
1) SQLITE_OPEN_FILEPROTECTION_* is not available in the sqlite3.h that ships with SQLCipher which will prevent apps using the SDK through cocoapod from building.
2) It looks like the flags are not used any way by sqlcipher


kchitalia:
:+1:

bhariharan:
:+1:

khawkins:
:+1:

PR #1256, opened 2015-11-30
Switching FMDB to submodule

bhariharan:


bhariharan:
I've done some initial cleanup of the project files and removed stale/non-existent dependencies, as a precursor to the dynamic frameworks work.

external/fmdb
```diff
@@ -1 +1 @@
-Subproject commit 243919a3529fa8de95052b2fc2994bfd931a618c
+Subproject commit 81d04e793839d3f3f7756004cd0f7b1d6db394ee
```
bhariharan: Had to expose some headers as public and tweak the public headers folder path in the project file - hence the changes in ```FMDB``` fork.

libs/SmartStore/Configuration/SmartStore-Common.xcconfig
```diff
@@ -11,8 +11,8 @@ GCC_PREFIX_HEADER = $(PROJECT_NAME)/$(PROJECT_NAME)-Prefix.pch
 
 PRODUCT_NAME = SmartStore
 
-OTHER_CFLAGS = -DSQLITE_HAS_CODEC -DFMDatabase=SF_FMDatabase -DFMStatement=SF_FMStatement -DFMDatabasePool=SF_FMDatabasePool -DFMDatabaseQueue=SF_FMDatabaseQueue -DFMResultSet=SF_FMResultSet -DFMDBBlockSQLiteCallBackFunction=SF_FMDBBlockSQLiteCallBackFunction $(inherited)
```
bhariharan: Removed special linker flags here, since we don't compile ```FMDB``` inline anymore.

libs/SmartStore/Configuration/SmartStore-Common.xcconfig
```diff
@@ -11,8 +11,8 @@ GCC_PREFIX_HEADER = $(PROJECT_NAME)/$(PROJECT_NAME)-Prefix.pch
...
-OTHER_LDFLAGS = -ObjC -lsqlcipher
```
bhariharan: ```-ObjC``` should come from the ```FMDB``` project with the addition of ```$(inherited)```.

libs/SmartStore/SmartStore.xcodeproj/project.pbxproj
```diff
@@ -760,17 +696,8 @@
 				GCC_PRECOMPILE_PREFIX_HEADER = YES;
 				GCC_PREFIX_HEADER = "SmartStore/SmartStore-Prefix.pch";
 				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
-				OTHER_CFLAGS = (
```
bhariharan: Removed these special flags here too.

libs/SmartStore/Configuration/SmartStore-Common.xcconfig
wmathurin: Same flags are in the pod spec

bhariharan: Removed them in [this](https://github.com/bhariharan/SalesforceMobileSDK-iOS/commit/9263978ade2ffb6990635713e4f97e3dd28f6844) commit.

bhariharan:
Built the ```Everything``` scheme and ran the ```SmartSyncExplorer``` app - looks good so far. Will fix the rest of the podspec as part of the switch to dynamic frameworks.

wmathurin:
:+1:

libs/SmartStore/Configuration/SmartStore-Common.xcconfig
khawkins: Does it?  I didn't think that projects "inherited" settings from other projects—just from their own configuration hierarchies.

bhariharan: I thought it did with dynamic frameworks, but I could be wrong. I'll add it back in for now, until I validate.

PR #1261, opened 2015-12-01
Misc fixes

wmathurin:


build/app_template_files/__ReactNativeTemplateAppName__/__ReactNativeTemplateAppName__/AppDelegate.m
```diff
@@ -24,7 +24,7 @@
 
 #import "AppDelegate.h"
 #import "InitialViewController.h"
-#import <React/Base/RCTRootView.h>
+#import <React/RCTRootView.h>
```
wmathurin: That's the path it has to be imported from with 0.16

SalesforceMobileSDK-iOS.podspec
```diff
@@ -18,7 +18,7 @@ Pod::Spec.new do |s|
 
   s.subspec 'SalesforceSDKCore' do |sdkcore|
 
-      sdkcore.dependency 'CocoaLumberjack'
+      sdkcore.dependency 'CocoaLumberjack', '~> 2.2.0'
```
wmathurin: We probably shouldn't not just run with the latest version

SalesforceMobileSDK-iOS.podspec
```diff
@@ -52,7 +52,8 @@ Pod::Spec.new do |s|
       smartstore.dependency 'SalesforceMobileSDK-iOS/SalesforceSDKCore'
       smartstore.dependency 'SQLCipher', '~> 3.1'
       smartstore.dependency 'SQLCipher/fts', '~> 3.1'
-      smartstore.source_files = 'libs/Smartstore/Smartstore/Classes/**/*.{h,m}', 'libs/Smartstore/Smartstore/Smartstore.h', 'external/fmdb/src/fmdb/*.{h,m}'
+      smartstore.dependency 'FMDB', '~> 2.5'
+      smartstore.source_files = 'libs/Smartstore/Smartstore/Classes/**/*.{h,m}', 'libs/Smartstore/Smartstore/Smartstore.h'
```
khawkins: The other change that seems like it was lost is the inclusion of `-DSQLITE_HAS_CODEC`.  That needs to be included somewhere.

khawkins: (Not sure at what point it was lost.)

khawkins:
Otherwise, this seems fine to me.

SalesforceMobileSDK-iOS.podspec
wmathurin: We lost it only three hours ago: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/9263978ade2ffb6990635713e4f97e3dd28f6844

I'll put it back

bhariharan:
:+1:

SalesforceMobileSDK-iOS.podspec
khawkins: Another thing that occurs to me: have we verified that this configuration will actually encrypt the database?  Given that we're depending on a standalone version of FMDB—i.e. a separate, independent compilation unit—it seems that FMDB should be the spec responsible for including `-DSQLITE_HAS_CODEC`, not SmartStore.  I have the feeling this is not going to do what we expect it to.

FMDB does have a subspec specifically for SQLCipher: `FMDB/SQLCipher`.  However, this too could be problematic in our case, since it has a dependency on the `SQLCipher` spec, whereas we want to use our own version, since the SQLCipher spec doesn't seem to be maintained.

I'm an inch away from suggesting that we need to re-incorporate FMDB directly back into our own code base.  The maintenance issues involved with using it as a standalone library, IMO, outweigh any benefits we have from separating what's otherwise such a small and integrated code base to our SmartStore offering.

I say we keep the fork to maintain alignment, but lose the standalone build of the project.  Thoughts?

bhariharan: It does sound like the pain of maintenance currently outweighs the benefits of keeping it standalone. If we get rid of it, we can remove the only static target dependency we have too. I'm okay with that approach.

khawkins: For what it's worth, I verified that encryption is in fact _not_ working for an app that uses the `SmartSync` pod spec in its current state.  I'm going to look at re-integrating FMDB into SmartStore, if there aren't objections.

bhariharan: We'll need to change the way Cordova consumes the binaries and the update
script too then.

On Tuesday, December 22, 2015, Kevin Hawkins <notifications@github.com>
wrote:

> In SalesforceMobileSDK-iOS.podspec
> <https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1261#discussion_r48317318>
> :
>
> > @@ -52,7 +52,8 @@ Pod::Spec.new do |s|
> >        smartstore.dependency 'SalesforceMobileSDK-iOS/SalesforceSDKCore'
> >        smartstore.dependency 'SQLCipher', '~> 3.1'
> >        smartstore.dependency 'SQLCipher/fts', '~> 3.1'
> > -      smartstore.source_files = 'libs/Smartstore/Smartstore/Classes/**/*.{h,m}', 'libs/Smartstore/Smartstore/Smartstore.h', 'external/fmdb/src/fmdb/*.{h,m}'
> > +      smartstore.dependency 'FMDB', '~> 2.5'
> > +      smartstore.source_files = 'libs/Smartstore/Smartstore/Classes/**/*.{h,m}', 'libs/Smartstore/Smartstore/Smartstore.h'
>
> For what it's worth, I verified that encryption is in fact *not* working
> for an app that uses the SmartSync pod spec in its current state. I'm
> going to look at re-integrating FMDB into SmartStore, if there aren't
> objections.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1261/files#r48317318>
> .
>


-- 
Sent from my iPhone

khawkins: Good times. :)  I guess the other alternative would be to write our own (sub)spec for FMDB, though that feels like it kind of defeats the purpose of having the agnostic dependency in the first place.  But it's another way to go.

bhariharan: Having our own subspec of FMDB might be easier to manage, given our current state, as opposed to re-integrating FMDB into SmartStore.

PR #1263, opened 2015-12-02
Switch to Cocoa touch framework targets

bhariharan:
First Tsunami of changes, so you guys can start reviewing.

What's done:

* Added run steps for podspec public header generation and umbrella header generation.
* Added header configs.
* Switched all targets to use the new dynamic targets.
* Cleaned up stale references.

What's remaining:

* A bunch of work (start by fixing compilation).
* SalesforceKit.
* podspec updates.


bhariharan:
And when it's all working, I'm hoping to remove the static targets and rename the dynamic ones to the original target name.

khawkins:
Seems reasonable to me, so far.

bhariharan:
@jeanbovet FYI

jeanbovet:
Looks good to me! Quick question: are you going to remove the static libraries or keep both static libraries and dynamic framework side by side?

khawkins:
I think the plan is to remove the static library targets, once all is said and done.  The SDK is moving to an iOS 8 minimum in 4.0, so there's no reason to hold onto legacy support?

bhariharan:
@jeanbovet I've removed the static targets and renamed the dynamic targets to the original static target names in the last few commits.

Items Remaining:
* Some more cleanup of dangling static references (almost there).
* Some more scheme cleanup (almost there).
* Linking of sample app targets and test targets (almost there).
* Run sample apps and tests in new structure. Basic sanity testing.
* ```SalesforceKit``` - to remove or to keep? Doesn't make much sense to keep it around anymore IMO since we have no static artifacts and the core has been combined in ```SalesforceSDKCore```. What say @wmathurin and @khawkins?
* ```podspec``` updates.

khawkins:
Because the core of what we provide has been consolidated into a single framework, I don't think the umbrella framework makes as much sense anymore.  The only other "external" frameworks we have are more satellite functionality—i.e. the organizational structure itself is now better suited to just taking what you need.  So I would think that we would want to dispense with the overhead of maintaining that structure, with the new layout.

jeanbovet:
@bhariharan @khawkins sounds great, love the direction it's going!

bhariharan:
What's done (in the last round of commits):

* Everything compiles and links fine now.
* Schemes are fixed - ```Everything``` scheme works too.
* Native and hybrid sample apps compile and link fine.
* Libraries and test targets compile and link fine.
* Removed as many dangling references as I could find.

What's left:

* ```SalesforceKit```.
* ```podspec``` updates.
* ```CordovaPlugin``` repo updates.
* Last round of cleanup of dangling references to old scheme names, static targets, etc.

@khawkins @wmathurin I'm thinking now's a good time to get this merged in. I'll send another pull request for ```SalesforceKit``` and ```podspec``` updates. Does that sound good?

wmathurin:
Do app runs and do tests pass?

bhariharan:
All apps are running fine (both native and hybrid). Haven't tried tests yet, but if they link they should be fine, since the changes are only in configs. Running tests is on my list.

bhariharan:
```SmartStore``` tests are running fine. The others should link up fine, but I'll run them all after the final round of cleanup.

PR #1264, opened 2015-12-03
React native app template more like the one for Android

wmathurin:
In 4.1 or later, we need to figure out how to generate for one platform and add support for the other platform afterwards - may be the react cli will provide that functionality


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -28,7 +28,6 @@
 #import "SFRootViewManager.h"
 #import "SFSDKWebUtils.h"
 #import "SFManagedPreferences.h"
-// FIXME-REORG #import "SFSmartStore.h"
```
wmathurin: That's unrelated.
There is still a FIXME-REORG left in this file. And we have a bug opened for it.

bhariharan:
:+1:

PR #1271, opened 2015-12-04
The same version of sqlcipher should be used regardless on how you consume the SDK

wmathurin:
Zetetic has not updated their pod spec since April 2014, so you get sqlcipher 3.1 when you use cocoapod and 3.3 otherwise.
Before 4.0, forceios generated apps didn't use cocoapod, now they do (unless you generate an hybrid application).

The same version of sqlcipher should be used regardless on how you consume the SDK (native vs hybrid, cocoapod or not).

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,7 +5,6 @@ external/cordova-ios
 external/fmdb
 external/ocmock
 external/react-native
-external/ThirdPartyDependencies
 external/shared/doc/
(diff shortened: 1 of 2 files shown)


khawkins:
Seems reasonable, in the absence of maintenance for the "official" SQLCipher podspec.

bhariharan:
:+1:

PR #1272, opened 2015-12-05
Fix project dependencies, target references

khawkins:
As with all Xcode project file updates, I'm not 100% sure why the previous iterations weren't working.  There are clearly some net-new configuration sections added to the project files where dependencies weren't showing up, so I can only assume that those must have been the missing components.  Anyway, the following should apply:

- All library and sample app targets should build from scratch.  I.e. all target dependencies are lined up.
- All warnings about missing target dependencies should be gone.
- Umbrella header files have been added to their respective projects, so warnings about umbrella headers should be gone too, except for SalesforceReact.  Looking at that next.

@bhariharan and @wmathurin : Could you try this out clean in your environments, and make sure everything builds and runs as expected for you?  That's the best validation test I know of for these crazy configuration changes. :)

### DIFF
--- a/hybrid/SampleApps/NoteSync/NoteSync.xcodeproj/project.pbxproj
+++ b/hybrid/SampleApps/NoteSync/NoteSync.xcodeproj/project.pbxproj
@@ -61,6 +61,7 @@
 		824A9DBF1724CAF900C9BD79 /* SystemConfiguration.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 824A9DBE1724CAF900C9BD79 /* SystemConfiguration.framework */; };
 		82638F0C19B8E1FD00FBB3C9 /* UIApplication+SalesforceHybridSDK.m in Sources */ = {isa = PBXBuildFile; fileRef = 82638F0B19B8E1FD00FBB3C9 /* UIApplication+SalesforceHybridSDK.m */; };
 		8272301C19D764160066A350 /* InitialViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 8272301B19D764160066A350 /* InitialViewController.m */; };
+		829DA2791C1263320040F5F1 /* SalesforceHybridSDK.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 829DA2721C1263010040F5F1 /* SalesforceHybridSDK.framework */; };
 		82AF5AA71A0DA9AD006C7A4B /* cordova_plugins.js in Copy Shared www Files */ = {isa = PBXBuildFile; fileRef = 82AF5AA51A0DA996006C7A4B /* cordova_plugins.js */; };
 		82AF5AE61A0DADA4006C7A4B /* CDVLogger.m in Sources */ = {isa = PBXBuildFile; fileRef = 82AF5AE01A0DADA4006C7A4B /* CDVLogger.m */; };
 		82AF5AEA1A0DAE72006C7A4B /* console-via-logger.js in Copy cordova-plugin-console plugin */ = {isa = PBXBuildFile; fileRef = 82AF5AE41A0DADA4006C7A4B /* console-via-logger.js */; };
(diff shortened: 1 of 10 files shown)


hybrid/SampleApps/NoteSync/NoteSync.xcodeproj/project.pbxproj
```diff
@@ -78,26 +78,33 @@
...
+		829DA2751C1263010040F5F1 /* PBXContainerItemProxy */ = {
 			isa = PBXContainerItemProxy;
-			containerPortal = 4FFEE6051BFE914200B7AA8A /* SalesforceHybridSDK.xcodeproj */;
+			containerPortal = 829DA26A1C1263010040F5F1 /* SalesforceHybridSDK.xcodeproj */;
 			proxyType = 2;
 			remoteGlobalIDString = 8222729D19C39F4500FC537E;
-			remoteInfo = HybridPluginTestAppTests;
+			remoteInfo = SalesforceHybridSDKTestAppTests;
+		};
+		829DA2771C1263260040F5F1 /* PBXContainerItemProxy */ = {
```
khawkins: This is a common net-new addition to the projects that weren't building before.  Not sure why it wouldn't have been there before, but Xcode added it, so I'm leaving it there.

hybrid/SampleApps/NoteSync/NoteSync.xcodeproj/project.pbxproj
```diff
@@ -645,9 +652,10 @@
 /* End PBXSourcesBuildPhase section */
 
 /* Begin PBXTargetDependency section */
-		CE4CE4D21C0E6838009F6029 /* PBXTargetDependency */ = {
+		829DA2781C1263260040F5F1 /* PBXTargetDependency */ = {
 			isa = PBXTargetDependency;
 			name = SalesforceHybridSDK;
+			targetProxy = 829DA2771C1263260040F5F1 /* PBXContainerItemProxy */;
```
khawkins: This is another common "missing" item before.  Probably links with the new configuration mentioned above?

wmathurin:
Getting a clone of your repo and trying builds etc.

bhariharan:
@khawkins I tried building ```AccountEditor``` from scratch (after wiping ```DerivedData```) and it complained:

```/Users/bhariharan/Desktop/Test/SalesforceMobileSDK-iOS/libs/SmartStore/SmartStore/SmartStore-Prefix.pch:9:13: 'SalesforceSDKCore/SFLogger.h' file not found```

I wonder if that's because of the ```AccountEditor``` scheme, or because of the project file.

bhariharan:
Never mind. False negative - my submodules weren't properly updated. Works now. :+1:

khawkins:
@bhariharan Ah good.  AccountEditor was one of the few projects I *didn't* change. :)

bhariharan:
Does the run step update the umbrella headers in their new location (under the project target)? I was looking at the ```podspec``` and wondering if we should just use the umbrella header there and remove the run step to update the other headers there.

wmathurin:
Seems to work as advertised:
- got a clean clone
- built one of the sample app
- switch target device and built again (worked - that wasn't working before)
- built everything
- built everything for testing
- ran some tests (smartstore)
- ran some apps (rest api explorer, sync explorer hybrid)

khawkins:
The umbrella headers haven't moved.  They just weren't a part of the projects before—they were only on the filesystem.  With regards to your podspec comment, do you just mean to include the umbrella header file in the podspec, minus the other public header files?

bhariharan:
Correct - just include the umbrella header so it's more readable and won't require the runstep that updates the podspec. I was planning to do something similar in Cordova's plugin.xml.

khawkins:
The issue with doing that is that CocoaPods segregates public from private headers, so the umbrella header file wouldn't be able to resolve the other header files.  The umbrella header file is just a convenience; all of the headers it works with are necessarily public as well.

bhariharan:
Never mind then :-(

Merge away! :+1:

PR #1273, opened 2015-12-05
SalesforceReact umbrella header, pod spec headers

khawkins:
Adding these to the project, as they weren't configured before.  Also cleaned up some of the file locations in the project, as they were a bit askew.


wmathurin:
:+1:

bhariharan:
:+1:

PR #1274, opened 2015-12-07
Subclass of SalesforceSDKManager for apps that use smartstore (takes …

wmathurin:
…care of removing all stores on logout)


libs/SmartStore/SmartStore/Classes/SalesforceSDKManagerWithSmartStore.m
```diff
@@ -0,0 +1,38 @@
...
+
+#import <SalesforceSDKCore/SFAuthenticationManager.h>
+#import "SFSmartStore.h"
+#import "SalesforceSDKManagerWithSmartStore.h"
+
+@implementation SalesforceSDKManagerWithSmartStore
+
+- (void)authManager:(SFAuthenticationManager *)manager willLogoutUser:(SFUserAccount *)user
+{
+    [SFSmartStore removeAllStoresForUser:user];
```
khawkins: Conceptually, should this call `[super authManager:manager willLogoutUser:user]` (if it exists), since the base class defines the delegate relationship?

khawkins:
Otherwise, this looks good to me. :+1:

libs/SmartStore/SmartStore/Classes/SalesforceSDKManagerWithSmartStore.m
wmathurin: It should but the fact that SalesforceSDKManager follows SFAuthenticationManagerDelegate is in SalesforceSDKManager+Internal.h.

wmathurin:
Calling super thanks to change https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/676e915e517d6e1eb37d1c785398ef791770e965

khawkins:
:+1:

bhariharan:
:+1:

PR #1278, opened 2015-12-08
Headers in SalesforceSDKCore should not import from <SalesforceSDKCore/...>

wmathurin:
Added default subspec so that when you "pod 'SalesforceMobileSDK-iOS'", you only get core


wmathurin:
With those changes, you can do use_framework in the podfile of an application but it will only consume SalesforceSDKCore, if you try to pull in more subspecs, it will not compile.

khawkins:
:+1:

PR #1279, opened 2015-12-08
use_frameworks! will not expose Public headers.  Use Private path.

khawkins:
`Private` is always there, regardless of using `use_frameworks!` or not.


wmathurin:
Awesome, I tried a forceios created app (with app type: native_swift) and pod file edited to go against a local clone of pod_private_headers and it worked great both with use_framework and without.

PR #1282, opened 2015-12-09
Reverting build.xml to build static libs and adding missing schemes

bhariharan:


bhariharan:
I added ```FMDB``` and ```CocoaLumberjack```, and removed ```SalesforceReact```, since we don't need it.

khawkins:
Seems okay to me....

build/build.xml
```diff
@@ -55,7 +81,7 @@
     <target name="buildSalesforceHybridSDK">
         <antcall target="buildLib">
             <param name="projectName" value="SalesforceHybridSDK" />
-            <param name="libName" value="SalesforceHybridSDK.framework" />
+            <param name="libName" value="libSalesforceHybridSDK.a" />
             <param name="headerDir" value="Headers/SalesforceHybridSDK" />
             <param name="schemeName" value="SalesforceHybridSDK" />
```
wmathurin: The scheme name is probably going to be SalesforceHybridSDK-Static (or whatever we decide to call it)

bhariharan: Yep - that's for the next round, when I bring in the static targets.

PR #1283, opened 2015-12-10
Adding static targets back

bhariharan:
Here's ```SalesforceSDKCore``` for review. The others are coming soon.

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK.xcodeproj/xcshareddata/xcschemes/SalesforceHybridSDKStatic.xcscheme
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK.xcodeproj/xcshareddata/xcschemes/SalesforceHybridSDKStatic.xcscheme
@@ -15,7 +15,7 @@
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "CEA884421C191795008D871B"
-               BuildableName = "libSalesforceHybridSDKStatic.a"
+               BuildableName = "libSalesforceHybridSDK.a"
                BlueprintName = "SalesforceHybridSDKStatic"
                ReferencedContainer = "container:SalesforceHybridSDK.xcodeproj">
             </BuildableReference>
@@ -46,7 +46,7 @@
          <BuildableReference
             BuildableIdentifier = "primary"
             BlueprintIdentifier = "CEA884421C191795008D871B"
-            BuildableName = "libSalesforceHybridSDKStatic.a"
+            BuildableName = "libSalesforceHybridSDK.a"
             BlueprintName = "SalesforceHybridSDKStatic"
             ReferencedContainer = "container:SalesforceHybridSDK.xcodeproj">
          </BuildableReference>
@@ -64,7 +64,7 @@
          <BuildableReference
             BuildableIdentifier = "primary"
             BlueprintIdentifier = "CEA884421C191795008D871B"
-            BuildableName = "libSalesforceHybridSDKStatic.a"
+            BuildableName = "libSalesforceHybridSDK.a"
             BlueprintName = "SalesforceHybridSDKStatic"
             ReferencedContainer = "container:SalesforceHybridSDK.xcodeproj">
          </BuildableReference>
@@ -75,6 +75,7 @@
    </AnalyzeAction>
    <ArchiveAction
       buildConfiguration = "Release"
+      customArchiveName = "SalesforceHybridSDK"
       revealArchiveInOrganizer = "YES">
    </ArchiveAction>
 </Scheme>
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK.xcodeproj/xcshareddata/xcschemes/SalesforceHybridSDKStatic.xcscheme
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK.xcodeproj/xcshareddata/xcschemes/SalesforceHybridSDKStatic.xcscheme
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "0710"
+   version = "1.3">
+   <BuildAction
+      parallelizeBuildables = "YES"
+      buildImplicitDependencies = "NO">
+      <BuildActionEntries>
+         <BuildActionEntry
+            buildForTesting = "YES"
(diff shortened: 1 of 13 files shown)


wmathurin:
where is the new scheme for SalesforceSDKCoreStatic?

bhariharan:
I haven't added schemes yet - I will once I create all the static targets.

bhariharan:
I have added all static targets. [Here](https://github.com/bhariharan/SalesforceMobileSDK-iOS/commit/b807104c32796948d6d840714ab9cc98241b2750) is the commit that adds the static schemes. I built them individually after wiping ```DerivedData``` each time, and they all seem to build and generate products as expected.

One interesting find from this exercise - since ```lib*.dylib``` has been replaced with ```lib*.tbd``` in ```iOS 9```, I tried to use them, but they don't work with static targets. :-( The workaround is to add linker flags for each of those libs to the static targets and remove the ```lib*.tbd``` from the linker phase. For example, ```-lz``` for ```libz.dylib``` and ```-lxml2``` for ```libxml2.dylib```. There's something new to learn about ```XCode``` and ```iOS``` everyday - good or bad.

bhariharan:
Updated scheme names used in ```build.xml``` [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/c3cadc0132a242c590845bed0aefb06c9bc82df7). This should be ready to merge now.

wmathurin:
:+1:

khawkins:
Static library targets should not link other libraries.  With static libs, the linking should be done at the app target level.

libs/SalesforceHybridSDK/SalesforceHybridSDK.xcodeproj/xcshareddata/xcschemes/SalesforceHybridSDKStatic.xcscheme
```diff
@@ -0,0 +1,80 @@
...
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "CEA884421C191795008D871B"
+               BuildableName = "libSalesforceHybridSDKStatic.a"
```
khawkins: All of the schemes are reverting these back to the product names for me.  E.g. `libSalesforceHybridSDK.a`.

libs/SalesforceRestAPI/SalesforceRestAPI.xcodeproj/project.pbxproj
```diff
@@ -258,6 +292,16 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 		};
+		CEA8838F1C19135D008D871B /* Frameworks */ = {
+			isa = PBXFrameworksBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+				CEA883A61C191461008D871B /* libSalesforceSDKCore.a in Frameworks */,
```
khawkins: So e.g. this and the `libSalesforceNetwork.a` link below should go away.  The apps will be responsible for linking all of the libraries, to avoid hard dependencies between any two libraries at compile time.

libs/SalesforceHybridSDK/SalesforceHybridSDK.xcodeproj/xcshareddata/xcschemes/SalesforceHybridSDKStatic.xcscheme
bhariharan: The binary should be product name, shouldn't it? There's no reason for the buildable to be based on the scheme name. I'll update that here.

khawkins: Exactly.  I think that's why Xcode changed them for me.  If you play around in the Schemes editor a little bit, it'll probably change them for you automatically. :)

libs/SalesforceRestAPI/SalesforceRestAPI.xcodeproj/project.pbxproj
bhariharan: So we need to remove our own library dependencies in the ```Link Binary with Libraries``` step for the static targets, but ```Target Dependencies``` remain the same, correct? I'll make those changes.

khawkins: That's correct.  Leave the target dependencies (so the dependent targets will build first, deploy their headers, etc.), but no "linking" phase for the library targets.

bhariharan:
@khawkins Fixed the buildable names [here](https://github.com/bhariharan/SalesforceMobileSDK-iOS/commit/fd7f1acc8482d2b6f0787308badb07f65a648a89). I hand-edited the schemes, since I couldn't find a way to update it through the IDE. For some reason, ```SmartStore``` already had the correct name, but the others didn't. Weird!

bhariharan:
Removed the library linking [here](https://github.com/bhariharan/SalesforceMobileSDK-iOS/commit/f2d105cc44272244b3a277cb8645904a53c11894).

wmathurin:
Are you able to Xcode build the static libs from the command line?

bhariharan:
Command line is not fully working yet, because we still depend on a dynamic ```CocoaLumberjack``` target, which won't work due to code signing issues. Fixing that is coming in the next pull request.

khawkins:
The SmartStoreStatic -> FMDB-iOS -> SalesforceSDKCore dependency chain is a bit dodgy.  We should probably say that SmartStoreStatic has two dependencies: FMDB-IOS and SalesforceSDKCoreStatic.  I think when the framework header output is fixed up, this target probably wouldn't build.

khawkins:
...although I'm not exactly sure what that does at runtime, even so.  I.e. if FMDB-IOS links SalesforceSDKCore.framework, does that mean that the framework has to be present at runtime?  I think it might.  We may have to revisit the FMDB targets a bit.

bhariharan:
Yep, already on it. The FMDB static target needs to depend on the SDKCore static target for binary generation.

khawkins:
Otherwise, this is looking okay to me.

PR #1284, opened 2015-12-10
Cleanup to (almost) pass lint

khawkins:
In order for SalesforceSDKCore to pass `pod spec lint`, the structure of the code files needed to be revisited, reorganized into hierarchical subspecs, basically because the support mechanism for specifying no-ARC on a per-file basis in the pod spec is fairly terrible.  There are three subspecs now:

1. `base` : Pretty much everything that `SFKeychainItemWrapper` depends on.  Specified as "ARC".
2. `no-arc` : `SFKeychainItemWrapper` files.  Depends on `base`.  Specified as "no ARC".
3. `arc` : Everything else.  Depends on `no-arc` and `base`.  Specified as "ARC"

Then some other more minor failures were cleaned up.  A couple of outstanding notes:

- My understanding is that `SalesforceReact`'s dependency on `React` has to be managed in a more user-driven way, with npm and specifying local pod spec locations.  Without including a dependency for `React`, `SalesforceReact` is always going to fail `pod spec lint`, because its dependent files don't exist.

- The script that we run on the Xcode projects to update the public header files is no longer going to work for SalesforceSDKCore, because of the subspec breakout.  It won't break, but it won't make updates.  Making the script subspec-aware is TBD.  We'll just have to keep them manually aligned in the meantime.


wmathurin:
Looks reasonable.

bhariharan:
:+1:

PR #1290, opened 2015-12-11
Only keeping static targets (for now?) in sdk.sh

wmathurin:


bhariharan:
Oh crap - I forgot about this file. Where's this used? Do we need this, now that we have an ```Everything``` scheme?

wmathurin:
It's meant to be like the Android tools/sdk.sh to build/test everything from the command line. I use it locally to make sure everything is fine.
The travis config also needs updating.

bhariharan:
All the command line scripts will have to use static targets for now I guess.

:+1:

wmathurin:
True. But since tests targets depends are not available, travis will be of very little use (not that it is now).
Anyway something we can look into improving post 4.0.

PR #1293, opened 2015-12-11
Deploying SalesforceSDKResources.bundle to SalesforceSDKCore.framework

khawkins:
With frameworks, the resources bundle rightly lives under the umbrella of the framework bundle itself, rather than a separate deployment.  Apps that still use static libraries should continue to deploy the bundle standalone.  Updates to `libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m` will find the bundle in either case.

Project file updates were to add the bundle deployment to SalesforceSDKCore.framework, and remove the deployment from the sample apps.  Incidental updates to the project files were due to the re-addition of the static targets, and otherwise have no impact.


wmathurin:
Looks fine.

bhariharan:
:+1:

PR #1295, opened 2015-12-14
Using 'use_frameworks' in swift template app

wmathurin:
Separate pod specs per SDK lib (using our own specs repo)
Getting headers from frameworks using import instead of through a bridging headers

If we decide to go forward with this, we should:
- have the specs repo be a submodule of the ios repo and have the build scripts update the individual specs 
- we need to decide whether we want to keep the combined spec or not


build/app_template_files/__NativeSwiftTemplateAppName__/Podfile
```diff
@@ -2,10 +2,21 @@
 # platform :ios, '8.0'
 
 target '__NativeSwiftTemplateAppName__' do
+source 'https://github.com/CocoaPods/Specs.git'
+source 'https://github.com/wmathurin/SalesforceMobileSDK-iOS-Specs.git'
```
wmathurin: The repo for our specs (for now on wmathurin, but would move to forcedotcom if we go forward)

build/app_template_files/__NativeSwiftTemplateAppName__/Podfile
```diff
@@ -2,10 +2,21 @@
...
 
-pod 'SalesforceMobileSDK-iOS', :subspecs => [
-    'SmartSync'
-]
+use_frameworks!
+pod 'SalesforceSDKCore'
+pod 'SalesforceNetwork'
+pod 'SalesforceRestAPI'
+pod 'SmartStore'
+pod 'SmartSync'
```
wmathurin: Separate pods (and therefore frameworks) per sdk lib

build/app_template_files/__NativeSwiftTemplateAppName__/__NativeSwiftTemplateAppName__/AppDelegate.swift
```diff
@@ -24,6 +24,7 @@ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
 
 import Foundation
 import UIKit
+import SalesforceSDKCore
```
wmathurin: Instead of bridging header

build/app_template_files/__NativeSwiftTemplateAppName__/__NativeSwiftTemplateAppName__/RootViewController.swift
```diff
@@ -23,6 +23,7 @@ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH
 */
 
 import Foundation
+import SalesforceRestAPI
```
wmathurin: Instead of bridging header

bhariharan:
Looks fine. Should we do this as the first item in ```4.1``` instead of now?

wmathurin:
Agreed, we should not hold up the release for that. At least we have a
solution now. It could wait for 4.1 (or we could do a 4.0.1).

build/app_template_files/__NativeSwiftTemplateAppName__/Podfile
khawkins: What does this mean for dependencies (e.g. CocoaLumberjack) that don't live in our specs repo?  Or do we envision adding local versions of those?  Or do they "just work" somehow?

wmathurin: That's why you need the CocoaPods specs repo as well (first source).
When looking for a pod, it tries the sources in order (kind of like a classpath).

khawkins: Ah, got it.  Cool. :+1:

wmathurin:
Closing here - will resubmit against master

PR #1312, opened 2016-01-01
Getting rid of the problematic SqliteAdditions

khawkins:
Not used anywhere in the SDK, and basically making non-modular framework header problems that outweigh whatever benefit these additions may bring.  `use_frameworks!` in a Podfile basically requires you to also add the workaround of including non-modular headers in your project, with this around.


bhariharan:
:+1: 

Do we still need ```CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES``` to be set to ```YES``` for each project, or can that go away too?

khawkins:
I'm moving for that to go away.  I *think* I've squashed them all (with this), but there's still some FMDB work to do.  But yes, that's the goal.

PR #1320, opened 2016-01-06
We need the split up specs at the root of the repo to be able to do:

wmathurin:
pod 'SalesforceSDKCore', :path => 'some clone of the repo'
or
pod 'SalesforceSDKCore', :git => 'some git version of the repo'


wmathurin:
Will resubmit for unstable

PR #1321, opened 2016-01-06
Support for endPoint param in react net bridge

wmathurin:


wmathurin:
will resubmit for unstable

PR #1322, opened 2016-01-06
We need the split up specs at the root of the repo to be able to do:

wmathurin:
pod 'SalesforceSDKCore', :path => 'some clone of the repo'
or
pod 'SalesforceSDKCore', :git => 'some git version of the repo'

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "external/CocoaLumberjack"]
 	path = external/CocoaLumberjack
 url=https://github.com/forcedotcom/CocoaLumberjack.git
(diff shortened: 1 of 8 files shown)


kchitalia:
:+1:

bhariharan:
Shouldn't this go against ```master``` for ```4.0.1```?

wmathurin:
We could cherry pick that in master.

khawkins:
:+1:

wmathurin:
It's not that useful in master - since master is (soon to be) released code.
It's very useful in unstable, to be able to try sample code against a local clone or some random branch etc.

PR #1326, opened 2016-01-08
Not compiling in generated app (anymore?)

wmathurin:


kchitalia:
:+1:

khawkins:
Not following.  This doesn't build in a certain environment?  It builds in the workspace.

wmathurin:
If you use forceios and generate a react native app, you end up with compilation errors.
The key one is:
/Users/wmathurin/Development/github/forcedotcom/SalesforceMobileSDK-iOS/libs/SmartStore/SmartStore/Classes/SFSmartStore.m:827:34: error: expected a type
- (NSString *)tableNameBySoupId:(sqlite_int64)soupId {

As a result you get a few like (since sqlite_int64 is unknown and treated as an id):
/Users/wmathurin/Development/github/forcedotcom/SalesforceMobileSDK-iOS/libs/SmartStore/SmartStore/Classes/SFSmartStore.m:814:53: error: implicit conversion of 'long' to 'id' is disallowed with ARC
            soupTableName = [self tableNameBySoupId:soupId];

The reason for that is that sqlite3_int64 should be used instead of sqlite_int64. The latter is deprecated won't be defined unless you pass in an extra flag to the compiler. And so you don't get it with the stock sqlcipher.

The proper fix is in https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/bb8df239b29ad429d9227ff539afa4bf3d2d29c1

khawkins:
Ah okay. :+1:

PR #1327, opened 2016-01-08
Pod changes

wmathurin:
- Not combined specs anymore
- Build updates split up specs
- Pod file have our specs repo as first source
- Native template use split up specs
- SQLCipher static library no longer bundled in forceios
- Pointing to latest ios-specs repo


shared/common/build/tools/update_podspec_headers.sh
```diff
@@ -30,6 +30,8 @@
 # set -x
 set -e
 
+VERSION=4.0.1 # XXX: update that version for every release
```
wmathurin: Is that defined somewhere?

khawkins: Not as such, that I can think of.  `SalesforceSDKConstants.h` is the closest match, but won't help in this context.

khawkins: We could start maintaining a "version number" file at the root of the repo, which could help in this scenario.

wmathurin:
Should the specs live in the main repo and be copied over to the specs repo at release time?
Otherwise, when new headers are created, you'll have changes in the submodule. But you won't be able to check-in from there because it goes to forcedotcom etc. We don't really care to update the specs repo until release day since you can have your pod-based application point to a local clone or to git directly for the unstable code. On release day, we could create the new folders in the spec repo and copy over the tagged version of the specs from the main repo.
Thoughts?

khawkins:
I'd be okay with that.  I think I'd actually prefer it, because it will be easier to maintain clean versioning of the specs.  With the setup right now, since spec updates have to live in completely new folders in the spec repo, the specs effectively get decoupled from their traditional SCM versioning, which makes it harder to go back and look at changes over time.

Since the specs repo is basically another example of git being used for something that's not really source control, I'd be happy to see us truly source control the specs in our main repo, and just use the specs repo as a copying ground (like the Cordova plugin).

Or in other words, :+1:  :)

khawkins:
Here's some other food for thought: should we just host all of the pod spec stuff in the main repo (in its own subfolders)?  That seems like it would alleviate the performance issue coming from multiple clones of the main repo at first install.

wmathurin:
Cool, done here: https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/b8e1469e076d5bb7f71b1b477c1e94f03850aa35

We need to update the release steps for ios and/or add a update.sh script in the specs repo (like we have in the plugin repo).

wmathurin:
Because the specs live in the main repo, we don't need to know the version anymore https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/156be629f01ae184ca46e4035e550f30e6de1cdc

khawkins:
@wmathurin Any thoughts on my previous comment?  Do we need to continue to maintain a separate "specs" repo at all?

wmathurin:
Some reasons I can think of not to do that:
1) pod 'libX', :path => local clone and pod 'libX', :git => main repo at arbitrary version would not work - they expect the pod spec at the root for libX (the source xxx are not used in that case).
2) Getting the source repo would be one more big download ;-)
3) The top level of the repo will be a bit messy:
lib1
lib2
lib3
libs
etc

khawkins:
Ah.  I didn't know if there was a way to basically subpath the spec stuff (to avoid the mess), and then refer to the subpath in the `source` commands.  If not, then I agree, it probably makes more mess than we want.

wmathurin:
I don't think there is a way to do that.
I'll merge this PR for now, we can always improve later if we find a way.

PR #1340, opened 2016-01-15
Everything scheme builds again

bhariharan:
For some reason, import of project header in a test file without first importing public headers referenced in the project header caused compilation errors
Now in tests, we import the framework master header before any project header

Had to enable import of non-modular headers in SalesforceHybridSDK because it was choking on the #import <Cordova/xxx.h>


bhariharan:
@wmathurin I cherrypicked your pull request on ```master``` to fix tests. Everything else can go on ```unstable``` going forward.

PR #1344, opened 2016-01-16
Tweaking framework target configs

khawkins:
Not much to it.  Just cleaning up some of the search paths where not necessary in framework targets.


wmathurin:
Looks reasonable to me.

bhariharan:
:+1:

PR #1346, opened 2016-01-18
FMDB submodule pointer, another random warning in SalesforceSDKCore

khawkins:
A library's plist shouldn't get added as a resource to the final product.  It's a build-time entity.


bhariharan:
:+1:

PR #1347, opened 2016-01-18
Flattening dependencies for samples, to embed frameworks

khawkins:
To summarize the awful project file updates:

- App targets need to embed the (non-system) frameworks they depend on, to form a proper app bundle, i.e. one that can be installed and run on a device.
- In order to embed a framework in an app bundle, it must be directly present—either as an Xcode project target or compiled framework library—in the app's project.

This necessitated flattening the Xcode project references for the sample app targets, which is the reason for all the flux in the project files.

Hybrid apps to follow sometime after this PR.


wmathurin:
Looks fine.

bhariharan:
:+1:

PR #1348, opened 2016-01-19
Embedding frameworks for hybrid sample apps

khawkins:
Here's the hybrid version of embedding frameworks in sample apps.


bhariharan:
:+1:

PR #1352, opened 2016-01-20
Cleaning up some test target configs

khawkins:
I'm sure it's not comprehensive, but it's a start!  Removing the framework search paths from the test targets—they don't need them.  Adding the runpath search paths—they do need them.  Standard config on framework test targets.


wmathurin:
Great :+1:

kchitalia:
:+1:

PR #1356, opened 2016-01-21
Reintroduce SalesforceKit into the Mobile SDK unstable branch

MikeNachbaur:


libs/SalesforceKit/SalesforceKit.modulemap
```diff
@@ -0,0 +1,54 @@
...
+
+    framework module SmartSync {
+        umbrella header "SmartSync.h"
+        
+        export *
+        module * { export * }
+        
+        use SalesforceRestAPI
+    }
+
```
bhariharan: We should probably add ```SalesforceHybridSDK``` and ```SalesforceReact``` frameworks here too for completeness.

MikeNachbaur: I didn't think the HybridSDK was usable outside of something like PhoneGap/Cordova.  In that case, wouldn't the templates be used directly instead of explicitly adding something like SalesforceKit?

SalesforceReact seemed similar to me in that regard.

libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common-Static.xcconfig
```diff
@@ -11,4 +11,4 @@
 INSTALL_PATH = /
 SKIP_INSTALL = YES
 
-HEADER_SEARCH_PATHS = $(BUILT_PRODUCTS_DIR)/Headers $(TARGET_BUILD_DIR)/Headers $(inherited)
+// HEADER_SEARCH_PATHS = $(BUILT_PRODUCTS_DIR)/Headers $(TARGET_BUILD_DIR)/Headers $(inherited)
```
bhariharan: Why is this commented out?

MikeNachbaur: I thought I'd removed it entirely, I'll add a commit for that.

Basically, the problem is that by adding BUILT_PRODUCTS_DIR/Headers directly to the import path, Xcode finds multiple headers defining the same symbol when building dependencies.  So the moment it encounters an #import of the umbrella header, it finds multiple alternatives and fails to compile.

Everything still compiles successfully without this line added.

libs/SalesforceKit/SalesforceKit.modulemap
bhariharan: Makes sense.

libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common-Static.xcconfig
bhariharan: I see. @khawkins has been doing some work on cleaning up our targets and fixing everything in the workspace, so I'll defer to him on this.

libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common-Static.xcconfig
```diff
@@ -10,5 +10,3 @@
 
 INSTALL_PATH = /
 SKIP_INSTALL = YES
-
-HEADER_SEARCH_PATHS = $(BUILT_PRODUCTS_DIR)/Headers $(TARGET_BUILD_DIR)/Headers $(inherited)
```
bhariharan: @khawkins has been doing some work on cleaning up our targets and fixing everything in the workspace, so I'll defer to him on this.

MikeNachbaur:
The only issue I'm having difficulty with, which is what's causing TravisCI to fail, is a strange "newline" error in sed when running the podspec script.  I was hoping someone on the Mobile SDK team could help point me in the right direction on that one.

libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common-Static.xcconfig
khawkins: I'll look into this one.

khawkins:
@MikeNachbaur The Travis CI stuff has been in a diminished state for a while, on a number of fronts.  If that doesn't get resolved as part of this request, I don't think it's the end of the world.

libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common-Static.xcconfig
khawkins: Looks like SalesforceSDKCoreStatic is otherwise picking up its necessary `HEADER_SEARCH_PATHS` from Packaging-Static.xcconfig, so this one's fine to ditch.

libs/SalesforceKit/SalesforceKit.modulemap
```diff
@@ -0,0 +1,54 @@
...
+    link "z"
+    link "sqlite3"
```
khawkins: I don't know if this will have an impact on the encryption/SQLCipher support for consumers of SalesforceKit.  But probably not, since SQLCipher is statically linked into FMDB in our case (and soon to be into SmartStore).

khawkins:
This all seems fine to me. :+1:

libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common-Static.xcconfig
bhariharan: :+1:

bhariharan:
Looks good to me too :+1:

TysonBunch:
Thanks @MikeNachbaur @khawkins @bhariharan !!!

PR #1359, opened 2016-01-21
We could have a non-logged in user (so currentUser would not be null …

wmathurin:
…but there would be no accessToken)


bhariharan:
:+1:

kchitalia:
:+1:

PR #1360, opened 2016-01-22
Fixing an issue with the recent moving of some header files

khawkins:
RestAPIExplorer was relying on an indirect import.


bhariharan:
:+1:

PR #1361, opened 2016-01-23
Going back to an inline FMDB model

khawkins:
SmartStore now points directly to the code files in the FMDB submodule, sets the proper SQLCipher build and linker flags, etc.

Left to do: I still need to validate that the pod specs work, both in independent apps and forceios-created apps.  Otherwise, tests seem to run fine, sample apps, etc.


bhariharan:
You'll also need to remove all references to ```fmdb``` targets and ```fmdb-iOS.a``` from the ```build/build.xml``` script.

khawkins:
Ah, thanks.  I overlooked those folders.  I'll update that now.

khawkins:
@bhariharan I made those changes, starting with 58f5b6c1091a5c6a658ed0e0a8517fa5f8a38a2b.

khawkins:
66c84ca15b7418027ec31c00ff94b49a0f8428f2 fixes up CocoaPods—I forgot to add the FMDB source to SmartStore's pod spec.  Also fixed a random unrelated warning in SFLogger.

bhariharan:
Looks good to me. :+1:

khawkins:
@bhariharan @wmathurin @kchitalia I thought I would mention that I did look at the CocoaPods specs repo to see if anything needed to be done as the result of this work, but as near as I can tell, we don't need to do anything there.  The update script doesn't explicitly look for named specs, and the existing FMDB spec (4.0.2) would stay around for posterity anyway.

wmathurin:
Yeah, it's a nice update.sh script ;-)

PR #1362, opened 2016-01-25
Non-leaf nodes support

wmathurin:
Latest shared - running new hybrid tests (queries against non-leaf nodes)
To make them work, had to make sure non-leaf nodes are json-ized in their index columns


bhariharan:
:+1:

khawkins:
Seems fine to me.

PR #1363, opened 2016-01-25
Should handle nil response without crashing

wmathurin:
Response can be nil (e.g. in the case of a delete)

Fix for https://github.com/forcedotcom/SmartSyncExplorerReactNative/issues/29


khawkins:
:+1:

PR #1364, opened 2016-01-25
Removing questionable cleansing methods from SalesforceSDKCore

khawkins:
The `cleansedArray` method does nothing.  The `cleansedDictionary` method isn't comprehensive in its data structure considerations, which seems like it would do more harm than good.  Neither method is used in the SDK.


wmathurin:
How does on get a nil in a NSArray in the first place?

khawkins:
It doesn't, except by an undefined "accident" of released object clean-up.  The other *n*% of the time, such an entity would point to garbage.  I.e. there's no "good state" at runtime where you would look for such a condition.

wmathurin:
:+1:

PR #1366, opened 2016-01-27
Fix for bug: forceios create with native apps does not use custom Con…

wmathurin:
…nected App Callback URIs


bhariharan:
:+1:

wmathurin:
I don't think it ever worked...

khawkins:
My favorite kind of bug. ;-)

PR #1368, opened 2016-01-28
Server Picker and Login host

kchitalia:
@bhariharan @khawkins @wmathurin Here are the changes for the server picker and login host story. I know its a lot of files, but most of it is just pulling in stuff from FoundationSDK.


wmathurin:
I see new public headers. Where are the changes for the podspec?

wmathurin:
Also are we getting rid of the settings bundle?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
```diff
@@ -50,4 +50,12 @@ + (NSString *)localizedString:(NSString *)localizationKey
     return NSLocalizedStringFromTableInBundle(localizationKey, @"Localizable", sdkBundle, nil);
 }
 
++ (UIImage *)imageNamed:(NSString *)name
+{
+    NSBundle *bundle = [NSBundle bundleWithIdentifier:@"com.salesforce.mobilesdk.SalesforceSDKCore"];
```
khawkins: We should stick with the paradigm used in `mainSdkBundle`: `[NSBundle bundleForClass:[self class]]`.  This will allow us to have the assets in the app bundle for apps that use static libraries, such as Cordova apps, and in the framework bundle for apps that utilize the framework.

libs/SalesforceSDKCore/Images.xcassets/Contents.json
```diff
@@ -0,0 +1,6 @@
+{
```
khawkins: A couple of thoughts on this.

1. We should consider whether this asset catalog should live in shared/resources/ with the other resources we currently have in the SDK.  The catalog is going to have to be pulled into the Cordova plugin, if that changes the consideration at all.
2. We should rename the catalog to something more specific—again, because it's going to end up logically separated in apps that use static libraries, like Cordova.  Something like SalesforceSDKAssets.xcassets, or something similar?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFLoginViewController.m
```diff
@@ -0,0 +1,217 @@
...
+    self.navBar.topItem.rightBarButtonItem = [[UIBarButtonItem alloc] initWithImage:image style:UIBarButtonItemStylePlain target:self action:@selector(showLoginHost:)];
+    [self.navBar applySFStyle];
+    [self.view addSubview:self.navBar];
+    [self setNeedsStatusBarAppearanceUpdate];
+}
+
+- (void)setupBackButton {
+    // setup left bar button
+    if ([self shouldShowBackButton]) {
+        UIImage *image = [SFSDKResourceUtils imageNamed:@"globalheader-back-arrow"];//SFFoundationImage(@"Login/globalheader-back-arrow");
```
bhariharan: Remove commented out code.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
bhariharan: +1

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFAppStyler.h
```diff
@@ -0,0 +1,145 @@
...
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+typedef NS_ENUM(NSUInteger, SFAppStylerFontStyle)  {
+    SFAppStylerFontStyleRegular = 0,
+    SFAppStylerFontStyleBold
+};
+
+/** Class responsible for app to set overall app style. All UI classes from SFFoundationSDK uses this class
```
khawkins: Not sure of the audience for SFFoundationSDK at this point.  Probably reconsider wording.

kchitalia:
Will start incorporating all the feedback items.

khawkins:
Strangely, when I try to run RestAPIExplorer, I get errors, because `SFInactivityTimerCenter` is using `@import` to import Foundation, but RestAPIExplorer has module support disabled.  The reason I say "strangely" is because I don't see any of your changes that triggered such a condition.  Easy enough to fix, though:

1. In the RestAPIExplorer target, set "Enable Modules (C and Objective-C)" to Yes.
2. The `@import` nomenclature can make it so that Objective-C++ cannot consume that code.  Let's revert that to `#import <Foundation/Foundation.h>` in `SFInactivityTimerCenter.h` and `SFKeychainItemWrapper.h`.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -115,7 +120,9 @@ - (void)initializeAppViewState
 - (void)setupRootViewController
 {
     RestAPIExplorerViewController *rootVC = [[RestAPIExplorerViewController alloc] initWithNibName:nil bundle:nil];
-    self.window.rootViewController = rootVC;
+    UINavigationController *navViewController = [[UINavigationController alloc] initWithRootViewController:rootVC];
+    [[SFAppStyler sharedInstance] styleNavigationBar:navViewController.navigationBar];
+    self.window.rootViewController = navViewController;
```
khawkins: This is pretty wonky.  It creates an artificial nav bar above an existing header bar in the root view controller.  The styling should be applied to the existing `UIToolBar` in the root view controller, if at all.  The nav controller configured here should be removed.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFAppStyler.h
```diff
@@ -0,0 +1,145 @@
...
+ to style relevant UI elementes
+ */
+@interface SFAppStyler : NSObject
```
khawkins: This is an interesting concept to have at the SDK level.  We have very little to style in the Mobile SDK—we have very few UI artifacts.  In that context, this whole system of styling feels pretty prescriptive, in a world where mobile developers are almost certainly going to have their own opinions / practices for the bulk styling concerns of their apps.

Would it not make more sense to simply make public the styling attributes of the few UI artifacts we *do* provide?

kchitalia: This is providing an easy way for setting the font, color of the nav bar, etc. I agree that developers will have their own practices. The way we have built this, it provides an option to override the defaults. 

@bhariharan @wmathurin Thoughts?

bhariharan:
@qingqingliu Could you review this too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFAppStyler.h
```diff
@@ -0,0 +1,145 @@
...
+ @param style Style to apply font to
+ */
+- (void)setFontName:(nullable NSString *)fontName forStyle:(SFAppStylerFontStyle)style;
+
+/** Returns font for a given size using the primary font name
+ 
+ @param fontSize Size of the font
+ @param style font style to use
+ */
+- ( UIFont * _Nonnull )fontWithSize:(CGFloat)fontSize style:(SFAppStylerFontStyle)style;
```
khawkins: We should standardize on either `_Nonnull` or `nonnull`, just for consistency.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFAppStyler.h
bhariharan: +1

libs/SalesforceSDKCore/Images.xcassets/Contents.json
kchitalia: In the shared/resoources/ we do have an Image asset catalog. I can add this as a separate asset catalog in there.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFAppStyler.h
bhariharan: I tend to agree with @khawkins. We have only 2 UI components really - the login screen and the passcode. The concept of an app styler seems like overkill for the Mobile SDK. I'm in favor of simplifying this flow by providing the ability to override those UI components, as opposed to a global styler.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFAppStyler.m
```diff
@@ -0,0 +1,210 @@
...
+    self.actionableFont = [self fontWithSize:14.0 style:SFAppStylerFontStyleRegular];
+}
+
+- (void)setFontName:(NSString * _Nullable)fontName forStyle:(SFAppStylerFontStyle)style {
+    if (fontName) {
+        self.fontMap[@(style)] = fontName;
+    } else {
+        [self.fontMap removeObjectForKey:@(style)];
+    }
+    [self updateDefaultSize];
```
khawkins: This is a bit odd to me.  Why do the different fonts get reset when you set a new font style-to-name configuration?  What if you've already customized the size of the fonts being reset?  You have to do it again?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFSDKLoginHostListViewController.m
```diff
@@ -0,0 +1,283 @@
...
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import "SFSDKLoginHostListViewController.h"
+#import "SFSDKNewLoginHostViewController.h"
+#import "SFSDKLoginHostStorage.h"
+#import "SFSDKLoginHost.h"
+#import "UINavigationController+SFAppStyler.h"
+#import <SalesforceSDKCore/SFAuthenticationManager.h>
```
bhariharan: I think these could be directly imported, like ```#import SFAuthenticationManager.h```, since they're all in the same framework library ```SalesforceSDKCore```. Let's change that for consistency.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFSDKLoginHostListViewController.m
```diff
@@ -0,0 +1,283 @@
...
+    CGRect r = [self.tableView rectForSection:0];
+    
+    CGSize size = CGSizeMake(380, r.size.height);
+    self.preferredContentSize = size;
+}
+
+- (void)viewDidLoad {
+    self.navigationItem.rightBarButtonItem = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemAdd target:self action:@selector(showAddLoginHost:)];
+    self.title = [SFSDKResourceUtils localizedString:@"LOGIN_CHOOSE_SERVER"];
+    //(@"LOGIN_CHOOSE_SERVER", @"choose server");
```
bhariharan: Remove commented out line.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFSDKLoginHostStorage.m
```diff
@@ -0,0 +1,149 @@
...
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import "SFSDKLoginHostStorage.h"
+#import "SFSDKLoginHost.h"
+#import <SalesforceSDKCore/SFManagedPreferences.h>
```
bhariharan: Change ```#import``` style here as well.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFSDKLoginHostStorage.m
```diff
@@ -0,0 +1,149 @@
...
+#import <SalesforceSDKCore/SFSDKResourceUtils.h>
+
+@interface SFSDKLoginHostStorage ()
+
+@property (nonatomic, strong) NSMutableArray *loginHostList;
+
+@end
+
+// Key under which the list of login hosts will be stored in the user defaults.
+static NSString * const SFSDKLoginHostList = @"ChatterLoginHostListPrefs";
```
bhariharan: Rename these keys to something more generic like ```SalesforceLoginHostListPrefs```. Remove references to ```Chatter```.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFSDKNewLoginHostViewController.m
```diff
@@ -0,0 +1,135 @@
...
+    
+    // Disable the scroll because there is enough space
+    // on both the iPhone and iPad to display the two editing rows.
+    self.tableView.scrollEnabled = NO;
+    
+    self.navigationItem.rightBarButtonItem = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemDone
+                                                                                           target:self action:@selector(addNewServer:)];
+    self.navigationItem.rightBarButtonItem.enabled = NO;
+    self.title = [SFSDKResourceUtils localizedString:@"LOGIN_ADD_SERVER"];
+    //SFFoundationLocalizedString(@"LOGIN_ADD_SERVER", @"ADD Server");
```
bhariharan: Remove commented out line here.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -167,6 +174,15 @@ - (void)handleUserSwitch:(SFUserAccount *)fromUser toUser:(SFUserAccount *)toUse
     }];
 }
 
+- (void) configureAppStyle {
+    SFAppStyler *styler = SFAppStyler.sharedInstance;
+    [styler setFontName:@"Helvetica" forStyle:SFAppStylerFontStyleRegular];
+    [styler setFontName:@"SalesforceSans-Bold" forStyle:SFAppStylerFontStyleBold];
+    
+    // Set primary color to different color to style the navigation header
```
bhariharan: Remove commented out code.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -31,6 +31,10 @@
 #import <SalesforceSDKCore/SFPushNotificationManager.h>
 #import <SalesforceSDKCore/SFDefaultUserManagementViewController.h>
 #import <SalesforceSDKCore/SalesforceSDKManager.h>
+#import <SalesforceSDKCore/SFAppStyler.h>
+#import <UIKit/UIKit.h>
+#import <SalesforceSDKCore/SFAuthenticationManager.h>
+#import <SalesforceSDKCore/SFLoginViewController.h>
```
wmathurin: Why are you important SFLoginViewController?

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -115,7 +120,9 @@ - (void)initializeAppViewState
 - (void)setupRootViewController
 {
     RestAPIExplorerViewController *rootVC = [[RestAPIExplorerViewController alloc] initWithNibName:nil bundle:nil];
-    self.window.rootViewController = rootVC;
+    UINavigationController *navViewController = [[UINavigationController alloc] initWithRootViewController:rootVC];
+    [[SFAppStyler sharedInstance] styleNavigationBar:navViewController.navigationBar];
```
wmathurin: So that's an example of how to leverage the new SFAppStyler?

kchitalia: Yes, in addition to the configureAppStyle method.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFAppStyler.m
```diff
@@ -0,0 +1,210 @@
...
+            default:
+                return [UIFont systemFontOfSize:fontSize];
+                break;
+        }
+    }
+}
+
+#pragma mark - Styling Methods
+
+- (nonnull UIImage *)headerBackgroundImage {
```
khawkins: Is there a particular benefit to creating the background color of a navigation bar as an image, rather than simply setting `UINavigationBar.barTintColor`?  Seems like the latter would be more efficient.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFLoginViewController.h
```diff
@@ -0,0 +1,33 @@
+/*
```
khawkins: The classes of SalesforceSDKCore are generally further sub-categorized on the filesystem.  And while these have been in Project Navigator, they're all physically dumped under Classes/ on the filesystem.

Those should be moved to the physical counterparts of their logical folder structure.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFLoginViewController.h
```diff
@@ -0,0 +1,33 @@
...
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <SalesforceSDKCore/SalesforceSDKCore.h>
+
+@interface SFLoginViewController : SFAuthorizingViewController
```
khawkins: It feels a little strange to override `SFAuthorizingViewController`, especially since the child class basically re-implements the few things its parent provides.  Should we just forego `SFAuthorizingViewController`?  The override logic in `SFAuthenticationManager` doesn't really make use of it—it expects customization to come in the form of custom display and dismiss blocks.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/SFLoginViewController.h
kchitalia: I noticed that too, I am going to move them to physical folders on the file system. Thanks for pointing it out.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
wmathurin: Should we do that in our template apps?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Styling/SFAppStyler.m
```diff
@@ -0,0 +1,95 @@
...
+
+#pragma mark - Styling Methods
+
+- (void)styleNavigationBar:(UINavigationBar *)navigationBar {
+    if (!navigationBar) {
+        return;
+    }
+    if (self.navBarColor) {
+        UIImage *backgroundImage = [self headerBackgroundImage];
+        [navigationBar setBackgroundImage:backgroundImage forBarMetrics:UIBarMetricsDefault];
```
kchitalia: Using the backgroundImage here as barTintColor shows a different shade of the color provided.

native/SampleApps/RestAPIExplorer/RestAPIExplorer.xcodeproj/project.pbxproj
```diff
@@ -944,6 +944,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ASSETCATALOG_COMPILER_LAUNCHIMAGE_NAME = LaunchImage;
 				BUNDLE_PREFIX = com.salesforce.mobilesdk;
+				CLANG_ENABLE_MODULES = YES;
```
kchitalia: Set this to yes to get rid of the non modules error.

native/SampleApps/RestAPIExplorer/RestAPIExplorer.xcodeproj/project.pbxproj
```diff
@@ -967,6 +968,7 @@
...
```
kchitalia: Set this to yes to get rid of the non modules error.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -83,6 +86,15 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
+    //
+    //Uncomment the code below to see how you can customize the color, textcolor, font and fontsize of the navigation bar
+    //
+    //SFAppStyler *styler = SFAppStyler.sharedInstance;
+    // Set primary color to different color to style the navigation header
+    //styler.navBarColor = [UIColor colorWithRed:0.051 green:0.765 blue:0.733 alpha:1.0];
+    //styler.navBarFont = [UIFont fontWithName:@"Helvetica" size:16.0];
+    //styler.navBarTextColor = [UIColor redColor];
+    //
     [[SalesforceSDKManager sharedManager] launch];
```
kchitalia: Leaving this code as commented to showcase how to use the SFAppStyler

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -1825,7 +1949,7 @@
 			shellPath = /bin/sh;
 			shellScript = "\"$SRCROOT/../../shared/common/build/tools/generate_headers.sh\" -o \"$TARGET_BUILD_DIR/$PRODUCT_NAME.framework/Headers/$PRODUCT_NAME.h\" -u \"$SRCROOT/$TARGET_NAME/$TARGET_NAME.h\"";
 		};
-		CE4CE4FB1C0E6E07009F6029 /* ShellScript */ = {
+		E1C80D751C5C5056001B3A21 /* ShellScript */ = {
```
kchitalia: @khawkins Adding the script for update_podspec_headers back as I was able to build successfully.

native/SampleApps/RestAPIExplorer/RestAPIExplorer.xcodeproj/project.pbxproj
wmathurin: I thought we tried hard getting rid of CLANG_ENABLE_MODULES.
Why is it needed again?

khawkins: This is different: it's actually ensuring that the project will use/expect modules where necessary.  The other setting we've been squashing is the `CLANG_ALLOW_NON_MODULAR_INCLUDES_IN_FRAMEWORK_MODULES` setting, which provides a stop gap from fixing underlying module errors.

libs/SalesforceNetwork/SalesforceNetwork.xcodeproj/xcshareddata/xcschemes/SalesforceNetworkiOS.xcscheme
```diff
@@ -0,0 +1,169 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "0640"
+   version = "1.3">
```
bhariharan: Why have we renamed this scheme to ```SalesforceNetworkiOS```? Could we revert this to ```SalesforceNetwork``` as it was before please?

libs/SalesforceNetwork/SalesforceNetwork.xcodeproj/xcshareddata/xcschemes/SalesforceNetworkiOS.xcscheme
```diff
@@ -0,0 +1,169 @@
...
+      <BuildActionEntries>
+         <BuildActionEntry
+            buildForTesting = "YES"
+            buildForRunning = "YES"
+            buildForProfiling = "YES"
+            buildForArchiving = "YES"
+            buildForAnalyzing = "YES">
+            <BuildableReference
+               BuildableIdentifier = "primary"
+               BlueprintIdentifier = "82D53D301A3618CD00E46F8F"
```
bhariharan: This scheme is all wrong anyway - it has references to projects that don't exist, such as ```SalesforceSDKCommon```.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Styling/SFAppStyler.h
```diff
@@ -0,0 +1,61 @@
...
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+typedef NS_ENUM(NSUInteger, SFAppStylerFontStyle)  {
```
bhariharan: Are we keeping ```SFAppStyler```? I thought we were going to localize the customizability to the login window itself, in those classes?

kchitalia: Removed SFAppStyler here: [https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/b20fd25e2704fa12c01d7eaba946dc2a117b534d](url)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostListViewController.m
```diff
@@ -0,0 +1,283 @@
...
+    // Refresh the UI and make sure the size is correct
+    [self.tableView reloadData];
+    [self makeLoginHostVisibleAtIndex:index];
+    [self resizeContentForPopover];
+    
+    [super viewDidLoad];
+    
+    [self.navigationItem.rightBarButtonItem setTintColor:[UIColor whiteColor]];
+    [self setEdgesForExtendedLayout:UIRectEdgeNone];
+    //Add check as this is available only in 9.0 and above
```
bhariharan: This comment appears to indicate that we should add an ```OS_VERSION``` check here to limit this to ```iOS 9``` and above.

kchitalia: The if condition handles that. If the OS version is less than 9 the if condition will return false. I have tested this with both iOS 8.4 and iOS 9.2, works fine on both. But if the convention on the iOS SDK code is to specifically use OS_VERSION, I can change it to that.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostListViewController.m
```diff
@@ -0,0 +1,284 @@
...
+    [self.tableView reloadData];
+    [self makeLoginHostVisibleAtIndex:index];
+    [self resizeContentForPopover];
+    
+    [super viewDidLoad];
+    
+    [self.navigationItem.rightBarButtonItem setTintColor:[UIColor whiteColor]];
+    [self setEdgesForExtendedLayout:UIRectEdgeNone];
+    
+    //TODO: Remove the check once we move to iOS 9 as minimum.
```
kchitalia: Added TODO here to remove the check once we move to iOS9 as minimum supported version.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
```diff
@@ -0,0 +1,50 @@
...
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import "SalesforceSDKCore.h"
+
+@interface SFLoginViewController : SFAuthorizingViewController
+
+/** Get a shared singleton of `SFLoginViewController` class
+ */
++(_Nonnull instancetype)sharedInstance;
```
khawkins: I'm rarely a fan of singleton view controllers.  Is there any reason for this one to be a singleton?

kchitalia: I was hesitant about this too, but since I moved the styling method for the navbar in here I created a singleton to access the view controller and override the styling of the navbar in the consuming app.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
```diff
@@ -0,0 +1,50 @@
...
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import "SalesforceSDKCore.h"
+
+@interface SFLoginViewController : SFAuthorizingViewController
```
khawkins: I think I had an old comment here that was probably lost with updates.  The gist of it was, is there any reason to keep `SFAuthorizingViewController`?  Our customization hooks for login pages follow a different pattern (customized blocks) from inheritance, so I see `SFLoginViewController` as more of a replacement for it.  Not sure I see value in maintaining the inheritance.  Thoughts?

kchitalia: I agree, if we dont use it anywhere we should get rid of it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
khawkins: That seems reasonable enough.  I have some lingering concerns about previous auth state hanging around in UI artifacts that don't die, but we can keep an eye out for those situations.  I'd be a fan of adding `[SFLoginViewController sharedInstance].oauthView = nil;` in the default `dismissBlock` set in `[SFAuthenticationManager init]`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
khawkins: We don't.  I think your stuff should simply replace it.

kchitalia: Sounds good. I'll remove it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
kchitalia: Will add it in.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
```diff
@@ -0,0 +1,55 @@
...
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import "SalesforceSDKCore.h"
+
+@interface SFLoginViewController : UIViewController
```
kchitalia: Removed inheritance from SFAuthorizingViewController

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -312,10 +312,11 @@ - (id)init
         self.authViewHandler = [[SFAuthenticationViewHandler alloc]
                                 initWithDisplayBlock:^(SFAuthenticationManager *authManager, UIWebView *authWebView) {
                                     if (weakSelf.authViewController == nil)
-                                        weakSelf.authViewController = [[SFAuthorizingViewController alloc] initWithNibName:nil bundle:nil];
+                                        weakSelf.authViewController = [SFLoginViewController sharedInstance];
                                     [weakSelf.authViewController setOauthView:authWebView];
                                     [[SFRootViewManager sharedManager] pushViewController:weakSelf.authViewController];
                                 } dismissBlock:^(SFAuthenticationManager *authViewManager) {
+                                    [SFLoginViewController sharedInstance].oauthView = nil;
```
kchitalia: @khawkins Setting oauthView to nil here.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -159,8 +159,6 @@
 		CE4CE35E1C0E526A009F6029 /* SFAuthErrorHandler.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F96FCF31BFD32130022F021 /* SFAuthErrorHandler.m */; };
 		CE4CE35F1C0E526A009F6029 /* SFAuthErrorHandlerList.h in Headers */ = {isa = PBXBuildFile; fileRef = 4F96FCF41BFD32130022F021 /* SFAuthErrorHandlerList.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		CE4CE3601C0E526A009F6029 /* SFAuthErrorHandlerList.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F96FCF51BFD32130022F021 /* SFAuthErrorHandlerList.m */; };
-		CE4CE3611C0E526A009F6029 /* SFAuthorizingViewController.h in Headers */ = {isa = PBXBuildFile; fileRef = 4F96FCF61BFD32130022F021 /* SFAuthorizingViewController.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		CE4CE3621C0E526A009F6029 /* SFAuthorizingViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F96FCF71BFD32130022F021 /* SFAuthorizingViewController.m */; };
```
kchitalia: @khawkins Removing SFAuthorizing ViewController here and in other files as well.

khawkins: Cool.  The only other place that I know of that it needs to be removed is in SalesforceSDKCore.podspec.

kchitalia: I'll have to replace it with SFLoginViewController right?

khawkins: Yep, good point.

khawkins:
I think this is mostly looking pretty good.  The only outstanding thing I can think of is @wmathurin 's original comment above, about whether we're also going to get rid of support for the Settings app.  I think we should, if for no other reason than to reduce the complexity of the authentication vectors.  It would definitely simplify some of the logic in `SalesforceSDKManager`.

Anyway, I can add some points of advice if we opt to go that way.

kchitalia:
I am thinking maybe that should be handled in a separate pull request?

khawkins:
Could be.  I'm okay with that.

bhariharan:
Looks good to me. Let's get this merged in and do the ```Settings``` removal in another pull request, along with ```podspec``` updates.

:+1:

wmathurin:
Sounds good to me too.

PR #1370, opened 2016-01-30
A script to test forceios.js 

wmathurin:
You can specify the fork/branch and app types you wish to test.
It clones the repo, run ant to build the forceios package, then npm installs it to a temporary directory.
Then for each app type specified, it forceios create an application and xcode builds it.
It will exit with an error if any of the steps fail.

For native application, it edits the pod file so that it points to the cloned repo (otherwise it would pull the pods from the current release).

For hybrid application however, it's pulling the currently published plugin, so it's not very useful for testing code from unstable. Maybe I should remove support for testing hybrid applications?

Anyway plenty of room for improvements.

### DIFF
--- a/node/test_forceios.js
+++ b/node/test_forceios.js
@@ -131,15 +131,15 @@ function createCompileApp(tmpDir, appType) {
     runProcess(forceiosPath + ' ' + forceiosArgs);
 
     // Pointing to repoDir in Podfile
-    shelljs.sed('-i', /pod ('Salesforce.*')/g, 'pod $1, :path => \'../SalesforceMobileSDK-iOS\'', path.join(tmpDir, 'nativeApp', 'Podfile'));
-    shelljs.sed('-i', /pod ('Smart.*')/g, 'pod $1, :path => \'../SalesforceMobileSDK-iOS\'', path.join(tmpDir, 'nativeApp', 'Podfile'));
+    shelljs.sed('-i', /pod ('Salesforce.*')/g, 'pod $1, :path => \'../SalesforceMobileSDK-iOS\'', path.join(tmpDir, appName, 'Podfile'));
+    shelljs.sed('-i', /pod ('Smart.*')/g, 'pod $1, :path => \'../SalesforceMobileSDK-iOS\'', path.join(tmpDir, appName, 'Podfile'));
 
     shelljs.pushd(path.join(tmpDir, appName));
     runProcess('pod update');    
     shelljs.popd();
 
     var workspacePath = path.join(tmpDir, appName, appName + '.xcworkspace');
-    runProcess('xcodebuild -workspace ' + workspacePath + ' -scheme Pods-' + appName)
+    runProcess('xcodebuild -workspace ' + workspacePath + ' -scheme Pods-' + appName + ' clean build')
 }
 
 //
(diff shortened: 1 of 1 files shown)


khawkins:
Good idea!  :+1:

node/test_forceios.js
```diff
@@ -0,0 +1,204 @@
...
+    // Pointing to repoDir in Podfile
+    shelljs.sed('-i', /pod ('Salesforce.*')/g, 'pod $1, :path => \'../SalesforceMobileSDK-iOS\'', path.join(tmpDir, 'nativeApp', 'Podfile'));
+    shelljs.sed('-i', /pod ('Smart.*')/g, 'pod $1, :path => \'../SalesforceMobileSDK-iOS\'', path.join(tmpDir, 'nativeApp', 'Podfile'));
+
+    shelljs.pushd(path.join(tmpDir, appName));
+    runProcess('pod update');    
+    shelljs.popd();
+
+    var workspacePath = path.join(tmpDir, appName, appName + '.xcworkspace');
+    runProcess('xcodebuild -workspace ' + workspacePath + ' -scheme Pods-' + appName)
```
khawkins: Shouldn't the scheme here be the actual app scheme, rather than the pod scheme created for its dependencies?

khawkins: Also, I'd expliclty add `clean build` to the end of your `xcodebuild` line.  Emphasis being the "clean" command.  That way, you (hopefully) won't encounter stale state in DerivedData from previous runs of the script for that app type.

wmathurin: Here is the tricky part. The app scheme is not there until you open the workspace in xcode.

```
$ xcodebuild -workspace nativeApp.xcworkspace -list
Information about workspace "nativeApp":
    Schemes:
        CocoaLumberjack
        Pods-nativeApp
        SalesforceNetwork
        SalesforceRestAPI
        SalesforceSDKCore
        SalesforceSDKCore-SalesforceSDKResources
        SalesforceSDKCore-Settings
        SmartStore
        SmartSync
        SQLCipher
$ open nativeApp.xcworkspace
$ xcodebuild -workspace nativeApp.xcworkspace -list
Information about workspace "nativeApp":
    Schemes:
        CocoaLumberjack
        Pods-nativeApp
        SalesforceNetwork
        SalesforceRestAPI
        SalesforceSDKCore
        SalesforceSDKCore-SalesforceSDKResources
        SalesforceSDKCore-Settings
        SmartStore
        SmartSync
        SQLCipher
        nativeApp
```

node/test_forceios.js
```diff
@@ -0,0 +1,204 @@
...
+}
+
+//
+// Create and compile hybrid app 
+//
+function createCompileHybridApp(tmpDir, appType) {
```
bhariharan: For testing hybrid, perhaps we could run ```sed``` on the local ```forcedroid``` we're testing and add the branch and fork specifier to the line where ```CordovaPlugin``` repo is specified. This will enable us to test it as well as native. Only catch is that we have to manually update the ```CordovaPlugin``` repo on that branch before running the script, but I think that should be okay.

bhariharan:
Looks good otherwise.

node/test_forceios.js
wmathurin: Good idea.
Doing it here https://github.com/wmathurin/SalesforceMobileSDK-iOS/commit/41d247340ea1f6b9580f43ed352c7c8f6e33c67d to test hybrid app with test_forceios.js

wmathurin:
I'm going to combine test_forceios.js and test_forcedroid.js and have the script live in the shared repo.
I'm going to send a PR soon.

PR #1371, opened 2016-02-01
Move completeOperationWithError to CSFAction.h

qingqingliu:
@bhariharan @khawkins review please.

Move "completeOperationWithError" to CSFAction.h so that subclass can override it to perform extra task.

### DIFF
--- a/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.h
+++ b/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.h
@@ -202,6 +202,20 @@
 
 - (BOOL)shouldRetryWithError:(NSError*)error;
 
+/** Subclassable method that will be called when the action is about to call responseBlock. 
+ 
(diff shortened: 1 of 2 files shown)


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.h
```diff
@@ -202,6 +202,20 @@
 
 - (BOOL)shouldRetryWithError:(NSError*)error;
 
+/** Subclassable method that will be called when the action is about to call responseBlock. 
+ 
+ @discussion
```
khawkins: We might want to make note that you must call the super class version of this as well, since it's responsible for cleaning up the operations.  Otherwise, :+1:

qingqingliu: @khawkins, updated PR to include the comment as you suggested to call super.

khawkins:
Thanks @qingqingliu !

PR #1372, opened 2016-02-01
Add Id and LastModifiedDate if they're missing

bhariharan:
Tests coming up.


wmathurin:
Looks good.
Shouldn't there be more code simplifications in the sample app?
Could we have a little test?

bhariharan:
Yes - adding test as we speak. Not much simplification in sample app other than we can afford to drop ```LastModifiedDate``` from the ```SOQL``` query for ```syncDown```.

wmathurin:
What about the Id? Can't we drop that too from the SOQL query?

wmathurin:
:+1:

bhariharan:
Added test [here](https://github.com/bhariharan/SalesforceMobileSDK-iOS/commit/6e7cc7c256520f7a30648150db9b0fc927ecfe2c). I'll drop ```Id``` too.

Issue #1218, opened 2015-10-19
Objective-c enum declarations should play nice with Swift


fobin:
This from SFSyncDownTarget:
```
typedef enum {
  SFSyncDownTargetQueryTypeMru,
  SFSyncDownTargetQueryTypeSosl,
  SFSyncDownTargetQueryTypeSoql,
  SFSyncDownTargetQueryTypeCustom
} SFSyncDownTargetQueryType;
```

Should be converted to NS_ENUM to function with Swift. And there are others like this also.

mgetzbw:
@wmathurin @bhariharan I will be slowly making pulls on this as I get to bits that are particularly egregious. If there are headers that you would prefer I focus on let me know, I am trying to be very conservative in what I do for now as I don't want to cause any implementation changes.

wmathurin:
Done here https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1380
Will be part of our next release.

PR #1373, opened 2016-02-02
Swift support, use NS_ENUM since this enum is not intended for options

mgetzbw:
contribution to #1218


bhariharan:
@mgetzbw Could you resubmit this against our ```unstable``` branch please?

mgetzbw:
@bhariharan done as #1374

PR #1375, opened 2016-02-03
Better swift support per #1218

mgetzbw:
Better swift support per #1218

* Set nullability to reduce the number of optional checks or ensure that swift programmers have to check for optionals in several cases where `nil` could result
* Use Objective-C lightweight generics to restrict the types that a swift programmer can pass to the API
* Update documentation to reflect nullability on return values


wmathurin:
Thanks.

PR #1377, opened 2016-02-05
Using react 0.19

wmathurin:
That was easy. It compiled without any changes.
I generated a react native app from the template (using test_force.js) and ran it from XCode without any problems.


bhariharan:
Wow....too easy. :+1:

wmathurin:
I'll try android now.

PR #1378, opened 2016-02-05
Using fixed version of shells (generation of hybrid local failing wit…

wmathurin:
…h shelljs 0.6.0)

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
 	"os" : [ "darwin" ],
 	"bin" : { "forceios" : "node/forceios.js" },
     "dependencies": {
-        "shelljs": ">=0.1.4"
+        "shelljs": "~0.5.3"
     },
     "scripts": {
         "prepublish": "external/shared/node/prepublish.js",


khawkins:
:+1:

Depending on the extent of updates to support a new version of shelljs, I wouldn't be averse to losing it altogether, either.  It's not particularly necessary on the iOS (OS X) side, where it's simply mimicking built-in unix functionality.

kchitalia:
:+1:

wmathurin:
I haven't looked at it too closely. But shelljs.cp where we copy path/* is now failing, complaining that it can't find path/*. 

It's the second time in a couple of months (last one was the sqlcipher release of 3.3) that an update to a third party library we depend on cause some of our stuff to break.
In the future, I recommend, we pin down the exact version we need wherever we use third party libraries.

bhariharan:
I agree with @khawkins. We could lose it on ```iOS``` and use straight up ```exec``` calls instead, since it's limited to ```OS X```. Usage of ```shelljs``` in ```forcedroid``` might be unavoidable though, in order to support ```Windows```.

bhariharan:
:+1:

PR #1379, opened 2016-02-06
Remove Settings bundle

kchitalia:


build/app_template_files/__NativeSwiftTemplateAppName__/__NativeSwiftTemplateAppName__.xcodeproj/project.pbxproj
```diff
@@ -15,11 +14,11 @@
 		4FD6C49F1B755242002F9F90 /* RootViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4FD6C49C1B755242002F9F90 /* RootViewController.swift */; };
 		4FD6C4A01B755242002F9F90 /* InitialViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4FD6C49D1B755242002F9F90 /* InitialViewController.swift */; };
 		C77FD628010EF64B9B2C4363 /* libPods-__NativeSwiftTemplateAppName__.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 9737303E047BCEC3E2EC6AE5 /* libPods-__NativeSwiftTemplateAppName__.a */; };
+		E1CE60881C655FE6007D6DCB /* SalesforceSDKAssets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = E1CE60871C655FE6007D6DCB /* SalesforceSDKAssets.xcassets */; };
```
kchitalia: Adding SDK Assets as part of the template app.

build/app_template_files/__NativeSwiftTemplateAppName__/__NativeSwiftTemplateAppName__/AppDelegate.swift
```diff
@@ -86,6 +86,15 @@ class AppDelegate : UIResponder, UIApplicationDelegate
...
+        //If you wish to customize the color, textcolor, font and fontsize of the navigation bar uncomment the
+        //code below.
+        //
+        //SFLoginViewController *loginViewController = [SFLoginViewController sharedInstance];
+        // Set primary color to different color to style the navigation header
+        //loginViewController.navBarColor = [UIColor colorWithRed:0.051 green:0.765 blue:0.733 alpha:1.0];
+        //loginViewController.navBarFont = [UIFont fontWithName:@"Helvetica" size:16.0];
+        //loginViewController.navBarTextColor = [UIColor blackColor];
+        //
         SalesforceSDKManager.sharedManager().launch()
```
kchitalia: Code to show how users can style the nav bar.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
```diff
@@ -52,10 +52,16 @@ + (NSString *)localizedString:(NSString *)localizationKey
 
 + (UIImage *)imageNamed:(NSString *)name
 {
-    NSBundle *bundle = [NSBundle bundleForClass:[self class]];
     NSAssert(name != nil, @"name must contain a value.");
+    //Get image from main bundle (this is to allow extracting SDK Assets when using template app as we include SDK Assets as part of the tempalte app)
```
kchitalia: I changed the logic in here. First we get the image from the main bundle of the app. If that fails we get the image from the SDK Asset catalog. I am doing this so that it works for both, sample apps in the sdk repo and apps generated via force ios. Apps generated via force ios will have the SDK asset catalog as part of the main bundle of the app.

wmathurin:
Settings.bundle and Images.xcassets are in shared/resources
And the template apps simply have pointers to them (and instead of having copies), we should do the same for SalesforceSDKAssets.xcassets.

wmathurin:
Also is the settings.bundle removed from all the sample apps?

kchitalia:
@wmathurin I will modify the way I have copied the SalesforceSDKAssets.xcassets.

 I removed the settings bundle from RestAPIExplorer, SmartSyncExplorer, SmartSyncExplorerHybrid, AccountEditor, NoteSync. Am I missing any sample app?

Also, I was able to create a sample app using forceios, all 3 flavors - native, native_swift, and native_react and see my changes for the login view controller.

wmathurin:
Those are all the sample apps we have in the iOS repo.

build/app_template_files/__NativeSwiftTemplateAppName__/__NativeSwiftTemplateAppName__/SalesforceSDKAssets.xcassets/login-window-gear.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
```
bhariharan: As @wmathurin mentioned, we probably want these to be symlinks to a common shared folder like ```Settings.bundle``` is/was.

build/app_template_files/__NativeTemplateAppName__/__NativeTemplateAppName__/SalesforceSDKAssets.xcassets/globalheader-back-arrow.imageset/Contents.json
```diff
@@ -0,0 +1,23 @@
+{
+  "images" : [
```
bhariharan: Same here - symlinks.

shared/resources/Settings.bundle/en.lproj/Root.strings
```diff
@@ -1,44 +0,0 @@
...
-"GROUP_TITLE_ACCOUNT"="Account";
-/* group label for Account settings */
-"LOGOUT_SELECTOR_TITLE"="Logout Now";
-/* allows user to log out of account*/
-"GROUP_TITLE_DEBUG"="Debug";
-/* a section title for debug settings (SEND_LOGS_SELECTOR_TITLE) */
-"SEND_LOGS_SELECTOR_TITLE"="Send Logs";
-/* if true, app will send debug logs to Salesforce Support */
-"GROUP_TITLE_CONNECTION"="Connection";
-/* Settings label for network connection */
```
bhariharan: Are these strings being used anywhere else?

bhariharan:
I believe we have some special logic in ```SalesforceSDKManager``` for host changing from ```Settings.bundle```, which can be removed, right @khawkins?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
khawkins: Why will the generated apps as part of the main bundle, if they're using frameworks?

shared/resources/Settings.bundle/en.lproj/Root.strings
khawkins: Nope.  Just for the Settings app.

khawkins:
Two places I see where the Settings bundle is still referenced:

1. `build/build.xml`, via the `sdkAppSettingsBundleName` property.  It's not being used, though, so you should just be able to remove that property.
2. In the configuration of the `SalesforceHybridSDKTestApp` target of the hybrid SDK.

shared/resources/Settings.bundle/en.lproj/Root.strings
kchitalia: Before I removed this, I did a search for all these strings and could not find it anywhere in the workspace.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
kchitalia: If we include these assets as part of the pod, the assets get packaged as a bundle versus a .car. So now we need two different ways (in code) of getting the images for the login screen. To avoid this we are including the SDK Assets(gear icon and back arrow icon) as part of the template app.

khawkins:
So as @bhariharan mentions, there are some other areas of the SDK that we should clean up, in correlation with the removal of support for the Settings bundle login settings.

### SFUserAccountManager

There are sections of support around the login host, for querying that value from the Settings app.  In particular, `[SFUserAccountManager appSettingsLoginHost]` is called by `[SFUserAccountManager loginHost]` and `[SFUserAccountManager updateLoginHost]`, to leverage any changes through that configuration.

### SalesforceSDKManager

`[SFUserAccountManager updateLoginHost]` ripples up to `SalesforceSDKManager`, which checks for login host changes and the logout switch being set, in `processLogoutAndUserSwitchSettings`.  This whole method can go away, since those settings will not otherwise be useful.  This in turn impacts the `handleAppForeground:` and `launch` methods, but hopefully that refactor should be straightforward.

kchitalia:
@khawkins Thanks for the pointers. I will refactor the code in those two places.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
khawkins: Hmmm.  This seems like a bigger problem than the template apps, then.  If it's an issue there, it's an issue with apps that consume the SDK directly through CocoaPods.  We're going to need to find a more general solution to supporting these image assets through CocoaPods.

khawkins:
@kchitalia Sure thing, let me know if you run into any questions with that, as I know my description is a bit open-ended.  I forgot one related method: `[SFUserAccountManager updateAppSettingsLoginHost:]` can also go away.  Basically, anything that uses the constants:

- `kAppSettingsLoginHost`
- `kAppSettingsLoginHostCustomValue`
- `kAppSettingsLoginHostIsCustom`

Up to and including those constants, can go away.  And to be clear, `[SFUserAccountManager updateLoginHost]` can also go away.  As can the `SFLoginHostUpdateResult` object definition, and any consumers.  I'll let you know if I come across more. :)  But the basic exercise is to just verify whether or not any of that related stuff is being used anywhere that seems like it still needs to be, and otherwise remove it all.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
kchitalia: Yep, the issue is with how an asset catalog gets packaged in a pod. I am
trying to see if instead of an asset catalog I can use a folder containing
.png files of the images. Thoughts?

Sent from my iPhone

On Feb 8, 2016, at 1:37 PM, Kevin Hawkins <notifications@github.com> wrote:

In
libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
<https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1379#discussion_r52223221>
:

>      NSAssert(name != nil, @"name must contain a value.");
> +    //Get image from main bundle (this is to allow extracting SDK Assets when using template app as we include SDK Assets as part of the tempalte app)

Hmmm. This seems like a bigger problem than the template apps, then. If
it's an issue there, it's an issue with apps that consume the SDK directly
through CocoaPods. We're going to need to find a more general solution to
supporting these image assets through CocoaPods.

—
Reply to this email directly or view it on GitHub
<https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1379/files#r52223221>
.

wmathurin: Have you looked at http://stackoverflow.com/questions/28126788/including-image-assets-when-releasing-a-cocoapod ?

khawkins: The solution is inaccurate.  There are stumbling blocks to getting asset catalogs even properly recognized in CocoaPods, and then [bugs/limitations in CocoaPods itself](https://github.com/CocoaPods/CocoaPods/issues/1546) that do not allow them to be compiled properly, once they exist in the Pod.

I think the proper support for the time being, given our heavy reliance on CocoaPods coupled with its lack of proper support for asset catalogs, is to deliver our image assets directly, through a resource bundle.  I.e. the old-fashioned way.

There's some elusive logic in between the lines out there for how you can make asset catalogs work with CocoaPods, but in my opinion, the moment you have to piece together 4 separate bug reports, 3 blog posts, and a Twitter post, to try to make a sane solution, it's time to fall back. :)

khawkins: Of course I say that so as to deliberately put my foot in my mouth, and thereby confirm that the solution works. ;-)  I had tried using the `resource_bundles` approach, without luck.  But when I add the following to the SalesforceSDKCore.podspec:
```
sdkcore.resource = 'shared/resources/SalesforceSDKAssets.xcassets'
```
That seems to do the right thing, both with and without `use_frameworks!`.  So it looks like that could be the solution.

wmathurin: Thanks for trying it out.

kchitalia: The only drawbacks of using "resources" is as mentioned here: 

http://guides.cocoapods.org/syntax/podspec.html#resources

The recommended way is to use resource bundles. I have tried using resources and it works, but we run the risk of running into name collisions.

khawkins: In this case, the collision would have to be on a file called SalesforceSDKAssets.xcassets.  IMO we're probably safe enough from collisions there.

PR #1381, opened 2016-02-10
Making up for xcodebuild's deficiencies

khawkins:
In Xcode 7.2 and later, xcodebuild will not build simulator builds without a further hint as to what the platform is that it's building against.  This PR adds the hint.  NB: I add it for device too, just so that there's not wonky conditional logic in the ant script.


wmathurin:
:+1:

PR #1382, opened 2016-02-10
Remove settings bundle

khawkins:
This picks up from @kchitalia 's [original PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1379), with the following updates:

- I added 'shared/resources/SalesforceSDKAssets.xcassets' as a resource to SalesforceSDKCore.podspec, as that seems to do the right thing in terms of including the asset catalog either in the framework, or standalone with the static library, depending on the Podfile configuration.
- I removed some of the outstanding Settings.bundle references in the repo—minor stuff, like dangling build script properties, etc.  Cf. SalesforceHybridSDK.xcodeproj, build.xml.
- I cleaned up the logic in `SFUserAccountManager` and `SalesforceSDKManager` around looking for login host changes and the logout setting in the Settings app.  Mostly just straight removal, but you can take a look at those two classes for more details.
- I reverted `[SFSDKResourceUtils imageNamed:]`, because, with the addition of the asset catalog to the podspec, the original functionality worked as is.

I believe this addresses the outstanding work, and comments, of the previous PR.  Let me know if I'm missing anything.  Cordova plugin PR to follow this one, but changes are validated there as well.


khawkins:
Oops, I just noticed that the SalesforceSDKAssets.xcassets entries are still present in the template apps, which they shouldn't be.  I'll look at removing those.

build/app_template_files/__NativeSwiftTemplateAppName__/__NativeSwiftTemplateAppName__/SalesforceSDKAssets.xcassets/Contents.json
```diff
@@ -0,0 +1,6 @@
+{
```
wmathurin: Shouldn't we have one copy of SalesforceSDKAssets.xcassets in the repo, and links to it in each template apps? (The way it used to be done with the Settings.bundle).

khawkins: The template apps no longer need any control over SalesforceSDKAssets.  That asset catalog should be delivered via the artifacts of the pod install.

khawkins: Or to put it another way, the asset catalog should be the responsibility of the SDK, as it is for independent CocoaPods installs for third party apps.

kchitalia: :+1:

kchitalia:
@khawkins You beat me to the punch. I was gonna say, you might have to remove the actual asset catalog from the template apps :)

khawkins:
@wmathurin gets the credit. ;-)  Anyway, I've removed references to and copies of the SalesforceSDKAssets from the template apps, and verified that they're still properly pulled from the asset catalog deployed as part of the pod install.  So I think this one's good to go, at least on that front.

wmathurin:
Looks good.
I ran test_force.js to check that native app could successfully be generated and compiled.
```
# in the SalesforceMobileSDK-Shared repo
./node/test_force.js --os=ios --test=native,native_swift,react_native --fork=khawkins --branch=remove_settings_bundle
```

bhariharan:
NIce! :+1:

PR #1384, opened 2016-02-10
Updating submodule pointer

khawkins:
@bhariharan @wmathurin I can update the Android side too; is there anything currently in Shared that would be incompatible there?


wmathurin:
There is nothing incompatible there. Thanks.

bhariharan:
Go for it! :+1:

PR #1387, opened 2016-02-12
Using UIAlertController instead of UIAlertView

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -1164,23 +1182,4 @@ - (void)identityCoordinator:(SFIdentityCoordinator *)coordinator didFailWithErro
     }
 }
 
-#pragma mark - UIAlertViewDelegate
-
-- (void)alertView:(UIAlertView *)alertView didDismissWithButtonIndex:(NSInteger)buttonIndex
-{
-    if (alertView == _statusAlert) {
-        _statusAlert = nil;
```
khawkins: This got lost in the new functionality.  Which means that the Retry functionality will only ever work once, because of the not-`nil` check in `showAlertWithTitle:message:buttonTitle:tag:`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -883,30 +882,49 @@ - (void)retrievedIdentityData
...
-        // Show alert and execute failure block.
-        _statusAlert = [[UIAlertView alloc] initWithTitle:[SFSDKResourceUtils localizedString:kAlertErrorTitleKey]
-                                                  message:[SFSDKResourceUtils localizedString:kAlertVersionMismatchErrorKey]
-                                                 delegate:self
-                                        cancelButtonTitle:[SFSDKResourceUtils localizedString:kAlertOkButtonKey]
-                                        otherButtonTitles: nil];
-        _statusAlert.tag = kConnectedAppVersionMismatchViewTag;
-        [_statusAlert show];
+        __weak SFAuthenticationManager *weakSelf = self;
+        _statusAlert = [UIAlertController alertControllerWithTitle:title
```
khawkins: Can we take this opportunity to swap out the backing variable use, in favor of the property accessor?  The latter buys us more automated variable management than this format.

wmathurin: Done here https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/154ba1346e57851385717bb7968cf874fea31683

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
wmathurin: Done here https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/154ba1346e57851385717bb7968cf874fea31683

bhariharan:
Looks good to me. :+1: 

Does ```Forgot Passcode``` work the same as before?

wmathurin:
I tried the forgot passcode flow and to the user it looks identical as before.

bhariharan:
:+1:

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
khawkins: Oh thanks.  I mostly meant for this property specifically, but thanks for covering all of the bases!

PR #1388, opened 2016-02-12
Occasional crash during logout or user switching

wmathurin:
Fixing exc_bad_access on post notification
Went looking for objects doing addObserver without doing removeObserver - and found at least one
http://stackoverflow.com/questions/15995138/nsnotificationcenter-postnotificationname-exec-badaccess


bhariharan:
:+1:

khawkins:
Ah, good catch!  I saw the `CSFNetwork` add/remove logic, but forgot to check `CSFSalesforceAction`. :+1:

Issue #1396, opened 2016-02-18
Custom host setting not working with new update


manish-metacube:
Hi All, 

I have tried MobileSDK 4.1, With new SDK, environment change is part of login screen. With new one if user enters custom host, It's not working anymore. Getting following error. 
`ERROR|SFAuthenticationManager|Error during authentication: Error Domain=WebKitErrorDomain Code=101 "(null)" `

One more problem with new interface, if user has made typo in custom host, he has no choice to correct it except deleting app as screen flash with above error continuously. 

Do we have any fix for that?

-Manish Kumar

bhariharan:
@manish-metacube We are aware of this issue and looking into it. Thanks for reporting it! Stay tuned for a fix.

brad36:
I don't see the error that OP sees. The issue I see is a blue screen, no way out for the user, when I try to change login host on 4.1 (pic attached):

![img_0028](https://cloud.githubusercontent.com/assets/5522047/13258420/10d2b976-da21-11e5-978e-a17c85bbb0e6.PNG)

bhariharan:
This issue has been fixed in ```master```. For the blue screen, you can click the ```Gear``` icon and change the server to a valid host. For an invalid custom host, you can use the ```Gear``` icon to create a new valid custom host. You won't need to uninstall the app.

brad36:
I'm getting in 4.1.1, changing to sandbox gives blue screen, then changing back to production does nothing, leaves the screen blue unchanged. User is left confused and bewildered. After about 10 minutes, then the login eventually appears. Clearly they can't wait 10 minutes. The only real workaround for the user appears to be app switcher, close app. Do you really consider this fixed, or am I misunderstanding something?

brad36:
Note the above is simply using the gear icon to try to change the login host for a fresh install of the app on 9.2.1, I'm not trying to do anything fancy

jeffreyjackson:
I'm able to reproduce this error as well.

brad36:
I'm still seeing this issue in 4.1.2

brad36:
Switching to Sandbox using the gear icon gives blue screen, the gear icon doesn't fix the blue screen. In fact, I can't even log in to sandbox unless I use the "Use Custom Domain" link. @bhariharan Are we talking about the same issue? Is this really fixed?

bhariharan:
@kchitalia Could you check and respond please?

kchitalia:
[...]

PR #1401, opened 2016-02-19
Fix error for custom server

kchitalia:
Fix for #1396


bhariharan:
:+1:

bhariharan:
Could you add some automation around this? It looks like this issue would've always been there.

PR #1403, opened 2016-02-22
more swift support

mgetzbw:
[implicit unwrapping should be avoided at all costs](https://developer.apple.com/swift/blog/?id=25) as it can introduce dangerous bugs. Default to nullable unless provable otherwise.

Concerns:
==========

* Several things I've marked as `nullable` probably shouldn't be but are because the implementation can return `nil` in pre `launch()` states (IMO calling these then without first assigning to them should be undefined behavior or should fail explicitly).
* There should probably be a tracking issue for all of the implicit unwrapping issues
* Changing nullability is technically a breaking change for swift users


mgetzbw:
ref #1412 and #1413

bhariharan:
Thanks @mgetzbw!

PR #1414, opened 2016-02-23
Fix for irrecoverable loop in 'Add Custom Server' flow

bhariharan:
This fixes the bad server name use case.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -168,7 +168,7 @@ static NSString * const kOAuthUserAgentUserDefaultsKey          = @"UserAgent";
     NSAssert([self.credentials.clientId length] > 0, @"credentials.clientId cannot be nil or empty");
     NSAssert([self.credentials.identifier length] > 0, @"credentials.identifier cannot be nil or empty");
     NSAssert(nil != self.delegate, @"cannot authenticate with nil delegate");
-    
+
     if (self.authenticating) {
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -886,6 +886,7 @@ - (void)webView:(UIWebView *)webView didFailLoadWithError:(NSError *)error {
     
     NSURL *requestUrl = [webView.request URL];
     NSString *errorUrlString = [NSString stringWithFormat:@"%@://%@%@", [requestUrl scheme], [requestUrl host], [requestUrl relativePath]];
+    [self.delegate oauthCoordinator:self didBeginAuthenticationWithView:self.view];
```
bhariharan: Showing the auth container in the failure use case, so that the user can change the login host.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -889,38 +891,40 @@ - (void)showAlertForConnectedAppVersionMismatchError
...
                                        } else if (tag == kConnectedAppVersionMismatchViewTag) {
+
                                            // The OAuth failure block should be followed, after acknowledging the version mismatch.
                                            [weakSelf execFailureBlocks];
                                        }
                                    }];
-        
         [self.statusAlert addAction:okAction];
-        [[SFRootViewManager sharedManager]  pushViewController:self.statusAlert];
+        UIAlertAction *cancelAction = [UIAlertAction actionWithTitle:secondButtonTitle
```
bhariharan: Adding an extra dismiss button for the popup, so that the loop is recoverable.

wmathurin:
:+1:

kchitalia:
:+1:

Issue #1417, opened 2016-02-24
Entries are not synchronized to server in order


coderReview:
When records are created offline they get local ids. But when the application is back online the order of the synchronized records is not the same as they were created.
NSSet was used and this class does not keep any order.

PR #1418, opened 2016-02-24
Synchronize with server with the same order as the entries were created

coderReview:
Related to issue #1417.


wmathurin:
Looks reasonable. Thanks.

We will make the same change in the Android and Windows implementation.

Issue #1413, opened 2016-02-23
Swift support: Use Objective-C lightweight Generics


mgetzbw:
[Swift when it imports Objective-c apis uses the generic annotations to strongly type collections](https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/WorkingWithCocoaDataTypes.html#//apple_ref/doc/uid/TP40014216-CH6-ID61). We should use this to limit what sort of data swift users of the SDK can pass to reduce the number of bugs and ensure we're getting the right data in.

kchitalia:
Closing this issue as PR has been merged.

Issue #1412, opened 2016-02-23
Swift support: Specify nullability


mgetzbw:
[Swift treats any api that does not explicitly specify nullability as implicitly unwrapped](https://developer.apple.com/swift/blog/?id=25), even when those apis can return `nil` and should be `Optional`. Implicit unwrapping can lead to bugs that a swift programmer simply cannot fix because the api is behaving as expected. We should specify nullability whenever possible to explicitly make it clear the intent of the API both in Objective-C and Swift.

kchitalia:
Closing this out as PR has been merged.

PR #1419, opened 2016-02-24
Swift Support and nullability

mgetzbw:
yet more work on #1412 and #1413 

Also switching from using `extern` to `FOUNDATION_EXTERN` on the very odd chance someone is using objective-c++, better to be compatible than sorry.


bhariharan:
:+1: Thanks!

PR #1427, opened 2016-02-29
Encode (serialize) the access and refresh tokens in the user account file

pbenschop:
Encode (serialize) the access and refresh tokens in the user account plist file to prevent repeated logins after upgrading from SDK 3.3.1 to SDK 4.1. Note that this repeated login problem does not occur if logging in "fresh" after a logout (the problem only occurs if logging in with an App such as Account Editor on SDK 3.3.1 first).


bhariharan:
@pbenschop Could you send us a pull request against the ```unstable``` branch? Thanks!

pbenschop:
No problem, please see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1428

PR #1428, opened 2016-02-29
Encode (serialize) the access and refresh tokens in the user account

pbenschop:
Encode (serialize) the access and refresh tokens in the user account plist file to prevent repeated logins after upgrading from SDK 3.3.1 to SDK 4.1. Note that this repeated login problem does not occur if logging in "fresh" after a logout (the problem only occurs if logging in with an App such as Account Editor on SDK 3.3.1 first).


bhariharan:
Thanks!

khawkins:
Nonononono.  This is a security hole, storing credentials on the filesystem in the clear.  This needs to be reverted.

khawkins:
The access and refresh tokens are stored in the keychain, in their own secure persistence.  Whatever problem this is attempting to solve, it needs to be reconsidered in the context of managing the OAuth credentials securely.

pbenschop:
Hi Kevin, when I stepped through the code it seemed to me these fields stored the keys to retrieve the tokens from the keychain instead of the actual token values. Did I misunderstand that?

khawkins:
The keys you're referring to are the encryption key components that will encrypt the access and refresh token values in the keychain.  They're all stored in the keychain.  The serialization and deserialization of the OAuth credentials is based on a keychain persistence model.

PR #1432, opened 2016-03-03
UI Test

wmathurin:
A singe test at this point  (a POC really)
The test does the following (through the UI):
- switch to sandbox
- logs in
- create three records 
- search for each record and open the detail screen for each one
- logs out


shared/test/test/SalesforceTestCase.swift
```diff
@@ -0,0 +1,51 @@
...
+        XCUIApplication().launch()
+        loginThroughUI()
+    }
+    
+    override func tearDown() {
+        super.tearDown()
+    }
+    
+    func loginThroughUI() {
+        loginDelegate.loginToSalesforce("wolf@cs1.mobilesdk.ee.org", password:"test123456", host:Host.sandbox)
```
wmathurin: I will move this out to a file similar to test_credentials.json

shared/test/pagehelpers/login/LoginHelper.swift
```diff
@@ -0,0 +1,59 @@
...
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+import Foundation
+import XCTest
+
+enum Host {
+    case production
```
bhariharan: Should we throw in a custom host here?

shared/test/test/SalesforceTestCase.swift
bhariharan: Let's use the standard test user. At least we won't be exposing admin privileges ;-)

bhariharan:
Looks good to me otherwise. :+1:

shared/test/test/SalesforceTestCase.swift
wmathurin: I can't remember the username / password. Could you email them to me?

bhariharan: I sent a reset password to your e-mail from that account.

PR #1434, opened 2016-03-03
Don't strip symbols for debug build and add bitcode flag for SalesforceKit-iOS-Aggregate

bbirman:


bhariharan:
@bbirman What does this do?

bbirman:
@bhariharan 
1) It lets us build apps with bitcode enabled when SalesforceKit.framework is included (right now we have to turn bitcode off on the app because otherwise building the .ipa errors on SalesforceKit)
2) For debug builds only, it includes the symbols so that we can debug from the app (that was already the case at the SalesforceKit build level, but they were being stripped when building the aggregate SalesforceKit target)

bhariharan:
:+1:

PR #1446, opened 2016-03-07
First round of reference edits

rwhitley:
First pass at a few classes. I'm keeping a list of content I've added that needs review. I'll send that out later for review in email or Gdocs.


bhariharan:
:+1:

PR #1456, opened 2016-03-16
The SFOAuthCredentials refactor missed the upgrade case.

khawkins:
Since `SFOAuthCredentials` became a base class in the 4.x timeframe, the upgrade case from < 4.x to 4.x would not deserialize the credentials properly.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
@@ -68,8 +68,13 @@ NSException * SFOAuthInvalidIdentifierException() {
 
 - (id)initWithCoder:(NSCoder *)coder {
     NSString *clusterClassName = [coder decodeObjectOfClass:[NSString class] forKey:kSFOAuthClusterImplementationKey];
-    Class clusterClass = NSClassFromString(clusterClassName) ?: self.class;
+    if (clusterClassName.length == 0) {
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -68,8 +68,13 @@ @implementation SFOAuthCredentials
 
 - (id)initWithCoder:(NSCoder *)coder {
     NSString *clusterClassName = [coder decodeObjectOfClass:[NSString class] forKey:kSFOAuthClusterImplementationKey];
-    Class clusterClass = NSClassFromString(clusterClassName) ?: self.class;
+    if (clusterClassName.length == 0) {
```
kchitalia: Will there ever be a case when clusterClasName is null?

khawkins: I'm taking advantage of the fact that messaging `length` from `nil` will return 0, to cover both cases.  It will most likely be `nil` in the case I'm coding for: a persisted OAuth credentials class that was persisted before we kept track of an implementation class in the persistence.  Is that what you're asking?

kchitalia: Yep. I think you got it covered :)

bhariharan:
:+1:

wmathurin:
:+1:

Issue #1430, opened 2016-03-03
SFRestRequest queryParams deosn't allow NSDictionary<AnyObject,AnyObject>  anymore


oolufemi:
This is a follow up to #1218 changes.

Prior to 4.1.1 I was able to call SFRestRequest with just NSDictionary with (AnyObject,AnyObject) but this has now been changed to NSDictionary with (String,String).

Specifically lines in SFRestRequest.m:
Line 115 -@property (nonatomic, strong) NSDictionary *queryParams;
Line 116 +@property (nullable, nonatomic, strong) NSDictionary<NSString*, NSString*> *queryParams;

We have defined our own rest interface methods and some of them have sub data. I.e ObjectA.ObjectB.FieldA. This doesn't work anymore. I want to know if this is the right approach. Or should I really be creating my own rest class using IOS's NSURLRequest and NSURLConnection classes?

wmathurin:
When did you need non-string query parameters?

oolufemi:
Thanks for looking into this.

Lets say I have a data structure defined for example as 
class Person
{
string id;
string name;
Type type;
}
class Type
{
string id
}

Then I define a salesforce doPost defined as
@HttpPost global static String doPost(Person person) {
return null;
}

The resultant JSON would be something like {"id":"P001","name":"foo boo", type:{"id":"T001"}}

We have defined a data structure with a lot of these type of coding and it worked before.

wmathurin:
Fixed in unstable - see referenced pull request above.

PR #1457, opened 2016-03-16
Query params map can have non-string values.

wmathurin:
Query params property is also used for post / patch requests: in which case, it gets serialized to a string using [SFJsonUtils JSONRepresentation:self.queryParams] and ends up in the body.

Addressing regressions reported here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1430


khawkins:
Seems reasonable. :+1:

bhariharan:
:+1:

kchitalia:
:+1:

PR #1458, opened 2016-03-17
Reverting changes from PR 1428

khawkins:
This is in anticipation of a more appropriate fix for the issue of users being logged out between upgrades from Mobile SDK 3.3 to 4.x.


wmathurin:
:+1:

PR #1461, opened 2016-03-21
Use main bundle as back up bundle for image

qingqingliu:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
```diff
@@ -54,8 +54,15 @@ + (UIImage *)imageNamed:(NSString *)name
...
     NSBundle *bundle = [NSBundle bundleForClass:[self class]];
     NSAssert(name != nil, @"name must contain a value.");
+    UIImage *image = [UIImage imageNamed:name inBundle:bundle compatibleWithTraitCollection:nil];
+    if (image) {
+        return image;
+    }
     
-    return [UIImage imageNamed:name inBundle:bundle compatibleWithTraitCollection:nil];
+    // get from main bundle
+    bundle = [NSBundle mainBundle];
```
qingqingliu: if image is not found in sdk bundle, use main bundle as the backup option

kchitalia:
Looks good to me. If I understand this correctly, its just a temporary work around, right?

qingqingliu:
@kchitalia, kind of. As I mentioned in the comments above, this fix would not needed to be removed either even after the bundle issue is fixed. Also, I would recommend you to work with Mike closely to see how MobileSDK can allow consumer of mobile sdk to override those resource (i.e. get from main bundle first and if not found there, get from SalesforceKit instead)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
khawkins: I'm not clear which use case this will support, above and beyond the code that exists.  `SFSDKResourceUtils` is meant to derive its resources from whatever container is most appropriate for *SalesforceSDKCore*.  Which means, in the framework scenario, SalesforceSDKCore.framework, and in the static library scenario, the main bundle.

It's current coding supports those two use cases.  Is there a third case we're not considering?

qingqingliu: @khawkins, this is because a bug we ran into when consuming Salesforce Mobile SDK using SalesforceKit instead. (which is what Service SDK does). Currently SalesforceKit does not contain images & localized strings properly and Kunal will fix it for 4.2 In between, we need to fix it so that when SalesforceKit is consumed, it can use images and localized string from main bundle as back up

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
```diff
@@ -54,8 +54,15 @@ + (UIImage *)imageNamed:(NSString *)name
 {
     NSBundle *bundle = [NSBundle bundleForClass:[self class]];
     NSAssert(name != nil, @"name must contain a value.");
+    UIImage *image = [UIImage imageNamed:name inBundle:bundle compatibleWithTraitCollection:nil];
+    if (image) {
+        return image;
+    }
     
-    return [UIImage imageNamed:name inBundle:bundle compatibleWithTraitCollection:nil];
```
qingqingliu: @khawkins, before this change, SFSDKResourceUtils is not supporting loading image from main bundle

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKResourceUtils.m
khawkins: Ah, that's the third case I was missing: the umbrella framework.  I would consider this fix temporary, then, until the resources are properly aligned for the umbrella framework?

qingqingliu: right @khawkins

kchitalia:
@qingqingliu I have set up time with Mike to discuss how Mobile SDK can allow consumers of Mobile SDK to override resources.

khawkins:
Cool, thanks for the clarification.  :+1:

bhariharan:
Seems reasonable until we fix ```SalesforceKit```. :+1:

PR #1467, opened 2016-03-28
True weak object collection support

khawkins:
Something that's been bugging me for a while.  `[NSValue valueWithNonretainedObject:]`, which we use to hold "weak" values in our multicast delegate support, does not actually provide weak values, i.e. values that go to `nil` when their reference count is exhausted.  It holds Unsafe Unretained values, which end up pointing at garbage when their reference count is exhausted.  This can lead to indeterminate behavior, including crashes, when enumerating these objects after they no longer exist.

This PR creates a container object for weak-referenced objects, and a category on `NSOrderedSet` to manipulate them in our delegate support.  That way, if/when the strong counterparts to delegate objects go away, enumerating these delegates will simply result in a no-op.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
@@ -59,6 +59,14 @@
 		4F7EB4AD1BFFCF2300768720 /* ViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F7EB4A41BFFCEF600768720 /* ViewController.m */; };
 		4FE5332A1BFFE70600814D2A /* Main_iPad.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 4F7EB4981BFFCEF600768720 /* Main_iPad.storyboard */; };
 		4FE5332B1BFFE70600814D2A /* Main_iPhone.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 4F7EB49A1BFFCEF600768720 /* Main_iPhone.storyboard */; };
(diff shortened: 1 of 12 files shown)


wmathurin:
Looks fine.

Couldn't you have a used a NSMapTable with weak keys where the keys would be the objects and the values would some dummy object.
(The equivalent of using a WeakHashMap<Object, Boolean> in Java).

https://developer.apple.com/library/ios/documentation/Cocoa/Reference/NSMapTable_class/

bhariharan:
:+1:

khawkins:
Or better yet, `NSHashTable`, which is the `NSSet` counterpart.  Alas, I hadn't revisited some of the newer collection containers before I started implementing.  Let me take a look at that solution and see if it fits.  Thanks for the pointer!

rwhitley:
Thanks for bringing this up. I just today changed the Mobile SDK trail
coding exercise to use [NSValue valueWithNonretainedObject:]. 🙄

On Mon, Mar 28, 2016 at 2:48 PM, Kevin Hawkins <notifications@github.com>
wrote:

> Or better yet, NSHashTable, which is the NSSet counterpart. Alas, I
> hadn't revisited some of the newer collection containers before I started
> implementing. Let me take a look at that solution and see if it fits.
> Thanks for the pointer!
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1467#issuecomment-202593662>
>



-- 
Richard

khawkins:
Closed in favor of a [new PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1640) with Cocoa's weak container classes utilized.

PR #1469, opened 2016-03-30
add nullability marker

qingqingliu:
@bhariharan @khawkins, adding nullability markers. Ran into an issue where customAttributes was missing the marker and then is assumed nonnull when consumed by Swift. So when it is nil, it is causing the Swift app to crash unless additional checking is added. (However end user might not know to add the nil checking if something is marked as nonnull)

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.h
@@ -33,12 +33,12 @@
 /**
  * The NSDictionary representation of this identity data.
  */
-@property (strong, nonatomic, readonly) NSDictionary *dictRepresentation;
+@property (nonnull, strong, nonatomic, readonly) NSDictionary *dictRepresentation;
 
 /**
  * The ID URL.
  */
-@property (strong, nonatomic, readonly) NSURL *idUrl;
+@property (nonnull, strong, nonatomic, readonly) NSURL *idUrl;
 
 /**
  * Whether or not this is the asserted user for this session.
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.m
```diff
@@ -83,11 +83,6 @@ - (id)initWithJsonDict:(NSDictionary *)jsonDict
     return self;
 }
 
-- (void)dealloc
-{
-    self.dictRepresentation = nil;
-}
```
qingqingliu: this is not needed anymore with ARC

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.h
```diff
@@ -220,22 +220,22 @@
...
  */
-@property (strong, nonatomic, readonly) NSDate *lastModifiedDate;
+@property (nonnull, strong, nonatomic, readonly) NSDate *lastModifiedDate;
 
 /**
  * Designated initializer for creating an instance of the SFIdentityData object.
  * @param jsonDict The JSON dictionary containing the user data.
  */
-- (id)initWithJsonDict:(NSDictionary *)jsonDict;
+- (nonnull instancetype)initWithJsonDict:(nonnull NSDictionary *)jsonDict;
```
qingqingliu: changed to use the new style of declare init (i.e. use instancetype instead of id)

khawkins:
Looks fine to me.  Thanks @qingqingliu !

wmathurin:
:+1:

bhariharan:
:+1:

PR #1472, opened 2016-04-01
Final (?) batch of SalesforceSDKCore edits

rwhitley:
Actually I found a few that I missed, so there'll be one more batch for SalesforceSDKCore.


wmathurin:
:+1:

PR #1476, opened 2016-04-04
Fix header visibility issues

qingqingliu:
@khawkins @bhariharan @wmathurin 

Fix public header visibility settings on public headers for static target as well. Those files are marked as public only on dynamic library target and not on the static target


bhariharan:
:+1:

PR #1477, opened 2016-04-05
Updates to the LLVM module map to account for unannounced changes in Xcode 7.3

MikeNachbaur:
Xcode 7.3 made changes to how LLVM manages library dependencies, so these changes are made to allow backwards compatibility to earlier versions of Xcode, while still working against Xcode 7.3.


bhariharan:
:+1:

PR #1479, opened 2016-04-05
Add SFApplicationHelper.h to public headers in podspec.

kchitalia:


rgoldiez:
@kchitalia - I pulled in this fix and got past my original issue, but running an issue with this file now (perhaps it's also relatively new)?

#import <SalesforceSDKCore/SFSDKLoginHost.h>

kchitalia:
@rgoldiez I'm sorry, which file?

bhariharan:
:+1:

wmathurin:
Looks like SFSDKLoginHost.h also need to be made public in the podspec.

rgoldiez:
This one too... i'm working thru the podspec also.

```
#import <SalesforceSDKCore/SFSDKLoginHostDelegate.h>
```

kchitalia:
@wmathurin I thought we had added all the login host related headers to the podspec.

kchitalia:
Check this out: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/b196de6effde6306b900ee4bfdd66f1a3ebbbf30

bhariharan:
@kchitalia That commit adds it only to the static target, not the podspec.

wmathurin:
That makes them public in the xcode project. Cocoapod relies exclusively on the podspec, so whatever is made public in the xcode project needs to also be made public in the podspec.
We have a script that keeps both in sync for all our libraries except for SalesforceSDK, where we need to do it manually.

rgoldiez:
Good news... I updated my podspec with the all of the LoginHost references and I'm able to build again... and I can now hide the login navbar.  I sincerely appreciate all the help.

kchitalia:
@wmathurin Ah, I see what you mean.

PR #1481, opened 2016-04-06
Add login host headers to podspec

kchitalia:


wmathurin:
:+1: 

I think have a new script to automatically update that list for the SalesforceSDKCore.podspec. We are bound to forget to update that list in the future.

khawkins:
@wmathurin I definitely agree, but it will require a different approach to the current script.  Right now, the script updates public header groups based on the public headers of the built library.  This works great in a "binary" header visibility paradigm—i.e. public headers and non-public headers.

But SalesforceSDKCore is necessarily built into three groups of public header files, to facilitate the non-ARC `SFKeychainItemWrapper` class in the podspec.  To organize the headers in this fashion requires more metadata to describe the header groupings than the build process is going to have.

We could create such a metadata mapping out of band, and let the script rely on that additional information.  Or, we could have the script parse the existing header groupings before overwriting them, and dump any new public headers into the "everything else" group (see the following).  Of the three groups, one is exclusively for `SFKeychainItemWrapper`, and another for its dependencies "below".  The third is "everything else". The chance of new header files falling into the first two groups is rare.

That's probably the most straightforward solution to get the existing script back online.

bhariharan:
:+1:

PR #1484, opened 2016-04-08
Fix Salesforce color extension

qingqingliu:
@bhariharan @wmathurin @khawkins @kchitalia 

Every time a new class gets added to the dynamic target, we need to also make sure we add to static target otherwise SalesforceKit will be broken. Can we pay special attention in the future for any public class added to MobileSDK when it is added to dynamic target? Until we make a decision on SalesforceKit, we have to maintain the parity.


MikeNachbaur:
:+1: Looks good.  We should really have a single target that the code is compiled in, and the dynamic framework should be based on that.  @kchitalia we've spoken about this in the past, so situations like this don't come up.

bhariharan:
:+1:

PR #1485, opened 2016-04-08
Fix SalesforceKit SQLCipher & bitcode issues

MikeNachbaur:
Also there were issues with bitcode signing of the libsqlcipher.a binary, which is why there's a new binary in this PR.  I used the included build script, but updated it to ensure bitcode was embedded.  I haven't made those changes to the target repo/branch however.


bhariharan:
:+1:

PR #1487, opened 2016-04-13
Fix for host added twice if present in list and info.plist

kchitalia:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
```diff
@@ -64,6 +64,11 @@ - (id)init {
     if (self) {
         self.loginHostList = [NSMutableArray array];
         
+        // Add the Production and Sandbox login hosts are defined. These two items cannot be deleted.
```
bhariharan: These will be replaced by runtime configuration if set by MDM, correct?

kchitalia:
Yes, that is correct. Line 77 will clear the list if hosts are enforced by MDM.

bhariharan:
:+1:

PR #1492, opened 2016-04-15
Exposing dataBuffer to allow client to gain access to the underlying …

dbreese:
…encrypted byte buffer when using InMemory encryption/decryption.

@bhariharan @qingqingliu


dbreese:
Going to close this because @jhneves ' PR (https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1493) will give us the capability to stream from file->nsdata too and then we don't have to expose a mutable data object which could be risky and scary.

PR #1498, opened 2016-04-19
Make SFCryptChunks compatible with SFCrypto

jhneves:
SFCrypto fills/truncate the key passed if the size is not correct. Applying the same logic in SFCryptChunks will guarantee interchangeability between SFCrypto and SFCryptChunks (SFEncryptStream and SFDecryptStream too).

Files encrypted/decrypted with SFCrypto can be encrypted/decrypted by SFCryptChunks and vice-versa.

@bhariharan @dbreese


bhariharan:
👍

PR #1501, opened 2016-04-21
Adding remote ghost records sync API

bhariharan:
Only outstanding item is adding tests.


bhariharan:
The implementation follows the same design as the ```Android``` side.

kchitalia:
One question, do we need to somehow show this functionality in the sample app?

bhariharan:
Not at this time. It's a simple API call. I'll be adding doc for the new API.

kchitalia:
Sounds good :+1:

bhariharan:
Added tests in latest commit.

PR #1502, opened 2016-04-25
Adding hybrid implementation of 'cleanReSyncGhosts' API

bhariharan:
```JS``` implementation coming up.

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFSmartSyncPlugin/SFSmartSyncPlugin.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFSmartSyncPlugin/SFSmartSyncPlugin.m
@@ -1,5 +1,5 @@
 /*
- Copyright (c) 2014, salesforce.com, inc. All rights reserved.
+ Copyright (c) 2014-present, salesforce.com, inc. All rights reserved.
  
  Redistribution and use of this software in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFSmartSyncPlugin/SFSmartSyncPlugin.m [hunk 1]
```diff
@@ -52,26 +51,25 @@ @interface SFSmartSyncPlugin ()
...
 }
 
-- (SFSmartSyncSyncManager *)globalSyncManager
+- (SFSmartSyncSyncManager *) globalSyncManager
 {
     return [SFSmartSyncSyncManager sharedInstanceForStore:[SFSmartStore sharedGlobalStoreWithName:kDefaultSmartStoreName]];
 }
 
-- (void)resetSyncManager
+- (void) resetSyncManager
```
kchitalia: Since this method is used for unit testing purposes only, should we move this to the test classes?

bhariharan: We can't. We need it exposed via the plugin. Also, it's not purely for unit testing. It can be used to reset the instance of ```SyncManager``` too.

kchitalia: Sounds good.

kchitalia:
:+1:

PR #1513, opened 2016-05-03
Remove unnecessary finalizeCrypt call when closing decrypt stream

jhneves:
This is not a big deal, but if the stream is closed before finished reading, it could potentially cause a crash. Also this call is unnecessary.
@bhariharan


wmathurin:
👍

bhariharan:
👍

PR #1516, opened 2016-05-06
infinite redirect loop fix

### LINKED ISSUE #1515: Infinite redirect loop when using remote hybrid configuration
Steps to reproduce:
1. Configure an hybrid remote app.
2. Start the application and do the oauth flow.
3. Stop the application.
4. Let the session expire (or manually revoke the session from Salesforce).
5. Start the application again.
When you do the last step, the SDK tries to use the sid previously stored to load the page through frontdoor.jsp. Then, a login redirect is catched because the sid is not valid anymore. This causes a refresh of the token, but after this step a new frontdoor.jsp url is generated. The retURL of this new url is the previous frontdoor.jsp url. This bug starts an infinite "frontdoor with valid sid" -> "frontdoor with invalid sid" -> token refresh loop

micheletriaca:
I've fixed the problem with this patch. resolve #1515


wmathurin:
Please resubmit against unstable (active development branch). We only update master when we do a release.

PR #1518, opened 2016-05-06
make app valid for iPad only

huminzhi:


wmathurin:
Please resubmit against unstable (active development branch).
We only update master when we do a release.

PR #1528, opened 2016-05-11
Fix clean up issue with sync tests

huminzhi:


sgoldberg-sfdc:
Change to +setup looks good and will enable that suite to pass when executed as part of the Everything target.

reporter.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/python
```
kchitalia: We might not need this. I am trying to see if using grep will suffice. Hang tight :)

wmathurin: You could simply do
```shell
cat output | sed 's/.*write-file.*/skipped/'
```

kchitalia:
👍

bhariharan:
👍

PR #1531, opened 2016-05-12
Change to Travis Script and write debug logs only if we build for Debug config

kchitalia:


kchitalia:
@huminzhi I am checking in the script for travis. We can look at filtering the logs further.

.travis.yml
```diff
@@ -4,14 +4,14 @@ branches:
      - unstable
 
 language: objective-c
-osx_image: xcode7
+osx_image: xcode7.2
```
bhariharan: We should probably use the latest (```7.3```).

.travis.yml
```diff
@@ -4,14 +4,14 @@ branches:
...
 
 env:
   matrix:
-    - SDK=iphonesimulator OS=8.4 NAME='iphone 6'
-    - SDK=iphonesimulator9.1 OS=9.1 NAME='iphone 6'
-    - SDK=iphonesimulator OS=8.4 NAME='ipad 2'
-    - SDK=iphonesimulator9.1 OS=9.1 NAME='ipad 2'
+    - SDK=iphonesimulator OS=8.4 NAME="iPhone 6"
+    - SDK=iphonesimulator9.2 OS=9.2 NAME="iPhone 6"
```
bhariharan: Should probably use the latest here too (```9.3```).

.travis.yml
kchitalia: There is a bug in travis CI that causes the build and tests to freeze if xcode version is 7.3 and iOS version is 9.3. Till the bug is fixed we'll have to downgrade to 7.2 and iOS 9.2

.travis.yml
kchitalia: Same here.

libs/SalesforceRestAPI/SalesforceRestAPITests/SalesforceRestAPITests.m
```diff
@@ -45,7 +45,11 @@ @implementation SalesforceRestAPITests
 + (void)setUp
 {
     @try {
-        [SFLogger setLogLevel:SFLogLevelDebug];
+        #if defined(DEBUG)
```
bhariharan: Don't we always want log output from tests?

kchitalia: Not if we are running them is release mode, right?

bhariharan: Release binaries are never built for the test apps or test targets AFAIK. I don't think we need these conditionals in the test classes.

kchitalia:
@bhariharan Removed the #if def from test classes.

bhariharan:
👍

PR #1535, opened 2016-05-16
fix MockNetworkTests

huminzhi:
@kchitalia @bhariharan test fix

### DIFF
--- a/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetworkTests/MockNetworkTests.m
+++ b/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetworkTests/MockNetworkTests.m
@@ -99,6 +99,7 @@
     XCTAssertFalse(network.refreshingAccessToken);
     XCTAssertNil(network.defaultConnectCommunityId);
     XCTAssertFalse(network.networkSuspended);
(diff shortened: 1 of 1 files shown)


bhariharan:
👍

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetworkTests/MockNetworkTests.m
```diff
@@ -99,22 +99,26 @@ - (void)testNetworkBasics {
...
     CSFSalesforceAction *action1 = [[CSFSalesforceAction alloc] initWithResponseBlock:nil];
     CSFSalesforceAction *action2 = [[CSFSalesforceAction alloc] initWithResponseBlock:nil];
     
     // Setting verbs and methods (values do not matter - but can't be nil otherwise it confuses duplicateActionInFlight/isEqualToAction
     action1.method = action2.method = @"GET";
     action1.verb = action2.verb = @"/xyz";
-
+    
     [network executeAction:action1];
+    [NSThread sleepForTimeInterval:1];
```
huminzhi: the network count is increased in a different thread, so without wait, it will always fail...  also need to suspend the network (line 102), otherwise, the action will just get done and removed from queue..

Issue #1510, opened 2016-04-28
Issue with [[SFRestAPI sharedInstance] sendRESTRequest:fileContentRequest failBlock:^(NSError *e) always fail


hiralnaik:
Hello there

We have an app with salesforce SDK. Recently We upgraded to Salesforce SDK from 4.0.0 to 4.1.1. After upgrading, we are having issue with download any file from the Salesforce. It used to work with previous Salesforce iOS SDK. We didn't change any code in our classes. Can you please help me with it ? or Did I miss anything in my code ?

Here is the method name that We used to download.
[SFRestAPI sharedInstance] sendRESTRequest:fileContentRequest failBlock:^(NSError *e).

Here is my code.
```
SFRestRequest *fileContentRequest = [[SFRestAPI sharedInstance] requestForFileContents:Id version:nil];
        [[SFRestAPI sharedInstance] sendRESTRequest:fileContentRequest failBlock:^(NSError *e)
         {
         }
         completeBlock:^(NSData *fileResponseData)
         {
         }];
```

**Error**:
`Error Domain=CSFNetworkErrorDomain Code=404 "The requested resource does not exist" UserInfo={action=<SFRestAPISalesforceAction: 0x7d83b6c0, GET "/chatter/files/XXXXXXXXXXXXX/content" (0.00)>, NSLocalizedFailureReason=NOT_FOUND, NSLocalizedDescription=The requested resource does not exist, isAuthenticationFailure=false}`

Thanks for your help.

wmathurin:
In 4.1.1, we change the default api version to v36.0 and it looks like the files end point changed (see https://developer.salesforce.com/docs/atlas.en-us.chatterapi.meta/chatterapi/connect_resources_files_content.htm).
If you change apiVersion on your SFRestAPI instance to v35.0, it should work again.

hiralnaik:
Yes I changed kSFRestDefaultAPIVersion constant to "v35.0" and it is working fine. So what will be the option here since I can't modify the SDK ?

wmathurin:
You should be able to do [SFRestAPI sharedInstance].apiVersion = @"v35.0"

wmathurin:
We'll fix the SDK for our next point release.

hiralnaik:
Thanks. I will do. Do you know when are you guys planning to release next sdk ?

wmathurin:
We are shooting for June.

hiralnaik:
[...]

PR #1536, opened 2016-05-16
Unstable

huminzhi:
@bhariharan @kchitalia @wmathurin @helenren FYI.

### DIFF
--- a/libs/SalesforceRestAPI/SalesforceRestAPI/Classes/SFRestAPI+Files.m
+++ b/libs/SalesforceRestAPI/SalesforceRestAPI/Classes/SFRestAPI+Files.m
@@ -40,28 +40,28 @@
 @implementation SFRestAPI (Files)
 
 - (SFRestRequest *) requestForOwnedFilesList:(NSString *)userId page:(NSUInteger)page {
-    NSString *path = [NSString stringWithFormat:@"/%@/chatter/users/%@/files", self.apiVersion, (userId == nil ? ME : userId)];
+    NSString *path = [NSString stringWithFormat:@"/%@/connect/files/users/%@", self.apiVersion, (userId == nil ? ME : userId)];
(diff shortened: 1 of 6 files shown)


kchitalia:
Is this related to the CordovaPlugin issue?

huminzhi:
@kchitalia sorry not yet..

kchitalia:
Ok, I got confused, sorry :). Looks like this will resolve https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1510

👍

bhariharan:
👍

libs/SalesforceRestAPI/SalesforceRestAPI/Classes/SFRestAPI+Files.m
```diff
@@ -40,41 +40,41 @@
 @implementation SFRestAPI (Files)
 
 - (SFRestRequest *) requestForOwnedFilesList:(NSString *)userId page:(NSUInteger)page {
-    NSString *path = [NSString stringWithFormat:@"/%@/chatter/users/%@/files", self.apiVersion, (userId == nil ? ME : userId)];
```
wmathurin: Should we do (api >36 ? @"/connect/files/..." : @"/chatter/users/...")?

bhariharan: Default is ```v36.0```, so we should be okay. There's no reason for users to downgrade API version on new SDK. Typically internal consumers upgrade to pre-rel.

wmathurin: That way the URL would be correct in all cases

bhariharan: I'll fix it.

Issue #883, opened 2015-01-14
Some classes are incorrectly importing UIKit when Foundation is more appropriate


NachoMan:
When trying to adapt some of the Salesforce Mobile SDK libraries to be cross-platform between iOS and Mac OS X, I discovered a number of classes are importing UIKit/UIKit.h even though it doesn't use any UIKit-specific features within those classes.  Instead they should only import Foundation/Foundation.h or another applicable framework to ease in porting these classes between platforms.

kchitalia:
This issue has been fixed.

PR #1537, opened 2016-05-16
Replace UIKit with Foundation.h where appropriate

kchitalia:
This is to take care of [this](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/883) issue. Will send in more PRs for other projects.


bhariharan:
👍

PR #1539, opened 2016-05-17
Fixing compilation error

bhariharan:


bhariharan:
@kchitalia This class requires ```UIKit```, because it uses ```UIImage```. ```SalesforceSDKCore``` won't compile without it.

kchitalia:
Thats interesting, I am able to build fine with UIKit removed. The tests are running fine too and sample app runs fine as well.

bhariharan:
Did you wipe ```DerivedData``` and build the ```Everything``` scheme?

PR #1540, opened 2016-05-17
Fixing another compilation error

bhariharan:
Something changed in ```Xcode 7.1``` and greater. Header search paths should now be empty for frameworks to work properly with ```Swift```. Everything compiles fine now.


native/SampleApps/SmartSyncExplorer/SmartSyncExplorer.xcodeproj/project.pbxproj
```diff
@@ -1212,6 +1212,7 @@
 				DEBUG_INFORMATION_FORMAT = dwarf;
 				ENABLE_STRICT_OBJC_MSGSEND = YES;
 				GCC_NO_COMMON_BLOCKS = YES;
+				HEADER_SEARCH_PATHS = "";
```
bhariharan: This is only for the test target.

PR #1541, opened 2016-05-17
IPv6 Compatible Version of Reachability

bhariharan:
Originally from ```Apple``` from [here](https://developer.apple.com/library/ios/samplecode/Reachability/Listings/Reachability_Reachability_h.html#//apple_ref/doc/uid/DTS40007324-Reachability_Reachability_h-DontLinkElementID_8).


wmathurin:
Looks fine.

PR #1547, opened 2016-05-18
Soup table will now have db indexes on created and last modified

wmathurin:
To get them for an existing soup, simply alter the soup passing in the original set of index specs


bhariharan:
Do we need any upgrade steps for existing soups?

👍

wmathurin:
No upgrade. if they care about the extra index, they can alterSoup passing in the original set of index specs. It should be pretty quick because it will all happen in SQL. There is a test for it.

bhariharan:
Sounds good. 👍

PR #1554, opened 2016-05-20
Adding error handling for MDM corner case

bhariharan:


bhariharan:
If the property to show only allowed hosts is set, but no hosts are set, we shouldn't fall back on ```Production``` and ```Sandbox```.

sgoldberg-sfdc:
LGTM

PR #1557, opened 2016-05-24
Added improved logging support and unit tests

MikeNachbaur:
This is meant to be able to use a centralized logging facility for multiple consumers of a shared Mobile SDK.  The idea is that different areas can include their own logging statements against a unique identifier, and log levels can be set to different thresholds for different areas.  For example, the log level could be cranked up for SmartStore, while other areas can be kept low.

Furthermore, since these identifiers are defined by opaque strings as opposed to enums, these can be extended by multiple consumers, including the app developer themselves.  And by using a reverse-DNS naming convention, you can set the log level for multiple identifiers based on a common prefix.

- Introduce the concept of a "Logging Identifier" to scope log statements, and set different log levels, for different areas of logic
- Begin migrating to logging macros to take advantage of CocoaLumberjack's performance capabilities (e.g. levels & flag bit masks)
- Replace and deprecate `context` methods, since they weren't useful
- Begin migrating methods over to instances methods, instead of class methods, to improve testability
- Update to a newer CocoaLumberjack allowing us to store our own private instance of DDLog, to prevent polluting the user's own instance
- Improve testability and support dependency injection on the DDLog instance
- Make logger types configurable
- Utilize a SFLogIdentifier private class, internally, to provide future expansion support for making per-functional-area logging configuration possible.

This pull request is meant to open a discussion, and to add further enhancements as needed.  In particular I would like to hear about the methods I've marked as deprecated, as well as the handling of assertions.


external/shared
```diff
@@ -1 +1 @@
-Subproject commit cce21846693a0774ce6d4f4ebd5f435f90ccbe24
+Subproject commit ac5d7efb6d2ce737bfe1eb0796f548ab28083b17
```
bhariharan: Revert changes to the submodule commit.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogIdentifierFilter.h
```diff
@@ -0,0 +1,13 @@
+//
```
bhariharan: Change copyright to ```Mobile SDK``` copyright. You can copy it from one of the other files.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogIdentifierFilter.m
```diff
@@ -0,0 +1,13 @@
+//
```
bhariharan: Change copyright to ```Mobile SDK``` copyright. You can copy it from one of the other files.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogStorage.h
```diff
@@ -0,0 +1,32 @@
+//
```
bhariharan: Change copyright to ```Mobile SDK``` copyright. You can copy it from one of the other files.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogger.h
```diff
@@ -35,20 +38,12 @@ extern NSString * const kSFLogLevelDebugString;
 extern NSString * const kSFLogLevelInfoString;
 extern NSString * const kSFLogLevelWarningString;
 extern NSString * const kSFLogLevelErrorString;
-
-typedef NS_ENUM(NSUInteger, SFLogLevel) {
```
bhariharan: Where did ```SFLogLevel``` go? Won't public API break if this is missing?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLoggerMacros.h
```diff
@@ -0,0 +1,119 @@
+//
+//  SFLoggerMacros.h
```
bhariharan: Change copyright to ```Mobile SDK``` copyright. You can copy it from one of the other files.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogger_Internal.h
```diff
@@ -0,0 +1,57 @@
+//
+//  SFLogger_Internal.h
```
bhariharan: Change copyright to ```Mobile SDK``` copyright. You can copy it from one of the other files.

libs/SalesforceSDKCore/SalesforceSDKCore/SFLoggerTests.m
```diff
@@ -0,0 +1,686 @@
+//
+//  SFLoggerTests.m
```
bhariharan: Change copyright to ```Mobile SDK``` copyright. You can copy it from one of the other files.

libs/SalesforceSDKCore/SalesforceSDKCore/SFLoggerTests.m
```diff
@@ -0,0 +1,686 @@
...
+//  SalesforceSDKCore
+//
```
bhariharan: Move this file under ```SalesforceSDKCoreTests``` folder and group.

kchitalia:
The intention here is to have consumers of the SDK use this logging mechanism as well?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogIdentifierFilter.h
MikeNachbaur: Wow, rookie mistakes. :-)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogger.h
MikeNachbaur: It's moved to SFLoggerMacros.h

MikeNachbaur:
@kchitalia Correct-ish.  The biggest goal of this is for other SDKs, that are built atop the MobileSDK, to be able to use it for its logging, as well as the MobileSDK itself.  For example, it could use logging identifiers such as `com.salesforce.MobileSDK.SmartStore`, `com.salesforce.MobileSDK.OAuth`, etc.  And then, you could change the logging level for individual components, or for all of the MobileSDK.

It could also be used by consumers as well though, so they could take advantage of logging, log rotation, etc.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFLoggerTests.m
```diff
@@ -0,0 +1,686 @@
+//
+//  SFLoggerTests.m
```
bhariharan: Copyright.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLoggerMacros.h
```diff
@@ -0,0 +1,135 @@
...
+
+#define SF_LOG_MAYBE(async, flg, ctx, tag, fnct, frmt, ...)                        \
+    do {                                                                           \
+        SFLogLevel level = SFLoggerContextLogLevels[MAX(1, ctx) - 1];              \
+        if (flg & level) {                                                         \
+            SF_LOG_MACRO(async, level, flg, ctx, tag, fnct, frmt, ## __VA_ARGS__); \
+        }                                                                          \
+    } while(0)
+
+//static void SF_LOG_MAYBE(BOOL async, SFLogFlag flg, NSUInteger ctx, id tag, const char * fnct, const char * frmt, ...) {
```
bhariharan: Could we get rid of this commented out block of code?

bhariharan:
@MikeNachbaur Looks pretty good. Just a couple of minor comments. Once we update the ```CocoaLumberjack``` pull request and get that merged, we can update the submodule commit here and get this merged.

kchitalia:
@MikeNachbaur If we want other SDKs and consumers to use this, we'll have to make this available via our cocoapod as well. Probably the podspec file needs to include these changes.

MikeNachbaur:
@kchitalia I'd certainly think that would be a good idea, but perhaps we should wait until v2 of these changes.  I'd like to clean up the legacy logging API features before exposing it to customers, since the older methods are a bit of an antipattern.  For the time being, perhaps we could start using this internally, and shake out the API a bit first.

bhariharan:
We do want to maintain parity on public headers exposed in the project being exposed on ```Cocoapods``` as well though.

bhariharan:
👍

PR #1559, opened 2016-05-25
New tests 

wmathurin:
Around the calculated fields for a query spec and smart sql to sql generation
Tests apply to recent changes (json1 / remove by query) but also older code.

One issue found and fixed (table alias for json1 columns in smart sql - same fix will need to be done on Android).
One issue found not yet fixed (joins with json1 columns that are not table/alias qualified - I'll try to fix this one before merging this pull request - the same fix will have to be done on Android).

### DIFF
--- a/libs/SmartStore/SmartStoreTests/SFQuerySpecTests.m
+++ b/libs/SmartStore/SmartStoreTests/SFQuerySpecTests.m
@@ -1,0 +1,146 @@
+/*
+ Copyright (c) 2016, salesforce.com, inc. All rights reserved.
+ 
+ Redistribution and use of this software in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this list of conditions
+ and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice, this list of
+ conditions and the following disclaimer in the documentation and/or other materials provided
+ with the distribution.
+ * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+ endorse or promote products derived from this software without specific prior written
+ permission of salesforce.com, inc.
+ 
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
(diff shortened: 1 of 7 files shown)


libs/SmartStore/SmartStoreTests/SFQuerySpecTests.m
```diff
@@ -0,0 +1,146 @@
...
+
+- (void) setUp
+{
+    [super setUp];
+    [SFLogger setLogLevel:SFLogLevelDebug];
+}
+
+
+#pragma mark - tests
+- (void) testAllQuerySmartSql
```
wmathurin: Since we introduced smart sql years ago, all query specs contain a smart sql field which is used for running the query. There is also a count smart sql field which is used when calling countQuery with the query spec. And with the remove by query story, we introduced a ids smart sql field used when doing a remove entries with the query spec.
All those have been indirectly tested in other places (by actually running queries), but now this test suite, looks at the actual smart sql calculated for various query spec types.

libs/SmartStore/SmartStoreTests/SFSmartSqlTests.m
```diff
@@ -167,6 +173,27 @@ - (void) testConvertSmartSqlWithSpecialColumnsAndSelfJoin
...
+    XCTAssertEqualObjects(@"select json_extract(TABLE_1.soup, '$.education') from TABLE_1 order by json_extract(TABLE_1.soup, '$.education')",
+                          [self.store convertSmartSql:@"select {employees}.{employees:education} from {employees} order by {employees}.{employees:education}"], @"Bad conversion");
+}
+
+- (void) testConvertSmartSqlWithJSON1AndTableAliases
+{
+    XCTAssertEqualObjects(@"select json_extract(e.soup, '$.education'), json_extract(soup, '$.building') from TABLE_1 as e, TABLE_2",
+                          [self.store convertSmartSql:@"select e.{employees:education}, {departments:building} from {employees} as e, {departments}"], @"Bad conversion");
+    
+    // XXX join query with json1 will only run if all the json1 columns are qualified by table or alias
```
wmathurin: If one writes a smart sql joining on json1 fields, because they do json_extract against the soup column which exist in all soup tables, it will not run. The work-around is for the user to do {soup}.{soup:json1path} or alias.{soup:json1path} (assuming the soup table was aliased in the FROM clause). We should look into making the smart sql to sql generation smarter to do json_extract(table.soup, ). It might not be doable without fully parsing the query though.

bhariharan:
👍

PR #1571, opened 2016-05-31
Fixes to console logging, and improve formatter to provide similar information to what NSLog provides

MikeNachbaur:


sgoldberg-sfdc:
Looks good other than some Unit Tests required updates. Thanks for adding process/thread info.

MikeNachbaur:
@sgoldberg-sfdc fortunately the unit tests were actually updated by virtue of that updated regex.  It's matching the proper expressions now, except that I'm not testing for the exact pid / thread name.

bhariharan:
👍

PR #1574, opened 2016-06-01
add back metadata tests

huminzhi:
@kchitalia @helenren @bhariharan @wmathurin Those tests are passing now after I use a new test account particularly for ios unit tests to reduce noise.


bhariharan:
Nice!

PR #1575, opened 2016-06-01
Since FTS query are done through a subquery instead of a join

wmathurin:
we no longer need to qualify the order by column

NB: should have gone with this change: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/95486afffcdc236de4161aa473dad70bc9df2180


bhariharan:
👍

PR #1577, opened 2016-06-02
Compiler warning fixes

MikeNachbaur:
If you use the MobileSDK with additional warnings, it results in a ton of header warnings which can get pretty annoying.  This updates those settings and fixes the issues that Xcode warns about.


bhariharan:
👍 Thanks @MikeNachbaur!

PR #1578, opened 2016-06-02
Don't attempt to capture `self` when all we want to know is the origi…

MikeNachbaur:
…nating class/selector

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogger.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogger.m
@@ -151,11 +151,11 @@ SFLogLevel SFLoggerContextLogLevels[SF_LOG_MAX_IDENTIFIER_COUNT];
 
 @implementation SFLogTag
 
-- (instancetype)initWithClass:(Class)originClass sender:(id)sender {
+- (instancetype)initWithClass:(Class)originClass selector:(SEL)selector {
     self = [self init];
     if (self) {
         _originClass = originClass;
-        _sender = sender;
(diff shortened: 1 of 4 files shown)


sgoldberg-sfdc:
Looks great to me.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogger.m
```diff
@@ -264,7 +264,7 @@ - (void)log:(SFLogLevel)level format:(NSString *)format, ... {
                 file:nil
             function:nil
                 line:0
-                 tag:[[SFLogTag alloc] initWithClass:self.class sender:self]
+                 tag:[[SFLogTag alloc] initWithClass:self.class selector:nil]
```
bhariharan: Should this be ```nil```?

bhariharan: Never mind. Got it.

bhariharan:
👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLogger.m
MikeNachbaur: Yes; we don't know what the selector is within this context.  This is to support the legacy NSObject category methods (which are inherently slower and provide less information than by using the macros).

Essentially we need to know "What's the class originating the log message, and what function / method is calling it?"  In this method here, we don't know that without analyzing the backtrace of the call.  Instead, by using the macros, it's able to glean that additional information properly.

PR #1560, opened 2016-05-25
Fix the bug where if id data is updated on the wrong account

qingqingliu:
@bhariharan @wmathurin review please

Fix a bug where existing flow is doing the following thing instead 

1) Call initiateIdentityDataRetrieval with Account A
2) Before ID data returns, switch to Account B
3) ID data returns for account A
4) Mobile SDK updates Account B with ID data for Account A, which is the bug

PR #1579, opened 2016-06-02
Fix regression caused by #1560

MikeNachbaur:
The bug here is the identity coordinator is comparing a value stored on a singleton instance.  This only works on the assumption that the identity coordinator will always be used with a user account that's the `currentUser` value.  If your SFUserAccount instance isn't assigned as the currentUser, it becomes impossible to load identity data for it.

Instead, this logic should be handled by the SFIdentityCoordinatorDelegate instance receiving updates about this class, and perform the necessary guard checks at that time.

I've discussed this with @qingqingliu and she agrees.


qingqingliu:
👍 , yes as this fix is to unblock service cloud SDK for embedded service and the bug it tries to fix is a real rare case, we can revisit that old bug once that change is in

MikeNachbaur:
Yes, or ensure the consumer of this class performs their own checks as necessary.

bhariharan:
👍

sgoldberg-sfdc:
Wow twice in the same day. I had just commented that out of my integrated version . !

PR #1581, opened 2016-06-03
Add test_force.js to travis CI

kchitalia:
For now adding only the native app types. For hybrid app types, the build will time out on the update.sh script in tools folder for cordova plugin. Will look into more details and add the hybrid app types as well.


wmathurin:
👍

Issue #1552, opened 2016-05-20
Using the app in a foreign Country with Roaming and Wifi turned off not working as expected.


rubenxifre:
We use the iOS Salesforce Mobile SDK to log in our app. This works perfectly. The issue happens when we try to use the app in a foreign country with the Roaming and Wifi turned off. At that point, when launching the app  we get an alert with the message "International roaming is currently off".

Our expected behaviour would be that when using the app in a foreign country with Roaming and Wifi turned off the app would work like if we were offline, which when using the app in our country and the phone has no internet connection the app works well offline. To make the app work well in this situation we have to turn on the Airplane Mode, which should not be needed.

Thanks in advance for the help!

bhariharan:
We don't do anything in the Mobile SDK to restrict network access based on any conditions. You'll need to debug the app to see where it's coming from. This forum is reserved for known bugs in the Mobile SDK. Please use our [Google+ community](https://plus.google.com/u/1/communities/114225252149514546445) for everything else.

rubenxifre:
Hi,

It's not that you restrict network access to the app when in a foreign country, with Roaming turned off, the issue is the flow followed when that happens.

I could work around that by adding the error code NSURLErrorInternationalRoamingOff in the method errorIsNetworkFailure of the SFAuthenticationManager. Then, when the app gets a RoamingOff error it will just proceed with the login. But I am not sure if this solution is robust and reliable. Is there not a proper way to add Error codes for which the app should log in anyways.

Thanks again.

bhariharan:
@rubenxifre I see. Could you send us a pull request against ```unstable```? Thanks!

PR #1584, opened 2016-06-06
Add international roaming as a cause of network failure

rubenxifre:
This was requested in issue: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1552

Roaming is not considered as a cause of network failure.


bhariharan:
👍

PR #1588, opened 2016-06-07
Full-text indexes now using FTS5 

wmathurin:
NB: In the future we should provide a way for developers to use advanced FTS features like selecting the tokenizer (https://www.sqlite.org/fts5.html#section_4_3) or sorting using bm25 (https://www.sqlite.org/fts5.html#section_5_1_1).

Added a way to select FTS4 or FTS5 -- not meant for app but used by tests to make sure no behavior changed.

For a soup created before 4.2, one simply needs to do an alterSoup passing in the original index specs, and the FTS virtual table will be recreated using FTS5 (there is a test for that).


bhariharan:
👍

kchitalia:
👍

PR #1589, opened 2016-06-08
If `kSFDeviceSalesforceId` didn't have a value then the following exc…

mattlacey:
…eption would be thrown when trying to get the device ID as an NSString reference.

Fatal Exception: NSInvalidArgumentException
*** -[NSPlaceholderString initWithString:]: nil argument


kchitalia:
👍

bhariharan:
@mattlacey Could you resubmit against ```unstable``` please?

mattlacey:
Ah yep! Will do that now.

PR #1592, opened 2016-06-09
Sync up failure bug fix

wmathurin:
iOS equivalent of: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1159

- we shouldn't go to the server for locally created records
- but if the app code unchecks the locally_created flag (e.g. after doing an local update or delete), the smart sync code should not fail the sync up
- added tests

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -510,12 +510,16 @@ static NSMutableDictionary *syncMgrList = nil;
     NSMutableDictionary* record = [[self.store retrieveEntries:@[idStr] fromSoup:soupName][0] mutableCopy];
     
     // Do we need to do a create, update or delete
+    BOOL locallyCreated = [record[kSyncManagerLocallyCreated] boolValue];
+    BOOL locallyUpdated = [record[kSyncManagerLocallyUpdated] boolValue];
+    BOOL locallyDeleted = [record[kSyncManagerLocallyDeleted] boolValue];
+    
     SFSyncUpTargetAction action = SFSyncUpTargetActionNone;
-    if ([record[kSyncManagerLocallyDeleted] boolValue])
+    if (locallyDeleted)
         action = SFSyncUpTargetActionDelete;
-    else if ([record[kSyncManagerLocallyCreated] boolValue])
+    else if (locallyCreated)
         action = SFSyncUpTargetActionCreate;
-    else if ([record[kSyncManagerLocallyUpdated] boolValue])
+    else if (locallyUpdated)
         action = SFSyncUpTargetActionUpdate;
     
     if (action == SFSyncUpTargetActionNone) {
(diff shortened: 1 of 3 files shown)


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -530,18 +534,15 @@ - (void)syncUpOneEntry:(SFSyncState*)sync
...
         __weak SFSmartSyncSyncManager *weakSelf = self;
         SFSyncUpRecordModificationResultBlock modificationBlock = ^(NSDate *localDate, NSDate *serverDate, NSError *error) {
             __strong SFSmartSyncSyncManager *strongSelf = weakSelf;
-            if (error) {
-                if (failBlock != NULL) {
-                    failBlock(error);
-                }
-            } else if ([localDate compare:serverDate] != NSOrderedAscending) {
-                // Local date is newer than or the same as the server date.
+            if (localDate == nil // We didn't capture the last modified date so we can't really enforce merge mode
```
wmathurin: The behavior was different that smartsync on android.
We should not fail to sync up if the modification date on the server cannot be established.

libs/SmartSync/SmartSync/Classes/Util/SFSyncUpTarget.m
```diff
@@ -133,10 +133,7 @@ - (void)fetchRecordModificationDates:(NSDictionary *)record
             if (nil != record) {
                 NSString *serverLastModifiedStr = record[self.modificationDateFieldName];
                 if (nil != serverLastModifiedStr) {
-                    NSDate *testServerModifiedDate = [SFSmartSyncObjectUtils getDateFromIsoDateString:serverLastModifiedStr];
-                    if (testServerModifiedDate != nil) {
-                        serverLastModifiedDate = testServerModifiedDate;
-                    }
+                    serverLastModifiedDate = [SFSmartSyncObjectUtils getDateFromIsoDateString:serverLastModifiedStr];
```
wmathurin: Caller can handle nil (see change above).

libs/SmartSync/SmartSync/Classes/Util/SFSyncUpTarget.m
```diff
@@ -123,7 +123,7 @@ - (void)fetchRecordModificationDates:(NSDictionary *)record
     SFRestFailBlock failBlock = ^(NSError *error) {
         [self log:SFLogLevelError format:@"REST request failed with error: %@", error];
         if (modificationResultBlock != NULL) {
-            modificationResultBlock(nil, nil, error);
+            modificationResultBlock(localLastModifiedDate, nil, error);
```
wmathurin: Passing nil for both or just the server date has the same result - but it's better to only pass nil for the ones that could not be obtained.

libs/SmartSync/SmartSyncTests/SyncManagerTests.m
```diff
@@ -714,6 +728,40 @@ - (void)testSyncUpWithLocallyCreatedRecords
...
 
 /**
+ * Create accounts locally, delete them locally, sync up with merge mode SFSyncStateMergeModeLeaveIfChanged, check smartstore
+ *
+ * Ideally an application that deletes locally created records should simply remove them from the smartstore
+ * But if records are kept in the smartstore and are flagged as created and deleted (or just deleted), then
+ * sync up should not throw any error and the records should end up being removed from the smartstore
+ *
+ */
+-(void) testSyncUpWithLocallyCreatedAndDeletedRecords
```
wmathurin: This test fails without the code changes above in the way reported in the bug.

libs/SmartSync/SmartSyncTests/SyncManagerTests.m
```diff
@@ -1074,7 +1123,7 @@ - (void)testCustomSyncUpWithFetchModificationDateFailure
     SFSyncUpTarget *customTarget = [[TestSyncUpTarget alloc] initWithRemoteModDateCompare:TestSyncUpTargetRemoteModDateGreaterThanLocal
                                                                                      sendRemoteModError:YES
                                                                                         sendSyncUpError:NO];
-    [self trySyncUp:3 actualChanges:1 target:customTarget mergeMode:SFSyncStateMergeModeLeaveIfChanged completionStatus:SFSyncStateStatusFailed];
+    [self trySyncUp:3 actualChanges:3 target:customTarget mergeMode:SFSyncStateMergeModeLeaveIfChanged completionStatus:SFSyncStateStatusDone];
```
wmathurin: The behavior was different from android before. Now it won't fail the sync up if the server date cannot be retrieved.

bhariharan:
👍

PR #1597, opened 2016-06-15
convert objc_unretainedPointer to (__bridge void*)

### LINKED ISSUE #1596: SDK fails to compile in XCode 8
Currently the SDK fails to compile in XCode 8 due to usages of `objc_unretainedPointer` which is deprecated and removed, they suggest replacing it with `(__bridge void*)(self)`

mgetzbw:
per fix-it for XCode 8, fixes #1596


bhariharan:
Thanks! 👍

PR #1601, opened 2016-06-17
Warning fixes from Xcode 8 analyzer

mgetzbw:
Fix undefined behavior concerning error handling, if `error` is nil and is dereferenced then undefined behavior occurs.


bhariharan:
👍

PR #1603, opened 2016-06-17
Undefined behavior fixes

mgetzbw:
* Fix issues where ivars were assigned without self being initialized
* Fix issue where nil could be passed to method requiring non-nil parameter resulting in crash


bhariharan:
👍

PR #1606, opened 2016-06-17
Deprecated method updates

mgetzbw:
* All deprecation fixes work on iOS 7 and newer
* stdatomic requires Xcode 7 (which is fine)


bhariharan:
👍

PR #1607, opened 2016-06-17
Enable using NS_STRING_ENUM

mgetzbw:
to make CSFActionKey a strongly typed `enum` in Swift 3, if we have an open ended value we should use `NS_EXTENSIBLE_STRING_ENUM` which wraps the values in a struct not an `enum`

* Define compatibility macro in headers

### DIFF
--- a/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.h
+++ b/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.h
@@ -271,10 +271,12 @@
 
 @end
 
-CSF_EXTERN NSString * const kCSFActionTimingTotalTimeKey;
-CSF_EXTERN NSString * const kCSFActionTimingNetworkTimeKey;
-CSF_EXTERN NSString * const kCSFActionTimingStartDelayKey;
-CSF_EXTERN NSString * const kCSFActionTimingPostProcessingKey;
+typedef NSString*const CSFActionTiming NS_STRING_ENUM;
+
+CSF_EXTERN CSFActionTiming kCSFActionTimingTotalTimeKey;
+CSF_EXTERN CSFActionTiming kCSFActionTimingNetworkTimeKey;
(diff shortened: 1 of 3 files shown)


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.h
```diff
@@ -294,6 +296,6 @@ CSF_EXTERN NSString * const kCSFActionTimingPostProcessingKey;
  
  @return The time interval for the requested key, or `0` if the key is invalid or if the request hasn't gathered enough information to supply that value yet.
  */
-- (NSTimeInterval)intervalForTimingKey:(NSString*)key;
+- (NSTimeInterval)intervalForTimingKey:(CSFActionKey)key;
```
bhariharan: Doesn't the implementation have to be updated as well in ```CSFAction.m```?

mgetzbw: Nope, if this was C++ it might have to be but objective-C is weakly typed and typedef is basically just a macro for type replacement. That enum macro is basically riding on top of that to work. For consistency I can change it there too if you would like.

bhariharan: Sounds good. Let's get it updated for consistency though.

mgetzbw: @bhariharan done

bhariharan:
👍

bhariharan:
@mgetzbw Looks like compilation is broken on the ```Everything``` scheme. Could you take a look? Thanks!

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.h
bhariharan: Found the culprit - this should be ```CSFActionTiming```.

PR #1610, opened 2016-06-17
minor mixups for swift

mgetzbw:
Looked into making the swift enum strings more closely tied... but the naming is weird due to the `k` prefix, because changing that would mean breaking compatibility I am deferring for now.


bhariharan:
👍

PR #1612, opened 2016-06-20
Fixing bug that caused testCleanResyncGhostsForMRUTarget unit test failure

wmathurin:
When passing a smart sql query spec to remove by query, it must be selecting on _soupEntryId 

NB: it's documented in the .h here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/unstable/libs/SmartStore/SmartStore/Classes/SFSmartStore.h#L348


bhariharan:
👍

PR #1613, opened 2016-06-20
Various native SmartSyncExplorer fixes:

wmathurin:
- in the detail screen, delete button becomes undelete when items is already marked as deleted
- after sync, search filter already typed is re-applied (before the filter would still be shown but all records would be listed)
- after fail sync, list is still refreshed because failure means not all records could be synched up, it doesn't mean that none could be synched up

### DIFF
--- a/native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/ContactDetailViewController.m
+++ b/native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/ContactDetailViewController.m
@@ -255,6 +255,12 @@
     [self.navigationController popViewControllerAnimated:YES];
 }
 
+- (void)undeleteContact {
+    [self.dataMgr undeleteLocalData:self.contact];
+    self.contactUpdated = YES;
+    [self.navigationController popViewControllerAnimated:YES];
+}
+
 - (UITextField *)contactTextField:(NSString *)propertyValue {
     UITextField *textField = [[UITextField alloc] initWithFrame:CGRectZero];
     textField.text = propertyValue;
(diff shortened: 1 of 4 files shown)


native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/ContactDetailViewController.m
```diff
@@ -255,20 +255,27 @@ - (void)deleteContact {
     [self.navigationController popViewControllerAnimated:YES];
 }
 
+- (void)undeleteContact {
```
wmathurin: Adding undelete support (already in the hybrid and react native versions of the application)

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/ContactListViewController.m
```diff
@@ -83,7 +84,7 @@ @implementation ContactListViewController
 - (id)initWithNibName:(NSString *)nibNameOrNil bundle:(NSBundle *)nibBundleOrNil {
     self = [super initWithNibName:nibNameOrNil bundle:nibBundleOrNil];
     if (self) {
-        self.dataMgr = [[SObjectDataManager alloc] initWithViewController:self dataSpec:[ContactSObjectData dataSpec]];
```
wmathurin: Data manager no longer has a reference to the parent view - instead, on refresh it expects a block

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/ContactListViewController.m
```diff
@@ -250,8 +259,13 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
 
 - (void)searchBar:(UISearchBar *)searchBar textDidChange:(NSString *)searchText {
     [self log:SFLogLevelDebug format:@"searching with text: %@", searchText];
+    self.searchText = searchText;
+    [self refreshList];
+}
+
+- (void) refreshList {
     __weak ContactListViewController *weakSelf = self;
-    [self.dataMgr filterOnSearchTerm:searchText completion:^{
```
wmathurin: On refresh, we apply the filter typed

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/ContactListViewController.m
```diff
@@ -361,17 +375,24 @@ - (void)syncUpDown {
...
+        [weakSelf refreshList];
+    };
+    
     [self.dataMgr updateRemoteData:^(SFSyncState *syncProgressDetails) {
         dispatch_async(dispatch_get_main_queue(), ^{
             weakSelf.navigationItem.rightBarButtonItem.enabled = YES;
+            // NB: when the sync failed it means not everything could be synched up
+            //     it doesn't mean nothing could be synched up
+            // Therefore we should be refreshing whether it was a complete success or not
+            [weakSelf.dataMgr refreshLocalData:completionBlock];
```
wmathurin: The comment in the code says it all

bhariharan:
👍

PR #1616, opened 2016-06-22
Added network logging identifier, and added explicit error for when App Transport Security error responses are received

MikeNachbaur:
Note that "Error" responses are on by default, so only certain log messages will be seen by default.

### DIFF
--- a/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
+++ b/libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
@@ -72,7 +72,9 @@ CSFActionTiming kCSFActionTimingPostProcessingKey = @"postProcessing";
 - (NSURL*)urlForActionWithError:(NSError**)error {
     NSURL *baseURL = self.baseURL;
     if (!baseURL) {
-        if (error){
+        NetworkWarn(@"Network action must have a base URL defined.");
+
+        if (error) {
             *error = [NSError errorWithDomain:CSFNetworkErrorDomain
                                          code:CSFNetworkURLCredentialsError
                                      userInfo:@{ NSLocalizedDescriptionKey: @"Network action must have a base URL defined",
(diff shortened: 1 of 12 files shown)


libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Utilities/CSFInternalDefines.m
```diff
@@ -32,6 +32,13 @@
 @import LaunchServices;
 #endif
 
+NSInteger kCSFNetworkLogContext = 0;
```
MikeNachbaur: Note that this is the recommended way to initialize a logging context from an identifier.  This way it gets initialized at load-time, but isn't bound to any particular class.

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Utilities/CSFDefines.h
```diff
@@ -40,6 +40,8 @@
 
 CSF_EXTERN NSString * const kCSFConnectVersion;
 
+CSF_EXTERN NSString * const CSFNetworkLogIdentifier;
```
MikeNachbaur: The log identifier is exposed so developers can change the log level using:

```
[[SFLogger sharedInstance] setLogLevel:SFLogLevelDebug forIdentifier:CSFNetworkLogIdentifier]
```

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Utilities/CSFDefines.m
```diff
@@ -26,3 +26,4 @@
 
 NSString * const CSFValidationErrorDomain = @"CSFValidationErrorDomain";
 NSString * const CSFNetworkInitializedNotification = @"CSFNetworkInitializedNotification";
+NSString * const CSFNetworkLogIdentifier = @"com.salesforce.MobileSDK.Network";
```
MikeNachbaur: Log identifier starts with `com.salesforce.MobileSDK` so that other MobileSDK libraries can add their own logging identifiers under a shared prefix.  This way, a developer can set the log level for the entire mobile SDK at once using:

```
[[SFLogger sharedInstance] setLogLevel:SFLogLevelDebug forIdentifiersWithPrefix:@"com.salesforce.MobileSDK"]
```

libs/SalesforceNetwork/SalesforceNetwork/SalesforceNetwork/Classes/Network/Action/CSFAction.m
```diff
@@ -427,6 +430,21 @@ - (void)sessionTask:(NSURLSessionTask*)task didCompleteWithError:(NSError*)error
     [self updateProgress];
 
     if (error) {
+        NetworkDebug(@"Received an error while processing %@: %@", self, error);
+
+        if ([error.domain isEqualToString:NSURLErrorDomain] &&
```
bhariharan: Nice!

bhariharan:
👍

Issue #1525, opened 2016-05-09
use_frameworks! in Podfile shouldn't break React-Native integration


ivanbogdanov:
steps to reproduce:
* run latest 'forceios create' with react_native template
* test it:  should run smoothly
* add  use_frameworks! to the project Podfile
* run 'pod update'
* result: the project doesn't compile

wmathurin:
Fixed in upcoming release (see linked pull request above).

PR #1618, opened 2016-06-23
Salesforce React pod spec changes - to make it work with frameworks

wmathurin:
Before that change app generated with forceios would not compile if use_frameworks! was used.
With that change, it works with and without use_frameworks!.

This fixes: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1525


bhariharan:
👍

PR #1631, opened 2016-07-07
Updating upgrade doc

bhariharan:


bhariharan:
@rwhitley No real changes here, just version number updates.

wmathurin:
Maybe for 5.0, we should also think of adding upgrade steps for react native apps.

bhariharan:
Sounds like a good idea for ```5.0```.

PR #1637, opened 2016-07-09
forceios UX friendly / updated Swift template for XCode 8

muenzpraeger:


bhariharan:
@muenzpraeger With your changes, can you verify if the ```Swift``` template is backward compatible with ```Xcode 7```? We need it to work with ```Xcode 7```.

muenzpraeger:
Nope, not backward compatible. You can "skip" it till GA of XCode 8 if you want.

Status is that the "old" template cannot be used in XCode 8 - and my modified version not in XCode 7.

An alternative could be to deliver on XCode 7 and one XCode 8 template and create the app skeleton to the current selected XCode version. What do you think?

bhariharan:
@muenzpraeger We will make ```Xcode 8``` our minimum when it goes GA. Until then, we will probably only support ```Xcode 7```. We have a whole body of work to support ```iOS 10``` anyway - we will tackle this as part of that. Adding an extra template in the interim is not desirable. Thanks for your contribution though.

muenzpraeger:
Makes sense. At least you can cherry-pick later if you'd like.

4.2.0 runs fine btw with iOS 10 so far (used the unstable branch till
yesterday ;-)).

PR #1640, opened 2016-07-11
Changing delegates implementation to weak container

khawkins:
Per comments on the [original PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1467/), I changed the implementation to use `NSHashTable`, which is a first-class weak container in Cocoa.  Same rationale applies: we should maintain containers of weak delegates, rather than unsafe_unretained, so that iterating released delegates will not cause indeterminate/crashing behavior.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecurityLockout.m
```diff
@@ -75,39 +75,41 @@ @implementation SFSecurityLockout
...
-    
-    sDelegates = [NSMutableOrderedSet orderedSet];
-    
-    [SFSecurityLockout setPasscodeViewControllerCreationBlock:^UIViewController *(SFPasscodeControllerMode mode, SFPasscodeConfigurationData configData) {
-        SFPasscodeViewController *pvc = nil;
-        if (mode == SFPasscodeControllerModeCreate) {
-            pvc = [[SFPasscodeViewController alloc] initForPasscodeCreation:configData];
-        } else if (mode == SFPasscodeControllerModeChange) {
-            pvc = [[SFPasscodeViewController alloc] initForPasscodeChange:configData];
+    if (self == [SFSecurityLockout class]) {
```
khawkins: Added this for sanity's sake, though heaven help us if anyone subclasses this. 😳

bhariharan:
👍

PR #1646, opened 2016-07-13
Magic link for iOS

huminzhi:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -906,6 +963,7 @@ - (BOOL)webView:(UIWebView *)webView shouldStartLoadWithRequest:(NSURLRequest *)
     BOOL result = YES;
     NSURL *requestUrl = [request URL];
     NSString *requestUrlString = [requestUrl absoluteString];
+    [self log:SFLogLevelDebug format:@"Start to load %@", requestUrlString];
```
bhariharan: Let's remove this log line. Security won't be too happy.

kchitalia:
Should we add unit tests around this flow?

huminzhi:
@kchitalia I will add a UI test for it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -557,15 +557,72 @@ - (void)beginUserAgentFlow {
...
+    NSString *bodyStr = [NSString stringWithFormat:@"grant_type=urn%%3Aietf%%3Aparams%%3Aoauth%%3Agrant-type%%3Ajwt-bearer&assertion=%@",self.credentials.jwt];
+    NSData *body = [bodyStr dataUsingEncoding:NSUTF8StringEncoding];
+    [request setHTTPBody:body];
+    [request setHTTPMethod:kHttpMethodPost];
+    [request setValue:kHttpPostContentType forHTTPHeaderField:kHttpHeaderContentType];
+    
+    [[self.session dataTaskWithRequest:request completionHandler:completionHandler] resume];
+}
+
+- (NSString *)URLEncodeStringFromString:(NSString *)string
```
khawkins: This method exists, in `NSString+SFAdditions`: `stringByURLEncoding`.  Can we use that?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -557,15 +557,72 @@ - (void)beginUserAgentFlow {
...
     
+    // JWT Flow
+    if (self.credentials.jwt && self.credentials.instanceUrl) {
+        [self swapJWTWithcompletionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+            bool swapOK = NO;
+            if (!error) {
+                NSError *jsonError = nil;
+                id json = nil;
+                
+                json = [NSJSONSerialization JSONObjectWithData:data options:0 error:&jsonError];
```
khawkins: Any reason not to use `SFJsonUtils` for this functionality?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -557,15 +557,72 @@ - (void)beginUserAgentFlow {
...
+
+- (NSString *)URLEncodeStringFromString:(NSString *)string
+{
+    static CFStringRef charset = CFSTR("!@#$%&*()+'\";:=,/?[] ");
+    CFStringRef str = (__bridge CFStringRef)string;
+    CFStringEncoding encoding = kCFStringEncodingUTF8;
+    return (NSString *)CFBridgingRelease(CFURLCreateStringByAddingPercentEscapes(NULL, str, NULL, charset, encoding));
+}
+
+- (void)doLoadURL:(NSString*)approvalUrl  withCookie:(bool)enableCookie {
```
khawkins: Minor: let's use `BOOL` instead of `bool`, for convention.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -557,15 +557,72 @@ - (void)beginUserAgentFlow {
...
+
+
+- (void)swapJWTWithcompletionHandler:(void (^)(NSData *data, NSURLResponse *response, NSError *error))completionHandler  {
+    NSString *url = [[NSString alloc] initWithFormat:@"%@://%@%@", self.credentials.protocol,
+                     self.credentials.instanceUrl.absoluteString,
+                     kSFOAuthEndPointToken];
+    
+    NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:url] cachePolicy:NSURLRequestReloadIgnoringCacheData
+                                                       timeoutInterval:self.timeout];
+    NSString *bodyStr = [NSString stringWithFormat:@"grant_type=urn%%3Aietf%%3Aparams%%3Aoauth%%3Agrant-type%%3Ajwt-bearer&assertion=%@",self.credentials.jwt];
```
khawkins: Can we create this by instantiating the unencoded string, then encoding it?  The original string will be easier to read/maintain.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -557,15 +557,72 @@ - (void)beginUserAgentFlow {
...
+}
+
+- (void)doLoadURL:(NSString*)approvalUrl  withCookie:(bool)enableCookie {
     if (self.credentials.logLevel < kSFOAuthLogLevelInfo) {
         [self log:SFLogLevelDebug format:@"SFOAuthCoordinator:beginUserAgentFlow with %@", approvalUrl];
     }
     
     NSMutableURLRequest *request = [NSMutableURLRequest requestWithURL:[NSURL URLWithString:approvalUrl]];
-	[request setHTTPShouldHandleCookies:NO]; // don't use shared cookies
+    if(!enableCookie) [request setHTTPShouldHandleCookies:enableCookie];
```
khawkins: Is this line necessary?

khawkins: Sorry, I should say: do we need the conditional?  This could just be `[request setHTTPShouldHandleCookies:enableCookie];`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
```diff
@@ -117,6 +117,13 @@ typedef NS_ENUM(NSInteger, SFOAuthCredentialsStorageType){
  @exception NSInternalInconsistencyException If accessed while the identifier property is `nil`.
  */
 @property (nonatomic, copy) NSString *activationCode;
+/** JWT.
+ 
+ JWT code used in the client breeze link flow.
+ @warning This property must not be modified while authenticating.
+ @warning This property should be set to nil after authentication.
+ */
+@property (nonatomic, copy) NSString *jwt;
```
khawkins: Let's add the `nullable` flag too, just because we'll have to at some point anyway.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -32,6 +32,7 @@
 NSString * const kSFOAuthServiceAccess          = @"com.salesforce.oauth.access";
 NSString * const kSFOAuthServiceRefresh         = @"com.salesforce.oauth.refresh";
 NSString * const kSFOAuthServiceActivation      = @"com.salesforce.oauth.activation";
+NSString * const kSFOAuthServiceJWT             = @"com.salesforce.oauth.jwt";
```
khawkins: This doesn't currently tie to anything.  Is it needed?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
huminzhi: @khawkins I am not so sure, but I saw some other places, line 419, line 722, also use this.. Shall I change them all?

khawkins: Ah, throwbacks to when OAuth lived at a lower layer than SalesforceSDKCore.  We can save that for another time, as it's consistent (as you point out) and not related to this PR.  My $0.02, anyway.

huminzhi: ok, I think I will leave it as it is for now..

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
huminzhi: good point, will remove it

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
huminzhi: sure.. I will also add for the rest

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
huminzhi: I am not so sure, but the original code is set it to NO, but for JWT flow, I have to set it to YES to have it work...

khawkins: I'm just saying that the value of `enableCookie` will set the argument properly, without the `if` clause, whether its value is `YES` or `NO`.

khawkins:
Looks great from my end, thanks @huminzhi !

bhariharan:
👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -557,15 +558,65 @@ - (void)beginUserAgentFlow {
...
+        }];
+    }
+    else {
+        [self doLoadURL:approvalUrl withCookie:NO];
+    }
+}
+
+
+- (void)swapJWTWithcompletionHandler:(void (^)(NSData *data, NSURLResponse *response, NSError *error))completionHandler  {
+    NSString *url = [[NSString alloc] initWithFormat:@"%@://%@%@", self.credentials.protocol,
```
khawkins: Sorry to come back to this late in the game, but I was reviewing the way that JWT Bearer flow works, and I'm not sure that `instanceUrl` is the appropriate value to use here.  According to [the docs](https://help.salesforce.com/HTViewHelpDoc?id=remoteaccess_oauth_jwt_flow.htm), the request should be made against the login endpoint (e.g. `login.salesforce.com`).

huminzhi: @khawkins that's right, but I believe for the token endpoint, both login endpoint and instance url endpoint should work.. The reason I use instanceURL here is it's the only thing we have... For the frontdoor flow, I need the instanceURL, so in the magic link, it gives  us 2 things: JWT and instanceURL.. but think about it again, you have a valid point, I should in this case set login URL as instanceURL or maybe ask the magicLink provide login URL also, so here it should login URL, I will send a fix soon..

PR #1647, opened 2016-07-13
External storage post merge work

wmathurin:
Code to prevent soup from having json1 index and external storage (and new test for it)
Test code duplication cleanup
New tests for externally stored soup with full text
New tests for alter soup with storage change and column type changes 

NB: The hybrid/react native work will come in a separate PR

### DIFF
--- a/libs/SmartStore/SmartStoreTests/SFSmartSqlWithExternalStorageTests.m
+++ b/libs/SmartStore/SmartStoreTests/SFSmartSqlWithExternalStorageTests.m
@@ -30,6 +30,29 @@
 #import <SalesforceSDKCore/SFJsonUtils.h>
 #import "SFSoupSpec.h"
 
+@interface SFSmartSqlTests ()
+- (SFUserAccount*) createUserAccount;
+- (void) testSharedInstance;
+- (void) testConvertSmartSqlWithInsertUpdateDelete;
+- (void) testSimpleConvertSmartSql;
+- (void) testConvertSmartSqlWithJoin;
+- (void) testConvertSmartSqlWithSelfJoin;
+- (void) testConvertSmartSqlWithSpecialColumns;
+- (void) testConvertSmartSqlWithSpecialColumnsAndJoin;
+- (void) testConvertSmartSqlWithSpecialColumnsAndSelfJoin;
+- (void) testConvertSmartSqlWithJSON1;
+- (void) testConvertSmartSqlWithJSON1AndTableQualifiedColumn;
+- (void) testConvertSmartSqlWithJSON1AndTableAliases;
+- (void) testSmartQueryDoingCount;
+- (void) testSmartQueryDoingSum;
+- (void) testSmartQueryReturningOneRowWithOneInteger;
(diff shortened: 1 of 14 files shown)


sgoldberg-sfdc:
👍  Thanks for cleaning that up.

libs/SmartStore/SmartStoreTests/SFSmartStoreFullTextSearchTests.h
```diff
@@ -27,5 +27,12 @@
 
 #import "SFSmartStoreTestCase.h"
 
-@interface SFSmartStoreFullTextSearchTests : SFSmartStoreTestCase 
+@interface SFSmartStoreFullTextSearchTests : SFSmartStoreTestCase
+
+#define kTestStore            @"testSmartStoreFullTextSearchStore"
```
bhariharan: I wonder if all the ```#define``` should be ```static const NSString *``` instead. Minor difference, but I think we do the latter in our code usually. I don't care much either way.

bhariharan:
Looks pretty good otherwise.

libs/SmartStore/SmartStoreTests/SFSmartStoreFullTextSearchTests.h
wmathurin: I have only be using #define in tests (and usually in the .m). I agree we should use const in the code. For tests, I don't think it matters.

bhariharan: 👍

kchitalia: Just for my understanding, what's the difference between  #define ans static const NSString * ?

bhariharan: @kchitalia ```#define``` is a preprocessor directive, so basically it's a macro - fancy way of doing ```Find and Replace```. Every occurrence of that key is substituted by the value. ```static const``` creates a single copy of that variable, and allocates memory to it, which is common to all instances of that object, so it's always in memory. That's basically the difference.

libs/SmartStore/SmartStoreTests/SFSmartSqlWithExternalStorageTests.m
```diff
@@ -30,6 +30,29 @@
 #import <SalesforceSDKCore/SFJsonUtils.h>
 #import "SFSoupSpec.h"
 
+@interface SFSmartSqlTests ()
```
khawkins: Not the end of the world, but in a scenario like this, why not put responsibility of defining the interface on the parent class, then reference that interface here?

wmathurin: That makes sense, I just made the change here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1650/commits/d8b65aacd8f24d6e52c4b058d24829de22283783

PR #1648, opened 2016-07-14
On some machine, forceios prints out hex

wmathurin:
Solution found here: http://stackoverflow.com/questions/7337564/why-does-console-logbuffer-give-me-a-hexadecimal-list


bhariharan:
👍

kchitalia:
Its interesting that this is seen only on some machines :)

PR #1651, opened 2016-07-15
Better soup table verification

wmathurin:


libs/SmartStore/SmartStoreTests/SFSmartStoreWithExternalStorageTests.m
```diff
@@ -279,8 +279,8 @@ -(void) testRemoveEntriesByQueryWithExternalStorage
         
         
         // Check soup
-        [self checkSoup:@[soupEltsCreated[1], soupEltsCreated[3]] shouldExist:YES store:store soupName:kSSExternalStorage_TestSoupName];
-        [self checkSoup:@[soupEltsCreated[0], soupEltsCreated[2]] shouldExist:NO store:store soupName:kSSExternalStorage_TestSoupName];
+        [self checkSoupTable:@[soupEltsCreated[1], soupEltsCreated[3]] shouldExist:YES store:store soupName:kSSExternalStorage_TestSoupName];
+        [self checkSoupTable:@[soupEltsCreated[0], soupEltsCreated[2]] shouldExist:NO store:store soupName:kSSExternalStorage_TestSoupName];
```
wmathurin: Previous implementation was calling retrieve which goes to the filesystem for externally stored soup.
Now checkSoupTable directly reads from the soup table in the database and makes sure indexed columns are populated as expected. Also make sure the given rows are found (or not).

bhariharan:
👍

kchitalia:
Definitely a better way to check. 👍

PR #1338, opened 2016-01-15
Everything scheme builds again

wmathurin:
For some reason, import of project header in a test file without first importing public headers referenced in the project header caused compilation errors
Now in tests, we import the framework master header before any project header

Had to enable import of non-modular headers in SalesforceHybridSDK because it was choking on the &#35;import &lt;Cordova/xxx&gt;

PR #1654, opened 2016-07-18
fix unit test on travisCI

huminzhi:


huminzhi:
This can fix one of the unit test issues

kchitalia:
LGTM

bhariharan:
@huminzhi How does this fix a test? The other two headers are public headers too, correct?

huminzhi:
right @bhariharan , it's something really weird and likely an xcode bug, according to @wmathurin we had similar issue before: 
https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1338
And from what I tried, it does fix the compiling issue on TravisCI (it's likely fixed in xcode7.3, however, on travisCI, we have to use xcode7.2 for another reason...) looking forward to move out of TravisCI...

PR #1659, opened 2016-07-20
Fix warnings

kchitalia:
These changes are not really for iOS 10 support, just fixing warnings. I tested and verifieid that both native and hybrid sample apps work with iOS 8 and iOS 9 too.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
@@ -246,7 +246,7 @@ typedef NS_ENUM(NSInteger, SFOAuthCredentialsStorageType){
  @param encrypted Determines if the sensitive data like refreshToken and accessToken should be encrypted
  @return An initialized authentication credential object.
  */
-- (_Nonnull instancetype)initWithIdentifier:( NSString * _Nullable)theIdentifier clientId:( NSString * _Nullable )theClientId encrypted:(BOOL)encrypted;
+- (_Nullable instancetype)initWithIdentifier:( NSString * _Nullable)theIdentifier clientId:( NSString * _Nullable )theClientId encrypted:(BOOL)encrypted;
 
 /** Initializes an authentication credential object with the given identifier and client ID. This is the designated initializer.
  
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
```diff
@@ -246,7 +246,7 @@ typedef NS_ENUM(NSInteger, SFOAuthCredentialsStorageType){
  @param encrypted Determines if the sensitive data like refreshToken and accessToken should be encrypted
  @return An initialized authentication credential object.
  */
-- (instancetype)initWithIdentifier:(NSString *)theIdentifier clientId:(NSString *)theClientId encrypted:(BOOL)encrypted;
+- (_Nonnull instancetype)initWithIdentifier:( NSString * _Nullable)theIdentifier clientId:( NSString * _Nullable )theClientId encrypted:(BOOL)encrypted;
```
khawkins: init* methods should almost always be `_Nullable`, as their instantiation design pattern allows for returning `nil` values.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
```diff
@@ -259,7 +259,7 @@ typedef NS_ENUM(NSInteger, SFOAuthCredentialsStorageType){
  @param type Indicates whether the OAuth credentials are stored in the keychain
  @return An initialized authentication credential object.
  */
-- (instancetype)initWithIdentifier:(NSString *)theIdentifier clientId:(NSString *)theClientId encrypted:(BOOL)encrypted storageType:(SFOAuthCredentialsStorageType)type;
+- (_Nullable instancetype)initWithIdentifier:(NSString * _Nullable )theIdentifier clientId:(NSString * _Nullable)theClientId encrypted:(BOOL)encrypted storageType:(SFOAuthCredentialsStorageType)type;
```
khawkins: Following the configuration through, `SFOAuthCredentials.identifier` is `nonnull`, which means `theIdentifier` should be too.

khawkins:
Otherwise, looks good to me.

PR #1666, opened 2016-07-21
Adding ReactNative support for external soup storage

bhariharan:


libs/SalesforceReact/SalesforceReact/Classes/SFSmartStoreReactBridge.m
```diff
@@ -67,31 +67,41 @@ @implementation SFSmartStoreReactBridge
...
+        NSError *error = nil;
+        BOOL result = [smartStore registerSoupWithSpec:soupSpec withIndexSpecs:indexSpecs error:&error];
+        if (result) {
+            callback(@[[NSNull null], soupName]);
+        } else {
+            callback(@[RCTMakeError(@"registerSoup failed", error, nil)]);
+        }
     } else {
-        callback(@[RCTMakeError(@"registerSoup failed", error, nil)]);
+        callback(@[RCTMakeError(@"registerSoup failed", nil, nil)]);
```
kchitalia: Not passing error anymore? If smartstore is nil, then we should somehow reflect that in the error.

bhariharan: We do pass in the error if the register call failed 2 lines above. ```SmartStore``` can never be ```nil``` realistically.

PR #1667, opened 2016-07-22
fix test timeout issue on travisCI

huminzhi:


libs/SmartStore/SmartStoreTests/SFSmartStoreWithExternalStorageTests.m
```diff
@@ -419,6 +419,9 @@ - (void)testExternalStorageIsNotEncryptedWhenDbIsNotEncrypted {
 
 - (void)testExternalStorageUpsertWithOneMBSizePayloadInRegression {
     NSInteger numberOfIterations = 500;
+    if ([[UIDevice currentDevice].systemVersion floatValue] < 9.3 ) {
+        numberOfIterations = 100;
+    }
```
kchitalia: Do we need a FIXME here too?

bhariharan:
@huminzhi I don't think we should do specific version conditionals in the code. 👎 

Let's run it only on ```iOS 9.3``` only instead, until this is resolved.

libs/SmartStore/SmartStoreTests/SFSmartStoreWithExternalStorageTests.m
huminzhi: @kchitalia with this fix, the test will work for both ios9.2 and ios9.3, so this code is good.
Without this change, it will fail in ios9.2 because ios9.2 will hit memory issue (somehow ios9.3 won't)

huminzhi:
@bhariharan I totally agree with you if we have the option to run on ios9.3 on travisCI.. 
The problem is there is some environment issue on travisCI so we can only run our tests in ios9.2 on travisCI... and the test is not really failing, rather timeout on travisCI.. so once we migrate to Cloudbee, I will immediately remove those lines.. how is that?
cc @kchitalia

bhariharan:
@huminzhi For now, could we just disable those 3 tests instead of conditionally running them?

huminzhi:
@bhariharan OK, temporarily bypassed

bhariharan:
👍

PR #1669, opened 2016-07-25
Remove appending of current UA

bhariharan:


bhariharan:
We never did append the current UA on ```Android```. This brings parity.

kchitalia:
👍

wmathurin:
Is this going to throw off some server side logic?

bhariharan:
It shouldn't. It's a legacy component. The server side logic has changed to check the values in our UA format to determine if it's coming from `mobile`.

PR #1670, opened 2016-07-25
Adding support for 'Remember User Name'

bhariharan:


kchitalia:
1. Do we already have unit tests for this?
2. Is this a bug we are fixing?

bhariharan:
This is in the ```WebView``` - there's no way to add unit tests for it. It's not a bug fix - it's an enhancement on the login page, owned by the auth team.

khawkins:
Seems fine, albeit with slight reservations about `removeAllCookies` retaining its naming convention, now that it no longer does. :)  Are there any scenarios where we might still want to be able to clear the entire state, including the user?

bhariharan:
I can't think of any use case. On ```Android```, we have the ability to do that explicitly with a ```Clear Cookies``` button in the menu, but we've never had that on ```iOS```.

kchitalia:
I tried the SmartSyncExplorerHybrid Sample App without this change. I checked the "Remember me" checkbox. I logged in to the App, then I logged out. The next time I try to login to the same login host, it remembers my user name. 

How will the app behave after this change?

bhariharan:
This is not for that flow. This is for the verification code on pre-approved device flow. Once a device is approved, you won't need to type in the verification code again if it has been whitelisted once.

kchitalia:
Ah. Got it.

PR #1676, opened 2016-08-02
notify magiclink fail so that user can customize the fail flow"

huminzhi:


huminzhi:
@bhariharan @khawkins Please help review.. This is so that the user can customize fail flow

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -574,14 +574,25 @@ - (void)beginUserAgentFlow {
                         swapOK = YES;
                         self.credentials.jwt = nil;
                     }
+                    else {
+                        if ([[dict valueForKey:@"error"] isEqual:kSFOAuthErrorTypeInvalidGrant]) {
+                            NSError *error = [NSError errorWithDomain:kSFOAuthEndPointToken
```
khawkins: To follow the pattern of OAuth errors returned to the client, we should:
1. Generate errors from `[SFOAuthCoordinator errorWithType:description:]`
2. Create the equivalent `NSString` constant type to go with the `kSFOAuthError` key in the user dictionary, and add the conditional logic to generate that error type/code—the errorWithType implementation mentioned above will make the pattern clear.

huminzhi: Thank you @khawkins , will make corresponding changes.

huminzhi:
@kchitalia @bhariharan just fixed the error pattern, please review, thanks!

kchitalia:
@huminzhi Just for my understanding, how will consumers of the SDK customize this fail flow?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -975,6 +975,11 @@ - (SFAuthErrorHandlerList *)populateDefaultAuthErrorHandlerList
                                                    [weakSelf execFailureBlocks];
                                                    return YES;
                                                }
+                                               if (error.code == kSFOAuthErrorJWTInvalidGrant) {
+                                                   [weakSelf log:SFLogLevelWarning format:@"JWT swap failed due to invalid grant.  Error code: %ld", (long)error.code];
+                                                   [weakSelf execFailureBlocks];
```
huminzhi: @kchitalia  So authenticationManager is the delegate for oauth failure.  And here as you can see, it will call "execFailureBlocks" which will eventually call [SalesforceSDKManager sharedManager].launchErrorAction,  so that user can just implement their desired flow in launchErrorAction..

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -575,13 +576,15 @@ - (void)beginUserAgentFlow {
                         self.credentials.jwt = nil;
                     }
                 }
+                if (!swapOK) {
```
khawkins: The only other minor thing I'd say is to move the `swapOK` definition [above](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1676/files#diff-a19763c4fa428fe97ccc907d35e442c3R563) into the `if` block, to reduce its scope.  Extremely minor. 😊

khawkins:
Otherwise, this PR looks fine to me @huminzhi !

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
kchitalia: Thank you for explaining that. Now I understand it better :)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
huminzhi: 👍 Great reviewer! @khawkins  Will change accordingly!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -560,8 +560,8 @@ - (void)beginUserAgentFlow {
     // JWT Flow
     if (self.credentials.jwt && self.credentials.instanceUrl) {
         [self swapJWTWithcompletionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
-            bool swapOK = NO;
             if (!error) {
+                bool swapOK = NO;
```
huminzhi: @khawkins reduced its scope!

khawkins:
LGTM

PR #1678, opened 2016-08-04
add drop table button in sample app 

huminzhi:
@wmathurin  this is enabling me to test a bug
@kchitalia @bhariharan @khawkins


kchitalia:
Should be fine as this is in the sample app. One question, how do you know if the table was successfully dropped?

wmathurin:
A unit test might be a better approach to testing such bug.

PR #1681, opened 2016-08-09
Fixing bug with app version and build number

bhariharan:
Looks like we were using the wrong field in our user agent, and hence getting build number instead of app version. This PR fetches both now.

From Apple's documentation [here](https://developer.apple.com/library/ios/technotes/tn2420/_index.html) - `Note that the Version Number is saved with the key CFBundleShortVersionString and the Build Number is associated with the key CFBundleVersion. You will find the Version Number (CFBundleShortVersionString) and Build Number (CFBundleVersion) described in many different places in the documentation and referred to using these key values`.


kchitalia:
:+1:

huminzhi:
👍 Thank you!

wmathurin:
Was it always that way or did it regress?

bhariharan:
@wmathurin It looks like it was always that way. I recently discovered it while working on `AILTN`.

wmathurin:
I guess we have not really queried the logs by app version.

PR #1689, opened 2016-08-15
Fixed parsing bug for notes-sync

trooper2013:
Fixed parsing issue for notes sync app. The app would crash on startFetch operation for the SyncDownTarget.


bhariharan:
👍

PR #1691, opened 2016-08-16
fix unit test getInfo

huminzhi:
@bhariharan  FYI


bhariharan:
@huminzhi You need to update this to the latest from `unstable` on `external/shared`.

bhariharan:
I took care of it [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1693).

PR #1699, opened 2016-08-18
Flowing use of login host throughout JWT flow

khawkins:
We missed some spots where instance URL was still driving the JWT setup, where login host (`domain`) should be used instead.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -569,9 +569,11 @@ static NSString * const kOAuthUserAgentUserDefaultsKey          = @"UserAgent";
                 json = [NSJSONSerialization JSONObjectWithData:data options:0 error:&jsonError];
                 if (nil == jsonError && [json isKindOfClass:[NSDictionary class]]) {
                     NSDictionary *dict = (NSDictionary *)json;
-                    if (dict[kSFOAuthAccessToken]) {
+                    [self updateCredentials:dict];
+                    if (self.credentials.accessToken && self.credentials.apiUrl) {
+                        NSString *baseUrlString = [self.credentials.apiUrl absoluteString];
                         NSString *escapedString = [approvalUrl stringByURLEncoding];
-                        NSString* approvalUrl = [NSString stringWithFormat:@"%@://%@/secur/frontdoor.jsp?sid=%@&retURL=%@", self.credentials.protocol, self.credentials.domain, dict[kSFOAuthAccessToken],escapedString];
+                        NSString* approvalUrl = [NSString stringWithFormat:@"%@/secur/frontdoor.jsp?sid=%@&retURL=%@", baseUrlString, self.credentials.accessToken, escapedString];
                         [self doLoadURL:approvalUrl withCookie:YES];
                         swapOK = YES;
                         self.credentials.jwt = nil;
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -170,8 +170,9 @@ - (void)dealloc {
 
 - (void)authenticate {
     NSAssert(nil != self.credentials, @"credentials cannot be nil");
-    NSAssert([self.credentials.clientId length] > 0, @"credentials.clientId cannot be nil or empty");
-    NSAssert([self.credentials.identifier length] > 0, @"credentials.identifier cannot be nil or empty");
+    NSAssert(self.credentials.clientId.length > 0, @"credentials.clientId cannot be nil or empty");
+    NSAssert(self.credentials.identifier.length > 0, @"credentials.identifier cannot be nil or empty");
+    NSAssert(self.credentials.domain.length > 0, @"credentials.domain cannot be nil or empty.");
```
khawkins: We were assuming that `domain` was set, pretty much all over the different auth flows.  May as well make it officially required at the beginning.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -558,7 +559,7 @@ - (void)beginUserAgentFlow {
     }
     
     // JWT Flow
-    if (self.credentials.jwt && self.credentials.instanceUrl) {
+    if (self.credentials.jwt) {
```
khawkins: This would have otherwise changed to checking the `nil`ness of `domain`, but we're checking it at the start, so no need.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -570,7 +571,7 @@ - (void)beginUserAgentFlow {
                     NSDictionary *dict = (NSDictionary *)json;
                     if (dict[kSFOAuthAccessToken]) {
                         NSString *escapedString = [approvalUrl stringByURLEncoding];
-                        NSString* approvalUrl = [NSString stringWithFormat:@"%@://%@/secur/frontdoor.jsp?sid=%@&retURL=%@", self.credentials.protocol, self.credentials.instanceUrl, dict[kSFOAuthAccessToken],escapedString];
+                        NSString* approvalUrl = [NSString stringWithFormat:@"%@://%@/secur/frontdoor.jsp?sid=%@&retURL=%@", self.credentials.protocol, self.credentials.domain, dict[kSFOAuthAccessToken],escapedString];
```
huminzhi: @khawkins so I tried to use domain here with the frontdoor url, something like login.salesforce.com, it won't work.. for frontdoor, it has to be the instanceUrl....

khawkins: Ah yes, I got that wrong.  We should be populating the rest of the available credentials properties from the output here, including `instance_url` and `community_url`, which should drive the frontdoor base URL.  I'll update accordingly.

bhariharan:
👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -568,9 +569,11 @@ - (void)beginUserAgentFlow {
                 json = [NSJSONSerialization JSONObjectWithData:data options:0 error:&jsonError];
                 if (nil == jsonError && [json isKindOfClass:[NSDictionary class]]) {
                     NSDictionary *dict = (NSDictionary *)json;
-                    if (dict[kSFOAuthAccessToken]) {
+                    [self updateCredentials:dict];
```
khawkins: The data that comes back from the swap call is the data that bootstraps the rest of the process.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -568,9 +569,11 @@ - (void)beginUserAgentFlow {
...
+                    if (self.credentials.accessToken && self.credentials.apiUrl) {
+                        NSString *baseUrlString = [self.credentials.apiUrl absoluteString];
```
khawkins: `apiUrl` will ensure Community URL is used if available, instance URL otherwise.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -829,28 +832,20 @@ - (NSString *)scopeQueryParamString
...
-    NSString *communityId = [params objectForKey:kSFOAuthCommunityId];
-    if (nil != communityId) {
-        self.credentials.communityId = communityId;
-    }
-    
-    NSString *communityUrl = [params objectForKey:kSFOAuthCommunityUrl];
-    if (nil != communityUrl) {
-        self.credentials.communityUrl = [NSURL URLWithString:communityUrl];
-    }
+    if (params[kSFOAuthAccessToken]) self.credentials.accessToken = params[kSFOAuthAccessToken];
```
khawkins: I don't know why we didn't change this long ago.  The data returned from OAuth responses basically indicates the need to address/change a data item by virtue of its inclusion in the response data.  I changed all of these to only mess with the value updates if there are actual response values.

PR #1709, opened 2016-08-26
Extensions merge

trooper2013:
Changes related to extension enablement.  Replaced [NSUserDefaults standardUserDefaults]  with
[NSUserDefaults msdkUserDefaults]


khawkins:
Can we rebase, to do away with the upstream unstable merges and flatten the commit history?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSUserDefaults+SFAdditions.h
```diff
@@ -0,0 +1,13 @@
+//
```
khawkins: Update all of these to use the standard Mobile SDK header/copyright from one of the other files.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSUserDefaults+SFAdditions.h
```diff
@@ -0,0 +1,13 @@
...
+//  NSUserDefaults+SFAdditions.h
+//  SalesforceSDKCore
+//
+//  Created by Raj Rao on 8/10/16.
+//  Copyright © 2016 salesforce.com. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface NSUserDefaults (SFAdditions)
```
khawkins: We should document the class and method(s) too.

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer.xcodeproj/project.pbxproj
```diff
@@ -867,6 +867,9 @@
 						CreatedOnToolsVersion = 7.1.1;
 						TestTargetID = 827C631019E6018C00027484;
 					};
+					827C631019E6018C00027484 = {
```
khawkins: Remove specific development team configuration.

khawkins: Actually, seems like this whole file can just be reverted.

libs/SalesforceSDKCore/SalesforceSDKCore/SalesforceSDKCore.h
```diff
@@ -2,7 +2,7 @@
  SalesforceSDKCore.h
  SalesforceSDKCore
 
- Created by Bharath Hariharan on Wed Aug 24 15:55:58 PDT 2016.
+ Created by Raj Rao on Fri Aug 26 10:03:31 PDT 2016.
```
khawkins: Looks like this file should be reverted.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -1069,6 +1073,8 @@
 		4F96FC511BFD32130022F021 /* Common */ = {
 			isa = PBXGroup;
 			children = (
+				B7BAFF491D70ADD50020F756 /* NSUserDefaults+SFAdditions.h */,
```
khawkins: Can we move these to be in alphabetical order with the rest of the code files in this folder?

khawkins:
The other thing we need to consider is the upgrade case.  I.e. if someone moves over to shared settings via app group, we probably need to migrate the previous settings as well.  I'm actually wondering if we need to consider the "downgrade" case too, where someone changes from using an app group to not using one.  Seems like we need data migration in both directions.

khawkins:
...and changing app group names.  There's actually a lot of data management that either needs to be considered from the standpoint of preserving the data, or documenting that data will _not_ be preserved in apps that change their app group configuration after initial data acquisition.

Probably a decision for the team to make.  One of those two paths needs to be considered, though.

bhariharan:
Moved [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1710/).

PR #1710, opened 2016-08-26
Extensions Sharing

trooper2013:
Changes related to extension enablement. Replaced [NSUserDefaults standardUserDefaults] with
[NSUserDefaults msdkUserDefaults]. Incorporated @khawkins  feedback from PR #1709

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
@@ -90,13 +90,12 @@
 		82D4642519C7C8170006BDFE /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 8220B74F16D804CA00EC3921 /* Foundation.framework */; };
 		82D4642619C7C8170006BDFE /* CoreGraphics.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 8280EBCC16E14F7000768DE8 /* CoreGraphics.framework */; };
 		82D4642719C7C8170006BDFE /* UIKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 8280EBB516E14E9F00768DE8 /* UIKit.framework */; };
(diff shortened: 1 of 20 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -1527,6 +1540,7 @@
 			isa = PBXHeadersBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				B76ACC521D5BD75E004BF7DB /* NSUserDefaults+SFAdditions.h in Headers */,
```
bhariharan: Make sure this header is a part of both targets of `SalesforceSDKCore` - the dynamic framework target and the static target.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSUserDefaults+SFAdditions.h
```diff
@@ -0,0 +1,35 @@
+/*
+ Copyright (c) 2015, salesforce.com, inc. All rights reserved.
```
bhariharan: `2016`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSUserDefaults+SFAdditions.m
```diff
@@ -0,0 +1,39 @@
+/*
+ Copyright (c) 2015, salesforce.com, inc. All rights reserved.
```
bhariharan: `2016`.

bhariharan:
Looks pretty good otherwise. We might need some upgrade/migration steps for existing apps.

bhariharan:
Oh, and one more thing - you'll need to add entries for this new file in `SalesforceSDKCore.podspec`, but we can do that later. I think I might have to do that for `AILTN` too, so I'll take care of that.

khawkins:
👍  We can figure out upgrade steps in the next round.

bhariharan:
👍

PR #1712, opened 2016-08-26
fix testAutoPopulateStartTime for 32bit

huminzhi:
Please help review @bhariharan


bhariharan:
This is not going to work. I'll discuss offline @huminzhi

PR #1719, opened 2016-09-06
Fixing app version on user agent

bhariharan:
Turns out the server side parsing depends on the space to break out fields.


wmathurin:
👍

khawkins:
👍

PR #1724, opened 2016-09-07
Fix ci

huminzhi:
include salesforceAnalytics tests
bypass OAuth/Passcode UI tests for now (focus on cloudbee setup now)


bhariharan:
👍

PR #1729, opened 2016-09-08
Replacing network calls with our network stack

bhariharan:
Tested end-to-end. This works.

To be done:

- Add a way to return completion status to the caller, so that it can determine whether to delete events or not (consequence of replacing a synchronous network call with our async stack).

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAILTNPublisher.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAILTNPublisher.m
@@ -39,10 +39,7 @@ static NSString* const kSchemaTypeKey = @"schemaType";
 static NSString* const kData = @"data";
 static NSString* const kLogLines = @"logLines";
 static NSString* const kPayload = @"payload";
-static NSString* const kRestApiPrefix = @"services/data";
-static NSString* const kApiVersion = @"v36.0";
 static NSString* const kRestApiSuffix = @"connect/proxy/app-analytics-logging";
-static NSString* const kBearer = @"Bearer %@";
 
 @implementation SFSDKAILTNPublisher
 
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAILTNPublisher.m
```diff
@@ -53,41 +50,18 @@ + (BOOL) publish:(NSArray *) events {
...
-    return NO;
+    // Adds GZIP compression.
+    NSString *bodyString = [[self class] dictionaryAsJSONString:bodyDictionary];
+    NSData *bodyData = [bodyString dataUsingEncoding:NSUTF8StringEncoding];
+    NSData *postData = [bodyData gzipDeflate];
+    [request setCustomRequestBodyData:postData contentType:@"application/json"];
+    [request setHeaderValue:@"gzip" forHeaderName:@"Content-Encoding"];
+    [request setHeaderValue:[NSString stringWithFormat:@"%lu", (unsigned long)[postData length]] forHeaderName:@"Content-Length"];
+    [[SFRestAPI sharedInstance] send:request delegate:self];
+    return YES;
```
bhariharan: Returning `YES` for now. I need to update the right status based on the delegate callbacks below. I'll probably have to add a completion block for this.

wmathurin: Maybe the network sdk (or SFRestAPI) should handle gzip-ing.
For instance, we could add a property on SFRestRequest and have SFRestAPI look for it and do what you are doing above. Or it could move lower.

wmathurin:
LGTM

bhariharan:
Added status handling [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1729/commits/97f91648f62e0d55e28b34c1d8dcd84f34ac2dad). Should be good to go. Events are showing up on `Splunk`.

bhariharan:
@wmathurin `GZIP` is very endpoint dependent, and I don't know how many of our `Connect` endpoints support `POST` with `GZIP` body.

wmathurin:
Looks like rest api accepts compression for requests and responses across the board (see https://developer.salesforce.com/docs/atlas.en-us.salesforce1api.meta/salesforce1api/apis_rest_best_practices.htm).
Maybe we should compress requests all the time (and we should also accept-encoding:gzip to get compressed responses - which I think we already do).

bhariharan:
Nice! Yes - we accept by default. Maybe we should send by default too. I can create another PR for that.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAILTNPublisher.m
```diff
@@ -53,41 +51,32 @@ + (BOOL) publish:(NSArray *) events {
...
+        finished = YES;
+            if (e) {
+                [SFLogger log:[self class] level:SFLogLevelError format:@"Upload failed %ld %@", (long)[e code], [e localizedDescription]];
+            }
+        } completeBlock:^(id response) {
+            finished = YES;
+            success = YES;
+        }
+    ];
+    while(!finished) {
```
khawkins: Turning a network request into a synchronous block is an anti-pattern, and should have a compelling argument for why it shouldn't be orchestrated in an asynchronous design pattern.  Looking through the code, I'm not seeing that argument.

Why are we running synchronous network requests here?

bhariharan: An asynchronous pattern here adds complexity to the code and some uncertainty in the order of requests/responses. The API contract of the publisher interface requires `publish` to return `BOOL` indicating success of the request. This allows the library to then queue the next publisher and use this status to determine whether to delete stored records or not. If we were to change this to an async pattern using completion blocks, it would make the caller's code a little more messy. For example, do we sequentially line up publishers or fire them all at the same time? Also, when the completion blocks are triggered, which publisher is it coming from? Also, the logic to remove records requires that all publishers were successful. This requires more logic in the completion blocks and counting how many publishers came back with a status (especially if we trigger them parallely). We would be pushing the sync portion one level lower to the `publish` method in the manager and adding code complexity. Hence, I went with the sync approach here, for this special case.

PR #1730, opened 2016-09-09
Remove unused bad API

bhariharan:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKWebUtils.m
```diff
@@ -51,37 +40,4 @@ + (void)configureUserAgent:(NSString *)userAgentString
...
 }
 
-+ (NSString *)currentUserAgentForApp
-{
-    [self stageUserAgentForApp];
-    
-    return gUserAgentForApp;
-}
-
-+ (void)stageUserAgentForApp
```
bhariharan: We don't use this anywhere - I removed it in `4.3.0`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKWebUtils.m
```diff
@@ -51,37 +40,4 @@ + (void)configureUserAgent:(NSString *)userAgentString
...
-    
-    WKWebView *webView = [[WKWebView alloc] initWithFrame:CGRectZero];
-
-   [webView evaluateJavaScript:@"navigator.userAgent"
-             completionHandler:^(id _Nullable val, NSError * _Nullable error) {
-                 gUserAgentForApp = val;
-                 finished = true;
-             }];
-
-   while( !finished ) {
```
bhariharan: Hackish approach alert! This flat out breaks on `iOS 10`. There's no good way to fetch user agent on `iOS` natively though.

khawkins:
👍

PR #1733, opened 2016-09-13
Fixing sample apps to run on iOS 10 simulators

khawkins:
For iOS 10 / Xcode 8, in order to properly access the keychain on the simulator, it seems that you must specify nominal keychain entitlements in the app configuration, to avoid missing keychain entitlements errors.

See [this OpenRadar bug](https://openradar.appspot.com/27422249) for more information.

Verified that the samples will continue to work on Xcode 7 as well, with warnings.


bhariharan:
Nice! 👍

PR #1735, opened 2016-09-14
Consolidate test targets - SalesforceSDKCore

khawkins:
Floating a trial balloon with SalesforceSDKCore, before I do the other frameworks.

Re-aligning unit test bundles directly with their frameworks under test, without the "unit test app" intermediary.  The previous setup was causing xcodebuild issues against the test app scheme, and this approach is more aligned with how framework projects roll out unit test bundles by default anyway.

NB: There are test failures here, but they don't seem to be associated with the move, as far as I can tell.


wmathurin:
Looks fine as far as I can tell.

PR #1737, opened 2016-09-14
Fixing up some OAuth, test issues

khawkins:
- Removing `[SFOAuthCredentials init]`, as its only purpose is to enable unit tests to not call the other initializers.
- Removing test to validate that the aforementioned `init` implementation was insufficient for object initialization.
- Fixing up some test data.


bhariharan:
👍

PR #1741, opened 2016-09-15
Revert "Consolidate test targets - SalesforceSDKCore"

khawkins:
I'm finding that the reliability of facilities such as `NSUserDefaults` is really problematic in the independent test bundle scenario (i.e. testing directly against frameworks), vs. testing against an intermediate app bundle container (i.e. testing the framework as part of an app's associated unit test setup).  So it would seem this change is not what we want at this time.


libs/SalesforceSDKCore/SalesforceSDKCoreTestApp/SalesforceSDKCoreTestApp-Prefix.pch
```diff
@@ -0,0 +1,17 @@
+//
+//  Prefix header
+//
+//  The contents of this file are implicitly included at the beginning of every source file.
+//
+
+#import <Availability.h>
+
+#ifndef __IPHONE_5_0
```
bhariharan: Whoa - `iPhone 5.0`!

khawkins: Heh, pretty old-school.  Newly generated PCH files are simplier:
```obj-c
#ifndef SalesforceSDKCoreTests_Prefix_pch
#define SalesforceSDKCoreTests_Prefix_pch

#import <SalesforceSDKCore/SFLogger.h>

#endif /* SalesforceSDKCoreTests_Prefix_pch */
```

khawkins:
All tests are passing.  I'm going to merge this in.

PR #1744, opened 2016-09-16
add back unit tests

huminzhi:


SalesforceMobileSDK.xcworkspace/xcshareddata/xcschemes/UnitTests.xcscheme
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <Scheme
-   LastUpgradeVersion = "0730"
+   LastUpgradeVersion = "0800"
```
bhariharan: @huminzhi I don't think we should update this just yet. Keep this as `0730`.

huminzhi: sure thing, let me revert back

PR #1745, opened 2016-09-20
Switching back to WKWebView as default

bhariharan:
Once we merge this and give it a few days to bake, if no one sees any issues, I'll send another PR to remove `UIWebView` altogether.


wmathurin:
👍

PR #1746, opened 2016-09-20
Bumping up Xcode and CocoaLumberjack versions

bhariharan:


bhariharan:
There seem to be no API changes in `CocoaLumberjack 2.4.0`. Other than minor bug fixes, the only big change I see is tying it down to `Swift 2.3` to enable `Xcode 8` compatibility. Sample apps are working and everything builds fine. FYI @trooper2013

PR #1748, opened 2016-09-20
Removing UIWebView support

bhariharan:


bhariharan:
Hybrid apps `AccountEditor` and `SmartSyncExplorerHybrid` are working fine.

robbmcmahan:
@bhariharan we have a UIWebview fallback plan and need to keep that capability

bhariharan:
@robbmcmahan That'll need to be implemented in the subclass in `S1`. With ATS restrictions coming, Apple's official stance is deprecation of `UIWebView` and ATS won't work there, so we won't be supporting it officially either.

bhariharan:
@robbmcmahan Also, the right hooks are available for `S1` to implement `UIWebViewDelegate` in a subclass.

wmathurin:
:+1:

robbmcmahan:
@bhariharan what about needing to pass the engine we are going to use to MSDK? Can we still pass the UIWebView engine?

bhariharan:
@robbmcmahan That API still exists [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1748/files#diff-bafe72208f1ff3afa7b2024a36312efeL198).

robbmcmahan:
ok, so this doesn't stop us from passing in the UI engine and using our own UIWebView. Gotcha

khawkins:
Looks fine to me.  There are some failing hybrid tests, but I'm not sure if it's related.

bhariharan:
All tests are passing, except a few in `SmartSyncPlugin`. Seems unrelated to this PR.

PR #1749, opened 2016-09-20
Bumping up min version to iOS 9

bhariharan:


install.sh
```diff
@@ -8,7 +8,7 @@
 
 # Check for iOS SDK minimum version
 IOS_MIN_VERSION_NUM=80
```
khawkins: Need this guy too.

wmathurin:
In the templates (e.g. https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/build/app_template_files/__NativeTemplateAppName__/Podfile#L2) should the commented out line say 9.0 also?

bhariharan:
Yup - I updated those already [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1749/files#diff-ff945aa11e522bee914a5ab75cf9adccL2). Find and replace ;-)

PR #1750, opened 2016-09-21
Refresh sync down

wmathurin:


libs/SmartSync/SmartSync/Classes/Util/SFRefreshSyncDownTarget.m
```diff
@@ -0,0 +1,259 @@
...
+    }
+    
+    __block NSMutableArray* remoteIds = [NSMutableArray new];
+    __block NSUInteger sliceSize = self.countIdsPerSoql;
+    __block NSUInteger countSlices = ceil((float)localIds.count / sliceSize);
+    __block NSUInteger slice = 0;
+    __block NSString* idFieldName = self.idFieldName;
+    __block SFSyncDownTargetFetchCompleteBlock fetchBlockRecurse = ^(NSArray *records) {};
+    SFSyncDownTargetFetchCompleteBlock fetchBlock = ^(NSArray* records) {
+        // NB with the recursive block, using weakSelf doesn't work (it goes to nil)
```
khawkins: This can/will happen with `__weak` if the owning container object—in this case, the instance of `SFRefreshSyncDownTarget`—gets released before one or more self references inside the block are dereferenced.  This is expected behavior, as classes aren't guaranteed to stay around inside a block, if the proper reference doesn't exist.

Weakifying/strongifying `self` inside of blocks is generally only necessary if the block *itself* will be retained on its owning object container.  That's not the case here.  But consider:

```obj-c
@interface BlockOwningClass ()

@property (nonatomic, copy) void (^myCopiedBlock)();
@property (nonatomic, copy) NSString *myString;

@end

@implementation BlockOwningClass

- (void)setAndCallMyBlock {
    self.myString = @"A string";
    self.myCopiedBlock = ^{
        NSLog(@"self.myString: %@", self.myString);
    }
    self.myCopiedBlock();
}

@end
```

This is likely to force a retain cycle, because the ownership of the block itself bumps the retain count, then the block captures `self`.

So the short answer here is, "capturing" `self` is okay in this case, and if any issue exists, it's possibly the expectation that the parent class should still be around/retained (somewhere) at the time this block is being called?

libs/SmartSync/SmartSync/Classes/Util/SFSoqlSyncDownTarget.m
```diff
@@ -147,27 +150,27 @@ - (void) getListOfRemoteIds:(SFSmartSyncSyncManager*)syncManager
     NSString* fromClause = [self.query substringFromIndex:rangeFirst.location];
     [soql appendString:fromClause];
     __block NSUInteger countFetched = 0;
-    __block NSUInteger totalSize = 0;
     __block NSMutableArray* allRecords = [[NSMutableArray alloc] init];
-    __block SFSyncDownTargetFetchCompleteBlock completionBlockRecurse = ^(NSArray *records) {};
-    __weak SFSoqlSyncDownTarget* weakSelf = self;
-    SFSyncDownTargetFetchCompleteBlock completionBlock = ^(NSArray* records) {
-        totalSize = self.totalSize;
+    __block SFSyncDownTargetFetchCompleteBlock fetchBlockRecurse = ^(NSArray *records) {};
```
khawkins: In addition to my previous comments, it wouldn't be bad for this to be a `__weak` reference as well, as it *is* possible to capture the block in a retain cycle with this pattern, otherwise.

libs/SmartSync/SmartSync/Classes/Util/SFMruSyncDownTarget.m
```diff
@@ -87,14 +87,15 @@ - (void) startFetch:(SFSmartSyncSyncManager*)syncManager
     __weak SFMruSyncDownTarget *weakSelf = self;
     SFRestRequest *request = [[SFRestAPI sharedInstance] requestForMetadataWithObjectType:self.objectType];
     [SFSmartSyncNetworkUtils sendRequestWithSmartSyncUserAgent:request failBlock:errorBlock completeBlock:^(NSDictionary* d) {
-        NSArray* recentItems = [weakSelf pluck:d[kRecentItems] key:self.idFieldName];
+        __strong typeof(self) strongSelf = weakSelf;
+        NSArray* recentItems = [strongSelf pluck:d[kRecentItems] key:self.idFieldName];
         NSString* inPredicate = [@[ self.idFieldName, @" IN ('", [recentItems componentsJoinedByString:@"', '"], @"')"]
                                  componentsJoinedByString:@""];
         NSString* soql = [[[[SFSmartSyncSoqlBuilder withFieldsArray:self.fieldlist]
```
wmathurin: Good catch - I did weakSelf->strongSelf string replacements but didn't check for self.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -275,46 +277,51 @@ - (void) syncDown:(SFSyncState*)sync updateSync:(SyncUpdateBlock)updateSync fail
...
-            updateSync(nil, totalSize == 0 ? 100 : 0, totalSize, kSyncManagerUnchanged);
-            if (totalSize == 0) {
-                return;
+        updateSync(nil, totalSize == 0 ? 100 : 0, totalSize, kSyncManagerUnchanged);
+        if (totalSize != 0) continueFetchBlockRecurse(records);
+    };
+
+    SFSyncDownTargetFetchCompleteBlock continueFetchBlock = ^(NSArray* records) {
+        __strong SFSmartSyncSyncManager *strongSelf = weakSelf;
+        if (records != nil) {
```
wmathurin: The sync down target is supposed to return nil when down - it might return empty arrays before (without being done) - for instance for a refresh-sync-down, we do soql calls with ids we found locally, the ids might be gone from the server and therefore some of the soql calls could possibly return no records.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -275,46 +277,51 @@ - (void) syncDown:(SFSyncState*)sync updateSync:(SyncUpdateBlock)updateSync fail
 
     __block NSUInteger countFetched = 0;
     __block NSUInteger totalSize = 0;
-    __block SFSyncDownTargetFetchCompleteBlock completeBlockRecurse = ^(NSArray *records) {};
+    __block NSUInteger progress = 0;
+    __block SFSyncDownTargetFetchCompleteBlock continueFetchBlockRecurse = ^(NSArray *records) {};
     __weak SFSmartSyncSyncManager *weakSelf = self;
```
wmathurin: I'll look for all the weakSelf/strongSelf declarations in the SDK

libs/SmartSync/SmartSync/Classes/Util/SFSoqlSyncDownTarget.m
wmathurin: If a do that Xcode complains that "assigning block literal to a weak variable; object will be released after assignment".

wmathurin:
I've addressed (most) comments with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1750/commits/4c1403f354d24f565e154ff91125155f6fc11da9 and https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1750/commits/c6831e401290b22982b29290d6d471ee80f40228

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
khawkins: Ah okay.

libs/SmartSync/SmartSync/Classes/Util/SFSoqlSyncDownTarget.m
khawkins: Oh weird.  I was playing with a similar setup and got the opposite: "you'll have a retain cycle".  I'm happy to let the compiler know best. :)  Blocks are tricky.

khawkins:
Looking good to me.

wmathurin:
@bhariharan can I merge?

bhariharan:
Looks good to me.

PR #1752, opened 2016-09-22
Strong/weak self

wmathurin:
* Always declaring weakSelf with: __weak typeof(self) weakSelf = self;
* Always declaring strongSelf with: __strong typeof(weakSelf) strongSelf = weakSelf;
* Using strongSelf where appropriate (see https://dhoerl.wordpress.com/2013/04/23/i-finally-figured-out-weakself-and-strongself/)


khawkins:
Cool, thanks for updating. 👍

PR #1753, opened 2016-09-22
Adding launch images for sample apps

bhariharan:
This removes the black band on top and bottom on `iOS 10`.


bhariharan:
@huminzhi FYI....this fixes the black border issue. I'll add launch images for the other sample apps too.

bhariharan:
No code changes.

PR #1756, opened 2016-09-26
fix Analytics unit tests for mulitple events

huminzhi:


libs/SalesforceAnalytics/SalesforceAnalyticsTests/EventStoreManagerTests.m
```diff
@@ -105,23 +105,6 @@ - (void) testFetchOneEvent {
 }
 
 /**
- * Test for fetching all stored events.
- */
-- (void) testFetchAllEvents {
```
huminzhi: This test is same as testStoreMultipleEvents, so I removed this one, and rename testStoreMultipleEvents to testStoreAndFetchMultipleEvents

libs/SalesforceAnalytics/SalesforceAnalyticsTests/EventStoreManagerTests.m
```diff
@@ -87,8 +87,8 @@ - (void) testStoreMultipleEvents {
     NSArray<SFSDKInstrumentationEvent *> *events = [self.storeManager fetchAllEvents];
     XCTAssertTrue(events != nil, @"List of events should not be nil");
     XCTAssertEqual(2, events.count, @"Number of events stored should be 2");
-    XCTAssertTrue([event1 isEqual:[events firstObject]], @"Stored event should be the same as generated event");
```
huminzhi: when fetch events from file system, it's not ordered by time stamp, so cannot guarantee the events are as the same sequence as created, thus use [events containsObject: ] to make sure the event exists in the store

libs/SalesforceAnalytics/SalesforceAnalyticsTests/EventStoreManagerTests.m
```diff
@@ -156,8 +139,6 @@ - (void) testDeleteMultipleEvents {
     NSArray<SFSDKInstrumentationEvent *> *eventsBeforeDel = [self.storeManager fetchAllEvents];
     XCTAssertTrue(eventsBeforeDel != nil, @"List of events should not be nil");
     XCTAssertEqual(2, eventsBeforeDel.count, @"Number of events stored should be 2");
-    XCTAssertTrue([event1 isEqual:[eventsBeforeDel firstObject]], @"Stored event should be the same as generated event");
```
huminzhi: this flow was verified in previous test, no need to check here again.

bhariharan:
👍

PR #1759, opened 2016-09-27
Implementing NSCopying for instrumentation event

khawkins:
For safety/convention during I/O.

### DIFF
--- a/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEvent.m
+++ b/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEvent.m
@@ -133,6 +133,56 @@
     return self;
 }
 
+#pragma mark - NSCopying
+
+- (instancetype)copyWithZone:(NSZone *)zone {
+    SFSDKInstrumentationEvent *eventCopy = [[[self class] allocWithZone:zone] init];
+    eventCopy.eventId = self.eventId;
+    eventCopy.startTime = self.startTime;
+    eventCopy.endTime = self.endTime;
+    eventCopy.name = self.name;
+    eventCopy.attributes = self.attributes;
+    eventCopy.sessionId = self.sessionId;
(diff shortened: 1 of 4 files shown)


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEvent.m
```diff
@@ -133,6 +133,56 @@ - (instancetype) initWithJson:(NSData *) jsonRepresentation {
     return self;
 }
 
+#pragma mark - NSCopying
+
+- (instancetype)copyWithZone:(NSZone *)zone {
```
khawkins: This is the only new method here.  The other changes in this file are just moving methods around to group them.

libs/SalesforceAnalytics/SalesforceAnalyticsTests/InstrumentationEventBuilderTests.m
```diff
@@ -151,41 +158,37 @@ - (void) testAutoPopulateStartTime {
...
+
+#pragma mark - Helper methods
+
+- (SFSDKInstrumentationEvent *)standardTestEvent {
+    return [SFSDKInstrumentationEventBuilder buildEventWithBuilderBlock:^(SFSDKInstrumentationEventBuilder *builder) {
         double curTime = 1000 * [[NSDate date] timeIntervalSince1970];
         NSString *eventName = [NSString stringWithFormat:kTestEventName, curTime];
         builder.name = eventName;
         builder.page = [[NSDictionary alloc] init];
+        builder.attributes = @{ @"Name1": @"Value1" };
```
bhariharan: Why `attributes`? This is unrelated, I presume?

khawkins: Ah, I was adding something for `attributes` just to have it populated, testing some copy scenarios.  This doesn't need to be here.  I'll remove it.

PR #1762, opened 2016-09-30
test fix and stabilitze

huminzhi:


shared/test/test/SalesforceNoSessionTestCase.swift
```diff
@@ -57,6 +58,7 @@ class SalesforceNoSessionTestCase: XCTestCase {
     
     override func tearDown() {
         super.tearDown()
+        searchScreen.logout()
```
huminzhi: always logout in teardown so that next test won't get impact by previous test

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
```diff
@@ -314,7 +314,7 @@ - (void)testCreateQuerySearchDelete {
 // - query updated object (requestForSearch) and make sure we don't get anything
 - (void)testCreateUpdateQuerySearchDelete {
     // create object
-    NSString *lastName = [NSString stringWithFormat:@"Doe-%@", [NSDate date]];
```
huminzhi: the "+" in the lastName is causing issue in query

khawkins: But shouldn't we correct that issue?  There's no reason there can't be a '+' in the query, right?

huminzhi: @khawkins so an alternative fix can be made on line 334 to correctly escape the query, but it's just another way to fix test issue. It's not a real bug..

khawkins: This is our bug, because the encoding of the query is the responsibility of the Mobile SDK.  It's actually a regression brought in from [this PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1754).

@bhariharan , in `CSFPrivateDefines` (and elsewhere in that PR), we've replaced the usage of `CFURLCreateStringByAddingPercentEscapes()` with `stringByAddingPercentEncodingWithAllowedCharacters:`, which is fine, but `[NSCharacterSet URLQueryAllowedCharacterSet]` isn't a complete set of characters to escape, unfortunately.  We should be doing a superset of that set and what we had before.

From [this SO answer](http://stackoverflow.com/a/29810106), I'd say that set looks like `#%<>[\]^``{|}&:/=+`.

bhariharan: Good catch @khawkins! @huminzhi We should probably close this PR and reopen the bug you filed :-)

PR #1765, opened 2016-10-04
Fixed Issue with CSFAction Equals implementation

trooper2013:
Added base URL to the equals implementation for CSFAction.


huminzhi:
wonderful! Thank you for also adding tests @trooper2013 !

PR #1772, opened 2016-10-13
Allow extensible category

bhariharan:
No code changes. Just broke it down, so that internal extensions can publish log lines directly.


sgoldberg-sfdc:
👍

khawkins:
👍

PR #1774, opened 2016-10-18
Use Fastlane/Danger for CI

huminzhi:


readme.md
```diff
@@ -1,4 +1,4 @@
-[![Build Status](https://travis-ci.org/forcedotcom/SalesforceMobileSDK-iOS.svg?branch=unstable)](https://travis-ci.org/forcedotcom/SalesforceMobileSDK-iOS)
+[![Build Status](https://gist.githubusercontent.com/huminzhi/74e0b81003314709ded9c23386ec28ba/raw/status.png)](https://mobilesdk.ci.data.com/job/MobileSDK-iOS)
```
bhariharan: Is this link externally accessible? Otherwise, this won't work.

huminzhi:
@bhariharan I revert back the readme as we need to discuss further, but the other changes are still valid. Please review. Thanks!

fastlane/Fastfile
```diff
@@ -0,0 +1,37 @@
+lane :CI do
```
bhariharan: @huminzhi Will this work with `Travis CI` too, if we decide to go with that?

DangerFile
```diff
@@ -0,0 +1,9 @@
+junit.parse "test_output/report.junit"
```
bhariharan: @huminzhi Will this work with `Travis CI` too, if we decide to go with that?

huminzhi: Yes, danger/fastlane are working with Travis CI also..

PR #1776, opened 2016-10-19
Salesforce SDK core pod spec cleanup

wmathurin:
base = arc-enabled source files required by no-arc
no-arc = source files that cannot be compiled with arc
arc = the rest
arc's excluded files = base + no-arc

**arc's public header is populated by the update_podspec_headers.sh automatically**
**We should no longer need to edit the pod spec manually**

Also updating deployment targets of template apps to 9.0
(NB: still to do, add entitlements so that they work in 10.0 simulator)


SalesforceSDKCore.podspec
```diff
@@ -25,9 +25,8 @@ Pod::Spec.new do |s|
       sdkcore.resource = 'shared/resources/SalesforceSDKAssets.xcassets'
 
       sdkcore.subspec 'base' do |sp|
```
bhariharan: Maybe we should add a comment about what goes in `base`, what goes in `arc`, etc., so we remember next time around? Basically, the description on this PR.

bhariharan:
Otherwise, looks good.

wmathurin:
@bhariharan good idea - done here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1776/commits/411d8c14dfe8aef21a8e073c5a8f6c4ac88d20fb

khawkins:
The only thing I don't know is if the podspec is going to be problematic including the public header files of the other two subspecs twice.  The update public headers script looks at _all_ of the public headers for the project.  Which means that blanket-including them in the `arc` subspec ensures that those public headers will be a superset of all the subspec public headers.

@wmathurin  does this cause any issues when you try to lint the podspec, or consume it?

khawkins:
And as an example, in case that note suffers from too much morning incoherence. 😉  `SFLogger.h` is included in the public headers for both the `base` subspec and `arc` subspec.

wmathurin:
I get it, I have the same concern.
I have not pod spec lint-ed it yet 
I have successfully consumed it in a pod generated application.

I have however just run into an issue when using use_frameworks!
More testing is needed.

At the end, I might have to make the update_podspec_headers.h smarter (it should make sure to remove from the list of headers any that are found in excluded files).

khawkins:
> I might have to make the update_podspec_headers.h smarter

Yeah, that's originally what I had wanted to do...but didn't come up with the cycles to do it.  We had also talked about moving away from the non-ARC keychain code in the 5.0 timeframe, which would obviate the need for this separation entirely.  Worth considering.

bhariharan:
Yes, I was going to arc-ify the Keychain wrapper class in 5.0, but haven't found the cycles yet. Maybe we should just bite the bullet and do that.

wmathurin:
I agree that arc-ifying the keychain wrapper is the way to go.
For now, I have modified the update_podspec_headers.h to  prevent duplication of headers between subspecs in SalesforceSDKCore.

The script makes sure to remove any headers that is part of excluded_files from the public_header_files list. See
https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1776/commits/361e6bd4c2a773f63f9e2bc6fd2fcf223f2ac159

bhariharan:
👍

bhariharan:
👍

khawkins:
👍

PR #1778, opened 2016-10-24
Don't remove security view on app foreground event.

jhneves:
This hook to dismiss the security view is faulty in scenarios where the app enters foreground but never becomes active (e.g. launch the app, then double tap the home button before the app actually completes the launch: the app will have entered foreground but never became active).

Dismissing the security view when the app enters foreground is not safe. It needs to wait until it the app actually becomes active.

@bhariharan @khawkins


huminzhi:
Can one of the admins verify this patch?

PR #1781, opened 2016-10-25
add lane for local

huminzhi:
Add a fast lane for sf dev local use


fastlane/Fastfile
```diff
@@ -35,3 +35,27 @@ lane :CI do
 	end
 
 end
+
+
+lane  :localtest do
+#setup test accounts (for sf dev only)
+	`git clone https://git.soma.salesforce.com/gist/d103e1ef29e1c11ee9a7a4733a1193d8.git`
```
huminzhi: these test credentials are our test org users, if we don't mind sharing them, we can put on public repo, however, I thought we don't want to share our test accounts to public..
I just want to make the life easier for us to run tests locally, because I know @trooper2013 has to update the test_account.json file every time before run unit tests, which is just not so convenient.. 
I am totally fine not to merge this change, but I will recommend our team: @trooper2013 @wmathurin @khawkins @helenren to make this change locally for your own convenience..

huminzhi: Sounds good.. @bhariharan , I am closing this pr then..

PR #1793, opened 2016-10-31
Reworking AILTN publisher to use async model

bhariharan:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAnalyticsPublisher.h
```diff
@@ -31,12 +31,14 @@
...
+
 /**
  * Publishes events to a network endpoint.
  *
  * @param events Events to be published.
- * @return True - if successful, False - otherwise.
+ * @param publishCompleteBlock Completion block invoked once network publish is complete.
  */
-+ (BOOL) publish:(nonnull NSArray *) events;
++ (void) publish:(nonnull NSArray *) events publishCompleteBlock:(nonnull PublishCompleteBlock) publishCompleteBlock;
```
bhariharan: I changed this API to pass in a completion block because it's too hard to make synchronous networking requests on `iOS`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAILTNPublisher.m
```diff
@@ -65,21 +66,14 @@ + (BOOL) publishLogLines:(NSDictionary *) bodyDictionary {
...
         if (e) {
             [SFLogger log:[self class] level:SFLogLevelError format:@"Upload failed %ld %@", (long)[e code], [e localizedDescription]];
         }
+        publishCompleteBlock(NO);
     } completeBlock:^(id response) {
-        finished = YES;
-        success = YES;
+        publishCompleteBlock(YES);
     }];
-    while(!finished) {
```
bhariharan: This is primarily what I wanted to get rid of. This run loop can potentially block a thread forever. I was using it just to pass synchronous result back, but I refactored this to call the completion block instead.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAnalyticsPublisher.h
```diff
@@ -31,12 +31,14 @@
 
 @protocol SFSDKAnalyticsPublisher <NSObject>
 
+typedef void (^ _Nonnull PublishCompleteBlock)(BOOL success);
```
bhariharan: New completion block signature.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -272,6 +278,22 @@ - (void) publishOnAppBackground {
     });
 }
 
+- (void) applyTransformAndPublish:(Class<SFSDKTransform>) curTransform events:(NSArray<SFSDKInstrumentationEvent *> *) events publishCompleteBlock:(PublishCompleteBlock) publishCompleteBlock {
```
bhariharan: New helper function to transform and publish.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -154,40 +154,46 @@ - (void) publishEvents:(NSArray<SFSDKInstrumentationEvent *> *) events {
...
-                        success = networkSuccess;
-                    }
+        for (SFSDKInstrumentationEvent *event in events) {
+            [eventIds addObject:event.eventId];
+        }
+        __block BOOL overallSuccess = YES;
+        __block BOOL overallCompletionStatus = NO;
+        __block NSMutableArray<Class<SFSDKTransform>> *remoteKeySet = [[self.remotes allKeys] mutableCopy];
+        __block Class<SFSDKTransform> curTransform = [remoteKeySet objectAtIndex:0];
+        PublishCompleteBlock publishCompleteBlock = ^void(BOOL success) {
```
bhariharan: Handling post processing in completion block now.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAnalyticsPublisher.h
khawkins: Should we add the Error too, for the case where there is an error?  The consumer could make use of it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
khawkins: This isn't getting called initially.  The process isn't going to be fired off.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -154,40 +154,46 @@ - (void) publishEvents:(NSArray<SFSDKInstrumentationEvent *> *) events {
...
+
+            /*
+             * Updates the success flag only if all previous requests have been
+             * successful. This ensures that the operation is marked success only
+             * if all publishers are successful.
+             */
+            if (success) {
+                overallSuccess = success;
```
khawkins: Unless I'm reading this wrong, `overallSuccess` is never _not_ going to be true.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -154,40 +154,46 @@ - (void) publishEvents:(NSArray<SFSDKInstrumentationEvent *> *) events {
...
+            if (!overallCompletionStatus) {
+                curTransform = [remoteKeySet objectAtIndex:0];
+                [self applyTransformAndPublish:curTransform events:events publishCompleteBlock:publishCompleteBlock];
+            } else {
+
+                /*
+                 * Deletes events from the event store if the network publishing was successful.
+                 */
+                if (overallSuccess) {
+                    [self.eventStoreManager deleteEvents:eventIds];
```
khawkins: The model is a little unclear to me, in this aggregate case.  It seems like, if some events fail in their publishing, none of the events will be deleted.  Which means that some events would conceivably be published twice?  Is that the intention?

bhariharan: Events are batched and sent as a single request, so either the request was successful or it failed. However, since we support multiple endpoints, it's possible that one request was successful but another failed. In this case, we will not delete events. They might be sent twice to a previously successful endpoint, but that's acceptable as opposed to loss of data on another endpoint/service.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
bhariharan: Yeah, I didn't commit some local changes. Coming up.

bhariharan:
Passing `error` back could be useful. I'll take a look at adding that in.

bhariharan:
@khawkins Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1793/commits/b42e3576e5b212c6303492e6d9a369d193719c6b).

khawkins:
Cool, LGTM.

PR #1794, opened 2016-11-01
App feature usage info in UserAgent

ivanbogdanov:


huminzhi:
Can one of the admins verify this patch?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -41,6 +41,13 @@ typedef NS_ENUM(NSUInteger, SFAppType) {
     kSFAppTypeReactNative
 };
 
+typedef NS_ENUM(NSUInteger, SFAppFeature) {
```
wmathurin: I don't think SalesforceSDKCore should know about the other libraries like SmartStore etc.
So instead appFeature could be a string.

bhariharan: I agree.

ivanbogdanov: Fixed

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -79,6 +84,7 @@ + (instancetype)sharedManager
     static SalesforceSDKManager *sdkManager = nil;
     dispatch_once(&pred , ^{
         uid = [[[UIDevice currentDevice] identifierForVendor] UUIDString];
+        features = [NSMutableSet setWithObjects:kSFAppFeatureOAuth, nil];
```
wmathurin: Should you set the oauth feature from the oauth code - I guess it's possible to do anonymous (unauthenticated) calls to the server.

khawkins: Agreed.  The OAuth setting should follow the same pattern that SmartStore does, as a consumer.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -41,6 +41,11 @@
 // Device id
 static NSString* uid = nil;
 
+static NSMutableSet *features;
```
khawkins: This variable is only ever used in the singleton context.  It should be represented as a property on the singleton.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -41,6 +41,11 @@
...
+
+NSString * const kSFAppFeatureOAuth = @"UA";
```
khawkins: This can be a `static` constant, if it's not going to be external.

ivanbogdanov:
Fixed. `UA` is reported on auth event

ivanbogdanov:
Fixed.

sgoldberg-sfdc:
Is there somewhere this concept of app feature is explained or documented?

ivanbogdanov:
Done: Removed extra empty line from `SFOAuthCoordinator.m`

ivanbogdanov:
@wmathurin @bhariharan

wmathurin:
👍

PR #1805, opened 2016-11-08
register Push Notifications App Feature

ivanbogdanov:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
```diff
@@ -103,6 +107,7 @@ + (SFPushNotificationManager *) sharedInstance
     dispatch_once(&pred, ^{
         mgr = [[super allocWithZone:nil] init];
     });
+    [[SalesforceSDKManager sharedManager] registerAppFeature:kSFAppFeaturePushNotifications];
```
bhariharan: I think this should be inside the `dispatch_once`, since we don't need it more than once.

ivanbogdanov:
Fixed

PR #1809, opened 2016-11-09
registering MDM App Feature

ivanbogdanov:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
```diff
@@ -78,6 +82,9 @@ - (id)init {
                                                           weakSelf.rawPreferences = [[NSUserDefaults standardUserDefaults] dictionaryForKey:kManagedConfigurationKey];
```
khawkins: Can you make the change below here, as well?  This notification handler refreshes the MDM preferences when user defaults values change.

ivanbogdanov: Fixed.

PR #1811, opened 2016-11-10
detect swift and report as app feature

ivanbogdanov:


ivanbogdanov:
@bhariharan @khawkins

PR #1812, opened 2016-11-10
Fix for  Failing tests

trooper2013:
Event for registersoup should  be done after the soup has been materialized ( tables & indices created)

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
@@ -1250,17 +1250,7 @@ NSString *const EXPLAIN_ROWS = @"rows";
     if (soupUsesExternalStorage && soupUsesJSON1) {
         @throw [NSException exceptionWithName:@"Can't have JSON1 index specs in externally stored soup" reason:nil userInfo:nil];
     }
-    NSMutableArray<NSString *> *features = [[NSMutableArray alloc] init];
-    if (soupUsesJSON1) {
-        [features addObject:@"JSON1"];
-    }
-    if (soupUsesExternalStorage) {
-        [features addObject:@"ExternalStorage"];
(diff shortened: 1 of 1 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -1351,6 +1341,19 @@ - (void)registerSoupWithSpec:(SFSoupSpec*)soupSpec withIndexSpecs:(NSArray*)inde
...
     
+    // LogEvent
+    NSMutableArray<NSString *> *features = [[NSMutableArray alloc] init];
+    if (soupUsesJSON1) {
+        [features addObject:@"JSON1"];
+    }
+    if (soupUsesExternalStorage) {
+        [features addObject:@"ExternalStorage"];
+    }
+    if ([self hasFts:soupSpec.soupName withDb:db]) {
```
wmathurin: Instead, we could define the following (right after defining soupUsesJSON1):
```objective-c
BOOL soupUsesFts = [SFSoupIndex hasFts:indexSpecs];
```
Then there is no need to move the block of code.

trooper2013: I did think of it.  I guess the logeevent should move then too right.
``` 
[[self class] logAnalyticsEventWithName:@"registerSoup" userAccount:self.user storeAttributes:nil features:features];
```
 The question then is if we should persist the event after completion of registration. Additionally  event logging code in one continuos code block vs. distributed over.

wmathurin: You are right, we should log the event if the registration successfully goes through not before.
For soupUsesFts, calling the static hasFts method on SFSoupIndex is the easiest/cheapest way.

PR #1814, opened 2016-11-10
Standardizing and using the new helper across the board

bhariharan:


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -284,14 +283,10 @@ - (void)authenticateWithCompletionBlock:(SFOAuthPluginAuthSuccessBlock)completio
...
             [self log:SFLogLevelError msg:@"OAuth plugin authentication request failed. Logging out."];
-            SFSDKSalesforceAnalyticsManager *manager = [SFSDKSalesforceAnalyticsManager sharedInstanceWithUser:[SFUserAccountManager sharedInstance].currentUser];
-            SFSDKInstrumentationEvent *event = [SFSDKInstrumentationEventBuilder buildEventWithBuilderBlock:^(SFSDKInstrumentationEventBuilder *builder) {
-                builder.name = [NSString stringWithFormat:@"Server Error: %ld, Logout Cause: %@", (long)error.code, error.localizedDescription];
-                builder.page = @{ @"context" : @"Login Authentication"};
-                builder.schemaType = SchemaTypeInteraction;
-                builder.eventType = EventTypeUser;
-            } analyticsManager:manager.analyticsManager];
-            [manager.analyticsManager.storeManager storeEvent:event];
+            NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
```
khawkins: For conciseness, why not just create and/or inline a literal `NSDictionary` with these values?

bhariharan: We could. I did this because I can see the list of `attributes` we want to capture growing in the future, which might make inline creation less readable. But I'm okay either way. I can change it to inline.

PR #1815, opened 2016-11-11
Adding feature for AILTN

bhariharan:


bhariharan:
@ivanbogdanov

ivanbogdanov:
Looks great! Thank you

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -44,6 +44,7 @@
 static NSString * const kEventStoresDirectory = @"event_stores";
 static NSString * const kEventStoreEncryptionKeyLabel = @"com.salesforce.eventStore.encryptionKey";
 static NSString * const kAnalyticsOnOffKey = @"ailtn_enabled";
+static NSString * const kSFAppFeatureAiltnEnabled = @"AI";
```
khawkins: Just thinking about this over the course of these additions: it seems like these constants should be owned/aggregated in a common location somewhere.  I.e. it'd be nice to go to a single `SFSDKFeatureConstants.h` file and see the delineation of all possible feature registrations that the SDK exposes.  Thoughts?

bhariharan: Yes, I agree with the idea. However, we can't have a higher level library's constants owned by a lower library, or vice versa. I guess middle ground would be for each library to have a single point of ownership.

khawkins: Well, we _could_ have the former.  I.e. they could be denormalized, since they don't otherwise compromise referential integrity—they're just strings, at the end of the day.

bhariharan: That's true. I'll merge this for now if that's okay and leave it to @ivanbogdanov, since he's working on adding app features.

khawkins: 👍

PR #1825, opened 2016-11-14
Removing ATS Exception Entries

bhariharan:


bhariharan:
Tested sample apps against `iOS 10` and `Sandbox`. Appears to work fine without `ATS` exceptions.

PR #1826, opened 2016-11-14
Comply with Cordovas use of default C-99 C language dialect

trooper2013:


shared/hybrid/AppDelegate+SalesforceHybridSDK.m
```diff
@@ -59,23 +59,23 @@ - (AppDelegate *)sfsdk_swizzled_init
     // Need to use SalesforceSDKManagerWithSmartStore when using smartstore
     [SalesforceSDKManager setInstanceClass:[SalesforceSDKManagerWithSmartStore class]];
     [SalesforceSDKManager sharedManager].appConfig = appConfig;
-    __weak typeof(self) weakSelf = self;
+    __weak __typeof(self) weakSelf = self;
     [SalesforceSDKManager sharedManager].postLaunchAction = ^(SFSDKLaunchAction launchActionList) {
-        __strong typeof(weakSelf) strongSelf = weakSelf;
-        [strongSelf log:SFLogLevelInfo format:@"Post-launch: launch actions taken: %@", [SalesforceSDKManager launchActionsStringRepresentation:launchActionList]];
-        [strongSelf setupRootViewController];
+      __strong __typeof(weakSelf) strongSelf = weakSelf;
```
khawkins: The spacing went wonky from the standard spacing that Xcode uses.  Can you Editor -> Structure -> Re-Indent these?

shared/hybrid/AppDelegate+SalesforceHybridSDK.m
```diff
@@ -59,23 +59,23 @@ - (AppDelegate *)sfsdk_swizzled_init
...
+      [strongSelf log:SFLogLevelInfo format:@"Post-launch: launch actions taken: %@", [SalesforceSDKManager launchActionsStringRepresentation:launchActionList]];
+      [strongSelf setupRootViewController];
     };
     [SalesforceSDKManager sharedManager].launchErrorAction = ^(NSError *error, SFSDKLaunchAction launchActionList) {
-        __strong typeof(weakSelf) strongSelf = weakSelf;
-        [strongSelf log:SFLogLevelError format:@"Error during SDK launch: %@", [error localizedDescription]];
-        [strongSelf initializeAppViewState];
-        [[SalesforceSDKManager sharedManager] launch];
+      __strong __typeof(weakSelf) strongSelf = weakSelf;
```
khawkins: Same here, re: spacing.

shared/hybrid/AppDelegate+SalesforceHybridSDK.m
```diff
@@ -59,23 +59,23 @@ - (AppDelegate *)sfsdk_swizzled_init
...
+      [strongSelf log:SFLogLevelError format:@"Error during SDK launch: %@", [error localizedDescription]];
+      [strongSelf initializeAppViewState];
+      [[SalesforceSDKManager sharedManager] launch];
     };
     [SalesforceSDKManager sharedManager].postLogoutAction = ^{
-        [weakSelf handleSdkManagerLogout];
+      [weakSelf handleSdkManagerLogout];
```
khawkins: Same here.  I'd just bulk-select the whole `sfsdk_swizzled_init` method and do Control-I on it.

wmathurin:
Looks good.

I always wondered if we should we have a AppDelegate+SalesforceNativeSDK.m to move most of the complexity out of AppDelegate.m? What do you think?

khawkins:
> I always wondered if we should we have a AppDelegate+SalesforceNativeSDK.m to move most of the complexity out of AppDelegate.m? What do you think?

I generally think that we should avoid swizzling in this manner, as much as possible.  It's a necessity with Cordova apps, based on the lack of access to the main `AppDelegate` entity.  But it's not a process I'd encourage in an iOS app that didn't need it.

PR #1833, opened 2016-11-20
Adding back missing delegate calls

bhariharan:


wmathurin:
When did we lose them?

bhariharan:
They were implemented only in the S1 fork of `Mobile SDK`. We missed them in their transition to vanilla `Mobile SDK`.

PR #1835, opened 2016-11-21
Bringing back UIWebView

bhariharan:
Original PR to remove `UIWebView` is [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1748/).


bhariharan:
Tested `AccountEditor` and `SmartSyncExplorerHybrid` on both `WKWebView` and `UIWebView`. Both appear to work fine.

PR #1836, opened 2016-11-21
Fixing hybrid apps login

wmathurin:


shared/hybrid/AppDelegate+SalesforceHybridSDK.m
```diff
@@ -95,7 +95,9 @@ - (BOOL)sfsdk_swizzled_application:(UIApplication *)application didFinishLaunchi
     
     [self initializeAppViewState];
     [[SalesforceSDKManager sharedManager] launch];
-    return [self sfsdk_swizzled_application:application didFinishLaunchingWithOptions:launchOptions];
+    return YES; // we don't want to run's Cordova didFinishLaunchingWithOptions - it creates another window with a webview
```
wmathurin: We used to return yes up until a couple of months back. However sample apps continue working because their AppDelegate.m was doing less than the AppDelegate.m that you actually get when doing cordova create (see below).

shared/hybrid/AppDelegate.m
```diff
@@ -29,8 +29,7 @@ @implementation AppDelegate
 
 - (BOOL)application:(UIApplication*)application didFinishLaunchingWithOptions:(NSDictionary*)launchOptions
 {
-    self.viewController = [[MainViewController alloc] init];
-    return YES;
+    return [super application:application didFinishLaunchingWithOptions:launchOptions];
```
wmathurin: This is what the AppDelegate.m generated by cordova actually looks like.
The super class (CDVAppDelegate.m) creates a window. You end up with our window and theirs, and theirs is the one visible.

PR #1837, opened 2016-11-22
Adding app features for UIWebView and localhost cache

bhariharan:


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -179,6 +182,9 @@ - (id) initWithConfig:(SFHybridViewConfig *) viewConfig useUIWebView:(BOOL) useU
     self = [super init];
     if (self) {
         self.useUIWebView = useUIWebView;
+        if (useUIWebView) {
```
bhariharan: Not setting anything for `WKWebView`, since it's the default. The absence of `UI` indicates `WKWebView` is being used.

wmathurin: Could we flip it and log WK vs nothing (WK is very unique two letter acronym that makes one think of WKWebView right away)?

wmathurin: May be WK vs WV - otherwise for UIWebView we will have to write a convoluted query that looks for hybrid apps and not WK (not query are always more expensive)?

bhariharan: The current way the query should be straightforward (look for hybrid apps with `UI`). I'm not logging `WK` because that's the default and will become the only option soon, so we don't really care about that. We only want to know which apps are specifically using our new API to switch to `UIWebView`. Maybe we could use different letters to specify `UIWebView` instead of `UI`?

wmathurin: We could do WV ?
For absolute numbers, the query would be simple.
But for %, we would have to look at ui one side, and for hybrid, ios app with app feature but not ui on the other side.

bhariharan: I'll change it to `WV`. For `%`, we could do query with `WV` over all of `hybrid`. I don't think we need a `not` query, since the remaining `%` will implicitly be using `WK`.

bhariharan:
Changed constant [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1837/commits/a9479ab4e8646de7f418d9fd6810569202e7d091).

PR #1838, opened 2016-11-22
Having the os restriction here means that we need to use a different …

wmathurin:
…package.json for iOS react native template and Android react native template

So the only place where we will have the os = darwin constraint is in the package.json for forceios

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,6 @@
 	"licenses" : [
 	    { "type": "Salesforce.com Mobile SDK License", "url": "https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/LICENSE.md" }
 	],
-	"os" : [ "darwin" ],
 	"repository": {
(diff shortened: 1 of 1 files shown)


helenren:
👍

bhariharan:
Maybe we don't need that explicit constraint for `forceios` - why would any developer attempt to run `forceios` on `win`? :-) We could add a note in the doc instead, if that's easier.

wmathurin:
forceios and forcedroid are the only ones users will actually npm install manually - all the other package.json are only used when pulling in dependencies.
I think it makes sense to prevent forceios installation on non mac os machines since it cannot possibly work. Beside we have had the {os:darwin} in the forceios package.json since day one.
It  ended up in the other ones because, I copied the package.json from forceios over and overloooked the {os:darwin} entry.

bhariharan:
Sounds good!  👍

PR #1841, opened 2016-11-29
fix resyncghosts callback timing

frontendloader:
The callback in the exported method clearResyncGhosts fires before the ghosts are done being cleared. We should expect this to happen after ghosts are cleared. That way if the callback method operates on the smartstore, it will not see any stale or ghost data.

### DIFF
--- a/libs/SalesforceReact/SalesforceReact/Classes/SFSmartSyncReactBridge.m
+++ b/libs/SalesforceReact/SalesforceReact/Classes/SFSmartSyncReactBridge.m
@@ -85,8 +85,9 @@ RCT_EXPORT_METHOD(cleanResyncGhosts:(NSDictionary *)args callback:(RCTResponseSe
 {
     NSNumber* syncId = (NSNumber*) [args nonNullObjectForKey:kSyncIdArg];
     [self log:SFLogLevelDebug format:@"cleanResyncGhosts with sync id: %@", syncId];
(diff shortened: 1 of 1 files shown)


libs/SalesforceReact/SalesforceReact/Classes/SFSmartSyncReactBridge.m
```diff
@@ -85,8 +85,9 @@ @implementation SFSmartSyncReactBridge
 {
     NSNumber* syncId = (NSNumber*) [args nonNullObjectForKey:kSyncIdArg];
     [self log:SFLogLevelDebug format:@"cleanResyncGhosts with sync id: %@", syncId];
-    [[self getSyncManagerInst:args] cleanResyncGhosts:syncId completionStatusBlock:nil];
-    callback(@[[NSNull null]]);
+    [[self getSyncManagerInst:isGlobal] cleanResyncGhosts:syncId completionStatusBlock:^void(SFSyncStateStatus syncStatus){
```
bhariharan: @frontendloader I believe this should be ```[[self getSyncManagerInst:args...```?

frontendloader: Updated PR with requested changes. Mixed up function signatures between master/unstable.

PR #1842, opened 2016-11-29
Allow networking pluging to POST non JSON body Content-Types

GrahO23:
The networking plugin doesn't support POSTs with Content-Types that are not JSON format.

This is an issues as many APIs including the Salesforce token end point ('/services/oauth2/token') don't support JSON Content Type.

As a fix I detect is the Content-Type on a post is "application/x-www-form-urlencoded; charset=UTF-8" and if so add the payload string to a NSDictionary with key  @"x-www-form-urlencoded".

This is then picked up in the POST request in SFRestRequest.m and converted to NSData and POST is completed.

To enable non-JSON Content-Type set the content header to in Javascript

eg

var payload  = "grant_type=refresh_token&client_id=xxx&refresh_token=xxxx"; cordova.require("com.salesforce.plugin.network").sendRequest('https://login.salesforce.com',
        '/services/oauth2/token',
        successCB,
        errorCB,
        "POST",
        payload
        headers
    );

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFNetworkPlugin/SFNetworkPlugin.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFNetworkPlugin/SFNetworkPlugin.m
@@ -46,6 +46,7 @@ static NSString * const kfileParams      = @"fileParams";
 static NSString * const kFileMimeType    = @"fileMimeType";
 static NSString * const kFileUrl         = @"fileUrl";
 static NSString * const kFileName        = @"fileName";
+static NSString * const kFormUrlUTF8EncodedType = @"application/x-www-form-urlencoded; charset=UTF-8";
 
 @implementation SFNetworkPlugin
 
@@ -57,6 +58,7 @@ static NSString * const kFileName        = @"fileName";
     NSString* path = [argsDict nonNullObjectForKey:kPathArg];
 
     NSDictionary* queryParams = [[NSDictionary alloc] init];
+    NSDictionary<NSString*, NSString*>* headerParams = [argsDict nonNullObjectForKey:kHeaderParams];
     id queryParamsObj = [argsDict nonNullObjectForKey:kQueryParams];
 
     /*
@@ -64,11 +66,19 @@ static NSString * const kFileName        = @"fileName";
      * for POST. These checks ensure that both cases are handled properly.
      */
     if ([queryParamsObj isKindOfClass:[NSString class]]) {
-        queryParams = [SFJsonUtils objectFromJSONString:queryParamsObj];
+        NSString* contentType = [headerParams objectForKey:@"Content-Type"];
+        if(contentType && [contentType isEqualToString:kFormUrlUTF8EncodedType]){
+            //Send POST body as encoded string and not JSON as not all REST Endpoints
+            //support JSON as a Content Type;
+            queryParams = [[NSDictionary alloc] initWithObjectsAndKeys:queryParamsObj, @"x-www-form-urlencoded", nil];
+        }
+        else{
+             queryParams = [SFJsonUtils objectFromJSONString:queryParamsObj];
+        }
     } else {
         queryParams = queryParamsObj;
     }
-    NSDictionary<NSString*, NSString*>* headerParams = [argsDict nonNullObjectForKey:kHeaderParams];
+
     NSDictionary<NSString*, NSDictionary*>* fileParams = [argsDict nonNullObjectForKey:kfileParams];
     SFRestRequest* request = [SFRestRequest requestWithMethod:method path:path queryParams:queryParams];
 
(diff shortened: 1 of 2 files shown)


huminzhi:
ok to test

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

## Current coverage for SalesforceAnalytics.framework is `73%`

---
## Current coverage for SalesforceSDKCore.framework is `51%`
Files changed | - | - 
--- | --- | ---
SFRestRequest.m | `67%` | :warning:
SFRestRequest.m | `67%` | :warning:

---
## Current coverage for CocoaLumberjack.framework is `32%`

---
## Current coverage for SmartStore.framework is `63%`

---
## Current coverage for SmartSync.framework is `76%`

---
## Current coverage for SalesforceHybridSDK.framework is `42%`
Files changed | - | - 
--- | --- | ---
SFNetworkPlugin.m | `72%` | :warning:
SFNetworkPlugin.m | `72%` | :warning:

---
## Current coverage for SalesforceReact.framework is `0%`

---
## Current coverage for RestAPIExplorer.app is `26%`

---
## Current coverage for SmartSyncExplorer.app is `19%`

---
## Current coverage for SmartSyncExplorerCommon.framework is `0%`

---
## Current coverage for RecentContactsTodayExtension.appex is `0%`

---
## Current coverage for AccountEditor.app is `0%`

---
## Current coverage for NoteSync.app is `0%`

---
## Current coverage for SmartSyncExplorerHybrid.app is `0%`

---
## Current coverage for SalesforceSDKCoreTestApp.app is `30%`

---
## Current coverage for SmartStoreTestApp.app is `61%`

---
## Current coverage for SmartSyncTestApp.app is `61%`

---
## Current coverage for SalesforceHybridSDKTestApp.app is `85%`

---
## Current coverage for SalesforceAnalyticsTestApp.app is `61%`

---

  Generated by :no_entry_sign: <a href="http://danger.systems/">danger</a>
</p>

GrahO23:
@ huminzhi Thanks 

Do I need to up the coverage for SalesforceSDKCore & SalesforceHybridSDK ?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -225,8 +225,17 @@ - (void)prepareRequestForSend
         // It's POSTish.  The Network SDK handles content-based requests more or less automatically,
         // but if you want to post a JSON object or other data, you have to manage the contents yourself.
         if (self.action.parameters.parameterStyle != CSFParameterStyleMultipart) {
-            // Standard POST data.  We'll assume we can just send it as JSON.
-            NSData *bodyData = [SFJsonUtils JSONDataRepresentation:self.queryParams];
+            NSData *bodyData = nil;
+            //Check if the POST should be sent as an encoded string rather than JSON
+            NSString* formEncoded = (NSString*)[self.queryParams objectForKey: @"x-www-form-urlencoded" ];
```
khawkins: There are far too many specialized content types to one-off each one in this area of the code.  It's what `[SFRestRequest setCustomRequestBodyString:contentType:]` and its other input analogs were made for.  See my previous comments.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFNetworkPlugin/SFNetworkPlugin.m
```diff
@@ -57,18 +58,27 @@ - (void) pgSendRequest:(CDVInvokedUrlCommand *) command
     NSString* path = [argsDict nonNullObjectForKey:kPathArg];
 
     NSDictionary* queryParams = [[NSDictionary alloc] init];
+    NSDictionary<NSString*, NSString*>* headerParams = [argsDict nonNullObjectForKey:kHeaderParams];
```
wmathurin: We need the ability to specify any http request headers (custom end point could expect specific headers and the client should have a way to specify them or for instance if-xxx headers for "smarter" end points). This is not the artifact of old functionality.

Content type is one of many "permanent"  http request headers (full list here: https://en.wikipedia.org/wiki/List_of_HTTP_header_fields#Request_fields), I don't see a reason to single it out and treat it as a separate parameter.

bhariharan:
Thanks @khawkins. It sounds like we should prioritize this work for a future release, where we can do this in a more generic way.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFNetworkPlugin/SFNetworkPlugin.m
khawkins: > This is not the artifact of old functionality.

Ah, that was me misreading the diff.  It looked like setting the value was new as of this request.  I missed it below.

> I don't see a reason to single it out and treat it as a separate parameter.

My reasoning is that, in a world where we're specifying arbitrary content as input, it makes more sense for the value not to be buried in another collection of properties as an afterthought.  Indeed, that's why Content Type is explicitly called out as a parameter to `SFRestRequest`'s initializer.  We can leave it in a collection, but it feels a bit removed from intention in that use case, to me.

khawkins: Or to think of it another way, IMO that value drives a creation process of the request that includes setting the proper header values, not the other way around.

GrahO23:
@bhariharan Thanks will close the request. Playing with the unstable Mobile SDK with forcejs and that plays really nice on WKWebview.

PR #1844, opened 2016-11-30
Fixing potential deadlock

bhariharan:
Long story short, this could lead to a potential deadlock if an anonymous user is being initialized. It happens in `Wave` in playground mode. The proper fix appears to be calling this after the `init` call completes to prevent the deadlock.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
@@ -117,9 +117,8 @@ static NSString * const kSFAppFeatureMultiUser   = @"MU";
     static SFUserAccountManager *userAccountManager = nil;
     dispatch_once(&pred, ^{
 		userAccountManager = [[self alloc] init];
-        [[NSNotificationCenter defaultCenter] postNotificationName:SFUserAccountManagerDidFinishUserInitNotification
-                                                            object:self];
 	});
+    [NSNotificationCenter postNotificationOnceWithName:SFUserAccountManagerDidFinishUserInitNotification object:nil userInfo:nil];
     return userAccountManager;
 }
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
@@ -42,6 +42,7 @@
 
 // Notifications
 NSString * const SFUserAccountManagerDidChangeCurrentUserNotification   = @"SFUserAccountManagerDidChangeCurrentUserNotification";
+NSString * const SFUserAccountManagerDidFinishUserInitNotification   = @"SFUserAccountManagerDidFinishUserInitNotification";
 
 NSString * const SFUserAccountManagerUserChangeKey      = @"change";
 
@@ -116,6 +117,8 @@ static NSString * const kSFAppFeatureMultiUser   = @"MU";
     static SFUserAccountManager *userAccountManager = nil;
     dispatch_once(&pred, ^{
 		userAccountManager = [[self alloc] init];
+        [[NSNotificationCenter defaultCenter] postNotificationName:SFUserAccountManagerDidFinishUserInitNotification
+                                                            object:self];
 	});
     return userAccountManager;
 }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
@@ -57,7 +57,7 @@ static NSString * const kSFDisableExternalPaste = @"DISABLE_EXTERNAL_PASTE";
 
 @synthesize rawPreferences = _rawPreferences;
 
-+ (instancetype) sharedPreferences {
++ (instancetype)sharedPreferences {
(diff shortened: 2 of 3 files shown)


bhariharan:
@aprabhu Could you give this a shot?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
```diff
@@ -84,55 +86,56 @@ - (id)init {
...
             [[SalesforceSDKManager sharedManager] registerAppFeature:kSFAppFeatureManagedByMDM];
         }
-        NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
-        if (self.rawPreferences) {
-            attributes[@"mdmIsActive"] = [NSNumber numberWithBool:YES];
-            attributes[@"mdmConfigs"] = self.rawPreferences;
-        } else {
-            attributes[@"mdmIsActive"] = [NSNumber numberWithBool:NO];
-        }
-        [SFSDKEventBuilderHelper createAndStoreEvent:@"mdmConfiguration" userAccount:nil className:NSStringFromClass([self class]) attributes:attributes];
```
khawkins: Or technically, making the new `storeAnalyticsEvent` call there (I presume we'd still need the refactor out into that method.

prabhuamol:
:

prabhuamol:
👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
prabhuamol: it has to moved out of SFManagedPreferences init (sharedIntance actually) since that init is called from inside the SFUserAccountManger sharedIntance init call. And internally its trying to init SFUserAccountManger from there.

khawkins: Makes sense.  That's a bummer of a circular dependency. 🙁

prabhuamol:
@bhariharan this is still causing a deadlock since storeAnalyticsEvent is called from loginHosts in SFManagedPreferences which will call createAndStoreEvent causing a deadlock since SFUserAccountManger sharedIntance hasn't returned yet.

We need to basically move this after SFUserAccountManager has been initialized

bhariharan:
The `NSNotification` approach seems to work @khawkins.

khawkins:
Cool!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
```diff
@@ -84,55 +85,56 @@ - (id)init {
...
+        NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+        [notificationCenter addObserver:self
+                               selector:@selector(storeAnalyticsEvent)
+                                   name:SFUserAccountManagerDidChangeCurrentUserNotification
```
prabhuamol: I think this should be SFUserAccountManagerDidFinishUserInitNotification

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -116,6 +117,8 @@ + (instancetype)sharedInstance {
     static SFUserAccountManager *userAccountManager = nil;
     dispatch_once(&pred, ^{
 		userAccountManager = [[self alloc] init];
+        [[NSNotificationCenter defaultCenter] postNotificationName:SFUserAccountManagerDidFinishUserInitNotification
```
prabhuamol: This also causes a deadlock since we are still within the dispatch_once. I tried locally and it is fixed by moving it outside of the dispatch_once and using 

[NSNotificationCenter postNotificationOnceWithName:SFUserAccountManagerDidFinishUserInitNotification object:nil userInfo:nil];

khawkins: > postNotificationOnceWithName:

I didn't even know we had such a beast.  Nice!

bhariharan:
@prabhuamol Good catch! Copy-paste fail. I moved it outside the `dispatch_once` too. Could you try now?

prabhuamol:
👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
```diff
@@ -84,55 +85,56 @@ - (id)init {
...
     return self.rawPreferences[kManagedKeyConnectedAppId];
 }
 
-- (NSString *)connectedAppCallbackUri {
+- (NSString *) connectedAppCallbackUri {
     return self.rawPreferences[kManagedKeyConnectedAppCallbackUri];
 }
 
-- (BOOL)shouldDisableExternalPasteDefinedByConnectedApp {
+- (BOOL) shouldDisableExternalPasteDefinedByConnectedApp {
```
khawkins: Is there a reason we're adding these spaces between the return type and method names?  The former is the general style convention for Objective-C coding.

bhariharan: @khawkins I'll revert the space changes.

prabhuamol:
@bhariharan yes merge away. works great

PR #1886, opened 2017-01-04
Only show leftBar Button in iPhone in LoginHostListView

liguopku:
@bbirman @khawkins


khawkins:
Please see my comments [here](https://gus.my.salesforce.com/0D5B000000QcFyF).  I don't understand why the cancel button shouldn't work on iPhone _and_ iPad.  And the screenshot is not exhibiting behavior that we'd see in the stock SDK.  Here's what the login host screen looks like in a similar scenario with the RestAPIExplorer sample app:

image

I think we need to better understand the differences in the code, before making these changes.

liguopku:
Thanks for the clarification! I will close this and make changes in our project instead.

PR #1894, opened 2017-01-17
Adding class name to instrumentation results

khawkins:
Makes `SFSDKInstrumentationPostExecutionData` a lot more modular.  Otherwise, you end up pretty tied to its defining `SFMethodInterceptor` instance to get the whole picture of the instrumented method.


khawkins:
@huminzhi The internal report seems to indicate an issue setting up the CI build (ruby gem installation failures).  Can you take a look?

PR #1895, opened 2017-01-18
Ensure select private headers are exposed under its own explicit module

MikeNachbaur:
There are some shared capabilities that are exposed by portions of the network logic that are used across other products and frameworks. These functions and constants aren't intended to be used by your average MobileSDK consumer, but their declarations still need to be exposed for other teams to be able to use it.

### DIFF
--- a/libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common.xcconfig
+++ b/libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common.xcconfig
@@ -15,6 +15,3 @@ LIBRARY_SEARCH_PATHS = $(inherited)
 
 // Make sure to use only API that are safe for application extensions
 APPLICATION_EXTENSION_API_ONLY = YES
-
-MODULEMAP_FILE = $(SRCROOT)/SalesforceSDKCore/SalesforceSDKCore.modulemap
-MODULEMAP_PRIVATE_FILE = $(SRCROOT)/SalesforceSDKCore/SalesforceSDKCore.private.modulemap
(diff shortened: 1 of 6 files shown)


libs/SalesforceSDKCore/Configuration/SalesforceSDKCore-Common.xcconfig
```diff
@@ -15,3 +15,6 @@ LIBRARY_SEARCH_PATHS = $(inherited)
 
 // Make sure to use only API that are safe for application extensions
 APPLICATION_EXTENSION_API_ONLY = YES
+
+MODULEMAP_FILE = $(SRCROOT)/SalesforceSDKCore/SalesforceSDKCore.modulemap
```
khawkins: Would these be a better fit in `SalesforceSDKCore-Common-Dynamic.xcconfig`?

MikeNachbaur: Fair enough, I didn't see the -Dynamic file there when I added this.  Admittedly, there's no harm in assigning a module map definition to a static target since Xcode will ignore it, but for accuracy, you're right.

huminzhi:
test me

huminzhi:
test me

huminzhi:
test me

huminzhi:
test me

khawkins:
Looks like "modularizing" `CSFPrivateDefines.h` triggered some header import issues in the test runs.  If you run:
```
$ xcodebuild -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceSDKCore \
-sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=10.2' \
clean build test
```
You'll get the following error:

/workspace/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Utilities/CSFInternalDefines.h:80:39: error: declaration of 'objc_property_t' must be imported from module 'ObjectiveC.runtime' before it is required
CSF_EXTERN BOOL CSFPropertyIsReadonly(objc_property_t property);

Adding `#import <objc/runtime.h>` to `CSFInternalDefines.h` seems to fix it.

khawkins:
test me

huminzhi:
CI passed, merge it now.. Thank you @khawkins @MikeNachbaur!

PR #1896, opened 2017-01-23
Updated frameworks to propagate their verison numbers properly up to the framework Info.plist

MikeNachbaur:
Also updated the build script to provide usage documentation on some flags that were missing.


configuration/Version.xcconfig
```diff
@@ -0,0 +1 @@
+CURRENT_PROJECT_VERSION = 5.0.0
```
bhariharan: Change this to `5.1.0`, since that's the upcoming release.

huminzhi:
test me

PR #1892, opened 2017-01-13
Simplifying boilerplate needed in custom sync up or sync down targets

wmathurin:
Sync up/down targets will always serialize their implementation class - the target type will only be used if no implementation class is provided (which should only happen if the sync up/down target was created with pre 5.1 code).

PR #1898, opened 2017-01-26
add slather in fast file"

huminzhi:


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -418,6 +418,9 @@ - (SFSyncState*)syncUpWithTarget:(SFSyncUpTarget *)target
                          options:(SFSyncOptions *)options
                         soupName:(NSString *)soupName
                      updateBlock:(SFSyncSyncManagerUpdateBlock)updateBlock {
+    if (target == nil) {
```
wmathurin: Who is calling this with nil? How was it working before?
I think something is failing before the call to this method (maybe a regression introduced by my recent changes https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1892). I think this change is just masking the issue.

bhariharan: +1 to what Wolf said. I don't think `target` should be `nil` here.

wmathurin: My earlier PR (https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1892) is indeed the cause of the regression. I'll have a PR to fix the issue shortly.

wmathurin: This should fix hybrid smartsync tests: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1901

huminzhi:
@wmathurin reverted my changes in this pr

huminzhi:
@helenren  I added slather in the fastlane file, we should get more details code cov info now..

PR #1899, opened 2017-01-26
Removed Anonymous Client Secret & Fixed Crashes/Race Condition between handling errors and dismissal of Snapshot

trooper2013:
Removed Anonymous Client Secret & Fixed Crashes/Race Condition between handling errors and dismissal of Snapshot. Crashes were related to changes being done to the UI state on background threads. The alert would get dismissed when normally run. Could not reproduce  run in a debugger. The fix for the timing issue is a short term fix. The long term solution remains to be addressed.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -85,7 +85,6 @@ static NSString * const kSFOAuthApprovalCode                    = @"code";
 static NSString * const kSFOAuthGrantTypeAuthorizationCode      = @"authorization_code";
 static NSString * const kSFOAuthResponseTypeActivatedClientCode = @"activated_client_code";
 static NSString * const kSFOAuthResponseClientSecret            = @"client_secret";
-static NSString * const kSFOAuthClientSecretAnonymous           = @"anonymous";
 
 // OAuth Error Descriptions
 // see https://na1.salesforce.com/help/doc/en/remoteaccess_oauth_refresh_token_flow.htm
@@ -683,7 +682,6 @@ static NSString * const kSFAppFeatureSafariBrowserForLogin   = @"BW";
         if (self.authInfo.authType == SFOAuthTypeAdvancedBrowser) {
             [params appendFormat:@"&%@=%@", kSFOAuthCodeVerifierParamName, self.codeVerifier];
             [logString appendFormat:@"&%@=REDACTED", kSFOAuthCodeVerifierParamName];
-            [params appendFormat:@"&%@=%@", kSFOAuthResponseClientSecret, kSFOAuthClientSecretAnonymous];
         }
         
         // Discard the approval code.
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -85,7 +85,6 @@
 static NSString * const kSFOAuthGrantTypeAuthorizationCode      = @"authorization_code";
 static NSString * const kSFOAuthResponseTypeActivatedClientCode = @"activated_client_code";
 static NSString * const kSFOAuthResponseClientSecret            = @"client_secret";
```
bhariharan: If the `client_secret` constant is not being used anymore, can we remove it too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -1051,6 +1061,7 @@ - (SFAuthErrorHandlerList *)populateDefaultAuthErrorHandlerList
     self.genericAuthErrorHandler = [[SFAuthErrorHandler alloc]
                                                  initWithName:kSFGenericFailureAuthErrorHandler
                                                  evalBlock:^BOOL(NSError *error, SFOAuthInfo *authInfo) {
+                                                     [[SalesforceSDKManager sharedManager] dismissSnapshot];
```
khawkins: It's a nitpick, but shouldn't this logic technically be in `showRetryAlertForAuthError:alertTag:`?  It's ultimately the presentation of the alert view that necessitates the dismissal of the snapshot view.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
khawkins: There's still an old IP bypass flow that utilizes the client_secret value, with an activation code.  I'm 95% sure that whole flow can go away too, but removing this value should wait on that work.

khawkins: I'm 100% sure that flow can go away, given that nothing sets the activationCode anymore, other than code to revoke it, which itself is never called either.  FWIW. 😊

trooper2013: it is being used

trooper2013: It is used for the approval code flow.

khawkins:
Thanks for digging into this, @trooper2013 !  Gnarly stuff.

huminzhi:
Thank you so much @trooper2013 to fix this Gnarly area!

bhariharan:
test me

huminzhi:
test me

PR #1901, opened 2017-01-26
Fix for sync up regression in hybrid app

wmathurin:
Hybrid apps don’t specify query type or implementation class by default. In the past, default sync up target was assumed in those cases. The recent refactor changed that and broke hybrid sync up.

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Util/SFSyncDownTarget.m
+++ b/libs/SmartSync/SmartSync/Classes/Util/SFSyncDownTarget.m
@@ -43,7 +43,7 @@ NSString * const kSFSyncTargetQueryTypeCustom = @"custom";
 #pragma mark - From/to dictionary
 
 + (SFSyncDownTarget*) newFromDict:(NSDictionary*)dict {
-    // We should have an implementation class unless sync down was created with SDK before 5.1 and is not custom
+    // We should have an implementation class or a target type
     NSString* implClassName = dict[kSFSyncTargetiOSImplKey];
     if (implClassName.length > 0) {
         Class customSyncDownClass = NSClassFromString(implClassName);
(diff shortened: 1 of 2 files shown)


libs/SmartSync/SmartSync/Classes/Util/SFSyncDownTarget.m
```diff
@@ -43,7 +43,7 @@ @implementation SFSyncDownTarget
 #pragma mark - From/to dictionary
 
 + (SFSyncDownTarget*) newFromDict:(NSDictionary*)dict {
-    // We should have an implementation class unless sync down was created with SDK before 5.1 and is not custom
+    // We should have an implementation class or a target type
```
wmathurin: Comment was not quite accurate - hybrid apps might not specify implementation class

libs/SmartSync/SmartSync/Classes/Util/SFSyncUpTarget.m
```diff
@@ -58,7 +58,7 @@ - (instancetype)init {
 #pragma mark - Serialization and factory methods
 
 + (instancetype)newFromDict:(NSDictionary*)dict {
-    // We should have an implementation class unless sync up was created with SDK before 5.1 and is not custom
+    // We should have an implementation class or a target type
```
wmathurin: Comment was not quite accurate - hybrid apps might not specify implementation class

libs/SmartSync/SmartSync/Classes/Util/SFSyncUpTarget.m
```diff
@@ -71,7 +71,9 @@ + (instancetype)newFromDict:(NSDictionary*)dict {
     }
     // No implementation class - using target type
     else {
-        switch ([self targetTypeFromString:dict[kSFSyncTargetTypeKey]]) {
+        // No target type - assume kSFSyncUpTargetTypeRestStandard (hybrid apps don't specify it a sync up target type by default)
```
wmathurin: That bit of logic was lost in https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1892

PR #1905, opened 2017-01-31
Alert View positioning issue

trooper2013:
More evil problems surface related to SFRootViewManager and Alerts. It is reported that when the alert is presented and app is sent to background and later resumed it is noticed that the alert is positioned off the screen. Unfortunately this has to do with the views being managed by the SFRootViewManager. I've temporarily patched the problem i.e we check if the presented view is a UIAlertController before presenting a snapshot. If so we dismiss it and continue to push the SnapshotView and then when the app is foregrounded we present the same alert after the SnapshotView is popped.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager+Internal.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager+Internal.h
@@ -27,7 +27,6 @@
 @interface SalesforceSDKManager () <SalesforceSDKManagerFlow, SFUserAccountManagerDelegate>
 {
     BOOL _isLaunching;
-    UIViewController* _defaultSnapshotViewController;
     UIViewController* _snapshotViewController;
 }
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFRootViewManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFRootViewManager.m
@@ -113,8 +113,8 @@
         UIViewController *currentViewController = strongSelf.mainWindow.rootViewController;
         while (currentViewController.presentedViewController != nil && !currentViewController.presentedViewController.isBeingDismissed) {
             if([currentViewController.presentedViewController isKindOfClass:[UIAlertController class]]) {
-                [currentViewController.presentedViewController dismissViewControllerAnimated:NO completion:nil];
                 strongSelf->_modalViewController = (UIAlertController *)currentViewController.presentedViewController;
+                [currentViewController.presentedViewController dismissViewControllerAnimated:NO completion:nil];
                 break;
             }
             currentViewController = currentViewController.presentedViewController;
@@ -178,7 +178,6 @@
                           [prevController presentViewController:strongSelf->_modalViewController animated:NO completion:^{
                             strongSelf->_modalViewController = nil;
                            }];
-                          return;
                       }
                       [strongSelf enumerateDelegates:^(id<SFRootViewManagerDelegate> delegate) {
                           if ([delegate respondsToSelector:@selector(rootViewManager:didPopViewControler:)]) {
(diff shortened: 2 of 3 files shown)


huminzhi:
test me

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFRootViewManager.m
```diff
@@ -125,6 +132,10 @@ - (void)pushViewController:(UIViewController *)viewController
                     }
                 }];
                 
+                if([currentViewController isKindOfClass:[UINavigationController class]]) {
+                    currentViewController =[((UINavigationController *) currentViewController) visibleViewController];
```
khawkins: Ah, good catch.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFRootViewManager.m
```diff
@@ -165,9 +179,16 @@ - (void)popViewController:(UIViewController *)viewController
                             [delegate rootViewManager:strongSelf didPopViewControler:viewController];
                         }
                     }];
+                    if(strongSelf->_modalViewController) {
```
khawkins: I would say that this action needs to preclude calls to `[SFRootViewManagerDelegate rootViewManager:didPopViewControler:]`, because that call is really meant to represent the logical conclusion of the root view manager process.  This should be refactored to reflect the intention of the code above it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager+Internal.h
```diff
@@ -27,7 +27,6 @@
 @interface SalesforceSDKManager () <SalesforceSDKManagerFlow, SFUserAccountManagerDelegate>
 {
     BOOL _isLaunching;
-    UIViewController* _defaultSnapshotViewController;
```
bhariharan: Isn't this meant to provide the ability to override the snapshot controller?

khawkins: I think the premise is, regardless of custom or default, there will only ever be one at a time, so why not just have one variable for it.

huminzhi:
test me

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFRootViewManager.m
```diff
@@ -160,14 +174,23 @@ - (void)popViewController:(UIViewController *)viewController
...
-                    [strongSelf enumerateDelegates:^(id<SFRootViewManagerDelegate> delegate) {
-                        if ([delegate respondsToSelector:@selector(rootViewManager:didPopViewControler:)]) {
-                            [delegate rootViewManager:strongSelf didPopViewControler:viewController];
-                        }
-                    }];
+                      if(strongSelf->_modalViewController) {
+                          [prevController presentViewController:strongSelf->_modalViewController animated:NO completion:^{
+                            strongSelf->_modalViewController = nil;
+                           }];
+                          return;
```
khawkins: (Sorry, previous comment was on the commit, which is a pain in posterity searching of PRs. 🙂)

Won't the `return` statement make it so that the `enumerateDelegates` call doesn't happen?

huminzhi:
test me

huminzhi:
test me

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager+Internal.h
bhariharan: Makes sense.

huminzhi:
test me

huminzhi:
test me

huminzhi:
test me

huminzhi:
test me

PR #1908, opened 2017-02-02
Fix retain cycle when an error occurs

MikeNachbaur:
The error userInfo description can capture `self` which creates a retain cycle.  I've removed the original key, and changed the key to instead use a description of itself, and changed the developer-facing constant to signify this change.


huminzhi:
test me

huminzhi:
test me

PR #1910, opened 2017-02-07
Ensure logging macros don't leak `self`

MikeNachbaur:
Since the macros reference `self`, if those logging macros are invoked from within a block copied to the heap, it's possible for a retain cycle to be created.  Instead, use the `__PRETTY_FUNCTION__` and `__FILE__` precompiler definitions.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLoggerMacros.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLoggerMacros.h
@@ -118,11 +118,11 @@ static BOOL const kSFASLLoggerEnabledDefault = NO;
 #define SFLogDebug(frmt, ...)      SFLogDebugToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
 #define SFLogVerbose(frmt, ...)  SFLogVerboseToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
 
-#define SFLogCError(frmt, ...)      SFLogError(frmt, ##__VA_ARGS__)
-#define SFLogCWarn(frmt, ...)        SFLogWarn(frmt, ##__VA_ARGS__)
-#define SFLogCInfo(frmt, ...)        SFLogInfo(frmt, ##__VA_ARGS__)
-#define SFLogCDebug(frmt, ...)      SFLogDebug(frmt, ##__VA_ARGS__)
-#define SFLogCVerbose(frmt, ...)  SFLogVerbose(frmt, ##__VA_ARGS__)
+#define SFLogCError(frmt, ...)    _Pragma ("GCC warning \"'SFLogCError' macro is deprecated, use SFLogError\"")     SFLogError(frmt, ##__VA_ARGS__)
+#define SFLogCWarn(frmt, ...)     _Pragma ("GCC warning \"'SFLogCWarn' macro is deprecated, use SFLogWarn\"")       SFLogWarn(frmt, ##__VA_ARGS__)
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLoggerMacros.h
```diff
@@ -106,23 +106,23 @@ static BOOL const kSFASLLoggerEnabledDefault = NO;
...
+#define SFLogInfo(frmt, ...)        SFLogInfoToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
+#define SFLogDebug(frmt, ...)      SFLogDebugToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
+#define SFLogVerbose(frmt, ...)  SFLogVerboseToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
 
-#define SFLogCError(frmt, ...)      SFLogErrorToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
-#define SFLogCWarn(frmt, ...)        SFLogWarnToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
-#define SFLogCInfo(frmt, ...)        SFLogInfoToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
-#define SFLogCDebug(frmt, ...)      SFLogDebugToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
-#define SFLogCVerbose(frmt, ...)  SFLogVerboseToContext(SFLoggerDefaultContext, nil, frmt, ##__VA_ARGS__)
+#define SFLogCError(frmt, ...)      SFLogError(frmt, ##__VA_ARGS__)
```
MikeNachbaur: These macros are now obsolete (the `*CError` ones).  Would you like me to remove them altogether, or leave these in for backwards compatibility purposes?

sgoldberg-sfdc: Were they deployed? If so we have no way of knowing if someone is using them or not. I think would have to deprecate them first.

MikeNachbaur: Yeah they were deployed, but it's doubtful many people will have used them.  Personally I'd rather err on the side of caution and keep these backwards-compatibility macros in, which is why I included it in the first place.  I just wanted to raise it as a suggestion to the MobileSDK team.

bhariharan:
@MikeNachbaur Looks like there are merge conflicts. Could you rebase against `dev`?

bhariharan:
test me

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLoggerMacros.h
bhariharan: Yup - we should mark them deprecated for now and we can remove them in our `6.0` release.

MikeNachbaur:
test me

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/Logging/SFLoggerMacros.h
MikeNachbaur: Added deprecation warnings...

huminzhi:
test me

huminzhi:
test me

bhariharan:
test me

PR #1915, opened 2017-02-10
Adopt new os_log functionality, replacing ASL

MikeNachbaur:
os_log is available on iOS 10+, and on iOS9 falls back to plain old NSLog (which backs into ASL).  This pull request adds integration for os_log for several benefits:

* More advanced integration of log identifiers, and log categories, for better and easier searching & filtering with the Console application.
* Linking logging data into the os_* methods for future support of activities & system trace (which I'll be adding soon)

These logging actions need to be done at the macro level because the system takes a lot of context about the calling location into account when performing system traces.  Furthermore, I would like to deprecate the `[self log:msg]` methods on NSObject, because not only do those incur additional performance penalties in calling objc_msgSend (e.g. why call an Objective-C method when it'll most likely be filtered out with the log level flag anyway?), but they don't provide the necessary context for the caller.

Instead I'd like people to adopt the SFLog* macros, or even better, create their own log identifiers with their own macros (e.g. for internal SDKs)


bhariharan:
test me

huminzhi:
test me

huminzhi:
Looks like some tests are failing with the change:

09:36:33 [09:36:33]: ▸ [35mRun test suite SalesforceRestAPITests encountered an error (Exceeded max restart count of 2)[0m
09:36:33 [09:36:33]: ▸ [35mFailing tests:[0m
09:36:33 [09:36:33]: ▸ [35m-[SFLoggerTests testObjectMethods][0m
09:36:33 [09:36:33]: ▸ [35m-[SFLoggerTests testObjectMethods][0m
09:36:33 [09:36:33]: ▸ [35m-[SFLoggerTests testObjectMethods][0m
09:36:33 [09:36:33]: ▸ [35m-[SFOAuthCoordinatorFlowTests testMultipleAuthenticationRequests][0m
09:36:33 [09:36:33]: ▸ [35m-[SFOAuthCoordinatorFlowTests testRefreshFlowInitiated][0m
09:36:33 [09:36:33]: ▸ [35m-[SFOAuthCoordinatorFlowTests testUserAgentFlowInitiated][0m
09:36:33 [09:36:33]: ▸ [35m-[SFOAuthSessionRefresherTests testBadInputData][0m
09:36:33 [09:36:33]: ▸ [35m-[SFOAuthSessionRefresherTests testFailedRefresh][0m
09:36:33 [09:36:33]: ▸ [35m-[SFUserAccountManagerTests testAnonymousUserFromPreviousVersion][0m
09:36:33 [09:36:33]: ▸ [35m** TEST FAILED **[0m

MikeNachbaur:
Interesting, those tests passed locally. I'll check it out on Tuesday (Monday is a holiday for me).

sgoldberg-sfdc:
My opinion on deprecating more methods: It feels like we just had a logging interface change a few months ago. Since logging is pervasive through apps, we should minimize the frequency of changes to the interface. 

Is there an end goal in mind for the logging utility? If so, and we are not there yet, I would prefer if we kept the old interfaces alive until we are.

MikeNachbaur:
@sgoldberg-sfdc the thing is, the original logging methods were a mistake to begin with. There's a high amount of overhead in dispatching through objc_msgSend (especially for log messages below the current log level threshold, meaning lots of `verbose` messages will still impact CPU performance even when filteres out), and calling through to another method / class pushes additional frames onto the stack, meaning you lose file/function/line number context. 

Also, the os/log.h methods introduced in iOS 10 use this context to provide information that can be upleveled to the crash log automatically. But that can only be achieved through the use of logging macros. 

To that end, that's why I'd like to progressively deprecate the logging category.

 First I'd introduced the alternative logging macros, compete with reverse-DNS log identifiers allowing different subsystems to have their own log levels (e.g. ServiceSDK components, Network classes, SmartStore, etc, could all have their own log levels set so you can improve the signal/noise ratio. 

This PR adds the beginnings of OS activity log / activity tracing to the logging subsystem. 

A subsequent PR would hopefully deprecate the legacy logging categories, followed by an eventual removal of it. 

My suggestion would be to read the docs I wrote on the logging header, about how to create your own log identifiers and logging macros.

MikeNachbaur:
test me

bhariharan:
test me

huminzhi:
test me

huminzhi:
test me

huminzhi:
test me

huminzhi:
test me

huminzhi:
@MikeNachbaur  still hit some test failures in SFLoggerTests.testExtraLoggers and SFLoggerTests.testLogFormatter,  a quick fix though

MikeNachbaur:
Yeah @huminzhi those stupid line-number logging tests are brittle.  I ran it locally and it worked, but it looks like something changed. *sigh*

huminzhi:
test me

PR #1916, opened 2017-02-11
Fix for swizzling from a community org

sgoldberg-sfdc:
- If swizzling from a community org we do not want to use the old communityid
- A new communityId if needed should be provided with the swizzle params.


sgoldberg-sfdc:
@bhariharan @khawkins

huminzhi:
test me

sgoldberg-sfdc:
@trooper2013 and @bhariharan , please include me in the discussion if you can.

PR #1919, opened 2017-02-14
fix forgot password link issue

huminzhi:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -474,7 +474,7 @@ - (SFUserAccount*)createUserAccount {
     SFUserAccount *newAcct = [[SFUserAccount alloc] initWithIdentifier:[self uniqueUserAccountIdentifier]];
     SFOAuthCredentials *creds = newAcct.credentials;
     creds.accessToken = nil;
-    creds.domain = self.loginHost;
+    creds.domain = [self.loginHost containsString:@"?display="]?self.loginHost:[self.loginHost stringByAppendingString:@"?display=page"];
```
bhariharan: Why are we doing this? We've always used `?display=touch` on mobile and in recent times I thought the display param was being deprecated. Also, I don't think domain should have a `display` param. Besides, how was it working on `UIWebView` and only failing on `WKWebView`?

khawkins: Seconding the question.  This looks like it's trying to create new API/functionality on the fly, but I'm not sure what that functionality is supposed to be.  Nor am I sufficiently comfortable that this is the place to implement it, whatever it might be.

PR #1920, opened 2017-02-15
A more robust way of handling equality between baseUrl and cachedAPIURL.

brichey:
@bhariharan @sgoldberg-sfdc @qingqingliu please review!

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
@@ -86,7 +86,7 @@ static NSString inline * CSFSalesforceErrorMessage(NSDictionary *errorDict) {
         if (_enqueuedNetwork) {
             if ([self shouldUpdateBaseUrl]) {
                 self.baseURL = self.enqueuedNetwork.account.credentials.apiUrl;
-                self.cachedAPIURL = self.enqueuedNetwork.account.credentials.apiUrl;
+                self.cachedAPIURL = self.baseURL;
             }
             [_enqueuedNetwork addObserver:self forKeyPath:kNetworkAccessTokenPath
                                  options:(NSKeyValueObservingOptionInitial |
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
```diff
@@ -289,7 +289,7 @@ - (void)observeValueForKeyPath:(NSString *)keyPath ofObject:(id)object change:(N
 - (BOOL)shouldUpdateBaseUrl {
     // only set base URL to apiURL if baseURL is not already specified as absolute URL with it's own host
     // this check is necessary as there are salesforce URL that is content server based and not API based
-    return (!self.baseURL.scheme && !self.baseURL.host) || (self.baseURL == self.cachedAPIURL);
+    return (!self.baseURL.scheme && !self.baseURL.host) || [self.baseURL isEqual:self.cachedAPIURL];
```
brichey: Both of these properties use "copy"...so there's a good chance "==" would not return true because they could be different instances.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
```diff
@@ -86,7 +86,7 @@ - (void)setEnqueuedNetwork:(CSFNetwork *) network {
         if (_enqueuedNetwork) {
             if ([self shouldUpdateBaseUrl]) {
                 self.baseURL = self.enqueuedNetwork.account.credentials.apiUrl;
-                self.cachedAPIURL = self.enqueuedNetwork.account.credentials.apiUrl;
+                self.cachedAPIURL = self.baseURL;
```
brichey: setBaseURL is overridden in CSFAction, and it has some logic for adding a trailing slash to the url. If that happens, then baseURL and cachedAPIURL would not be equal. That's why it's better to set cachedAPIURL directly to self.baseURL.

bhariharan:
test me

bhariharan:
test me

PR #1926, opened 2017-02-16
Simpler Method for Calculating Base URL

jarbogast:
We've had a lot of changes lately to exactly when the baseURL for an action could/should be updated. That logic was just getting more and more complicated, so much so that it was too complicated for me to understand.

I wrote a few tests around this area and those tests led me to what I believe are the real 'rules' that we want to enforce for the baseURL.

1. If someone has manually set the baseURL, then respect that URL
2. If no baseURL has been manually set, use the API URL from the account on the network on which the action is enqueued.

Then I expressed those rules more simply in a custom getter for baseURL.

Is my thinking on point here?

@khawkins @bhariharan @qingqingliu


libs/SalesforceSDKCore/SalesforceSDKCore/CSFSalesforceActionTests.m
```diff
@@ -0,0 +1,83 @@
+//
+//  CSFSalesforceActionTests.m
+//  SalesforceSDKCore
+//
+//  Created by Jonathan Arbogast on 2/16/17.
+//  Copyright © 2017 salesforce.com. All rights reserved.
+//
+
+@import XCTest;
```
bhariharan: @jarbogast Move this file [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/tree/dev/libs/SalesforceSDKCore/SalesforceSDKCoreTests), where all the other tests live.

libs/SalesforceSDKCore/SalesforceSDKCore/CSFSalesforceActionTests.m
```diff
@@ -0,0 +1,83 @@
+//
```
bhariharan: Update copyright to this:
```
/*
 Copyright (c) 2017-present, salesforce.com, inc. All rights reserved.
 
 Redistribution and use of this software in source and binary forms, with or without modification,
 are permitted provided that the following conditions are met:
 * Redistributions of source code must retain the above copyright notice, this list of conditions
 and the following disclaimer.
 * Redistributions in binary form must reproduce the above copyright notice, this list of
 conditions and the following disclaimer in the documentation and/or other materials provided
 with the distribution.
 * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
 endorse or promote products derived from this software without specific prior written
 permission of salesforce.com, inc.
 
 THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
 IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
 FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
 CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
 DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
 DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
 WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
 */
```

jarbogast:
Thanks @bhariharan!

I also realize now that I'm not enforcing the 'must have a trailing slash' rule in all cases. Also, that's a property marked with copy semantics so I should probably respect that :)

I'll ping everyone again when I'm done cleaning up all those items.

huminzhi:
@jarbogast I love this simplified logic!!!

jarbogast:
Ok @bhariharan. I addressed your feedback from the last review!

I also moved the override for baseURL into CSFSalesforceAction instead of CSFAction. That was the only place it was using the network's account's API Url previously.

I also broke out the existing logic for appending a trailing slash into a separate method and added a few tests as a sanity check.

jarbogast:
My only concern now with these changes is that I did break KVO for anyone who would be observing the baseURL property of a CSFSalesforceAction. In my mind, that would be a questionable thing to do, but I guess its always possible.

bhariharan:
test me

sgoldberg-sfdc:
So fresh and so clean! 👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/NSURL+SFStringUtils.m
```diff
@@ -116,4 +116,14 @@ + (void)appendPathComponents:(NSArray*)pathComponents toMutableUrlString:(NSMuta
     }
 }
 
+- (NSURL *)slashTerminatedUrl {
+    NSMutableString *urlString = [self.absoluteString mutableCopy];
+
+    if (NSMaxRange([self.path rangeOfString:@"/" options:NSBackwardsSearch]) < self.path.length - 1) {
```
khawkins: That seems like a long way around to `![self.path hasSuffix:@"/"]`, but okay. 🙂

bhariharan:
test me

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/NSURL+SFStringUtils.m
jarbogast: I actually put that code in there! But I seem to remember it actually broke one of the unit test cases I had written for the existing functionality. Overall this snippet was a little confusing to me because we are checking the path property, but then modifying the absoluteString.

So I basically left it alone and backed away slowly :)

huminzhi:
test me

huminzhi:
test me

huminzhi:
@jarbogast looks good to me, about the tests, can we have one test for orgSplit and one for the case @qingqingliu mentioned about static resources? 
For orgSplit, we can change the instanceURL of credential AFTER enqueue the request and make sure the baseURL getter is also reflecting the change (I know the logic is much simplified now, but still trying to have some thing to catch regression of orgSplit so that I can sleep better :)

jarbogast:
@huminzhi No worries. I will absolutely look into adding those other two test cases since we found out recently that they're pretty important! Actually mocking that out will be a little interesting, but I'm sure I can find a way to make it happen :)

In a bigger picture sense, there's still some things here that would be nice to address in an ideal world. 

- We still have this 'weird' accessor for baseUrl. Its still a little too magical to me. The tests I have and the tests I'm going to add hint at this. Why does enqueuing an action lead to a change in the baseURL? Its just a little unexpected.
- Network Actions also still know about the 'network' they're being executed on. This feels a little weird to me as well. Kinda feels like a view knowing about its view controller or some other child entity knowing about its parent.
- We also have Actions (Operations) that are very mutable. My caveman brain wants Operations that get initialized with 'all the things' they need, and then none of those change for the lifetime of the Operation. If we need to replay an Action for some reason, I was expecting that we'd create a new Action with the new set of data that it needs (in this case maybe a new baseURL).
- At some point I also saw Actions called start on themselves. That also breaks my caveman brain understanding about how Operations and Operation Queues are supposed to work.

We certainly won't address those things here in this PR(!), but just wanted to mention those as some of the 'surprises' I found as someone who is used to working with Operations, but new to this specific codebase.

PR #1933, opened 2017-02-22
Fix NSTimer Race Condition

sgoldberg-sfdc:
- Original code was developed to ensure thread is stopped on same thread it is created on
- This can also be accomplished by ensuring it is always run on the main thread (which in practice it most often was)
- If the timer had been created on a background thread this could result in a race condition between the performselector call and the thread terminating.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -135,7 +135,6 @@ static NSString * const kSFAppFeatureSafariBrowserForLogin   = @"BW";
 @synthesize approvalCode                = _approvalCode;
 @synthesize scopes                      = _scopes;
 @synthesize refreshFlowConnectionTimer  = _refreshFlowConnectionTimer;
-@synthesize refreshTimerThread          = _refreshTimerThread;
 @synthesize advancedAuthConfiguration   = _advancedAuthConfiguration;
 @synthesize advancedAuthState           = _advancedAuthState;
 @synthesize codeVerifier                = _codeVerifier;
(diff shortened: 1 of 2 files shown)


sgoldberg-sfdc:
@bhariharan @khawkins

bhariharan:
test me

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -932,39 +931,43 @@ - (void)resetWebUserAgent
 
 - (void)startRefreshFlowConnectionTimer
 {
-    self.refreshTimerThread = [NSThread currentThread];
+    if (![NSThread isMainThread]) {
+        __weak typeof(self) weakSelf = self;
```
khawkins: `dispatch_async()` will not capture `self`, and in this case it's possible to lose the reference to `self` as a result.  I'd remove the casting here.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -932,39 +931,43 @@ - (void)resetWebUserAgent
...
+                                                                          repeats:NO];
 }
 
 - (void)stopRefreshFlowConnectionTimer
 {
-    if (self.refreshFlowConnectionTimer != nil && self.refreshTimerThread != nil) {
-        [self performSelector:@selector(invalidateRefreshTimer) onThread:self.refreshTimerThread withObject:nil waitUntilDone:YES];
-        [self cleanupRefreshTimer];
+    if (![NSThread isMainThread]) {
+        __weak typeof(self) weakSelf = self;
```
khawkins: Same here.

sgoldberg-sfdc:
Removed weak/strong dance per @khawkins

bhariharan:
test me

sgoldberg-sfdc:
All green!

sgoldberg-sfdc:
@khawkins  and @bhariharan - On second thought, I think it would be better to replace this NStimer with a dispatch_after or dispatch_source_set_timer().

PR #1934, opened 2017-02-24
Danger

huminzhi:
just for test ci


fastlane/Fastfile
```diff
@@ -35,3 +35,27 @@ lane :CI do
 	end
 
 end
+
+
+lane  :localtest do
+#setup test accounts (for sf dev only)
+	`git clone https://git.soma.salesforce.com/gist/d103e1ef29e1c11ee9a7a4733a1193d8.git`
```
bhariharan: I'd prefer not to have internal references in our external facing environment. We can do this locally on our boxes if needed.

PR #1941, opened 2017-03-02
Fixing bug with ‘baseUrl’ computation

bhariharan:


sgoldberg-sfdc:
Is there a unit test for that case? seems an easy one to test for.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFAction.m
```diff
@@ -97,12 +97,17 @@ - (NSURL*)urlForActionWithError:(NSError**)error {
...
-    if ([baseURL.absoluteString hasSuffix:@"/"] && [path hasPrefix:@"/"]) {
-        [path deleteCharactersInRange:NSMakeRange(0, 1)];
+
+    // 'baseUrl' needs to account for URLs with extra trailing slashes, like custom community URLs.
+    NSMutableString *baseUrlString = [NSMutableString stringWithFormat:@"%@", baseURL.absoluteString];
+    if ([baseUrlString hasSuffix:@"/"]) {
+        [baseUrlString deleteCharactersInRange:NSMakeRange(0, 1)];
     }
-    
-    NSURL *url = [NSURL URLWithString:path relativeToURL:baseURL];
```
bhariharan: This doesn't account for the case where `baseURL` is more than the domain itself and has something after the trailing slash, like a custom community URL.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFAction.m
```diff
@@ -97,12 +97,17 @@ - (NSURL*)urlForActionWithError:(NSError**)error {
...
+    if (![path hasPrefix:@"/"]) {
+        baseUrlString = [NSMutableString stringWithFormat:@"%@/", baseUrlString];
+    }
+    NSString *urlString = [NSString stringWithFormat:@"%@%@", baseUrlString, path];
```
wmathurin: Can't we do that with [pathWithComponents](https://developer.apple.com/reference/foundation/nsstring/1414489-pathcomponents?language=objc)?

wmathurin: Never mind, I just saw the pathWithComponents only works with file paths—not, for example, string representations of URLs.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFAction.m
```diff
@@ -97,12 +97,17 @@ - (NSURL*)urlForActionWithError:(NSError**)error {
         }
         return nil;
     }
-    
-    if ([baseURL.absoluteString hasSuffix:@"/"] && [path hasPrefix:@"/"]) {
-        [path deleteCharactersInRange:NSMakeRange(0, 1)];
+
+    // 'baseUrl' needs to account for URLs with extra trailing slashes, like custom community URLs.
+    NSMutableString *baseUrlString = [NSMutableString stringWithFormat:@"%@", baseURL.absoluteString];
```
wmathurin: Couldn't we do
```
NSMutableString *baseUrlString = [NSMutableString stringWithString:baseURL.absoluteString];
```

bhariharan:
Simplified the logic a bit and added a test.

PR #1944, opened 2017-03-07
- Fix deadlock caused by recursive call on [SFUserAccountManager sharedInstance]

samipdesai:


samipdesai:
@bhariharan please review.

sgoldberg-sfdc:
@samipdesai I do not see the recursion, can you give more details on what was happening prior to the fix.

samipdesai:
@sgoldberg-sfdc see this. https://rink.hockeyapp.net/manage/apps/71830/app_versions/1850/crash_reasons/159868042

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
```diff
@@ -63,6 +63,10 @@ + (instancetype)sharedPreferences {
     dispatch_once(&pred, ^{
         preferences = [[self alloc] init];
     });
+    
+    dispatch_once(&pred, ^{
```
khawkins: This is not doing what you think.  Given that `pred` is "exhaused" in the prior `dispatch_once` call, the second call is not being made at all.

samipdesai: yes indeed. i have no clue how that worked the first time. this mean we need to change it in other places like SFuserAccountManager as well.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
```diff
@@ -94,6 +95,13 @@ - (id)init {
     return self;
 }
 
+- (void)initializeManagedPreferences {
+    self.rawPreferences = [[NSUserDefaults msdkUserDefaults] dictionaryForKey:kManagedConfigurationKey];
+    if (self.rawPreferences) {
```
khawkins: Why don't we just move this block to `[SalesforceSDKManager configureManagedSettings]`, and avoid the deadlock altogether.  This is ultimately for a feature contained in `SalesforceSDKManager` anyway.

khawkins: But ultimately, @bhariharan, @trooper2013, and @wmathurin I think these app feature loggings would be better managed through an intermediary, that basically fronted the storage of these "features" (e.g. in `NSUserDefaults`) on the one hand, and the retrieval from e.g. `SalesforceSDKManager` on the other.  That way, these singleton state issues would never come into play.  My $0.02.

PR #1945, opened 2017-03-08
See iphone9,1 and iphone9,2 on feedback emails

liguopku:
add models for iphone7 and 7plus based on the following website.
https://www.theiphonewiki.com/wiki/Models


bhariharan:
build

PR #1947, opened 2017-03-09
Still Need to Load Legacy Host List

jarbogast:
On a recent app upgrade, all our users lost their list of custom login hosts.

Turns out the keys used to store these hosts changed and so those login hosts got 'orphaned' in user defaults. This change attempts to salvage those login hosts from their 'legacy' location in user defaults.

Still trying to come up with the best way to test drive this without blowing up the existing public interface :)

@bhariharan @khawkins

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
@@ -38,14 +38,13 @@
 
 @end
 
-// Key under which the list of login hosts will be stored in the user defaults.
 static NSString * const SFSDKLoginHostList = @"SalesforceLoginHostListPrefs";
-
-// Key for the host.
 static NSString * const SFSDKLoginHostKey = @"SalesforceLoginHostKey";
-
-// Key for the name.
 static NSString * const SFSDKLoginHostNameKey = @"SalesforceLoginHostNameKey";
+
+static NSString * const SFSDKLoginHostListLegacy = @"ChatterLoginHostListPrefs";
+static NSString * const SFSDKLoginHostLegacyKey = @"ChatterLoginHostKey";
+static NSString * const SFSDKLoginHostNameLegacyKey = @"ChatterLoginHostNameKey";
 
 @implementation SFSDKLoginHostStorage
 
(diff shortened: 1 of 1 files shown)


huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

## Current coverage for SalesforceSDKCore.framework is `49%`
Files changed | - | - 
--- | --- | ---
SFSDKLoginHostStorage.m | `61%` | :warning:

---

---

</p>

urthakkar:
Thanks @jarbogast for the PR, we will look into it.  However, please note that we support upgrades only from the previous version and you are very likely to run into such issues if you are skipping versions.

khawkins:
@jarbogast I would prefer to consider a solution more primarily focused on the S1 side, given that these keys/values have never otherwise been represented in the Mobile SDK.

jarbogast:
@khawkins Those keys **did** exist in MobileSDK! This file got moved from Login to Login/LoginHost, and we lost the older history on it :(

@urthakkar Ok, sounds like we mis-used the SDK by not upgrading to each successive release. We'll have to keep that in mind for S1 in general that not upgrading to each release will cause issues!

khawkins:
Oh my bad, so they did!  I didn't initially see them in the repo history.  Then let's reopen this and evaluate it on its merits.  I'd personally be okay supporting an upgrade case where one erroneously wasn't supplied before.

jarbogast:
Ok, I'll work on adding some tests!

khawkins:
...although it never made it off of a feature branch.  I.e. it was reverted out of review before coming into `dev` ([#1368](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1368)).

I dunno...I'm not a super fan of supporting legacy that doesn't actually exist for public SDK customers—it's just another avenue for something to go wrong, that lives on in virtual perpetuity.

I'll chat with you offline to see what we can do.

jarbogast:
Ugghh. You're right. We somehow used something that never made it into a public release. What an archaeological dig this has been.

Going to think about alternate solutions as well.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
```diff
@@ -38,15 +38,14 @@ @interface SFSDKLoginHostStorage ()
...
-// Key under which the list of login hosts will be stored in the user defaults.
 static NSString * const SFSDKLoginHostList = @"SalesforceLoginHostListPrefs";
-
-// Key for the host.
 static NSString * const SFSDKLoginHostKey = @"SalesforceLoginHostKey";
-
-// Key for the name.
 static NSString * const SFSDKLoginHostNameKey = @"SalesforceLoginHostNameKey";
 
+static NSString * const SFSDKLoginHostListLegacy = @"ChatterLoginHostListPrefs";
```
bhariharan: I'd rather not have specific internal key references here such as `Chatter...`, especially since this never made it to public SDK. Let's consider a solution implemented in the app layer instead.

jarbogast: I agree :) I actually forgot this was still open!

PR #1948, opened 2017-03-10
- fix deadlock due to dispatch_once on [SFUserAccountManager init].

samipdesai:
https://rink.hockeyapp.net/manage/apps/71830/app_versions/1856/crashes/10746012307

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
@@ -49,7 +49,7 @@ static NSString * const kSFDisableExternalPaste = @"DISABLE_EXTERNAL_PASTE";
 @interface SFManagedPreferences ()
 
 @property (nonatomic, strong, readwrite) NSDictionary *rawPreferences;
-@property (nonatomic, strong) NSOperationQueue *syncQueue;
+@property (nonatomic, strong) dispatch_queue_t syncQueue;
 
 @end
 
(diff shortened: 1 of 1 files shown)


samipdesai:
@bhariharan @khawkins . i am seeing another deadlock see the hockeyApp link for reference.

huminzhi:
build

khawkins:
Not to say that we shouldn't find more graceful failure fallbacks than deadlocks to these situations, but the reason that they're coming up in your case is that `SalesforceSDKManager` is not being initialized/`launch`ed as the first access to the Mobile SDK, which is an anti-pattern.  The SDK manager launch process is meant to precede all other access to SDK components.

samipdesai:
Thanks @khawkins! i just spoke with @sgoldberg-sfdc and I am currently investigating if it possible to change things on our side as well.

huminzhi:
build

khawkins:
I've got a WIP to remove `SalesforceSDKManager` from the responsibility chain of these feature settings.  Stay tuned.

huminzhi:
build

khawkins:
@samipdesai I think [#1949](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1949) should alleviate the kinds of deadlocks you're seeing, when it's merged.

samipdesai:
@khawkins thats a much better fix.  quick question. will this have any detrimental effects if i cherry pick the commits on to master for out team fork? We need to address this in an upcoming release.

khawkins:
I wouldn't imagine, as it just reorganized the existing functionality into different ownership.

huminzhi:
build

samipdesai:
I will close this in favor of @khawkins PR.

PR #1949, opened 2017-03-10
App feature registration separated from SalesforceSDKManager

khawkins:
We keep coming across issues where the fact that `SalesforceSDKManager` owns the registration and un-registration of the app feature analytics markers is causing bad behavior, deadlocks, etc.  The SDK manager shouldn't really own that process anyway.  I moved it out into a basic static class with no dependencies, which the SDK manager and feature registers can collectively depend on.


huminzhi:
build

huminzhi:
build

sgoldberg-sfdc:
Great refactor. 👍

PR #1950, opened 2017-03-14
Fix bug with sid not being set for communityUrl

bhariharan:


bhariharan:
`apiUrl` falls back to `instanceUrl` in the org context (regular login). However, it uses `communityUrl` in the community login context.

PR #1954, opened 2017-03-17
Safeguard against nil key in certain race conditions

bhariharan:


sgoldberg-sfdc:
Hmm I don't know... let me think about this one.

khawkins:
:shipit:

PR #1958, opened 2017-03-22
Guarding logout to only get processed once

khawkins:
If multiple threads enter `logoutUser:` with the same user, this should ensure that the logout logic is only processed once.

@trooper2013 @bhariharan I formed this up so that, when merged with the user account management PR, you should just be able to accept your version of the files.  I think it's totally compatible with the changes there, but let me know if you see any disparities.  Attn: @sgoldberg-sfdc

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
@@ -26,7 +26,7 @@
 #import "SFApplication.h"
 #import "SFAuthenticationManager+Internal.h"
 #import "SalesforceSDKManager+Internal.h"
-#import "SFUserAccount.h"
+#import "SFUserAccount+Internal.h"
 #import "SFUserAccountManager.h"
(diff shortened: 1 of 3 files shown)


sgoldberg-sfdc:
I will check this out tomorrow. I have seen some weird behavior in the logs though. It looks like part of the logout workflow is to re-call "launch" on the MobileSDKManager.  Is this expected use of the SDK Manager?

In at least one of the crash instances it looks to me like the race condition is between the logout and a new bootstrapping of the MobileSDK. They all appear to involve revoked refresh tokens.

khawkins:
That sounds on its surface like an S1 bootstrap state issue.  There's nothing in the Mobile SDK that's prescriptive about post-logout actions.  Our default template apps implement a pattern in `AppDelegate` that responds by logging in again if there are no other users, but that's wholly dependent on the consumer.

khawkins:
build

khawkins:
Not sure what the CI status is here.  Let me know and/or merge away when it's okay to do so.

bhariharan:
@khawkins Our CI env is quite broken at the moment. I'm going to merge this. @sgoldberg-sfdc Feel free to cherry-pick after it's merged.

PR #1964, opened 2017-03-23
Check for nil userKey

rafael-chavez:
@bhariharan @bhariharan


rafael-chavez:
actually i see that you have already addressed this @khawkins https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1958

PR #1978, opened 2017-03-30
Fixing bug where all users’ AILTN logs are being uploaded

bhariharan:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -288,6 +288,11 @@ - (BOOL) readAnalyticsPolicy {
 }
 
 - (void) publishOnAppBackground {
+
+    // Publishing should only happen for the current user, not for all users signed in.
+    if (![self.userAccount.idData isEqual:[SFUserAccountManager sharedInstance].currentUser.idData]) {
```
khawkins: Shouldn't this comparison be on `accountIdentity`?

bhariharan: Argh. Forgot to commit. Yes.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -288,6 +288,11 @@ - (BOOL) readAnalyticsPolicy {
 }
 
 - (void) publishOnAppBackground {
+
+    // Publishing should only happen for the current user, not for all users signed in.
+    if (![self.userAccount.accountIdentity isEqual:[SFUserAccountManager sharedInstance].currentUser.accountIdentity]) {
+        return;
+    }
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
wmathurin: What if the dispatch_async is called and the user switch, would the events still be published under the wrong user then?

khawkins:
The only other thing that I'd say is that this is an artificial limitation that should be addressed more comprehensively at some point.  To wit, we've siloed some of the analytics management by user, while de facto relying on `currentUser` in other parts of the system (notably, here: publishing).

Ideally, it should all be siloed by user, so that current user context doesn't matter for things like publishing.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
bhariharan: This call gets triggered after app is backgrounded - user switch can't happen then.

bhariharan:
@khawkins I agree - there's a larger body of work I plan on doing in the `5.2` timeframe on this.

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

## Current coverage for SalesforceAnalytics.framework is `66%`

---

</p>

PR #1979, opened 2017-03-31
Warnings - Inconsistent prototype/implementation

sgoldberg-sfdc:
-  The implementation of some of the initWithJSON methods accept either NSArray or NSDictionary
-  The header reflected only NSDictionary leading to compile warnings


sgoldberg-sfdc:
@bhariharan @khawkins @trooper2013

PR #1980, opened 2017-04-04
Replaced KVO with Notifications

trooper2013:
- Refactored UserAccountManager to use notifications for a given user.
- Changed Notification observers and senders to use SFUserAccountManagerDidChangeUserDataNotification.
- Changed SFOAuthCoordinator code to collect and reset changed values. The changes are applied through the SFUserAccountManager.
- Changed CSFNetwork+Salesforce & CSFSalesforceAction to reflect changes in the pattern.
- Added AccountManager Notifications tests.


bhariharan:
@sgoldberg-sfdc FYI

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -738,7 +743,12 @@ - (void)finalizeAuthCompletion
 {
     // Apply the credentials that will ensure there is a user and that this
     // current user as the proper credentials.
-    SFUserAccount *user = [[SFUserAccountManager sharedInstance] applyCredentials:self.coordinator.credentials];
+    SFUserAccountChange change = [self userAccountChangeFromDict:self.coordinator.credentialsChangeSet];
```
khawkins: This seems like an area of responsibility for `SFUserAccountManager`.  Why not make this assessment there?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -589,29 +601,49 @@ - (BOOL)hasCommunityChanged {
...
+    [self notifyUserChange:SFUserAccountManagerDidChangeUserDataNotification withUser:user andChange:change];
+}
+
+- (void)notifyUserChange:(NSString *)notificationName withUser:(SFUserAccount *)user andChange:(SFUserAccountChange)change {
+    if (user) {
+        [[NSNotificationCenter defaultCenter] postNotificationName:notificationName
+                                                            object:self
+                                                          userInfo:@{
+                                                                  SFUserAccountManagerUserChangeKey: @(change),
+                                                                SFUserAccountManagerUserChangeUserIdKey: user.accountIdentity.userId,
```
khawkins: Why are we not simply sending the `SFUserAccount` that changed?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
```diff
@@ -39,17 +39,18 @@ - (void)setDefaultConnectCommunityId:(NSString *)defaultConnectCommunityId {
 - (void)setupSalesforceObserver {
     [[NSNotificationCenter defaultCenter] addObserver:self
                                              selector:@selector(userAccountManagerDidChangeCurrentUser:)
```
khawkins: Can we change the name of the selector as well?  It doesn't reflect the nature of the notification anymore.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
```diff
@@ -39,17 +39,18 @@ - (void)setDefaultConnectCommunityId:(NSString *)defaultConnectCommunityId {
...
 }
 
 #pragma mark SFAuthenticationManagerDelegate
-
 - (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
     SFUserAccountManager *accountManager = (SFUserAccountManager*)notification.object;
-    if ([accountManager isKindOfClass:[SFUserAccountManager class]]) {
+    SFUserAccountChange change = (SFUserAccountChange)[notification.userInfo[SFUserAccountManagerUserChangeKey] integerValue];
+    if ([accountManager isKindOfClass:[SFUserAccountManager class]]
+            && (change & SFUserAccountChangeCurrentUser)) {
```
khawkins: This is adding dependencies on `currentUser` that do not otherwise exist.  `CSFNetwork` instances are siloed by user.  It shouldn't care about changes to any other user than the account that's configured for this instance.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
```diff
@@ -39,17 +39,18 @@ - (void)setDefaultConnectCommunityId:(NSString *)defaultConnectCommunityId {
...
         if ([accountManager.currentUserIdentity isEqual:self.account.accountIdentity] &&
-            ![accountManager.currentCommunityId isEqualToString:self.defaultConnectCommunityId])
+                ![accountManager.currentCommunityId isEqualToString:self.defaultConnectCommunityId])
```
khawkins: As mentioned before, it would be better to be able to do a simple user-to-user comparison, by passing the user account in the notification.

libs/SalesforceSDKCore/SalesforceSDKCore/SFUserAccountManagerNotificationsTests.m
```diff
@@ -0,0 +1,561 @@
+//
```
bhariharan: Use our copyright and move this under `SalesforceSDKCoreTests` folder.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -589,29 +601,49 @@ - (BOOL)hasCommunityChanged {
...
 }
 
-- (void)userChanged:(SFUserAccountChange)change {
-    if (![self.lastChangedOrgId isEqualToString:self.currentUser.credentials.organizationId]) {
-        self.lastChangedOrgId = self.currentUser.credentials.organizationId;
+- (void)userChanged:(SFUserAccount *)user change:(SFUserAccountChange)change {
+    [self notifyUserChange:SFUserAccountManagerDidChangeUserDataNotification withUser:user andChange:change];
+}
+
+- (void)notifyCurrentUserChange:(SFUserAccountChange)change {
```
khawkins: Ah.  This feels like an undesirable mashing up of two different notifications.  I'm wondering if we should just keep two notifications—they really fulfill two different use cases.

One is listening for the fact that the current user has switched, which is really about changing to an entirely different entity, so granular parsing of what changed below that is really beside the point.  It's a totally different user.  It's all presumably changed, at that point.

The other is listening for specific changes on a given user, where granular denoting of what changed is desirable.

They're really two different things.  Mashing them together distorts the purpose of the notification.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -589,29 +601,49 @@ - (BOOL)hasCommunityChanged {
...
+    SFUserAccount *user  = [self currentUser];
+    if (![self.lastChangedOrgId isEqualToString:user.credentials.organizationId]) {
+        self.lastChangedOrgId = user.credentials.organizationId;
         change |= SFUserAccountChangeOrgId;
         change &= ~SFUserAccountChangeUnknown; // clear the unknown bit
     }
 
     if (![self.lastChangedUserId isEqualToString:self.currentUser.credentials.userId]) {
-        self.lastChangedUserId = self.currentUser.credentials.userId;
+        self.lastChangedUserId = user.credentials.userId;
```
khawkins: Plus, it would be super sweet to get rid of all of these disconnected state management variables that get out of sync at the drop of a hat.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
sgoldberg-sfdc: Agree

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
```diff
@@ -324,18 +296,37 @@ + (BOOL)isNetworkError:(nullable NSError *)error {
...
+
+        NSString *userId = notification.userInfo[SFUserAccountManagerUserChangeUserIdKey];
+        NSString *orgId = notification.userInfo[SFUserAccountManagerUserChangeOrgIdKey];
+        SFUserAccountIdentity *identity = [[SFUserAccountIdentity alloc] initWithUserId:userId orgId:orgId];
+        SFUserAccount *userAccount = [[SFUserAccountManager sharedInstance] userAccountForUserIdentity:identity];
+        SFUserAccountChange change = (SFUserAccountChange)[notification.userInfo[SFUserAccountManagerUserChangeKey] integerValue];
+
+        if (change & SFUserAccountChangeCurrentUser) {
+            if (![accountManager.currentUserIdentity isEqual:self.enqueuedNetwork.account.accountIdentity]) {
+                self.enqueuedNetwork.networkSuspended = YES;
```
MikeNachbaur: We shouldn't ever set the `networkSuspended` property internally; this is meant to be used by the developer consuming the network stack to suspend/resume the network on their own.

This will also break instances where clean-up operations, or background operations need to be performed on one user while the account manager switches to a different user identity.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
```diff
@@ -324,18 +296,37 @@ + (BOOL)isNetworkError:(nullable NSError *)error {
...
-        if (accountManager.currentCommunityId != self.enqueuedNetwork.defaultConnectCommunityId) {
-            self.enqueuedNetwork.defaultConnectCommunityId = accountManager.currentCommunityId;
+
+        NSString *userId = notification.userInfo[SFUserAccountManagerUserChangeUserIdKey];
+        NSString *orgId = notification.userInfo[SFUserAccountManagerUserChangeOrgIdKey];
+        SFUserAccountIdentity *identity = [[SFUserAccountIdentity alloc] initWithUserId:userId orgId:orgId];
+        SFUserAccount *userAccount = [[SFUserAccountManager sharedInstance] userAccountForUserIdentity:identity];
+        SFUserAccountChange change = (SFUserAccountChange)[notification.userInfo[SFUserAccountManagerUserChangeKey] integerValue];
+
+        if (change & SFUserAccountChangeCurrentUser) {
```
MikeNachbaur: Is there any reason why the salesforce actions need to know that the "Current User" changes?  The network instances shouldn't care who the current user is; rather, if the application using the network decides to make network requests after switching a user, it's their prerogative to utilize the network via that new user.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
```diff
@@ -324,18 +296,37 @@ + (BOOL)isNetworkError:(nullable NSError *)error {
 }
 
 #pragma mark SFAuthenticationManagerDelegate
-
-- (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
+- (void)userAccountManagerDidChangeUserDataNotification:(NSNotification*)notification {
     SFUserAccountManager *accountManager = (SFUserAccountManager*)notification.object;
```
MikeNachbaur: Is there any reason why the notification's object is the `SFUserAccountManager`? Wouldn't it make more sense for the object to be the user account that's being changed?  That way, the action can subscribe to the notification based on that object.  e.g.

```
[[NSNotificationCenter defaultCenter] addObserver:self
                                   selector:@selector(userAccountManagerDidChangeUserDataNotification:)
                                       name:SFUserAccountManagerDidChangeUserDataNotification
                                     object:self.enqueuedNetwork.account];
```

That way the actions will only receive notifications for changes to the account they care about.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -853,17 +853,59 @@ - (NSString *)scopeQueryParamString {
...
+    }
+
+    if (params[kSFOAuthIssuedAt]) {
+        self.credentials.issuedAt = [[self class] timestampStringToDate:params[kSFOAuthIssuedAt]];
+    }
+
+    if (params[kSFOAuthInstanceUrl]) {
+        NSURL *newInstanceUrl = [NSURL URLWithString:params[kSFOAuthInstanceUrl]];
+
+        if(![newInstanceUrl isEqual:self.credentials.instanceUrl])
```
MikeNachbaur: The lack of parenthesis around these conditional blocks creeps me out, especially with the subsequent inline conditionals used.  We all remember the OpenSSL bug...

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -853,17 +853,59 @@ - (NSString *)scopeQueryParamString {
  - communityId
  - communityUrl
  */
-
 - (void) updateCredentials:(NSDictionary *) params {
+    NSMutableDictionary *changeSet = [NSMutableDictionary new];
+    if (params[kSFOAuthAccessToken]) {
+        if(![params[kSFOAuthAccessToken] isEqualToString:self.credentials.accessToken])
```
MikeNachbaur: Inconsistent / risky mixture of conditionals without parenthesis, combined with inline conditions...

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -853,17 +853,59 @@ - (NSString *)scopeQueryParamString {
...
+                    self.credentials.communityId:[NSNull null],
+                    params[kSFOAuthCommunityId]];
+        self.credentials.communityId = params[kSFOAuthCommunityId];
+    }
+
+    if (params[kSFOAuthCommunityUrl]) {
+        NSURL *newCommUrl = [NSURL URLWithString:params[kSFOAuthCommunityUrl]];
+        if(![newCommUrl isEqual:self.credentials.communityUrl])
+            changeSet[kSFOAuthCommunityUrl] = @[self.credentials.communityUrl?
+                    [self.credentials.communityUrl absoluteString]:[NSNull null],
```
MikeNachbaur: I'm wondering why you need to pass `NSNull` at all?  Instead of using a dictionary with string constant keys, and a nested array whose position represents old/new values (if I'm reading this code correctly), it might be better to write this up as a proper model object.

```
@interface SFOAuthCredentialChange : NSObject

@property (nonatomic, copy, readonly) NSString *type;
@property (nonatomic, strong, readonly) id oldValue;
@property (nonatomic, strong, readonly) id newValue;

@end
```

Then it could be used as: `NSArray<SFOAuthCredentialChange*> *changes`.  Or something along those lines?  It would certainly make it easier because then you wouldn't have to compare against `[NSNull null]`, you can override `isEqual` as needed, etc.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -76,6 +77,10 @@
 // Oauth
 static NSString * const kSFUserAccountOAuthLoginHostDefault = @"login.salesforce.com"; // last resort default OAuth host
 static NSString * const kSFUserAccountOAuthLoginHost = @"SFDCOAuthLoginHost";
+static NSString * const kSFAuthAccessToken = @"access_token";
```
MikeNachbaur: Might I suggest making these `extern` keys instead?  I could see situations where these constants can be useful to a consuming app.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.h
```diff
@@ -283,7 +294,7 @@ FOUNDATION_EXTERN NSString * const kSFLoginHostChangedNotificationUpdatedHostKey
  if you don't know what kind of change was made to this object and this method
  will try to determine that.
  */
-- (void)userChanged:(SFUserAccountChange)change;
+- (void)userChanged:(SFUserAccount *)user change:(SFUserAccountChange)change;
```
MikeNachbaur: You should update the headerdoc for this method to reflect the new property added.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
MikeNachbaur: Also, under some circumstances, consumers may not actually care that the current user has switched.  Largely that feels like an app-oriented thing.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
```diff
@@ -38,20 +40,53 @@ - (void)setDefaultConnectCommunityId:(NSString *)defaultConnectCommunityId {
...
+                                                 name:SFUserAccountManagerDidChangeUserNotification
                                                object:nil];
 }
 
 #pragma mark SFAuthenticationManagerDelegate
+- (void)userAccountManagerDidChangeUser:(NSNotification*)notification {
 
-- (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
     SFUserAccountManager *accountManager = (SFUserAccountManager*)notification.object;
+    SFUserAccountChange change = (SFUserAccountChange)[notification.userInfo[SFUserAccountManagerUserChangeKey] integerValue];
```
khawkins: I'm having a hard time understanding what this logic is meant to accomplish.  So I'm going to walk through a bit, and see if I can unpack some of this.

I know that, prior to this PR, this was implemented:
```obj-c
- (void)userAccountManagerDidChangeCurrentUser:(NSNotification*)notification {
    SFUserAccountManager *accountManager = (SFUserAccountManager*)notification.object;
    if ([accountManager isKindOfClass:[SFUserAccountManager class]]) {
        if ([accountManager.currentUserIdentity isEqual:self.account.accountIdentity] &&
            ![accountManager.currentCommunityId isEqualToString:self.defaultConnectCommunityId])
        {
            self.defaultConnectCommunityId = accountManager.currentCommunityId;
        }
    }
}
```
My guess is that the original implementation was something of a hack—the only good way to listen to user changes was to listen to _currentUser_ changes, because that's all `SFUserAccountManager` posted on.  Since `CSFNetwork` instances are per-user, it then put the additional sanity guard in: "Is this my user?"  And then proceeded accordingly.

In our new paradigm, we're reporting on user changes for _any_ user.  So "changed current user" listening can simply go away in this class now, in favor of user-specific changes.

And I see what @MikeNachbaur is saying now, with regards to the notification "sender" of `SFUserAccountManagerDidChangeUserDataNotification`: this should really be a notification that the specific changing user "owns".  If we amend the notification to make the user account the owner, then it gives consumers the option to only listen to that user (useful in cases like this one), while still having the `nil` fallback if facilities want to listen to changes for any user.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
```diff
@@ -38,20 +40,53 @@ - (void)setDefaultConnectCommunityId:(NSString *)defaultConnectCommunityId {
...
+                                               object:nil];
+
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(userAccountManagerDidChangeUser:)
+                                                 name:SFUserAccountManagerDidChangeUserNotification
                                                object:nil];
 }
 
 #pragma mark SFAuthenticationManagerDelegate
+- (void)userAccountManagerDidChangeUser:(NSNotification*)notification {
```
khawkins: So let's ditch this notification handling entirely.  It's superfluous now that we have the two notification types.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
```diff
@@ -38,20 +40,53 @@ - (void)setDefaultConnectCommunityId:(NSString *)defaultConnectCommunityId {
...
+    SFUserAccountDataChange change = (SFUserAccountDataChange)[notification.userInfo[SFUserAccountManagerUserChangeKey] integerValue];
     if ([accountManager isKindOfClass:[SFUserAccountManager class]]) {
-        if ([accountManager.currentUserIdentity isEqual:self.account.accountIdentity] &&
-            ![accountManager.currentCommunityId isEqualToString:self.defaultConnectCommunityId])
-        {
-            self.defaultConnectCommunityId = accountManager.currentCommunityId;
+        SFUserAccount *userAccount = notification.userInfo[SFUserAccountManagerUserChangeUserKey];
+        if([self.account.accountIdentity isEqual:userAccount.accountIdentity]) {
+
+            if ([accountManager.currentUserIdentity isEqual:self.account.accountIdentity] &&
```
khawkins: This part of the compound condition should not be here.  Actionability is only dependent on the notification coming from the user that matches this instance's account—which is checked immediately before this.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
```diff
@@ -323,20 +292,23 @@ + (BOOL)isNetworkError:(nullable NSError *)error {
...
-    SFUserAccountManager *accountManager = (SFUserAccountManager*)notification.object;
-    if ([accountManager isKindOfClass:[SFUserAccountManager class]]) {
-        if (![accountManager.currentUserIdentity isEqual:self.enqueuedNetwork.account.accountIdentity]) {
-            self.enqueuedNetwork.networkSuspended = YES;
-        } else {
-            [self.enqueuedNetwork resetSession];
-            self.enqueuedNetwork.networkSuspended = NO;
+        [self willChangeValueForKey:@"isReady"];
+        if (change & SFUserAccountDataChangeCommunityId) {
+            self.enqueuedNetwork.defaultConnectCommunityId = userAccount.communityId;
```
khawkins: This is redundant to the logic in `CSFNetwork`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
```diff
@@ -323,20 +292,23 @@ + (BOOL)isNetworkError:(nullable NSError *)error {
...
         }
-        if (accountManager.currentCommunityId != self.enqueuedNetwork.defaultConnectCommunityId) {
-            self.enqueuedNetwork.defaultConnectCommunityId = accountManager.currentCommunityId;
+        SFUserAccountDataChange credsChanged = SFUserAccountDataChangeInstanceURL | SFUserAccountDataChangeAccessToken;
+        if ((change & credsChanged) == credsChanged) {
+            self.credentialsReady = YES;
```
khawkins: This isn't sufficient.  `credentialsReady = YES` depends on `accessToken` and `instanceUrl` being set.  It should be `NO` otherwise.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
khawkins: +1. This is the model object I was talking about to replace the more generic dictionary approach of `credentialsChangeSet`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -392,20 +396,48 @@ - (BOOL)deleteAccountForUser:(SFUserAccount *)user error:(NSError **)error {
...
 - (SFUserAccount *)applyCredentials:(SFOAuthCredentials*)credentials {
+    return [self applyCredentials:credentials withIdData:nil];
+}
+
+- (SFUserAccount *)applyCredentials:(SFOAuthCredentials*)credentials withIdData:(SFIdentityData *) identityData {
+    return [self applyCredentials:credentials withIdData:identityData andNotification:YES];
+}
 
-    SFUserAccountChange change = SFUserAccountChangeCredentials;
+- (SFUserAccount *)applyCredentials:(SFOAuthCredentials*)credentials withIdData:(SFIdentityData *) identityData andNotification:(BOOL) shouldSendNotification{
```
khawkins: Why would you not send the notification?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -589,29 +628,54 @@ - (BOOL)hasCommunityChanged {
...
-    }
+- (void)userChanged:(SFUserAccount *)user change:(SFUserAccountDataChange)change {
+    [self notifyUserDataChange:SFUserAccountManagerDidChangeUserDataNotification withUser:user andChange:change];
+}
 
-    if (![self.lastChangedUserId isEqualToString:self.currentUser.credentials.userId]) {
-        self.lastChangedUserId = self.currentUser.credentials.userId;
-        change |= SFUserAccountChangeUserId;
-        change &= ~SFUserAccountChangeUnknown; // clear the unknown bit
+- (void)notifyCurrentUserChange:(SFUserAccountDataChange)change {
```
khawkins: Nothing calls this method.  Can it be removed?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
trooper2013: @khawkins So we should associate the UserAccount to the notification? The Notification is being posted by SFUserAccountManager.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
trooper2013: @khawkins the only other changes I see here are ID data & communityId? Should they affect the readiness of the operation?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
trooper2013: @khawkins Can be used by  anyone consuming this API directly and not wanting notifications but simply addition of the user account.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Queue/CSFNetwork+Salesforce.m
khawkins: It's being posted by SFUserAccountManager, but reasonably speaking, it's being posted on behalf of `SFUserAccount`.  I think it makes sense to designate the user account as the owner of the notification.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Action/CSFSalesforceAction.m
khawkins: It's not the set of properties involved that I'm talking about; it's their values.  If one or both of these properties have changed to `nil` (`NSNull`, I guess), `credentialsReady` should be `NO`.  It's not sufficient to simply look at the fact that the values changed; we have to look at what they changed to.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
khawkins: Okay.  I'd argue that there's a general refactor point for isolating account creation from account notification, but it works this way too.

khawkins:
Looking pretty good to me, over all. 👍

libs/SalesforceSDKCore/SalesforceSDKCore/SFUserAccountManagerNotificationsTests.m
```diff
@@ -0,0 +1,512 @@
+/*
+ Copyright (c) 2017-present, salesforce.com, inc. All rights reserved.
```
bhariharan: @trooper2013 Move this file under `SalesforceSDKCoreTests`.

trooper2013: @bhariharan How does Xcode even do this? Logically it still shows under tests group. I'll move it. Thanks for catching it.

bhariharan: You have to move it in the filesystem in `Finder`. The logical grouping in `Xcode` has no association with the physical location on the filesystem.

PR #1983, opened 2017-04-06
Logging out user if invalid grant error received 

wmathurin:
(which means user token was revoked on the server)

Note:
That was taken care of by [CSFSalesforceOAuthRefresh.m](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/OAuth/CSFSalesforceOAuthRefresh.m#L79) when we were using the network SDK.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
```diff
@@ -95,13 +96,35 @@ - (void)completeWithSuccess:(SFOAuthCredentials *)credentials {
...
+    }
+    else if (self.errorBlock) {
         dispatch_async(dispatch_get_main_queue(), ^{
             self.errorBlock(error);
         });
     }
 }
 
+- (void)logoutUser {
+    SFUserAccountIdentity *userIdentity = [[SFUserAccountIdentity alloc] initWithUserId:self.coordinator.credentials.userId orgId:self.coordinator.credentials.organizationId];
```
khawkins: This is kind of going in the wrong direction from this class's original scope.  The idea is that this class is merely a conduit for the OAuth negotiations, and its consumer would ultimately be responsible for the results.

I think this functionality should be handled upstream of this class.

bhariharan: We have multiple consumers of this class though. Wouldn't it be nice for the session refresher to handle the results of that, instead of each consumer handling it, especially in the logout scenario?

khawkins: Specifically, I think a better place to handle this is in the failure block of `[SFRestAPI replayRequestIfRequired:response:error:request:delegate:]`

wmathurin: Moved to SFRestAPI [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1983/commits/43969a23bce1e52575de5f85421a270bd0f263a1)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -278,6 +286,13 @@ - (void)replayRequestIfRequired:(NSData *)data response:(NSURLResponse *)respons
     }
 }
 
+- (void)createAndStoreLogoutEvent:(NSError *)error {
+    NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
+    attributes[@"errorCode"] = [NSNumber numberWithInteger:error.code];
+    attributes[@"errorDescription"] = error.localizedDescription;
+    [SFSDKEventBuilderHelper createAndStoreEvent:@"userLogout" userAccount:nil className:NSStringFromClass([self class]) attributes:attributes];
```
bhariharan: Pass in `user` here for `userAccount`, since it's called before `logout`.

wmathurin: That makes sense.
Should I fix it [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/OAuth/CSFSalesforceOAuthRefresh.m#L86) too?

bhariharan: Yes please.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
khawkins: Sorry, I fell off the map yesterday before I could hit Send. 😐  This is what I meant to say!  I'll take a look at what has transpired since then.

----

> Wouldn't it be nice for the session refresher to handle the results of that

I think it could be nice for a single access point to handle that, yes.  But doing it in this manner short-circuits the error handling process of the upstream consumer, yanking the user context out from under it before it has a chance to clean up.

We used to do things more like this in iOS before, and kept running into issues where a standard error flow would dead-end in this case, leaving unfinished processes hanging around with a lack of state.

So if we can convenience-consolidate this functionality at a higher level (like `SFAuthenticationManager`, for example) and make it easier for consumers to call, I'm all for that.  But I think it's otherwise disjointed to the current design pattern to logout in the middle of an error flow, as opposed to the end of it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -240,7 +240,7 @@ - (void)replayRequestIfRequired:(NSData *)data response:(NSURLResponse *)respons
     // Checks if the access token has expired.
     NSInteger statusCode = [(NSHTTPURLResponse *)response statusCode];
     if (statusCode == 401 || statusCode == 403) {
-        SFUserAccount *user = [SFUserAccountManager sharedInstance].currentUser;
+        __block SFUserAccount *user = [SFUserAccountManager sharedInstance].currentUser;
```
khawkins: Why `__block` here?  Doesn't look like the user value changes in this process?

khawkins:
Otherwise, this looks fine to me.  It's got me thinking about consolidated refresh functionality in `SFAuthenticationManager`, and some of the unfinished work around partitioning authentication across different credentials that we've talked about previously (with @trooper2013 's refactoring).

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

Files changed | - | - 
--- | --- | ---
SFSDKLoginHostListViewController.m | `0%` | :skull:
CSFSalesforceOAuthRefresh.m | `25%` | :no_entry_sign:
SFRestAPI.m | `85%` | :white_check_mark:
SFOAuthSessionRefresher.m | `86%` | :white_check_mark:

---

---

</p>

PR #1989, opened 2017-04-11
More granular management of login state

khawkins:
The current implementation provides a pretty small "exclusive" window for not allowing multiple logouts.  Realistically, logout of a user shouldn't be allowed until that account has made it back into a logged in state.

Cc: @sgoldberg-sfdc

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
@@ -429,11 +429,11 @@ static Class InstanceClass = nil;
         return;
     }
 
-    if (user.isUserLoggingOut) {
-        [self log:SFLogLevelInfo msg:@"logoutUser: user is already in the process of logout."];
+    if (user.loginState != SFUserAccountLoginStateLoggedIn) {
+        [self log:SFLogLevelInfo format:@"%@ User login state is not logged in (%d). No action taken.", NSStringFromSelector(_cmd), user.loginState];
         return;
     }
-    user.userLoggingOut = YES;
+    user.loginState = SFUserAccountLoginStateLoggingOut;
     [self log:SFLogLevelInfo format:@"Logging out user '%@'.", user.userName];
     NSDictionary *userInfo = @{ @"account": user };
     [[NSNotificationCenter defaultCenter] postNotificationName:kSFUserWillLogoutNotification
@@ -477,7 +477,7 @@ static Class InstanceClass = nil;
             }
         }];
     }
-    user.userLoggingOut = NO;
+    user.loginState = SFUserAccountLoginStateNotLoggedIn;
 }
 
 - (void)cancelAuthentication
@@ -742,6 +742,7 @@ static Class InstanceClass = nil;
     // current user as the proper credentials.
     SFUserAccount *user = [[SFUserAccountManager sharedInstance] applyCredentials:self.coordinator.credentials
                                                                        withIdData:self.idCoordinator.idData];
+    user.loginState = SFUserAccountLoginStateLoggedIn;
  
     // Notify the session is ready
     [self willChangeValueForKey:@"haveValidSession"];
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
@@ -429,13 +429,12 @@ static Class InstanceClass = nil;
         return;
     }
 
-    @synchronized (user) {
-        if (user.loginState != SFUserAccountLoginStateLoggedIn) {
-            [self log:SFLogLevelInfo format:@"%@ User login state is not logged in (%d). No action taken.", NSStringFromSelector(_cmd), user.loginState];
-            return;
-        }
-        user.loginState = SFUserAccountLoginStateLoggingOut;
+    BOOL loggingOutTransitionSucceeded = [user transitionToLoginState:SFUserAccountLoginStateLoggingOut];
+    if (!loggingOutTransitionSucceeded) {
+        // SFUserAccount already logs the transition failure.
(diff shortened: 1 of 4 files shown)


sgoldberg-sfdc:
I like it. But should login be allowed if the user is not logged out? 

Can we make the state ivar thread safe?

khawkins:
I can make the ivar thread safe.  The user login state is trickier.  In 5.1, authentication is predicated on credentials, not users.  So user accounts are configured after the fact of authentication.

There's work that needs to be done in `SFAuthenticationManager` to consider this possibility in a multi-user scenario, probably along the lines of gating refresh by UserID/OrgID, and otherwise only allowing one "new" login at a time.  But in either case, that's probably something that will happen at the credentials level, and not the user account level.

khawkins:
It's probably going to make more sense to synchronize the block in which the multi-logout is determined.  I don't know that there's value specifically in writing exclusive block code for the ivar itself.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -429,11 +429,13 @@ - (void)logoutUser:(SFUserAccount *)user
         return;
     }
 
-    if (user.isUserLoggingOut) {
-        [self log:SFLogLevelInfo msg:@"logoutUser: user is already in the process of logout."];
-        return;
+    @synchronized (user) {
```
khawkins: @sgoldberg-sfdc I think this should cover what you're looking for?

sgoldberg-sfdc: I tend to be a bit more heavy handed and ensure all writes to the ivar are protected but you could be correct that this check is all that must be protected. Without knowing what the actual bizarre states the app could be in it is hard to say. For instance, could there be an authentication occurring at same time as a logout of the same user?

sgoldberg-sfdc:
> I can make the ivar thread safe. The user login state is trickier. In 5.1, authentication is predicated on credentials, not users. So user accounts are configured after the fact of authentication.

That makes me wonder if the user is the correct owner of the state. Does 5.1 track which users are currently authenticated, such that it can validate their current state based on credentials prior to starting an authentication.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
khawkins: There could be, in the sense that authentication in 5.1 doesn't really know anything about user accounts until credentials have been obtained.  So when you're in the middle of a "new" login, you have no idea what user account will ultimately be impacted by the results.

So I guess you could argue for other guards on the user account.  For instance, if an account is in a "logging out" state, presumably you wouldn't want to switch from there to a "logged in" state, as strange as that scenario may be.

sgoldberg-sfdc: I am working off little sleep atm, but this might be something I need to get down on paper to see if any edge cases pop out. I will look again tomorrow.

khawkins:
> That makes me wonder if the user is the correct owner of the state. Does 5.1 track which users are currently authenticated, such that it can validate their current state based on credentials prior to starting an authentication.

In the event of a "new" authentication, there's nothing to validate against.  And in the event of an authentication with existing credentials, it's entirely valid that an auth refresh could be happening for that user.

The SDK doesn't maintain an exclusivity of credentials per user.  If authentication completes with a set of credentials that match an existing user account, that account is configured with the updated credentials.  If not, a new user is created.  Beyond that, users are not interrogated prior to authentication.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -84,7 +84,8 @@ - (instancetype)initWithCredentials:(SFOAuthCredentials*) credentials {
     self = [super init];
     if (self) {
         _observingCredentials = NO;
-        self.credentials = credentials;
+        _credentials = credentials;
+        _loginState = (credentials.refreshToken.length > 0 ? SFUserAccountLoginStateLoggedIn : SFUserAccountLoginStateNotLoggedIn);
         _syncQueue = dispatch_queue_create(kSyncQueue, NULL);
     }
```
sgoldberg-sfdc: So if there is a refresh token we assume the user is logged in, although it may be expired. In that case the user will be logged out after attempting to use it. Makes sense.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
sgoldberg-sfdc: Kevin, Personally I would prefer if we used a single write-multi read pattern around the state, but I believe you are correct and this will address the crashing issue I am investigating. 

We can discuss extending exclusion to logging in, as well as documenting the expected behavior/use patterns  elsewhere.

khawkins: > I would prefer if we used a single write-multi read pattern around the state

We could do that, encapsulated in the user account object. But it will require additional "blessed" state transition checks. E.g. you can go from LoggedIn to LoggingOut, but not LoggingOut to LoggingOut, etc.

I can take a look at doing that.  Shouldn't be a big change.  Stay tuned.

khawkins:
@sgoldberg-sfdc I made the login state transitions more bullet proof, moved the atomic action into `SFUserAccount`.  Let me know what you think.

sgoldberg-sfdc:
:+1 It is single read/single right now, which is good for the purpose.

bhariharan:
@khawkins @sgoldberg-sfdc Good to merge?

khawkins:
What could possibly go wrong? 😉  If you all are good, I'm good.

PR #2006, opened 2017-04-25
Fix predicate

bhariharan:


sgoldberg-sfdc:
Also, in 5.0 the notification meant the users were loaded. That is no longer the case in 5.1. I am not sure it still has any value.

PR #1999, opened 2017-04-21
Cookie handling for WKWebView

trooper2013:
Switched to using WKWebsiteDataStore for handling cookies. WKWebview has gaping holes in functionality related to cookie handling, hence the SFAuthenticationManager  was reworked to handle cookie deletion successfully. Problems have been noticed by consumers of SDK in use with their custom SSO pages, where Logout completed but the cookie deletion was either incomplete or out of sync with the WKWebView.

PR #2007, opened 2017-04-26
Revert "Cookie handling for WKWebView"

bhariharan:
Reverts forcedotcom/SalesforceMobileSDK-iOS#1999


bhariharan:
@trooper2013 Let's revert this from `5.1` for now, since there's a known race condition and too many unknowns. We can bring it back into `dev` after `5.1` drops to give it time to bake and figure out how to fix the race condition and make it more deterministic.

PR #2025, opened 2017-05-03
CSFOutput Crashes if subclass has a non-encodable ivar

sgoldberg-sfdc:
-  We needed to add some concurrency control to a subclass
-  I implemented a Multiple Reader/Single Writer pattern in order to achieve this goal
-  Unfortunately, the current implementation of CSFOutput assumes every ivar is serializable via NSCopying

I believe this patch will handle the situation where the encoding specifier is '@' signifying an Objective-C object, yet we could not obtain a class for it. This is the situation for OS_dispatch_queue.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Model/CSFOutput.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Model/CSFOutput.m
@@ -92,6 +92,7 @@ static NSString * const kCSFInputCustomArrayAttributes = @"__CSFOutput_Array_Sto
             
             Ivar ivar = class_getInstanceVariable(ivarInfo[@"class"], [ivarName UTF8String]);
             Class ivarClass = CSFClassFromEncoding(ivarInfo[@"encoding"]);
+            NSString * encoding = ivarInfo[@"encoding"];
             if (ivarClass) {
                 id result = [decoder decodeObjectOfClass:ivarClass forKey:propertyName];
                 if ([result isKindOfClass:[CSFOutput class]]) {
@@ -99,7 +100,7 @@ static NSString * const kCSFInputCustomArrayAttributes = @"__CSFOutput_Array_Sto
                     resultOutput.parentObject = self;
                 }
                 object_setIvar(self, ivar, result);
-            } else if (ivarInfo[@"encoding"]) {
+            } else if (encoding && ![encoding hasPrefix:@"@"]) {
                 const void * ivarPtr = (__bridge void*)(self) + ivar_getOffset(ivar);
                 [decoder decodeValueOfObjCType:[ivarInfo[@"encoding"] UTF8String] at:(void *)ivarPtr];
             }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Model/CSFOutput.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Network/Model/CSFOutput.m
@@ -70,11 +70,10 @@ static NSString * const kCSFInputCustomArrayAttributes = @"__CSFOutput_Array_Sto
 
         Ivar ivar = class_getInstanceVariable(ivarInfo[@"class"], [ivarName UTF8String]);
         Class ivarClass = CSFClassFromEncoding(ivarInfo[@"encoding"]);
-        NSString * encoding = ivarInfo[@"encoding"];
         if (ivarClass) {
             id value = object_getIvar(self, ivar);
             [encoder encodeObject:value forKey:propertyName];
-        } else if (encoding) {
+        } else if (ivarInfo[@"encoding"]) {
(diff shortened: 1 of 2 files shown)


sgoldberg-sfdc:
FYI: @bhariharan @trooper2013 @khawkins 

@MikeNachbaur ,
I am not sure if I correctly interpreted the expectations around CSFClassFromEncoding and CSFClassIvars, or why this class did not limit itself to encoding properties.

MikeNachbaur:
@sgoldberg-sfdc yeah this was a bug that I've already fixed, but it's currently stuck in a branch that hasn't been approved to be merged in yet.  That batch request / tech-debt branch I presented to the DRB weeks ago.

sgoldberg-sfdc:
@MikeNachbaur what was the solution in your fork. I would rather cherry pick from it if it is more robust of a solution.

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

Files changed | - | - 
--- | --- | ---
CSFOutput.m | `61%` | :warning:

---

---

</p>

MikeNachbaur:
@sgoldberg-sfdc I actually fixed the problem in the root function CSFClassFromEncoding:
https://github.com/MikeNachbaur/SalesforceMobileSDK-iOS/blob/NetworkImprovements/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Utilities/CSFInternalDefines.m#L104

Basically the problem is if a property is defined to have a type of `id`, then it has an encoding type of `@`, not `@"ClassName"`.  So instead it falls back to defining the result class type as `NSObject`.  Non-`@` encodings result in a `nil` class, causing it to go through the primitive ivar handling path.

MikeNachbaur:
I'd still like to revisit merging my network improvements / trust work into the dev branch sometime soon, particularly since the bulk of the account management refactoring changes have happened.

sgoldberg-sfdc:
Thanks, Ill look at your change.

sgoldberg-sfdc:
Cherry picked @MikeNachbaur change.

sgoldberg-sfdc:
@MikeNachbaur the change does not address this case. The "encoding" in the case of an OS_dispatch_queue is "@\"NSObject<OS_dispatch_queue>\"" which has a length greater than 2. 

Did the logic in my initial PR make sense to you, or do you think that would catch some false positives?

MikeNachbaur:
Hmm, so my question is, why are you receiving a dispatch queue in your CSFOutput class?  The automatic property synthesis handling is meant for consuming JSON data from a network response.  So it's unlikely you're getting an `OS_dispatch_queue` from a JSON response.

If you're storing / returning a dispatch queue, for convenience sake, from a CSFOutput subclass, I'd suggest either telling CSFOutput to skip that property (return `nil` from `storageKeyPathForPropertyName:` for that property), or implement that as a simple getter method instead of as an `@property`.

For example, a common practice would be to have methods that return an instance, or subclass, of `CSFAction` fully configured to traverse to associated URLs returned in an output payload (e.g. nextPageUrl, etc).

sgoldberg-sfdc:
They are not stored as properties at all, but as ivars. The dispatch_queue is not part of the JSON output, but a mechanism used to implement multi-reader/single-writer to make the output , an array in this case, thread safe.

I will look at storageKeyPathForPropertyName but since these are ivar and you are pulling every ivar, not just exposed properties in your serialization, I am not sure it will make a difference. 

I have to question the use of the superclass to serialize subclass properties and ivars, rather than push this responsibility down to the subclass, and let the superclass serialize it's own ivars/properties. I do not believe anyone needs this dynamism.

MikeNachbaur:
I see what you're doing there.  Well, if your class in this case isn't actually storing property values on its own, but is rather containing other objects that it wants to handle thread safe access, you don't actually need to subclass from CSFOutput at all.  You only need to conform to CSFActionModel.  

CSFOutput is meant to be a way to create model representations of complex JSON trees simply without all the boilerplate of read / check-for-null / parse / convert-to-NSDate, etc.  But using CSFOutput isn't a requirement to interact with the network.

That being said, it couldn't hurt to put in your original fix in that case, since this shouldn't crash in any circumstance anyway.  What I'd suggest though is in addition to your check to ensure the encoding doesn't begin with an `@`, it might make sense to only encode to disk ivar values that have an associated `@property`.  Otherwise it might encode more state than it should.  If there's extra ivars that should be encoded, then it should be the responsibility of that subclass to encode those ivars as it sees fit.

sgoldberg-sfdc:
@MikeNachbaur I agree it makes more sense to encode ivar backed properties, not every ivar. 

I was uncertain if I should make such a large change without knowing your intention for the CSFOutput as it looks like the serialization of all ivars was intentional. I would much prefer a solution like that than a bandaid that might not catch everything, or might catch things it shouldn't.

Let me revisit the encode routine.

PR #2029, opened 2017-05-04
Rework of WKWebView Cookie handling

trooper2013:
@khawkins @bhariharan Please review cookie cleanup code. @sgoldberg-sfdc you were seeing problems in logout previously leading to a race-condition, resulting from the fact that cancelAuthentication during logout wasn't being invoked from the completion block. If you could give this a try again, it would be great.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
@@ -454,33 +454,41 @@ static Class InstanceClass = nil;
         [[SFPushNotificationManager sharedInstance] unregisterSalesforceNotifications:user];
         [userAccountManager deleteAccountForUser:user error:nil];
         [self revokeRefreshToken:user];
+        // NB: There's no real action that can be taken if this login state transition fails.  At any rate,
+        // it's an unlikely scenario.
+        [user transitionToLoginState:SFUserAccountLoginStateNotLoggedIn];
     }else {
         // Otherwise, the current user is being logged out.  Supply the user account to the
         // "Will Logout" notification before the credentials are revoked.  This will ensure
         // that databases and other resources keyed off of the userID can be destroyed/cleaned up.
-        if ([SFPushNotificationManager sharedInstance].deviceSalesforceId) {
-            [[SFPushNotificationManager sharedInstance] unregisterSalesforceNotifications];
-        }
-        [self cancelAuthentication];
-        [self clearAccountState:YES];
-
-        [self willChangeValueForKey:@"haveValidSession"];
-        [userAccountManager deleteAccountForUser:user error:nil];
-        [self revokeRefreshToken:user];
-        userAccountManager.currentUser = nil;
-        [self didChangeValueForKey:@"haveValidSession"];
-
-        NSNotification *logoutNotification = [NSNotification notificationWithName:kSFUserLogoutNotification object:self];
-        [[NSNotificationCenter defaultCenter] postNotification:logoutNotification];
-        [self enumerateDelegates:^(id<SFAuthenticationManagerDelegate> delegate) {
-            if ([delegate respondsToSelector:@selector(authManagerDidLogout:)]) {
-                [delegate authManagerDidLogout:weakSelf];
+        __weak typeof (self) weakSelf = self;
+        [self clearAccountState:YES withCompletion:^{
+            __weak typeof (weakSelf) strongSelf = weakSelf;
+            
+            if ([SFPushNotificationManager sharedInstance].deviceSalesforceId) {
+                [[SFPushNotificationManager sharedInstance] unregisterSalesforceNotifications];
             }
+
+            [strongSelf cancelAuthentication];
+            [strongSelf willChangeValueForKey:@"haveValidSession"];
+            [userAccountManager deleteAccountForUser:user error:nil];
+            [strongSelf revokeRefreshToken:user];
+            userAccountManager.currentUser = nil;
+            [strongSelf didChangeValueForKey:@"haveValidSession"];
+
+            NSNotification *logoutNotification = [NSNotification notificationWithName:kSFUserLogoutNotification object:strongSelf];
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -454,33 +454,41 @@ - (void)logoutUser:(SFUserAccount *)user
...
-        [self revokeRefreshToken:user];
-        userAccountManager.currentUser = nil;
-        [self didChangeValueForKey:@"haveValidSession"];
-
-        NSNotification *logoutNotification = [NSNotification notificationWithName:kSFUserLogoutNotification object:self];
-        [[NSNotificationCenter defaultCenter] postNotification:logoutNotification];
-        [self enumerateDelegates:^(id<SFAuthenticationManagerDelegate> delegate) {
-            if ([delegate respondsToSelector:@selector(authManagerDidLogout:)]) {
-                [delegate authManagerDidLogout:weakSelf];
+        __weak typeof (self) weakSelf = self;
```
khawkins: This should be `__strong`, not `__weak`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -454,33 +454,41 @@ - (void)logoutUser:(SFUserAccount *)user
...
+        [self clearAccountState:YES withCompletion:^{
+            __weak typeof (weakSelf) strongSelf = weakSelf;
+            
+            if ([SFPushNotificationManager sharedInstance].deviceSalesforceId) {
+                [[SFPushNotificationManager sharedInstance] unregisterSalesforceNotifications];
             }
+
+            [strongSelf cancelAuthentication];
```
khawkins: Why are we canceling authentication?  What if authentication for another user is in progress?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -454,33 +454,41 @@ - (void)logoutUser:(SFUserAccount *)user
...
-        userAccountManager.currentUser = nil;
-        [self didChangeValueForKey:@"haveValidSession"];
-
-        NSNotification *logoutNotification = [NSNotification notificationWithName:kSFUserLogoutNotification object:self];
-        [[NSNotificationCenter defaultCenter] postNotification:logoutNotification];
-        [self enumerateDelegates:^(id<SFAuthenticationManagerDelegate> delegate) {
-            if ([delegate respondsToSelector:@selector(authManagerDidLogout:)]) {
-                [delegate authManagerDidLogout:weakSelf];
+        __weak typeof (self) weakSelf = self;
+        [self clearAccountState:YES withCompletion:^{
```
khawkins: This inherently changes `logoutUser:` from a synchronous method to an asynchronous one.  That seems like:

- A pretty big change to make in a point release.
- Something that we probably haven't considered all of the ramifications for.

Users historically expect logout to be finished logging out the user at the end of the method call.  I'd personally prefer that we fire and forget the ultimate call to `[WKWebsiteDataStore removeDataOfTypes: modifiedSince:completionHandler:]`, and leave the SDK functionality synchronous.  IMO there are far fewer potential consequences for that approach than making all of our surrounding code asynchronous to facilitate it.

trooper2013: @khawkins Logout should be asynchronous ideally. We've had race-conditions reported by one customer using SSO pages in particular. 1) How about firing a specific notification when  cookie cleanup is complete instead?  2) Document the need to invoke removeAllCookies::withCompletion  and have customers invoke it in such circumstances.

khawkins: Can you justify why logout should be asynchronous ideally?  I think it should ideally be synchronous.  I.e. when the method's done, the user is logged out.  We've already had a number of issues with multiple threads coming _into_ the logout method(s) and creating shared state issues that needed to be addressed. Adding the complexity of non-deterministic logout completion opens up the potential for a host of other issues, including race conditions for users going back through the login process before logout has completed.

Sorry, I simply do not agree that logout is ideally asynchronous.  And I don't see a compelling reason for making it so.

khawkins: I have another thought on this.  Based on the work that @ptiarks did to create a unified `WKProcessPool` for the SDK web views, what if we just ditch/re-create the process pool at logout/user context switch, and effectively start over.  That's what we want anyway, right?  There shouldn't be any residual web context allowed to persist once a user has been logged out, or another user has taken over the web view context.  Thoughts?

Cc: @jhneves

bhariharan: I agree with Kevin. I'd prefer that logout be synchronous and cookie clearing be fire-and-forget. We can add documentation for SSO users to explicitly clear cookies first using our APIs before attempting to login again. It's not a good idea to increase complexity by making logout asynchronous for the 1% SSO use case.

bhariharan: That's an excellent approach @khawkins. I'm in support of that.

khawkins: And to be fair, @jhneves actually spurred that thought, not me.  Credit is all his. 🙂

trooper2013: @khawkins @bhariharan I see what you are saying. The WKProcessPool should help and the fix is much simpler too.  Thanks @jhneves. I'll close this PR.

PR #2030, opened 2017-05-04
Fixing bug with unknown response type

bhariharan:
Sometimes we get a response that just contains `<>`.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
@@ -267,6 +267,9 @@ __strong static NSDateFormatter *httpDateFormatter = nil;
             NSError *parsingError;
             NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:data options:0 error:&parsingError];
             if (parsingError) {
+                if (data == nil || [data isEqualToData:[[NSData alloc] init]]) {
+                    data = nil;
+                }
                 [delegate request:request didLoadResponse:data];
             } else {
                 [delegate request:request didLoadResponse:jsonDict];
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
@@ -267,7 +267,7 @@ __strong static NSDateFormatter *httpDateFormatter = nil;
             NSError *parsingError;
             NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:data options:0 error:&parsingError];
             if (parsingError) {
-                if (data == nil || [data isEqualToData:[[NSData alloc] init]]) {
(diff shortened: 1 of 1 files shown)


huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

Files changed | - | - 
--- | --- | ---
SFRestAPI+Blocks.m | `86%` | :white_check_mark:

---

---

</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Blocks.m
```diff
@@ -231,23 +231,27 @@ - (SFRestRequest *) performRequestWithMethod:(SFRestMethod)method path:(NSString
 #pragma mark - response delegate
 
 - (void) sendActionForRequest:(SFRestRequest *)request success:(BOOL)success withObject:(id)object {
-    if( success ) {
-        // This block def basically generalizes the SFRestDictionaryResponseBlock and SFRestArrayResponseBlock
-        // block typedefs, so that we can handle either.
+
+    // Ensures that the response is a well-known type.
+    if (![object isKindOfClass:[NSDictionary class]] && ![object isKindOfClass:[NSArray class]] && ![object isKindOfClass:[NSString class]]) {
```
khawkins: But &lt;&gt; is an `NSString` too?  This feels a bit like it's masking whatever the root cause of the issue is.  Why can the consumer not handle output that's otherwise considered valid (if not exactly usual) from the service?

bhariharan: I got to the bottom of the issue. The response here is an empty `NSData` object, the debugger shows it as `<>`, but it's really an empty instance of `NSData`. It fell through the cracks because we have a `nil` check but no check for an empty object. I should have a better fix soon.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -267,6 +267,9 @@ - (void)replayRequestIfRequired:(NSData *)data response:(NSURLResponse *)respons
             NSError *parsingError;
             NSDictionary *jsonDict = [NSJSONSerialization JSONObjectWithData:data options:0 error:&parsingError];
             if (parsingError) {
+                if (data == nil || [data isEqualToData:[[NSData alloc] init]]) {
```
khawkins: Can we check the `data.length` property?  Shouldn't need to compare to an empty object.

khawkins: Given that `nil` data would zero out integer properties as well, you could simplify this whole conditional to `if (data.length == 0)`.

wmathurin:
Should we add a test?

bhariharan:
@khawkins Done.

PR #2034, opened 2017-05-09
Refactored WKProcessPool usage and cookie removal

trooper2013:
@khawkins @bhariharan Moved shared WKProcessPool to a static Factory class, added convenience class methods  in SDKManager & refactored cookie removal to a fire & forget mechanism.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWKProcessPoolFactory.h
```diff
@@ -0,0 +1,35 @@
...
+#import <Foundation/Foundation.h>
+
+@class WKProcessPool;
+@interface SFSDKWKProcessPoolFactory : NSObject
+
+/**
+ Gets or sets an instance of WKProcessPool that will be used during instantiation of any WKWebView instances @discussion
+  Use the default instance or provide an instance to share state between WKWebView instances
+ */
+@property (class, strong, nonatomic) WKProcessPool *sharedProcessPool;
```
khawkins: I'll be damned.  I totally missed this addition to Xcode 8!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWKProcessPoolFactory.m
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
+
++ (WKProcessPool *)sharedProcessPool {
+    return _processPool;
+}
+
++ (void)setSharedProcessPool:(WKProcessPool *)sharedProcessPool {
+    if (_processPool != sharedProcessPool) {
+        _processPool = sharedProcessPool ?: [[WKProcessPool alloc] init];
```
khawkins: We had some internal discussion about this from before, and I'd like to suggest we implement a slightly different pattern.  Namely, this is a pattern that would more conventionally be flipped to the getter, where, if `_processPool` is `nil`, it would return a new `WKProcessPool` instance.  That would also alleviate the need for the `initialize` method.

It just cleans up a lot of code, and follows a more conventional pattern.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -115,6 +115,19 @@ typedef void (^SFSnapshotViewControllerDismissalBlock)(UIViewController* snapsho
 + (nonnull NSString *)ailtnAppName;
 
 /**
+   Gets an instance of WKProcessPool that will be used during instantiation of any WKWebView instances
+  @discussion
+  Use the default instance to share state between WKWebView instances
+  */
++ (nonnull WKProcessPool *) processPool;
```
khawkins: Given that we've flipped out of dependency on `SalesforceSDKManager` for the process pool, I would very much like to do away with the dependency in the SDK manager, and simply access the process pool directly from the consumers.

`SalesforceSDKManager` tends to carry some baggage with it when it's included around in other classes, arguably creating circular dependencies in others.  Since we have a standalone facility in this implementation, let's use it directly.

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

### Tests: 

Classname | Name|
--- | ---|
SmartSyncTestSuite | testSyncSObjectDetectConflictDelete|
SmartSyncTestSuite | testSyncSObjectDetectConflictUpdate|


## Current coverage for SalesforceHybridSDK.framework is `54%`
Files changed | - | - 
--- | --- | ---
SFHybridViewController.m | `27%` | :no_entry_sign:

---
## Current coverage for Cordova.framework is `33%`

---

---
## Current coverage for SalesforceSDKCore.framework is `22%`
Files changed | - | - 
--- | --- | ---
SFOAuthCoordinator.m | `18%` | :skull:
SalesforceSDKManager.m | `26%` | :no_entry_sign:
SFAuthenticationManager.m | `35%` | :no_entry_sign:

---
## Current coverage for SmartStore.framework is `54%`

---
## Current coverage for SmartSync.framework is `36%`

---

</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -624,39 +624,49 @@ - (void)setOauthClientId:(NSString *)newClientId
...
+    __weak typeof(self) weakSelf = self;
+    [self removeCookiesFromDomains:@[@".salesforce.com", @".force.com", @".cloudforce.com"] withCompletion:^{
+        [weakSelf addSidCookieForInstance];
+    }];
 }
 
-+ (void)removeCookies:(NSArray *)cookieNames fromDomains:(NSArray *)domainNames
-{
-    NSAssert(cookieNames != nil && [cookieNames count] > 0, @"No cookie names given to delete.");
++ (void)removeCookiesFromDomains:(NSArray *)domainNames withCompletion:(nullable void(^)())completionBlock {
```
khawkins: Rather than introduce new interfaces here, I would prefer to do the following:

(
- Keep the `removeCookies:fromDomains:` interface.

_and_

- Fire and forget the calls to `[WKWebsiteDataStore removeDataOfTypes:forDataRecords:completionHandler:]`

_and_

- _Deprecate_ the method.

)

_OR_

- Do away with the method altogether.

The only thing that uses this method is `resetSessionCookie`, which should be amended to simply clear the shared process pool, because its current functionality does not work anyway: Adding the sid cookie back to `NSHTTPCookieStorage` is effectively a no-op for `WKWebView`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -624,39 +624,49 @@ - (void)setOauthClientId:(NSString *)newClientId
...
+                                                                    completionHandler:^{
+                                                                        if (completionBlock)
+                                                                            completionBlock();
+                                                                    }];
+                     }];
+    
 }
 
-+ (void)removeAllCookies
++ (void)removeAllCookiesWithCompletion:(void(^)())completionBlock
```
khawkins: Same for this method.  Either keep the original synchronous interface and deprecate it, or do away with it altogether.  Nothing uses this method, and we need to get out of the business of exposing these, one way or another.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -927,7 +937,8 @@ - (void)clearAccountState:(BOOL)clearAccountData {
         [self.coordinator.view removeFromSuperview];
     }
     
-    [SFAuthenticationManager removeAllCookies];
+    SFSDKWKProcessPoolFactory.sharedProcessPool = nil;
```
khawkins: See?  What did I tell you about that `nil` check in the setter? 😉

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
bhariharan: Agreed. Let's do away with the method and clean up good.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWKProcessPoolFactory.m
```diff
@@ -0,0 +1,46 @@
...
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#import <WebKit/WebKit.h>
+#import "SFSDKWKProcessPoolFactory.h"
+
+@implementation SFSDKWKProcessPoolFactory
+
+static WKProcessPool *_processPool = nil;
+
++ (void)initialize {
```
khawkins: With the lazy loading of `_processPool` in the getter, this method can go away.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -624,39 +624,36 @@ - (void)setOauthClientId:(NSString *)newClientId
 
 + (void)resetSessionCookie
 {
-    [self removeCookies:@[@"sid"]
-            fromDomains:@[@".salesforce.com", @".force.com", @".cloudforce.com"]];
-    [self addSidCookieForInstance];
+    __weak typeof(self) weakSelf = self;
```
khawkins: Ugh.  I just realized, going through these code changes, that we still support `UIWebView` in addition to `WKWebView` in the hybrid client. 🙁  Which makes all of this problematic.

wmathurin: We only use WKWebView for the authentication web view though.

khawkins: Yeah, but all of this cookie cleanup applies to both auth and hybrid views.

It seems like, for the methods we keep on this front, they need to clean up both the `NSHTTPCookieStorage` data  _and_ the `WKProcessPool` instance—I'd posit that we should get out of the business of cleaning up the `WKWebsiteDataStore` altogether for the WK side, in favor of simply clearing the `WKProcessPool` that we share.

wmathurin: True, true. I was narrowly thinking of that sso bug which got all that work started (and which involved left over cookies in the oauth webview).

trooper2013: @khawkins @wmathurin We can do both for now i.e use the NSHttpCookieStorage and reset the pool. Ill make these changes.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -36,6 +36,7 @@
 #import "SFSwiftDetectUtil.h"
 #import "SFUserAccountManager.h"
 #import "SFSDKAppFeatureMarkers.h"
+#import "SFSDKWKProcessPoolFactory.h"
```
khawkins: You can remove this line now.

khawkins:
Okay, after talking with @trooper2013 , it's clear that there may be scenarios where you don't want to simply blow away the `WKProcessPool`, because it can impact app cache and other data of the web view.  We want to make sure that we're only doing that as part of a user context switch (logout, and/or switch to different user).

@trooper2013 I'll reach out to you offline to talk about our options.  I think it will be beneficial to step through them interactively.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWebViewStateManager.h
```diff
@@ -24,7 +24,18 @@
...
 
 @class WKProcessPool;
-@interface SFSDKWKProcessPoolFactory : NSObject
+
+@interface SFSDKWebViewStateManager : NSObject
+
+/**
+  Reset some associated state with WKWebView and UIWebView e.g. Remove Cookies then add SID cookie
+ */
++ (void)resetSessionWithAccessToken:(NSString *) accessToken andProtocol:(BOOL) isSecure;
```
khawkins: Couple of minor asks:

- Can this be `resetSessionWithNewAccessToken`?  The current wording makes it sound like you're resetting the session associated with the access token.
- `andProtocol` would be better described as e.g. `isSecureProtocol`.  The former makes it sound like you're passing in the protocol.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWebViewStateManager.m
```diff
@@ -0,0 +1,132 @@
...
+                         if (deletedRecords.count > 0)
+                             [[WKWebsiteDataStore defaultDataStore] removeDataOfTypes:websiteDataTypes
+                                                                       forDataRecords:deletedRecords
+                                                                    completionHandler:^{
+                                                                        if (completionBlock)
+                                                                            completionBlock();
+                                                                    }];
+                     }];
+}
++ (void)addSidCookieForDomain:(NSString*)domain withAccessToken:accessToken andProtocol:(BOOL)isSecure
```
khawkins: Same here with the `andProtocol`.  It's misleading.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWebViewStateManager.m
```diff
@@ -0,0 +1,132 @@
...
+
++ (WKProcessPool *)sharedProcessPool {
+    if (!_processPool) {
+        _processPool = [[WKProcessPool alloc] init];
+    }
+    return _processPool;
+}
+
++ (void)setSharedProcessPool:(WKProcessPool *)sharedProcessPool {
+    _processPool = sharedProcessPool;
```
khawkins: I'd keep the standard setter pattern here.  I.e.:

```obj-c
if (sharedProcessPool != _processPool) {
    _processPool = sharedProcessPool;
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWebViewStateManager.h
trooper2013: @khawkins I meant to change that after first iteration on my end. Guess it slipped. I will make that change.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWebViewStateManager.m
trooper2013: @khawkins oops!

PR #2036, opened 2017-05-09
Replacing Direct Usage of NSURLSession with SFNetwork

bhariharan:
This pull request replaces areas where we use `NSURLSession` directly with `SFNetwork`, such as auth flow and push notification registration/un-registration.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -40,6 +40,7 @@
 #import "SFSDKEventBuilderHelper.h"
 #import "SFSDKAppFeatureMarkers.h"
 #import "SalesforceSDKManager.h"
+#import "SFNetwork.h"
 
 // Public constants
 
@@ -935,20 +936,20 @@ static NSString * const kSFAppFeatureSafariBrowserForLogin   = @"BW";
 
 - (NSURLSession*)session {
     if (_session == nil) {
-        _session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration ephemeralSessionConfiguration]];
+        SFNetwork *network = [[SFNetwork alloc] init];
+        _session = network.activeSession;
     }
     return _session;
 }
 
 #pragma mark - WKNavigationDelegate (User-Agent Token Flow)
 - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler {
-
     NSURL *url = navigationAction.request.URL;
     NSString *requestUrl = [url absoluteString];
     if ([self isRedirectURL:requestUrl]) {
         [self handleUserAgentResponse:url];
         decisionHandler(WKNavigationActionPolicyCancel);
-    }else {
+    } else {
         decisionHandler(WKNavigationActionPolicyAllow);
     }
 }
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -159,9 +159,11 @@ - (NSURLRequest *)prepareRequestForSend {
         // Sets HTTP method on the request.
         [self.request setHTTPMethod:[SFRestRequest httpMethodFromSFRestMethod:self.method]];
 
-        // Sets OAuth Bearer token header on the request.
-        NSString *bearer = [NSString stringWithFormat:@"Bearer %@", user.credentials.accessToken];
-        [self.request setValue:bearer forHTTPHeaderField:@"Authorization"];
+        // Sets OAuth Bearer token header on the request (if not already present).
+        if (self.request.allHTTPHeaderFields && self.request.allHTTPHeaderFields.allKeys && ![self.request.allHTTPHeaderFields.allKeys containsObject:@"Authorization"]) {
```
bhariharan: Minor enhancement: The `Bearer` token header should be set only if it's not already set by the client. Current user's token should be the fallback if not set - it shouldn't overwrite a token that's being sent in by the client.

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

Files changed | - | - 
--- | --- | ---
SFPushNotificationManager.m | `28%` | :no_entry_sign:
SFOAuthCoordinator.m | `32%` | :no_entry_sign:
SFAuthenticationManager.m | `41%` | :no_entry_sign:
SFIdentityCoordinator.m | `68%` | :warning:
SFRestRequest.m | `75%` | :warning:

---

---

</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -935,20 +936,20 @@ + (NSString *)advancedAuthStateDesc:(SFOAuthAdvancedAuthState)authState
 
 - (NSURLSession*)session {
     if (_session == nil) {
-        _session = [NSURLSession sessionWithConfiguration:[NSURLSessionConfiguration ephemeralSessionConfiguration]];
+        SFNetwork *network = [[SFNetwork alloc] init];
+        _session = network.activeSession;
```
khawkins: Is there value to doing this here, if it's not actually going to make use of `SFNetwork` functionality?

bhariharan: @khawkins I didn't really look at what was being done with it. I followed the path of least resistance with search and replace ;-)

bhariharan: The value here is to consolidate all requests that go through the `Mobile SDK` through a single network stack. This ensures that all requests have the same `NSURLConfiguration`, which is important in some corner cases. I recently discovered that there's a server side configuration that allows the server to reject requests that originate from a different IP subnet. This would fail if `TwinPrime` configuration was configured in `SFNetwork` but these requests used their own `NSURLConfiguration`, because the IP addresses would be different and the server would reject the valid `Bearer` token.

khawkins: To rephrase then: "Why?" 🙂  What's the value of using one instance of an `NSURLSession` over another?

khawkins: Ah, our responses crossed in the night.  That makes sense.  We may want to follow up in a future PR to integrate more with `SFNetwork` in these spots, which in `SFIdentityCoordinator` for instance could benefit from the built-in auth refresh, etc.  But this is a good start.  Thanks for the explanation.

PR #2046, opened 2017-05-16
switchToUser: with the same user should be a no-op

khawkins:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -596,20 +596,24 @@ - (void)switchToNewUser {
...
-
-    SFUserAccount *prevUser = self.currentUser;
-    [self setCurrentUser:newCurrentUser];
-
-    [self enumerateDelegates:^(id<SFUserAccountManagerDelegate> delegate) {
-        if ([delegate respondsToSelector:@selector(userAccountManager:didSwitchFromUser:toUser:)]) {
-            [delegate userAccountManager:self didSwitchFromUser:prevUser toUser:newCurrentUser];
-        }
-    }];
+    if (self.currentUser == newCurrentUser) {
```
bhariharan: Shouldn't we use `isEqual` here? `isEqual` should compare if the `orgId`, `userId` and optionally `communityId` are the same to determine equality.

bhariharan: Shouldn't we use `isEqual` here? `isEqual` should compare if the `orgId`, `userId` and optionally `communityId` are the same to determine equality.

huminzhi:
## Danger has errored
[!] Invalid `DangerFile` file: dlopen(/usr/local/lib/ruby/gems/2.4.0/gems/ox-2.4.9/ext/ox/ox.bundle, 9): Library not loaded: /usr/local/opt/ruby/lib/libruby.2.4.0.dylib
  Referenced from: /usr/local/lib/ruby/gems/2.4.0/gems/ox-2.4.9/ext/ox/ox.bundle
  Reason: image not found - /usr/local/lib/ruby/gems/2.4.0/gems/ox-2.4.9/ext/ox/ox.bundle
```
 #  from DangerFile:1
```

</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
khawkins: It was a conscious choice, but I'm wondering too if I should have gone the other way.  The premise is that, with 5.1 and beyond, we should no longer have the possibility of multiple `SFUserAccount` instances with the same identity.  But in the event that we do, for whatever reason, it may make more sense to "fail fast" and disallow that in the switch as well.

I'll make the change.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -596,20 +596,24 @@ - (void)switchToNewUser {
...
-    SFUserAccount *prevUser = self.currentUser;
-    [self setCurrentUser:newCurrentUser];
-
-    [self enumerateDelegates:^(id<SFUserAccountManagerDelegate> delegate) {
-        if ([delegate respondsToSelector:@selector(userAccountManager:didSwitchFromUser:toUser:)]) {
-            [delegate userAccountManager:self didSwitchFromUser:prevUser toUser:newCurrentUser];
-        }
-    }];
+    if ((self.currentUser.accountIdentity == nil && newCurrentUser.accountIdentity == nil) || [self.currentUser.accountIdentity isEqual:newCurrentUser.accountIdentity])  {
+        [self log:SFLogLevelWarning format:@"%@ new user identity is the same as the current user (%@/%@).  No action taken.", NSStringFromSelector(_cmd), newCurrentUser.credentials.organizationId, newCurrentUser.credentials.userId];
```
bhariharan: @khawkins Is this safe to do? Could `credentials` be `nil`?

khawkins: Yeah.  Any `nil` will propagate down the chain, so in that case you'd get a string of `(null)/(null)` (I believe), which IMO is also informative.

khawkins:
test me

PR #2054, opened 2017-05-19
Add missing static target membership

bhariharan:
FYI @wmathurin


bhariharan:
The file was included in the dynamic target but not in the static target.

PR #2058, opened 2017-05-23
Fix CSFOutput to match encode to initwithcoder

sgoldberg-sfdc:
- My previous change to CSFOutput failed to match the encode routine with the initWithCoder resulting in an incorrectly encoded object.
  - Reverted previous change to CSFClassFromEncoding as it had no impact on the problem at hand.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Utilities/CSFInternalDefines.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Utilities/CSFInternalDefines.m
@@ -76,8 +76,12 @@ CSFParameterStyle CSFRequiredParameterStyleForHTTPMethod(NSString *method) {
 Class CSFClassFromEncoding(NSString *encoding) {
     Class result = nil;
     if ([encoding hasPrefix:@"@"]) {
-        NSString *className = [[encoding substringFromIndex:2] stringByReplacingOccurrencesOfString:@"\"" withString:@""];
-        result = NSClassFromString(className);
+        if (encoding.length > 2) {
+            NSString *className = [[encoding substringFromIndex:2] stringByReplacingOccurrencesOfString:@"\"" withString:@""];
+            result = NSClassFromString(className);
+        } else {
+            result = [NSObject class];
+        }
     }
     return result;
 }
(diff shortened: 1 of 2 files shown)


sgoldberg-sfdc:
@bhariharan @khawkins @trooper2013

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

Files changed | - | - 
--- | --- | ---
CSFInternalDefines.m | `56%` | :warning:
CSFOutput.m | `61%` | :warning:

---

---

</p>

sgoldberg-sfdc:
Don't see any test failures, what is wrong?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Network/Utilities/CSFInternalDefines.m
```diff
@@ -76,12 +76,8 @@ CSFParameterStyle CSFRequiredParameterStyleForHTTPMethod(NSString *method) {
 Class CSFClassFromEncoding(NSString *encoding) {
     Class result = nil;
     if ([encoding hasPrefix:@"@"]) {
-        if (encoding.length > 2) {
-            NSString *className = [[encoding substringFromIndex:2] stringByReplacingOccurrencesOfString:@"\"" withString:@""];
-            result = NSClassFromString(className);
-        } else {
-            result = [NSObject class];
-        }
+        NSString *className = [[encoding substringFromIndex:2] stringByReplacingOccurrencesOfString:@"\"" withString:@""];
```
khawkins: Seems dangerous to lose the length check here.

sgoldberg-sfdc: If we were losing something that had been there for a while I would agree. I added the length check at Michaels recommendation with my previous PR as a potential solution to the issue I reported. Since it did not do anything for that problem, I would prefer to remove it and Michael can add it himself if he feels it is required for some other problem.

khawkins: I'm specifically concerned about the call to `[encoding substringFromIndex:2]` without bounds checking that `encoding` is actually 3 or more characters in length.  I'm find with the business logic change.

PR #2062, opened 2017-05-25
Changing login host should not trigger user change

khawkins:
Changing the current user should be a deliberate action on the part of the SDK consumer.  When switching login hosts from the login screen, all we really want to do is restart the authentication that was already in progress, with the new login host configuration.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
@@ -26,7 +26,24 @@
  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#import "SalesforceSDKCore.h"
+@class SFLoginViewController;
+@class SFSDKLoginHost;
+
+/**
+ * Delegate protocol for the owner of SFLoginViewController.
+ */
+@protocol SFLoginViewControllerDelegate <NSObject>
+
+@optional
+
+/**
+ * Notifies the delegate that the selected login host has been changed.
+ * @param loginViewController The instance sending this message.
+ * @param newLoginHost The updated login host.
+ */
+- (void)loginViewController:(nonnull SFLoginViewController *)loginViewController didChangeLoginHost:(nonnull SFSDKLoginHost *)newLoginHost;
+
+@end
 
 
 /** The Salesforce login screen view.
@@ -36,6 +53,11 @@
 /** Returns a shared singleton of `SFLoginViewController` class.
  */
 +(_Nonnull instancetype)sharedInstance;
+
+/**
+ * The delegate representing the owner of this object.
+ */
+@property (nonatomic, weak, nullable) id<SFLoginViewControllerDelegate> delegate;
 
 /**
  * Outlet to the OAuth web view.
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -32,9 +32,12 @@
 #import "SFSDKLoginHostListViewController.h"
 #import "SFSDKLoginHostDelegate.h"
 #import "UIColor+SFColors.h"
+#import "SFSDKResourceUtils.h"
+#import "SFUserAccountManager.h"
+#import "SFAuthenticationManager.h"
 
 
-@interface SFLoginViewController () <SFSDKLoginHostDelegate, SFUserAccountManagerDelegate, SFAuthenticationManagerDelegate>
+@interface SFLoginViewController () <SFSDKLoginHostDelegate, SFUserAccountManagerDelegate>
 
 @property (nonatomic, strong) UINavigationBar *navBar;
 
@@ -236,6 +239,12 @@
     [self hideHostListView:YES];
 }
 
+- (void)hostListViewController:(SFSDKLoginHostListViewController *)hostListViewController didChangeLoginHost:(SFSDKLoginHost *)newLoginHost {
+    if ([self.delegate respondsToSelector:@selector(loginViewController:didChangeLoginHost:)]) {
+        [self.delegate loginViewController:self didChangeLoginHost:newLoginHost];
+    }
+}
+
 #pragma mark - Login Host
 
 - (void)showHostListView {
@@ -248,7 +257,7 @@
     [self dismissViewControllerAnimated:animated completion:nil];
 }
 
-#pragma mark - SF Authentication Manager
+#pragma mark - SFUserAccountManagerDelegate
 
 - (void)userAccountManager:(SFUserAccountManager *)userAccountManager
         willSwitchFromUser:(SFUserAccount *)fromUser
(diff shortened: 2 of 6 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -846,6 +846,18 @@ - (void)setupWithCredentials:(SFOAuthCredentials*) credentials {
...
 
+- (void)restartAuthentication {
+    @synchronized (self.authBlockList) {
+        if (!self.authenticating) {
+            [self log:SFLogLevelWarning format:@"%@: Authentication manager is not currently authenticating.  No action taken.", NSStringFromSelector(_cmd)];
+            return;
+        }
+        [self log:SFLogLevelInfo format:@"%@: Restarting in-progress authentication process.", NSStringFromSelector(_cmd)];
+        [self.coordinator stopAuthentication];
+        [self loginWithCredentials:[self createOAuthCredentials]];
```
bhariharan: How do we know if the authentication in progress didn't have its own `credentials` passed in? Instead of calling `[self createOAuthCredentials]`, we should be passing in the `credentials` originally passed in here when we restart. For example, in the hybrid view controller, we explicitly pass in credentials.

khawkins: Passing in credentials is effectively synonymous with an auth refresh, which wouldn't be expected to happen in conjunction with this method being called.

I take your point about the idea of making this generic enough to restart any in-progress authentication.  But if we do that, we're either going to have to change this interface to provide the auth configuration that needs to change—in this case, the login host—or use existing methods and more or less break encapsulation by specifying all of that information directly to the existing credentials object from the login host picker, before calling into authentication again.

There may be another way or middle ground, but I haven't seen it yet.

khawkins: The thing I keep coming back to is that this is effectively something I only want to run in the context of the login host picker.  I wonder if a better approach would be to change the host picker logic to raise a notification with the login host changes, which `SFAuthenticationManager` would listen for, and handle the rest of the re-configuration internally.  @bhariharan what are your thoughts on that approach?

bhariharan: Ideally, what I'd like to see is some sort of `SFLoginHostManager` that's completely independent where the value is pulled from. In this case, for now, in `- (void)loginWithCredentials:(SFOAuthCredentials *) credentials`, could we store `credentials` in a property that can be used in `restartAuthentication`? The property in question would be set only when auth is in flight and unset when it completes. It would also be internal, so external consumers can't misuse the property. Would that work?

khawkins: The property is already effectively stored, internally.  The issue is that, for this use case, you have to change it prior to restarting authentication, because the credentials object owns the login host at the time that authentication actually takes place.

So my approach for changing it was to create new credentials, which reconfigures with the (updated) login host, since in the use case we care about, the previous credentials are temporary anyway.

If we would want to re-use the existing credentials, we'd need another way to tell configure the object to use the new login host.  But in that scenario it doesn't really make sense, because existing credentials should be permanently tied to whatever login host they had in the first place.

Which is basically what I was getting at in my second point: this process should be reasonably hard-wired between a login host change from the user interface and an associated auth restart—there's no reason to initiate that auth restart in any other scenario I can think of.  Which is what led me to consider whether we should just use a callback approach (delegate, notification), which would satisfy that specificity of process.

huminzhi:
## Danger has errored
[!] Invalid `DangerFile` file: dlopen(/usr/local/lib/ruby/gems/2.4.0/gems/ox-2.4.9/ext/ox/ox.bundle, 9): Library not loaded: /usr/local/opt/ruby/lib/libruby.2.4.0.dylib
  Referenced from: /usr/local/lib/ruby/gems/2.4.0/gems/ox-2.4.9/ext/ox/ox.bundle
  Reason: image not found - /usr/local/lib/ruby/gems/2.4.0/gems/ox-2.4.9/ext/ox/ox.bundle
```
 #  from DangerFile:1
```

</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
bhariharan: That makes sense. Besides, I just noticed that this is an internal API. I'm comfortable with the notification approach for this.

khawkins: 👍 I'll take a look at that approach.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.h
```diff
@@ -26,7 +26,24 @@
  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#import "SalesforceSDKCore.h"
```
khawkins: This caused the a ripple effect of needing to import header files in places they weren't, but the reality is that this class wasn't actually making use of the broad-brush header import, and that kind of thing tends to have unintended consequences over time.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -32,9 +32,12 @@
 #import "SFSDKLoginHostListViewController.h"
 #import "SFSDKLoginHostDelegate.h"
 #import "UIColor+SFColors.h"
+#import "SFSDKResourceUtils.h"
+#import "SFUserAccountManager.h"
+#import "SFAuthenticationManager.h"
 
 
-@interface SFLoginViewController () <SFSDKLoginHostDelegate, SFUserAccountManagerDelegate, SFAuthenticationManagerDelegate>
```
khawkins: `SFAuthenticationManagerDelegate` wasn't actually implemented, so I took it out.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -203,12 +211,6 @@ @interface SFAuthenticationManager () <SFSDKLoginHostDelegate>
 - (void)cleanupStatusAlert;
 
 /**
- Method to present the authorizing view controller with the given auth webView.
- @param webView The auth webView to present.
- */
-- (void)presentAuthViewController:(WKWebView *)webView;
```
khawkins: This method is internally defined, but called nowhere.  I took it out.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
khawkins: To follow up, this now only gets called internal to the class, as the result of receiving a login host change callback from the login view controller (see below).  That keeps its usage a) specific and b) internalized/isolated.

PR #2067, opened 2017-05-31
Sync up related records (parent with children) for iOS

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.h
```diff
@@ -377,6 +377,10 @@ extern NSString* const kSFRestIfUnmodifiedSince;
 /// @name Other utility methods
 ///---------------------------------------------------------------------------------------
 
++ (BOOL)isStatusCodeSuccess:(NSUInteger)statusCode;
+
++ (BOOL)isStatusCodeNotFound:(NSUInteger)statusCode;
```
wmathurin: Hmm let me fix that

wmathurin: Neve mind

libs/SmartSync/SmartSync/Classes/Target/SFParentChildrenSyncUpTarget.m
```diff
@@ -0,0 +1,591 @@
...
+        }
+    }
+}
+
+- (NSDictionary *)parseIdsFromResponse:(NSDictionary *)refIdToResponses {
+    NSMutableDictionary *refIdToId = [NSMutableDictionary new];
+    for (NSString *refId in [refIdToResponses allKeys]) {
+        NSDictionary *response = refIdToResponses[refId];
+        if ([((NSNumber *) response[@"httpStatusCode"]) unsignedIntegerValue] == 201) {
+            NSString *serverId = response[@"body"][@"id"];
```
bhariharan: Could this be custom ID field?

bhariharan: Should this be `idFieldName`?

wmathurin: Actually no.
With standard/custom sobjects, the identity field when you get objects or query them with soql is "Id" (capital i). However when you create one sobject, the identity field in the response is "id" (small case i) - see https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_sobject_create.htm.

We don't have a way to specify an alternative identity field for the create response. If they have one, they will need to override the createOnServer method. Since alternate identity fields are only useful if you use other end points for your CRUD operations, you should be overriding that method anyway.

bhariharan: Sounds good.

PR #2068, opened 2017-06-01
Use existing credentials for authenticateWithCompletionBlock

trooper2013:


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -313,7 +317,16 @@ - (void)authenticateWithCompletionBlock:(SFOAuthPluginAuthSuccessBlock)completio
         } else if (failureBlock != NULL) {
             failureBlock(authInfo, error);
         }
-    }];
+    };
+
+    if (![SFUserAccountManager sharedInstance].currentUser) {
```
bhariharan: I don't see how this could happen (`SFHybridViewController` being present, but `currentUser` being yanked out from underneath), but we can leave that check in there for now until we refactor this class in `6.0`.

sgoldberg-sfdc:
👍

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -287,24 +287,28 @@ - (SFHybridViewConfig *)hybridViewConfig
...
      * Reconfigure user agent. Basically this ensures that Cordova whitelisting won't apply to the
      * WKWebView that hosts the login screen (important for SSO outside of Salesforce domains).
      */
     [SFSDKWebUtils configureUserAgent:[self sfHybridViewUserAgentString]];
-    [[SFAuthenticationManager sharedManager] loginWithCompletion:^(SFOAuthInfo *authInfo, SFUserAccount *userAccount) {
-        [self authenticationCompletion:nil authInfo:authInfo];
+    __weak __typeof(self) weakSelf = self;
+    SFOAuthFlowSuccessCallbackBlock authCompletionBlock = ^(SFOAuthInfo *authInfo, SFUserAccount *userAccount) {
+        __strong typeof(weakSelf) strongSelf = weakSelf;
+        [strongSelf authenticationCompletion:nil authInfo:authInfo];
```
khawkins: The only other thing we need to do is set `currentUser` here, based on the returned user account.

sgoldberg-sfdc: So this method really just forces the login screen to display? Does it make sense that the plugin would ever invoke auth vs a refresh?

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
khawkins: It could happen in the hybrid local case, where authentication through the plugin is happening on demand from the local user.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
khawkins: See my comment above.  In the hybrid local case, authentication can be deferred to a time determined by the user/developer.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
bhariharan: That would be driven through `SalesforceOAuthPlugin` though, which would trigger a different code path.

khawkins: This is the code path that's ultimately triggered by `SalesforceOAuthPlugin`.

PR #2070, opened 2017-06-01
Fixing bug that caused REST request looping on 401

bhariharan:
In some corner cases, it is possible to receive a `401` back from the service even after a successful refresh flow. One such use case is if an admin user temporarily suspends his/her account. Another possibility is if the token is valid but the correct API scopes weren't granted. Most of these situations indicate config errors, but nonetheless, our API shouldn't infinitely loop on the request-refresh cycle.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
@@ -180,6 +180,10 @@ __strong static NSDateFormatter *httpDateFormatter = nil;
 #pragma mark - send method
 
 - (void)send:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate {
+    [self send:request delegate:delegate shouldRetry:YES];
+}
+
+- (void)send:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate shouldRetry:(BOOL)shouldRetry {
     if (nil != delegate) {
         request.delegate = delegate;
     }
@@ -192,7 +196,7 @@ __strong static NSDateFormatter *httpDateFormatter = nil;
         [self log:SFLogLevelInfo msg:@"No auth credentials found. Authenticating before sending request."];
         [[SFAuthenticationManager sharedManager] loginWithCompletion:^(SFOAuthInfo *authInfo, SFUserAccount *userAccount) {
             __strong typeof(weakSelf) strongSelf = weakSelf;
-            [strongSelf enqueueRequest:request delegate:delegate];
+            [strongSelf enqueueRequest:request delegate:delegate shouldRetry:shouldRetry];
         } failure:^(SFOAuthInfo *authInfo, NSError *error) {
             [self log:SFLogLevelError format:@"Authentication failed in SFRestAPI: %@. Logging out.", error];
             NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
@@ -204,11 +208,11 @@ __strong static NSDateFormatter *httpDateFormatter = nil;
     } else {
 
         // Auth credentials exist. Just send the request.
-        [self enqueueRequest:request delegate:delegate];
+        [self enqueueRequest:request delegate:delegate shouldRetry:shouldRetry];
     }
 }
 
-- (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate {
+- (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate shouldRetry:(BOOL)shouldRetry {
     __weak __typeof(self) weakSelf = self;
     NSURLRequest *finalRequest = [request prepareRequestForSend];
     if (finalRequest) {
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -229,37 +233,43 @@ - (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)del
...
-            if ([refreshError.domain isEqualToString:kSFOAuthErrorDomain] && refreshError.code == kSFOAuthErrorInvalidGrant) {
-                [strongSelf log:SFLogLevelInfo format:@"%@ Invalid grant error received, triggering logout.", NSStringFromSelector(_cmd)];
-                // make sure we call logoutUser on main thread
-                dispatch_async(dispatch_get_main_queue(), ^{
-                    [strongSelf createAndStoreLogoutEvent:error user:user];
-                    [[SFAuthenticationManager sharedManager] logoutUser:user];
-                });
-            }
-        }];
+        if (shouldRetry) {
```
bhariharan: Added `if (shouldRetry)` here.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -229,37 +233,43 @@ - (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)del
...
+
+                    // Make sure we call logout on the main thread.
+                    dispatch_async(dispatch_get_main_queue(), ^{
+                        [strongSelf createAndStoreLogoutEvent:error user:user];
+                        [[SFAuthenticationManager sharedManager] logoutUser:user];
+                    });
+                }
+            }];
+        } else {
+            NSError *retryError = [[NSError alloc] initWithDomain:response.URL.absoluteString code:statusCode userInfo:nil];
```
bhariharan: Sending back response status code if we're done retrying.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -229,37 +233,43 @@ - (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)del
...
-        }];
+        if (shouldRetry) {
+            SFUserAccount *user = [SFUserAccountManager sharedInstance].currentUser;
+            [self log:SFLogLevelInfo format:@"%@: REST request failed due to expired credentials. Attempting to refresh credentials.", NSStringFromSelector(_cmd)];
+            self.oauthSessionRefresher = [[SFOAuthSessionRefresher alloc] initWithCredentials:user.credentials];
+            __weak __typeof(self) weakSelf = self;
+            [self.oauthSessionRefresher refreshSessionWithCompletion:^(SFOAuthCredentials *updatedCredentials) {
+                __strong typeof(weakSelf) strongSelf = weakSelf;
+                [strongSelf log:SFLogLevelInfo format:@"%@: Credentials refresh successful. Replaying original REST request.", NSStringFromSelector(_cmd)];
+                [strongSelf send:request delegate:delegate shouldRetry:NO];
```
bhariharan: Calling send with `shouldRetry = NO`, since we want to retry only once.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -180,6 +180,10 @@ - (void)userAccountManager:(SFUserAccountManager *)userAccountManager
 #pragma mark - send method
 
 - (void)send:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate {
+    [self send:request delegate:delegate shouldRetry:YES];
```
bhariharan: By default, it retries once.

bhariharan:
@trooper2013

bhariharan:
No - `Android` does only 1 retry I believe. This seems to have been lost when I swapped out `NetworkSDK`.

bhariharan:
CI is stuck as always. Merging.

khawkins:
Late to the party, sorry!  How were we dealing with this before?  In the cases you mention, is there not more specific data in the response to rule out those special cases?

It seems like we weren't running into this in our previous network incarnation.  If possible, it'd be great to take the lessons learned there and make our comparisons more specific—a blanket retry interval seems pretty broad-brush.

bhariharan:
There's no specific error message to compare against. Admins can temporarily deactivate an access token, which will return `401` until the admin reactivates it. There are some other cases. In general, we've always followed this approach on `Android`:
- If access token has expired, attempt to refresh the token.
- If response indicates that refresh token has expired, logout the user. Else, attempt to replay the request.
- If the replayed request returns `401` or any other error, pass that right back to the caller.
I don't think we should be retrying more than once. I don't remember the specific code that enforced this in the `NetworkSDK` version, but it did the same (retry only once) AFAIK.

Issue #2057, opened 2017-05-22
Cannot disable logging


mikecaulley:
1. Version of Mobile SDK Used:  5.1 
2. Issue found in Native App or Hybrid App: Native App, Smart Sync Example
3. OS Version: 10.3.2
4. Device: iPhone 7
5. Steps to reproduce: Edit the debug level line in AppDelegate init to always have 
            [SFLogger sharedLogger].logLevel = SFLogLevelOff;
6. Actual behavior: I still see log information in the output window.
7. Expected Behavior: I expect not to see any logging information.

trooper2013:
@mikecaulley Do you see log statements  from salesforce components? e.g. <Datetime  AppName  LEVEL com.salesforce.*>  We can control logging for mobile-sdk components only.

mikecaulley:
@trooper2013 I see logs like:

```
2017-05-22 16:08:50.445844-0400 SmartSyncExplorer[84959:1647906] syncDown:{
  "status" : "NEW",
  "target" : {
    "idFieldName" : "Id",
    "modificationDateFieldName" : "LastModifiedDate",
    "iOSImpl" : "SFSoqlSyncDownTarget",
    "type" : "soql",
    "query" : "SELECT LastModifiedDate, Id,FirstName,LastName,Title,MobilePhone,Email,Department,HomePhone FROM Contact LIMIT 10000"
  },
  "_soupEntryId" : 9,
  "maxTimeStamp" : 0,
  "totalSize" : -1,
  "type" : "syncDown",
  "progress" : 0,
  "options" : {
    "mergeMode" : "LEAVE_IF_CHANGED"
  },
  "soupName" : "contacts"
}
2017-05-22 16:08:50.446961-0400 SmartSyncExplorer[84959:1647906] Sync update:{
  "status" : "RUNNING",
  "target" : {
    "idFieldName" : "Id",
    "modificationDateFieldName" : "LastModifiedDate",
    "iOSImpl" : "SFSoqlSyncDownTarget",
    "type" : "soql",
    "query" : "SELECT LastModifiedDate, Id,FirstName,LastName,Title,MobilePhone,Email,Department,HomePhone FROM Contact LIMIT 10000"
  },
  "_soupEntryId" : 9,
  "maxTimeStamp" : 0,
  "totalSize" : -1,
  "type" : "syncDown",
  "progress" : 0,
  "options" : {
    "mergeMode" : "LEAVE_IF_CHANGED"
[...]

PR #2076, opened 2017-06-07
Logic tweak to the logging macro to prevent logging at a broader log …

MikeNachbaur:
…level than requested


wmathurin:
Would it fix: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/2057

MikeNachbaur:
Yes @wmathurin that's definitely caused by the code that my PR aims to fix.

wmathurin:
Thanks @MikeNachbaur

Issue #2066, opened 2017-05-30
Getting malformed query when we try to pass datetime field in SOQL query through performSOQLQuery method


SrinivasGit049:
Please fill out the following details:

1. Version of Mobile SDK Used: 5.1
2. Issue found in Native App or Hybrid App:NATIVE
3. OS Version:iOS 10.3
4. Device:iPad Air simulator
5. Steps to reproduce:
Try to send an soql query with having date value of following type with 
ex:[SFRestAPI sharedInstance]performSOQLQuery:@"Select Name from Account where LastModifiedDate > 2017-03-21T12:11:06.000+0000"  failure:  success];

6. Actual behavior:
performSOQLQuery failure block is  is getting called with malformed query error
7. Expected Behavior:
Completion handler should be called with result of query containing records
8. Error Log:
errorCode = "MALFORMED_QUERY";
        message = "\nLastModifiedDate > 2017-03-21T12:11:06.000 0000\n                                         ^\nERROR at Row:1:Column:780\nline 1:780 no viable alternative at character ' '";

This was working fine in SFSDK 4.x but broken in 5.1

bhariharan:
Thanks for reporting. I'll take a look.

wmathurin:
I confirmed the bug.
It looks like the + is not url encoded (even though we use [NSURLQueryItems](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m#L150).

wmathurin:
Others have reported the same issue (for instance [here](https://stackoverflow.com/questions/41273994/special-characters-not-being-encoded-properly-inside-urlqueryitems). 
We will figure out a work around for our next release.

wmathurin:
Fixed in dev.

PR #2081, opened 2017-06-13
Some special characters not url escaped

wmathurin:
Fix for: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/2066

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
@@ -24,9 +24,8 @@
 
 #import "SFRestRequest+Internal.h"
 #import "SFRestAPI+Internal.h"
-#import "SalesforceSDKConstants.h"
 #import "SFJsonUtils.h"
-#import "SalesforceSDKCore.h"
+#import <SalesforceSDKCore/NSString+SFAdditions.h>
 
 NSString * const kSFDefaultRestEndpoint = @"/services/data";
 
@@ -279,10 +278,12 @@ NSString * const kSFDefaultRestEndpoint = @"/services/data";
     NSMutableString *params = [NSMutableString new];
     if (components) {
         [params appendString:@"?"];
+        NSUInteger countParams = [components count];
         for (NSString *paramName in [components allKeys]) {
             [params appendString:paramName];
             [params appendString:@"="];
             [params appendString:[components[paramName] stringByURLEncoding]];
+            if (countParams > 1) [params appendString:@"&"];
         }
     }
     return params;
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
@@ -25,7 +25,7 @@
 #import "SFRestRequest+Internal.h"
 #import "SFRestAPI+Internal.h"
 #import "SFJsonUtils.h"
-#import <SalesforceSDKCore/NSString+SFAdditions.h>
+#import "NSString+SFAdditions.h"
 
 NSString * const kSFDefaultRestEndpoint = @"/services/data";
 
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -26,6 +26,7 @@
 #import "SFRestAPI+Internal.h"
 #import "SalesforceSDKConstants.h"
 #import "SFJsonUtils.h"
+#import "SalesforceSDKCore.h"
```
bhariharan: Why the extra import?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -281,4 +275,16 @@ + (SFRestMethod)sfRestMethodFromHTTPMethod:(NSString *)httpMethod {
...
 }
 
++ (NSString *)toQueryString:(NSDictionary *)components {
+    NSMutableString *params = [NSMutableString new];
+    if (components) {
+        [params appendString:@"?"];
+        for (NSString *paramName in [components allKeys]) {
+            [params appendString:paramName];
+            [params appendString:@"="];
+            [params appendString:[components[paramName] stringByURLEncoding]];
```
bhariharan: Don't we need to append a `&` at the end of this block before we add more params?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
wmathurin: Because stringByURLEncoding is defined in NSString+SFAdditions.h.
Should I import
```
#import <SalesforceSDKCore/NSString+SFAdditions.h>
```
instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
wmathurin: You are right!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
bhariharan: Yeah, probably? I don't know what the conventions are regarding importing umbrella header within the project.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -24,8 +24,8 @@
 
 #import "SFRestRequest+Internal.h"
 #import "SFRestAPI+Internal.h"
-#import "SalesforceSDKConstants.h"
 #import "SFJsonUtils.h"
+#import <SalesforceSDKCore/NSString+SFAdditions.h>
```
bhariharan: Minor: Could you change this notation to `#import "NSString+SFAdditions.h"`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -281,4 +274,18 @@ + (SFRestMethod)sfRestMethodFromHTTPMethod:(NSString *)httpMethod {
...
+    NSMutableString *params = [NSMutableString new];
+    if (components) {
+        [params appendString:@"?"];
+        NSUInteger countParams = [components count];
+        for (NSString *paramName in [components allKeys]) {
+            [params appendString:paramName];
+            [params appendString:@"="];
+            [params appendString:[components[paramName] stringByURLEncoding]];
+            if (countParams > 1) [params appendString:@"&"];
+        }
```
bhariharan: You need to remove the last `&` after the `for` block.

bhariharan: I think you can remove the `if` block here and just remove the last `&` after the `for` block.

paul-moon:
Is there an ETA on when the next release will be going out?  This fix is critical for the app I'm working on.

wmathurin:
In the next couple of months (safe harbor). 
We recommend you cherry pick the fix in your build if you can't wait.

paul-moon:
I'm currently using cocoapods for my dependencies and i was hoping to avoid a fork but it looks like that won't be possible.  thanks

PR #2085, opened 2017-06-19
Cleaning up RestAPI library

bhariharan:
@trooper2013 Here's the cleanup we discussed.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
@@ -24,7 +24,6 @@
 
 #import "SFRestAPI+Internal.h"
 #import "SFRestRequest+Internal.h"
-#import "SFOAuthCoordinator.h"
 #import "SFUserAccount.h"
 #import "SFAuthenticationManager.h"
 #import "SFSDKWebUtils.h"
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -197,6 +186,7 @@ - (void)send:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate shou
         [[SFAuthenticationManager sharedManager] loginWithCompletion:^(SFOAuthInfo *authInfo, SFUserAccount *userAccount) {
             __strong typeof(weakSelf) strongSelf = weakSelf;
             [strongSelf enqueueRequest:request delegate:delegate shouldRetry:shouldRetry];
+            [SFUserAccountManager sharedInstance].currentUser = userAccount;
```
khawkins: NB: what do you guys think about making this authentication block go away in the future (6.0?), in favor of allowing unauthenticated requests to fail in this scenario?  In our quest to minimize the setting of `currentUser`, this "convenience authentication" block gives me pause.  It means that the user has bypassed the SDK manager launch process, and IMO that's something that should be discouraged.  Thoughts?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/SFSDKTestRequestListener.m
```diff
@@ -116,9 +114,9 @@ - (NSString *)waitForCompletion {
 - (void)configureAccountServiceDelegate
 {
     if (_serviceType == SFAccountManagerServiceTypeIdentity) {
-        _authMgr.idCoordinator.delegate = self;
+        [SFAuthenticationManager sharedManager].idCoordinator.delegate = self;
```
khawkins: Should these not follow a pattern of saving and restoring the original value?  Seems like it could have an impact on downstream tests, potentially.

khawkins: ...but I guess they've been this way the whole time, so maybe not.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
bhariharan: Yes - I support that. Ideally, I'd like for us to make `SFUserAccountManager` the way to do anything auth related - login/logout/switch user.

PR #2086, opened 2017-06-20
Adding persistence for logging prefs

bhariharan:


huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

## Current coverage for SalesforceAnalytics.framework is `72%`
Files changed | - | - 
--- | --- | ---
SFSDKLogger.m | `98%` | :white_check_mark:

---

</p>

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.m
```diff
@@ -81,25 +83,90 @@ - (instancetype)initWithComponent:(NSString *)componentName {
...
     DDTTYLogger *consoleLogger = [DDTTYLogger sharedInstance];
     consoleLogger.colorsEnabled = YES;
     [self.logger addLogger:consoleLogger withLevel:logLevel];
     [self.logger addLogger:self.fileLogger withLevel:logLevel];
-    _logLevel = logLevel;
+}
+
+- (void)storeFileLoggingPolicy:(BOOL)enabled {
+    @synchronized (self) {
+        NSUserDefaults *defs = [NSUserDefaults standardUserDefaults];
```
khawkins: These calls should use `msdkUserDefaults`, no?

khawkins:
Minor comments.  Overall, looks fine to me.

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.m
bhariharan: We can't. That's in the core lib, this is one level lower.

khawkins: Should we consider the existing location of the user defaults functionality then?  Otherwise, we're basically saying that logging will not work for apps with extensions.  @trooper2013 do I have that right?

bhariharan: I don't think we can move that stuff to `SalesforceAnalytics` because they depend on `SFUserAccount`. The analytics library does not have any knowledge of user accounts.

khawkins: There shouldn't be any dependency on user accounts, that I know of.  The only dependency that `msdkUserDefaults` has is on `SFSDKDatasharingHelper`, which itself has no SDK dependencies other than a couple of lines of `SFLogger`, which presumably we'd rewrite into the new logging functionality.

It tends to get away from the namesake of "Analytics", but it'd be nice to have that support across the different app configurations.

bhariharan: You're right. I was looking at the wrong class. It doesn't seem too bad - I'll take a stab at moving it over for better consistency of user defaults.

bhariharan: Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2089).

PR #2087, opened 2017-06-21
Adding helper log APIs

bhariharan:


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.m
```diff
@@ -169,4 +196,21 @@ - (DDLogLevel)readLogLevel {
     return logLevel;
 }
 
+static inline DDLogFlag DDLogFlagForLogLevel(DDLogLevel level) {
+    switch (level) {
+        case DDLogLevelOff:
```
wmathurin: How does one turn it off then? (It looks like off is treated like error).

bhariharan: You turn it off by setting the `logLevel` property. This is simply returning the associated flag level. It doesn't affect the actual log level.

wmathurin: Shouldn't it be 0 for off?

Looking [here](http://cocoadocs.org/docsets/CocoaLumberjack/2.0.0/Constants/DDLogLevel.html), the level is the bit mask and flags are individual bits. What's the meaning of that function then?

bhariharan: You're right - it should probably be 0 for correctness. Looking at the implementation, it looks like flags are for fine-grained logging or for custom levels by using bitmask directly. Since the operation is bitwise `&` and the `level` is 0 for `OFF`, wouldn't the end result be 0 anyway no matter what the flag is? `DDLogFlag` is meant for fine-tuning log levels (which we don't really use in this case). See doc [here](http://cocoadocs.org/docsets/CocoaLumberjack/2.0.0/Constants/DDLogFlag.html). I guess the flag value should also be `DDLogLevelOff` in this case.

Good find on the `Cocoapod` doc - I was going crazy reading the doc within `Xcode` and looking at code to figure out the inner workings.

khawkins: > wouldn't the end result be 0 anyway no matter what the flag is?

It seems like it.

> You're right - it should probably be 0 for correctness.

That's probably the best bet.

khawkins:
Over all, looks good to me.

PR #2090, opened 2017-06-22
Migrating SalesforceReact to new logger framework

bhariharan:


libs/SalesforceReact/SalesforceReact/Classes/SFSDKReactLogger.h
```diff
@@ -0,0 +1,74 @@
...
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <SalesforceAnalytics/SFSDKLogger.h>
+
+@interface NSObject (SFSDKReactLogger)
```
wmathurin: If a class is importing SFSDKReactLogger.h and say SFSDKSmartStoreLogger.h, and then does ```[self d:xxx]```, what happens?

bhariharan: Good question. I really don't know. I'd have to try it I guess.

wmathurin: The compile might complain or maybe the last one wins...

bhariharan: I don't anticipate that happening though. These are meant for our components. Why would an app import both or even use our component logger for that matter?

wmathurin: What if I want to set the log level for SmartStore and React to DEBUG, wouldn't I import SFSDKReactLogger.h and SFSmartStoreLogger.h?

wmathurin: Maybe the NSObject extensions should be in a separate header?

bhariharan: Okay - I could look at moving the extensions to a common header.

huminzhi:
## Danger has errored
[!] Invalid `DangerFile` file: No Junit file was found at test_output/report.junit
```
 #  from DangerFile:1
```

</p>

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFSDKHybridLogger.h
```diff
@@ -0,0 +1,74 @@
...
+ */
+
+#import <SalesforceAnalytics/SFSDKLogger.h>
+
+@interface NSObject (SFSDKHybridLogger)
+
+/**
+ * Used to get and set the current log level associated with this logger.
+ */
+@property (nonatomic, readwrite, assign, getter=getLogLevel) DDLogLevel logLevel;
```
khawkins: Minor: in Objective-C, stylistically, "get" isn't usually preferred as a prefix to property getters.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFSDKHybridLogger.h
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Logs an error log line.
+ *
+ * @param message Log message.
+ */
+- (void)e:(nonnull NSString *)message;
```
khawkins: These are pretty descriptively terse and ambiguous at the `NSObject` level.  I'd prefer something more along the lines of `logError`.  A developer isn't necessarily going to understand what `[self e:msg]` means, by looking at it.

libs/SalesforceReact/SalesforceReact/Classes/SFSDKReactLogger.h
bhariharan: I think categories are problematic in this case. I'm going to change them to class methods instead.

bhariharan:
Should be good now.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFSDKHybridLogger.h
```diff
@@ -0,0 +1,88 @@
...
+ */
++ (void)setLogLevel:(DDLogLevel)logLevel;
+
+/**
+ * Logs an error log line.
+ *
+ * @param class Class.
+ * @param message Log message.
+ */
++ (void)e:(nonnull Class)class message:(nonnull NSString *)message;
```
wmathurin: Should we make the e,w,i,v,d methods [variadic functions](https://stackoverflow.com/questions/4804674/how-to-create-variable-argument-methods-in-objective-c), to safe the callers the ned to do ```[NSString stringWithFormat:...]```

bhariharan: I can take a look at that.

bhariharan:
I switched them over to variadic functions, but kept the original log lines as `stringWithFormat:` - it's too much work to change them.

PR #2091, opened 2017-06-23
Safari view controller for advanced oauth flow

wmathurin:
Using safari view controller instead of safari.

In this PR, keeping the same user experience (minus the swizzles) so you get an alert first, if you choose cancel, you get the host picker. Also if you choose done while the safari view controller is shown then you are back to the host picker.

Note: Rest api explorer now uses advanced oauth flow (it is using a new connected app id that has com.salesforce.mobilesdk.sample.restapiexplorer://oauth/success as call back url).

Next, we can look at putting the SFSafariViewController in a UINavigationController showing the same nav bar we have in SFLoginViewController - but there is some work to get there.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
```diff
@@ -220,6 +221,13 @@ typedef void (^SFOAuthBrowserFlowCallbackBlock)(BOOL);
 - (void)oauthCoordinator:(SFOAuthCoordinator *)coordinator willBeginBrowserAuthentication:(SFOAuthBrowserFlowCallbackBlock)callbackBlock;
 
 /**
+ Sent to notify the delegate that a browser authentication flow was cancelled out of by the user.
+ 
+ @param coordinator   The SFOAuthCoordinator instance processing this message.
+ */
+- (void)oauthCoordinatorDidCancelBrowserAuthentication:(SFOAuthCoordinator *)coordinator;
```
wmathurin: This gets called when the user exits the safari view controller by clicking the done button.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -509,17 +510,11 @@ - (void)continueNativeBrowserFlow {
...
-        NSError *launchError = [[self class] errorWithType:kSFOAuthErrorTypeBrowserLaunchFailed description:@"The native browser failed to launch for advanced authentication."];
-        dispatch_async(dispatch_get_main_queue(), ^{
-            [self notifyDelegateOfFailure:launchError authInfo:self.authInfo];
-        });
-    } else {
-        self.advancedAuthState = SFOAuthAdvancedAuthStateBrowserRequestInitiated;
-    }
-    
+
+    SFSafariViewController *svc = [[SFSafariViewController alloc] initWithURL:nativeBrowserUrl];
```
wmathurin: That's all it takes.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -1024,6 +1019,15 @@ - (void)webView:(WKWebView *)webView runJavaScriptConfirmPanelWithMessage:(NSStr
     }
 }
 
+#pragma mark - SFSafariViewControllerDelegate
+-(void)safariViewControllerDidFinish:(SFSafariViewController *)controller {
+    if ([self.delegate respondsToSelector:@selector(oauthCoordinatorDidCancelBrowserAuthentication:)]) {
```
wmathurin: Handling of done.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -852,11 +852,7 @@ - (void)setupWithCredentials:(SFOAuthCredentials*) credentials {
 
 - (void)restartAuthentication {
     @synchronized (self.authBlockList) {
-        if (!self.authenticating) {
-            [self log:SFLogLevelWarning format:@"%@: Authentication manager is not currently authenticating.  No action taken.", NSStringFromSelector(_cmd)];
-            return;
-        }
-        [self log:SFLogLevelInfo format:@"%@: Restarting in-progress authentication process.", NSStringFromSelector(_cmd)];
+        [self log:SFLogLevelInfo format:@"%@: Restarting authentication process.", NSStringFromSelector(_cmd)];
```
wmathurin: To me restart should do a start if it's not running.
(In the browser auth flow, we consider the authentication over if cancelled, as a result calling restart was not doing anything).
If we don't change the behavior, I could introduce a separate startAuthentication method.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -1400,6 +1403,11 @@ - (void)oauthCoordinator:(SFOAuthCoordinator *)coordinator displayConfirmationMe
     [[SFRootViewManager sharedManager] pushViewController:alert];
 }
 
+- (void)oauthCoordinatorDidCancelBrowserAuthentication:(SFOAuthCoordinator*)coordinator
+{
+    [self delegateDidCancelBrowserFlow];
```
wmathurin: Hitting done when the safari view controller is shown and hitting cancel on the alert that precedes safari being launched should have the same effect.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -36,8 +36,8 @@
 #import <SalesforceSDKCore/SFAuthenticationManager.h>
 
 // Fill these in when creating a new Connected Application on Force.com
-static NSString * const RemoteAccessConsumerKey = @"3MVG9Iu66FKeHhINkB1l7xt7kR8czFcCTUhgoA8Ol2Ltf1eYHOU4SqQRSEitYFDUpqRWcoQ2.dBv_a1Dyu5xa";
-static NSString * const OAuthRedirectURI        = @"testsfdc:///mobilesdk/detect/oauth/done";
+static NSString * const RemoteAccessConsumerKey = @"3MVG98dostKihXN53TYStBIiS8NRTXcbDzn9nHPb3piMElfQDD.kTyHeXjKV9JNUbe5sZeSQ4CVY1Onzpq21N";
```
wmathurin: That way Rest API explorer will be our go to application for testing advanced oauth flows.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/RestAPIExplorer-Info.plist
```diff
@@ -29,7 +40,7 @@
 	<key>NSMainNibFile~ipad</key>
 	<string></string>
 	<key>SFDCOAuthLoginHost</key>
-	<string>login.salesforce.com</string>
+	<string>test.salesforce.com</string>
```
wmathurin: I can revert that line but I always use it against sandbox and I got tired of switching hosts.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -509,17 +510,11 @@ - (void)continueNativeBrowserFlow {
...
-            [self notifyDelegateOfFailure:launchError authInfo:self.authInfo];
-        });
-    } else {
-        self.advancedAuthState = SFOAuthAdvancedAuthStateBrowserRequestInitiated;
-    }
-    
+
+    SFSafariViewController *svc = [[SFSafariViewController alloc] initWithURL:nativeBrowserUrl];
+    svc.delegate = self;
+    [[SFRootViewManager sharedManager] pushViewController:svc];
```
khawkins: This breaks the configurability of root view management in the SDK.  Everywhere else where `SFRootViewManager` is employed, there is configurable logic to replace it with an alternative approach to "uber" VC management.

We need to provide a mechanism for configuring how the VC will be displayed.  I would suggest:

1. Creating an `SFOAuthCoordinatorDelegate` method (similar to `oauthCoordinator:didBeginAuthenticationWithView:`) that hands this VC off to `SFAuthenticationManager`.
2. Providing customization hooks similar to (or embedded in) `SFAuthenticationViewHandler` within `SFAuthenticationManager`, to allow customizable presentation of the VC.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
khawkins: I don't think this should be optional.  Delegates of the OAuth coordinator need to be able to handle this event, if/when it occurs.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
khawkins: I'm not seeing in the flow where we consider authentication over if canceled.  Where does that cancelation occur?

wmathurin: This gets called from [here](https://github.com/wmathurin/SalesforceMobileSDK-iOS/blob/805fbd1e344a0c630e879af09a1fae58fa253eb7/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m#L1025). Is that what you are asking?

khawkins: Oh I commented in the wrong place, which makes my comment pretty confusing.  I'll add it to the proper code above and amend.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
khawkins: I'm not seeing in the flow where we consider authentication over if canceled. Where does that cancelation occur?  At least in your embedded VC flow.  And I actually prefer your behavior—not "finishing" authentication in this case—because it otherwise breaks the contract passed in to the public login methods by the consumer.

The restart process was actually implemented to solve these kinds of auth-internal use cases.  I.e. don't "finish" the authentication, but do a soft reset, and then the next iteration can still call the appropriate completion handlers that were configured originally.

So I'd opt for keeping the `self.authenticating` guard.

bhariharan:
What has changed in the `shared` submodule?

wmathurin:
@bhariharan Nothing changed in shared, it's back to where it should be with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2091/commits/2fe6aa8380d3b83f94e4e9b586b7f439eb42c295

huminzhi:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

Files changed | - | - 
--- | --- | ---
SFOAuthCoordinator.m | `32%` | :no_entry_sign:
SFAuthenticationManager.m | `42%` | :no_entry_sign:

---
## Current coverage for SalesforceAnalytics.framework is `32%`

---

</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
wmathurin: I introduced a new handler (SFAuthenticationSafariControllerHandler) here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2091/commits/2fe6aa8380d3b83f94e4e9b586b7f439eb42c295

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
wmathurin: restartAuthentication is new in 5.2, and is called only from loginViewController:didChangeLoginHost:. That happens when you are changing host during a normal login flow.

I wanted to call it also from hostListViewController:didChangeLoginHost:. That one happens when you are changing flow after cancelling out of an advanced login flow.

In the first case authenticating is happening but not in the second case.

Either I change restartAuthentication to always start an authentication if none is happening.
Or I could introduce a new method startAuthentication and call it in the second case.

khawkins: I went through the flow, and the use case I missed is that we're still showing the initial alert about Safari opening to authenticate.  Shouldn't we just remove that with the embedded view?  I don't see that it brings value when we switch to an inline authentication process.

The process I was commenting on was from the "Safari VC close" use case, in which case the authentication is still technically in progress, and should be fine.

khawkins: And as I mentioned in my first comment, because the "Cancel the initial alert" use case cancels the in-progress authentication, it breaks the contract created by the consumer.

With every other action of our asynchronous public login methods, there are only two paths out: completion or failure.  And those run through the respective handler blocks passed into those methods. `self.authenticating` is just a convenience property to determine whether there have been auth callback container object(s) established.  No auth callback containers means the login* methods either haven't been called, or have completed.

In the event of a successful "restarted" authentication in this use case, what's the tie-back to the original caller?  There isn't one.  Broken contract.

If this is an opportunity to close the gap on that aberration, I think we should do it—in addition to the fact that I still don't see any further value to having the alert anyway.

sgoldberg-sfdc:
Thank you for doing this. I had not expected these changes so soon.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -850,13 +852,21 @@ - (void)setupWithCredentials:(SFOAuthCredentials*) credentials {
     self.idCoordinator.delegate = self;
 }
 
-- (void)restartAuthentication {
+- (void)restartAuthentication:(BOOL)onlyIfAuthenticating {
```
wmathurin: @khawkins how do you like that?
(remember it's private method only used from two places loginViewController:didChangeLoginHost: and hostListViewController:didChangeLoginHost:

wmathurin:
Safari VC auth flow doesn't play nice with multi user support:
- if you try to login a second user against the same host, you won't be able to, because the browser retained cookies and will take you straight to the oauth success callback url and therefore back into the app
- if you decide you no longer want to add a new user, there is no way to cancel out, getting out of the safari vc, takes you to the host list and there is no way out of there
NB: both issues also exist with safari-based advanced auth flow.

sgoldberg-sfdc:
NB: both issues also exist with safari-based advanced auth flow.

I have observed this myself. I am not aware of any customer requests to support that use case, are you?

wmathurin:
@sgoldberg-sfdc Good point.

wmathurin:
Here are a few reasons for keeping the alert (until maybe we do something more ambition with the Safari VC in a new VC wih nav bar):
1) it keeps the behavior the same as before (so minimizes doc update for a point release)
2) it  warns you that the UI is about to change significantly
3) in the case of multi user, it lets you at least log in a second user if it uses another host

If we really want to drop the alert (in this release or the next), I would prefer to do it in a subsequent PR.
Thoughts?

bhariharan:
@wmathurin I'm in favor of keeping the alert dialog around for this release. It's probably best to avoid drastic changes this close to `5.2`, especially given that we don't have a good test environment for cert based auth.

khawkins:
> in the case of multi user, it lets you at least log in a second user if it uses another host

My main concern with this is that it does so in a way that's disconnected from the original request.  A consumer has no way to respond to this subsequent authentication.

This is likely an existing bug with the current flow as well—I haven't looked into it, but I don't see why it wouldn't be—so I suppose we can keep it that way.  But I'd lobby for fixing this underlying issue, perhaps in the 6.0 timeframe, so that this use case is properly supported.

wmathurin:
Test me

PR #2105, opened 2017-06-29
Replacing log lines in tests and sample apps

bhariharan:
Also marked `SFLogger` deprecated.


bhariharan:
@wmathurin Updated pod specs in [this commit](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2105/commits/b3e63e52308eed1c687c32e2baee074c3ee1d74d).

libs/SalesforceHybridSDK/SalesforceHybridSDKTestApp/Classes/AppDelegate.m
```diff
@@ -197,15 +189,20 @@ - (void)initializeAppViewState
     self.window.rootViewController = self.viewController;
     [self.window makeKeyAndVisible];
 }
-                                                            
 
-@end
+- (void)logWithLevel:(DDLogLevel)level format:(NSString *)format, ... {
```
wmathurin: You had to introduce such a method in many places.
Could we have a default logger for all those cases?

bhariharan: Yeah, I did it for the convenience of replacement (path of least resistance) since they were only test files. Should I create a shared `TestLogger`?

wmathurin: It's also in sample apps code.
Should we have a default logger in the SDK?

khawkins: > Should we have a default logger in the SDK?

I've been wondering the same thing, as we've gone through these reviews.  Seems like, along with the component-specific loggers, it would be useful—possibly the most common case, even—to have a "default" component for logging.

bhariharan: Ah, I see. So you're suggesting the component name be the same for tests and sample apps. I could create a `default` logger for that - sure.

khawkins: I would go a step further, and suggest that we implement an `[SFSDKLogger sharedInstance]` method, that's basically a convenience to `[SFSDKLogger sharedInstanceWithComponent:@"default"]`.  Thoughts?

bhariharan: Not a bad idea. I could do that. However, this will change the behavior here. I was logging them all with different component names. This will make them all use the same component name - `default`. Is that what we want for tests? I guess it's not a big deal - they're just tests.

khawkins: I see what you mean.  I guess whether or not it matters overly much in this case depends on the extent to which we use those log messages to aid in the differentiation of their source.  I'm not sure.  Anyway, I like a default logger in either case. 😉

bhariharan: Okay - I'll add a default logger and replace most of these calls with the default logger.

bhariharan:
@khawkins Is [this](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2105/commits/ed020557b22ef925133c3ab043064a768712da94) what you were thinking? I can switch the tests to use this logger and simplify the code.

khawkins:
Yeah, pretty much.  Seems like a convenience that many developers would make use of. 👍

bhariharan:
Tests and sample apps use the default logger now.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -48,9 +48,9 @@ - (id)init
     self = [super init];
     if (self) {
         #if defined(DEBUG)
-            [SFLogger sharedLogger].logLevel = SFLogLevelDebug;
+            [SFSDKLogger sharedDefaultInstance].logLevel = DDLogLevelDebug;
```
wmathurin: What about the non default loggers, will they be set to debug/info also?

bhariharan: I discussed with Urvin. The default we want in release mode is `Error` and debug mode is `Debug` for all loggers. I should remove these from here.

bhariharan:
Removed default logger's level being set [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2105/commits/aa9942f92317d1707fd934da174189b32450c69e). We want to go with `Error` and `Debug` for `Release` and `Debug` respectively.

wmathurin:
I see, you have that logic [there](https://github.com/bhariharan/SalesforceMobileSDK-iOS/blob/aa9942f92317d1707fd934da174189b32450c69e/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.m#L205) now

PR #2125, opened 2017-07-11
SmartStore compilation on XCode 9

wmathurin:
Using sqlite3.h that comes with sqlcipher

### DIFF
--- a/external/ThirdPartyDependencies/sqlcipher/sqlite3.h
+++ b/external/ThirdPartyDependencies/sqlcipher/sqlite3.h
@@ -1,0 +1,8688 @@
+/*
+** 2001 September 15
+**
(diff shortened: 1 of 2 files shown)


wmathurin:
I'm still looking at the cocoapods side of things.

bhariharan:
Is it not bundled as part of the `podspec` for `SQLCipher`?

wmathurin:
It is bundled. But FMDB is including from &lt;sqlite3.h&gt; and picks up the system one. If I change the include to "sqlite3.h" it works fine.
Trying to figure out the right xconfig setting in SmartStore.podspec to make it work.

wmathurin:
Will address the cocoapod side with separate PRs.

PR #2145, opened 2017-07-20
Centralized SFSDKLogger code

khawkins:
For the team's consideration.  We have a lot of duplicated logger code in the new logging system, that seems ripe for centralizing through an inheritance model.  This code implements that approach with `SFSDKAnalyticsLogger`.  If this is agreeable, I can pretty quickly bang out the rest of the logging to follow suit.

### DIFF
--- a/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.h
+++ b/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.h
@@ -59,6 +59,14 @@
 @property (nonatomic, readwrite, assign, getter=isFileLoggingEnabled) BOOL fileLoggingEnabled;
 
 /**
+ * By default, returns the default shared instance. Child classes implement this method to return an instance with
+ * their defined component name.
+ *
+ * @return Instance of this class.
+ */
++ (nonnull instancetype)sharedInstance;
+
+/**
  * Returns an instance of this class associated with the default component.
  *
  * @return Instance of this class.
@@ -94,12 +102,30 @@
 - (void)e:(nonnull Class)class message:(nonnull NSString *)message;
 
 /**
+ * Logs an error log line.
+ *
+ * @param class Class.
+ * @param format The format message, and optional arguments to expand in the format.
+ * @param ... The arguments to the message format string.
+ */
+- (void)e:(nonnull Class)class format:(nonnull NSString *)format, ...;
+
+/**
  * Logs a warning log line.
  *
  * @param class Class.
  * @param message Log message.
  */
 - (void)w:(nonnull Class)class message:(nonnull NSString *)message;
+
+/**
+ * Logs a warning log line.
+ *
+ * @param class Class.
+ * @param format The format message, and optional arguments to expand in the format.
+ * @param ... The arguments to the message format string.
+ */
+- (void)w:(nonnull Class)class format:(nonnull NSString *)format, ...;
 
 /**
  * Logs an info log line.
@@ -110,6 +136,15 @@
 - (void)i:(nonnull Class)class message:(nonnull NSString *)message;
 
 /**
+ * Logs an info log line.
+ *
+ * @param class Class.
+ * @param format The format message, and optional arguments to expand in the format.
+ * @param ... The arguments to the message format string.
+ */
+- (void)i:(nonnull Class)class format:(nonnull NSString *)format, ...;
+
+/**
  * Logs a verbose log line.
  *
  * @param class Class.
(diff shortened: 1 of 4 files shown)


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.h
```diff
@@ -94,6 +102,15 @@
 - (void)e:(nonnull Class)class message:(nonnull NSString *)message;
 
 /**
+ * Logs an error log line.
+ *
+ * @param class Class.
+ * @param format The format message, and optional arguments to expand in the format.
+ * @param ... The arguments to the message format string.
+ */
+- (void)e:(nonnull Class)class format:(nonnull NSString *)format, ...;
```
khawkins: These are just added for symmetry.  Seems like both method signatures should exist for each log level.

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.h
```diff
@@ -144,4 +197,103 @@
  */
 - (void)log:(nonnull Class)class level:(DDLogLevel)level format:(nonnull NSString *)format, ...;
 
+/**
```
khawkins: All of these are class-level conveniences to make working at the class level cleaner.

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.m
```diff
@@ -69,6 +69,10 @@ + (instancetype)sharedInstanceWithComponent:(NSString *)componentName {
     }
 }
 
++ (instancetype)sharedInstance {
+    return [self sharedInstanceWithComponent:kDefaultComponentName];
```
khawkins: NB: `sharedDefaultInstance` could just go away with this implementation—it'd simply be the default implementation of `sharedInstance`.

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Util/SFSDKAnalyticsLogger.h
```diff
@@ -29,65 +29,8 @@
 
 #import "SFSDKLogger.h"
 
-@interface SFSDKAnalyticsLogger : NSObject
+extern NSString * _Nonnull const kSFSDKAnalyticsComponentName;
 
-/**
```
khawkins: Basically, all of the component-specific code can simply be pulled up into the parent class, leaving only the component configuration.

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Logger/SFSDKLogger.h
bhariharan: I'd rather not have class-level conveniences, at least not for setting and getting log level. The behavior can be misleading to developers especially based on our 5.2 documentation. For example, a developer might use `[SFSDKSmartStoreLogger setLogLevel:...]` thinking that he/she is setting the log level for `SmartStore`, but it's actually being set for the default component. Same goes for any of the class level logging methods.

khawkins: They will be set for the proper component.  The key is the `sharedInstance` override.  Each of our established components (like `SFSDKAnalyticsLogger` is here) override this method, to return their own component.  Same for the log messages.

The only time that they'd go to the default is if you e.g. did `[SFSDKLogger setLogLevel:]`, which IMO is also intuitive—the default is simply the base class.  Thoughts?

bhariharan: You're right. My bad. That's why I shouldn't be doing code reviews at 6 A.M. with no sleep the previous night.

Looks good to me. 👍

khawkins: Cool.  I'll do the follow-on components in the next PR.

wmathurin: Interesting. That would not work in Java (you can't override static methods).
For a good read on why java is that way see [this](https://stackoverflow.com/questions/2223386/why-doesnt-java-allow-overriding-of-static-methods).

khawkins: I contemplated this when I was initially looking at solutions.  In a more classic OO design pattern, it seems that you'd separate your `LoggingConfiguration` class (which would be more of an instance class) from your `Logger` class (which would be more of a service class that owned `LoggingConfiguration` instances).  In the current design, those roles are both aggregated in the same class.

While I think I'd like the clean separation in that approach, it seemed like a bigger departure from the current implementation.  It's an idea I could poke at, though, if you guys think that would be better.

PR #2151, opened 2017-07-24
Replacing SFRootViewManager with multi-window based solution

trooper2013:
@khawkins @bhariharan  Please review. 

There were quite a few quirks with the multi-window approach. We have tested for all of the mentioned problems. UIKit doesn't quite play nice with multiple windows during rotation, accessibility-features & special overlays (context menus etc.). Tested native & hybrid apps with and without passcode setting, with advanced Auth as well.

NOTE:
1. WindowLevels if set too high result in NO-SHOW on status bar. I don't see any value in having the window levels to be higher that UIWindowLevelAlert.

2. Multiple UIWindows that are visible cause quite a bit of confusion in the UIKIT w.r.t accessibility and other floating overlay options.  It results in overlays and accessibility being attached to the wrong windows. 

3. During rotations all UIWindow frames except for the keyWindow is updated by UIKit. Looks like the CGAffineTransforms are executed but not having a root-view controller at this point creates awkward behavior. 

Overview 
The crux of this PR revolves around  the SFSDKWindowManager & SFSDKWindowContainer. The SDK now creates and manages  snapshotWindow, passcodeWindow & authWindow. These are listed in descending order of their Z-order levels. 

1.  The Manager manages the creation of  windows and bring2front/send2back functionality.  

2. The SFSDKWindowContainer is a wrapper around a UIWindow (SFSDKWindow) and has the functionality to manage its view  stack.

3. SFRootController is used for  the SDK managed windows. All this controller does is traverses the view stack and invokes shouldAutorotate, supportedInterfaceOrientations etc..

4. SFSDKWindow at this point doesn't do much, but it could be used to add overriding behavior for SDK managed windows in the future. It provides a way for us to identify SDK created windows if required. 

Addl. Notes : 

1. Ive noticed that using setVisible property on UIWindow causes a black screen during transitions between windows.  Need to investigate if theres a better workaround. 

2. Should we provide a way to release the references to all these UIWindows? On the same note should we hold a weak reference to a SFSDKWindow in the SFContainer and re-create if needed? 
 
3. Do we need additional customization points  for these windows for  sdk consumers?


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -206,7 +206,7 @@ typedef void (^SFSnapshotViewControllerDismissalBlock)(UIViewController* snapsho
 
 /**
  The block to execute to present the snapshot viewcontroller.
- If this property is not set, SFRootViewManager will be used to present the snapshot.
+ If this property is not set, SFWindowManager will be used to present the snapshot in the snpshot window.
```
khawkins: Minor: looks like this should be `SFSDKWindowManager`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.h
```diff
@@ -0,0 +1,157 @@
...
+ */
+- (void)pushViewController:(UIViewController *)controller window:(SFSDKWindowContainer *)window withCompletion:(void (^)(void))completion;
+
+/** Api to pop viewcontroller from a given window.
+ */
+- (void)popViewController:(UIViewController *)controller window:(SFSDKWindowContainer *)window withCompletion:(void (^)(void))completion;
+
+/** Returns the SFSDKWindowContainer window representing the mainWindow that has been set
+ */
+- (SFSDKWindowContainer *)mainWindow;
```
khawkins: Why are this and `setMainWindow:` defined explicitly, rather than exposing a property?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -235,8 +231,8 @@ - (BOOL)launch
     [SFSDKCoreLogger i:[self class] format:@"Launching the Salesforce SDK."];
     _isLaunching = YES;
     self.launchActions = SFSDKLaunchActionNone;
-    if ([SFRootViewManager sharedManager].mainWindow == nil) {
-        [SFRootViewManager sharedManager].mainWindow = [SFApplicationHelper sharedApplication].windows[0];
+    if ([SFSDKWindowManager sharedManager].mainWindow == nil) {
```
khawkins: We're using `mainWindow` interchangeably with property notation and selector notation, and inconsistently so, including its definition.  Can we standardize on property notation?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowContainer.h
```diff
@@ -0,0 +1,176 @@
...
+
+@protocol SFSDKWindowContainerDelegate<NSObject>
+
+@optional
+
+/**
+ Called when the window is going to be made key & visible
+ @param window The window that is going to be made key & visible
+ */
+- (void)windowWillMakeKeyVisible:(SFSDKWindowContainer *)window;
```
khawkins: New classes should define nullability constraints on all methods and properties.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowContainer.h
```diff
@@ -0,0 +1,176 @@
...
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#import <Foundation/Foundation.h>
+#import <UIKit/UIKit.h>
+
+typedef NS_OPTIONS(NSUInteger, SFSDKWindowType) {
```
khawkins: This is not the proper data structure.  `NS_OPTIONS` should be used for bitwise flags that can be considered as different components of the same value.  This should be an `NS_ENUM`.

libs/SalesforceSDKCore/SalesforceSDKCore/SFSDKWindowManagerTests.m
```diff
@@ -0,0 +1,65 @@
+//
+//  SFSDKWindowManagerTests.m
```
khawkins: Ideally, we could abstract the windowing interfaces to drive more unit tests around the logic for window ordering, priorities, etc.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,304 @@
...
+    SFSDKWindowContainer *_prevActiveWindow;
+    SFSDKWindowContainer *_currentWindow;
+}
+@property (nonatomic, strong) NSMutableOrderedSet *delegates;
+
+@end
+
+@implementation SFSDKWindowManager
+
+static const CGFloat SFWindowLevelAuth      = 300;
```
khawkins: Wouldn't it be prudent to put all "non-main" window levels that you care about somewhere between `UIWindowLevelStatusBar` and `UIWindowAlert`, at a minimum?

khawkins: Also, these values should not be set to literal values—they should be defined in terms of the UIKit window level constants.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,304 @@
...
+        _prevActiveWindow = _currentWindow;
+    
+    _currentWindow = windowContainer;
+    [windowContainer makeKeyVisible];
+    
+    //send the rest of the windows to back
+    for ( NSString *wName in self.namedWindows.keyEnumerator ) {
+        SFSDKWindowContainer *wContainer = [self.namedWindows objectForKey:wName];
+        if ( ![wName isEqualToString:windowContainer.windowName] ) {
+            [wContainer sendToBack];
```
khawkins: This is confusing to me.  On the one hand, we're creating distinct window levels for our named windows.  But on the other, we're still managing visibility of the windows through other mechanisms.  Why have both?  We can we not just e.g. manage window visibility with a combination of window levels and `makeKeyAndVisible`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindow.m
```diff
@@ -0,0 +1,109 @@
...
+
+@end
+
+@implementation SFSDKWindow
+
+- (instancetype) initWithFrame:(CGRect) frame {
+    self = [super initWithFrame:frame];
+    
+    if (self) {
+        self.hidden = YES;
```
khawkins: It seems like overkill to manage a child class of `UIWindow`, simply for setting its default visibility.  Can we not just use `UIWindow` and configure it in this case?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
trooper2013: @khawkins Merely setting distinct window levels doesn't work as intended. If the send2back is not done (set the other window levels to lower), I've observed that the current key window  becomes unresponsive although it is the first responder & is visible. Comment this block of code you will notice the effects.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
trooper2013: @khawkins These are truly just offsets from UIWindowLevelNormal. I will rename them. We should have all window levels lower than UIWindowLevelStatusBar except that of snapshot. Windows with UIWindowLevelStatusBar and higher wont show the status bar.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindow.m
trooper2013: @khawkins I can do away with it. I decided to extend UIWindow inorder to provide the sdk with ability to sift through app windows and determine the ones created by the Window Manager. I left it in place for future extensions. We don't need it now, but I can take it out.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
khawkins: > Windows with UIWindowLevelStatusBar and higher wont show the status bar.

👍 

Ah, looking at the constants usage, I realize these are just offsets.  Yeah, maybe rename them to indicate that the values are offset values, rather than assumed absolute window levels.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindow.m
khawkins: That could be a reasonable purpose.  I'm okay with that—just seemed like overkill if this was the end of the line.

huminzhi:
## Danger has errored
```
 #  from DangerFile:5
 #  
 >  xcov.report(
 #     scheme: 'UnitTests',
```

</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,308 @@
...
+        result = YES;
+    }
+    return result;
+}
+
+- (SFSDKWindowContainer *)windowWithName:(NSString *)name {
+    return [self.namedWindows objectForKey:name];
+}
+
+- (void)bringToFront:(SFSDKWindowContainer *)windowContainer {
```
khawkins: The logic that we're doing here is at odds with the reasoning for exposing windows of different window levels in the first place.  The idea is that, with windows in a structured window level hierarchy, you don't manipulate the ordering of windows.  Instead, you let the framework own responsibility for which window should have the focus, and merely toggle the visibility of your windows as they come into or go out of scope.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowContainer.m
```diff
@@ -0,0 +1,289 @@
...
+
+- (void)pushViewController:(UIViewController *)controller {
+    [self pushViewController:controller animated:NO completion:nil];
+}
+
+- (void)popViewController:(UIViewController *)controller {
+    [self popViewController:controller animated:NO completion:nil];
+}
+
+- (void)pushViewController:(UIViewController *)viewController animated:(BOOL)flag completion:(void (^)(void))completion {
```
khawkins: I'm not a big fan of these methods.  The idea of establishing a multi-window hierarchy was in part to get rid of fancy view controller management, in favor of a more established best practices for consumers to manage their own view hierarchies.

In this paradigm, I feel that we should leave view controller management to consumers of this functionality.  Even for internal consumption, we should know enough about our view hierarchies to explicitly manage what gets shown, where.  And if we don't, we need to ask why.

Leaving this functionality around encourages the kind of deviation from best practices that we've historically seen in this realm.  Consumers should be responsible for their view state management.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowContainer.m
```diff
@@ -0,0 +1,289 @@
...
+        return;
+    }
+    [self enumerateDelegates:^(id<SFSDKWindowContainerDelegate> delegate) {
+        if ([delegate respondsToSelector:@selector(windowWillMakeKeyVisible:)]) {
+            [delegate windowWillMakeKeyVisible:weakSelf];
+        }
+    }];
+    
+    self.window.windowLevel= fabs(self.window.windowLevel);
+    [self.window setHidden:NO];
```
khawkins: Toggling on `hidden` precludes the ability to animate these changes in and out.  I'd recommend using `alpha` to drive the visibility of the window, which allows for animation.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
khawkins: Whatever form this method and its counterpart take—making a window visible or not—should a) be asynchronous, as with all UI display interactions, and b) ideally provide the option for animation, to control the presentation experience.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,308 @@
...
+#import "SFSDKWindowManager.h"
+#import "SFSDKWindowContainer.h"
+#import "SFSDKRootController.h"
+#import "SFApplicationHelper.h"
+#import "SFSecurityLockout.h"
+@interface SFSDKWindowManager() <SFSDKWindowContainerDelegate> {
+    SFSDKWindowContainer *_prevActiveWindow;
+    SFSDKWindowContainer *_currentWindow;
+}
+@property (nonatomic, strong) NSMutableOrderedSet *delegates;
```
khawkins: Having delegate containers that hold strong references is problematic.  This should be switched to an `NSHashTable` pattern.  See `SFUserAccountManager` for an example, and the following PRs [[1](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1640), [2](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1467)] for an explanation.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,308 @@
...
+                                              valueOptions:NSMapTableStrongMemory];
+        _delegates = [NSMutableOrderedSet orderedSet];
+    }
+    return self;
+    
+}
+
+- (SFSDKWindowContainer *)mainWindow {
+    if (![self.namedWindows objectForKey:kSFMainWindowKey]) {
+        [self setMainUIWindow:[SFApplicationHelper sharedApplication].windows[0]];
```
khawkins: I'd suggest that this logic also include the possibility of the `UIApplicationDelegate.window` property.  So something like:
```obj-c
if ([SFApplicationHelper sharedApplication].delegate.window != nil) {
    [self setMainUIWindow:[SFApplicationHelper sharedApplication].delegate.window];
} else if ([SFApplicationHelper sharedApplication].windows[0] != nil) {
    [self setMainUIWindow:[SFApplicationHelper sharedApplication].windows[0]];
} else {
    // We should probably consider an adverse case where there are no windows defined.
}
```

libs/SmartStore/SmartStore/Classes/SFSmartStoreInspectorViewController.m
```diff
@@ -28,7 +28,6 @@
 
 #import "SFSmartStoreInspectorViewController.h"
```
wmathurin: I wonder if SmartStore inspector should go to its own window also.
(Right now it's not available in React Native apps because I ran into issues when simply pushing the VC on top of RN's VC).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,301 @@
...
+#import "SFApplicationHelper.h"
+#import "SFSecurityLockout.h"
+
+@interface SFSDKWindowManager()<SFSDKWindowContainerDelegate>
+
+@property (nonatomic, strong) NSHashTable *delegates;
+@property (nonatomic, strong,readonly) NSMapTable<NSString *,SFSDKWindowContainer *> * _Nonnull namedWindows;
+
+- (void)makeTransparentWithCompletion:(SFSDKWindowContainer *)window completion:(void (^)(void))completion;
+- (void)makeOpagueWithCompletion:(SFSDKWindowContainer *)window completion:(void (^)(void))completion;
```
wmathurin: Typo: should be `opaque` not `opague`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,301 @@
...
+        if ([delegate respondsToSelector:@selector(windowManager:didDisableWindow:)]){
+            [delegate windowManager:self didDisableWindow:window];
+        }
+    }];
+    if (completion)
+        completion();
+}
+
+- (void)makeOpagueWithCompletion:(SFSDKWindowContainer *)window completion:(void (^)(void))completion {
+    window.window.alpha = 1.0; //make Opague
```
wmathurin: So you make one window opaque and then loop through all window looking for a non-transparent window and make the first one found the key window. So if the app created other (non-transparent) windows, couldn't one of those be made the key window instead of the one you explicitly enabled?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,301 @@
...
+- (BOOL)isKeyboard:(UIWindow *) window {
+    return ([NSStringFromClass([window class]) hasPrefix:@"UIRemoteKeyboardWindow"]
+            || [NSStringFromClass([window class])hasPrefix:@"UITextEffectsWindow"]);
+}
+
+- (void)updateKeyWindow {
+    if ([self.snapshotWindow isEnabled])
+        return;
+    
+    for (NSInteger i = [SFApplicationHelper sharedApplication].windows.count - 1; i >= 0; i--) {
```
wmathurin: What order are the windows in?

libs/SmartStore/SmartStore/Classes/SFSmartStoreInspectorViewController.m
trooper2013: @wmathurin We could do so, the WindowManager has provisions for creating new windows.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
trooper2013: @wmathurin The idea here is to ensure that the Window with the highest Z-order level which is visible, is made the key window. Any app related windows will get control only when the SDK requirements are met.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
trooper2013: @wmathurin The list is in ascending order of levels.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,301 @@
...
+    } else {
+        [self makeTransparentWithCompletion:window completion:completion];
+    }
+    
+}
+
+
+#pragma mark - private methods
+- (void)makeTransparentWithCompletion:(SFSDKWindowContainer *)window completion:(void (^)(void))completion {
+    window.window.alpha = 0.0; //make Transparent
```
wmathurin: The original alpha value is not remembered. What if the app created a semi-transparent window? Instead we want to show one of our windows, we should make everything else transparent and remember what the key window was and what all the alpha values were. And then we we are done, we should revert the changes and get back to the state the application was in.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,301 @@
...
+#pragma mark - SFSDKWindowContainerDelegate
+- (void)windowEnable:(SFSDKWindowContainer *_Nonnull)window animated:(BOOL)animated withCompletion:(void (^)(void))completion{
+
+    [self enumerateDelegates:^(id<SFSDKWindowManagerDelegate> delegate) {
+        if ([delegate respondsToSelector:@selector(windowManager:willEnableWindow:)]){
+            [delegate windowManager:self willEnableWindow:window];
+        }
+    }];
+    
+    if (animated) {
```
wmathurin: Instead of just making the window we want to show opaque, could we also make all the other ones transparent. That way the window we are trying to enable, would be the one that ends up being shown/active.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -0,0 +1,301 @@
...
+        if ([delegate respondsToSelector:@selector(windowManager:didDisableWindow:)]){
+            [delegate windowManager:self didDisableWindow:window];
+        }
+    }];
+    if (completion)
+        completion();
+}
+
+- (void)makeOpaqueWithCompletion:(SFSDKWindowContainer *)window completion:(void (^)(void))completion {
+    window.window.alpha = 1.0; //make Opaque
```
wmathurin: Same comment as above. Could we remember the old value, and revert to that once we are done showing the window we wanted to show?

libs/SmartStore/SmartStore/Classes/SFSmartStoreInspectorViewController.m
wmathurin: But right now it looks like it knows of all the other windows and we don't want a SmartStore only window to be referenced in the SalesforceSDKCode library though.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
trooper2013: @wmathurin We do not tinker with main Applications Window and its visibility. Making all windows transparent implies changing the behavior of applications windows. It is quite possible that more than one window is visible within a application. Generally speaking,  this logic applies only to the windows the sdk creates i.e. the sdk windows are made visible or invisible when conditions are met.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
trooper2013: @wmathurin based on the previous explanation, the logic should be applied to only sdk windows. For us the requirement is clear, i.e. visible or not. As much as this code looks generic the alpha attributes are only applied to snapshot, passcode  and auth windows.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
wmathurin: updateKeyWindow loops through all the windows in the application and could end up making another window the key window (not the one we were trying to enable).

wmathurin: Maybe updateKeyWindow could be passed the window we are trying to enable and throw an error or a warning if the visible top window found is a different one?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
wmathurin: That's fine. Although I don't see any code making sure that we are not changing opacity on some other window.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -206,7 +206,7 @@ typedef void (^SFSnapshotViewControllerDismissalBlock)(UIViewController* snapsho
 
 /**
  The block to execute to present the snapshot viewcontroller.
- If this property is not set, SFRootViewManager will be used to present the snapshot.
+ If this property is not set, SFSDKWindowManager will be used to present the snapshot in the snpshot window.
```
bhariharan: Minor typo: `snapshot` instead of `snpshot`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -235,8 +231,8 @@ - (BOOL)launch
...
+        [[SFSDKWindowManager sharedManager] setMainUIWindow:[SFApplicationHelper sharedApplication].windows[0]];
```
bhariharan: Like Kevin's comment above, we probably want to change this to `[[SFSDKWindowManager sharedManager].mainWindow = ...`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -1002,12 +1002,23 @@ - (void)showAlertWithTitle:(nullable NSString *)title message:(nullable NSString
         
         [self.statusAlert addAction:cancelAction];
         dispatch_async(dispatch_get_main_queue(), ^{
-            [[SFRootViewManager sharedManager] pushViewController:weakSelf.statusAlert];
+            
+            SFSDKWindowManager.sharedManager.authWindow.viewController = [self blankViewController];
+            [SFSDKWindowManager.sharedManager.authWindow enable:YES withCompletion:^{
+                [SFSDKWindowManager.sharedManager.authWindow.viewController presentViewController:weakSelf.statusAlert animated:NO completion:^{
```
bhariharan: It looks like this completion block is empty. Can we replace it with `nil` instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -1395,7 +1408,13 @@ - (void)oauthCoordinator:(SFOAuthCoordinator *)coordinator displayAlertMessage:(
                                                           }];
     
     [alert addAction:defaultAction];
-    [[SFRootViewManager sharedManager] pushViewController:alert];
+    __weak typeof(self) weakSelf = self;
+    SFSDKWindowManager.sharedManager.authWindow.viewController = [self blankViewController];
+    [SFSDKWindowManager.sharedManager.authWindow enable:NO withCompletion:^{
+            [SFSDKWindowManager.sharedManager.authWindow.viewController presentViewController:weakSelf.statusAlert animated:NO completion:^{
```
bhariharan: It looks like this completion block is empty. Can we replace it with `nil` instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFAuthenticationManager.m
```diff
@@ -1416,8 +1435,14 @@ - (void)oauthCoordinator:(SFOAuthCoordinator *)coordinator displayConfirmationMe
                                                           }];
     [alert addAction:cancelAction];
 
-    [[SFRootViewManager sharedManager] pushViewController:alert];
-}
+    SFSDKWindowManager.sharedManager.authWindow.viewController = [self blankViewController];
+    [SFSDKWindowManager.sharedManager.authWindow enable:NO withCompletion:^{
+        [SFSDKWindowManager.sharedManager.authWindow.viewController presentViewController:alert animated:NO completion:^{
```
bhariharan: It looks like this completion block is empty. Can we replace it with `nil` instead?

libs/SalesforceSDKCore/SalesforceSDKCore/SFSDKWindowManagerTests.m
```diff
@@ -0,0 +1,212 @@
...
+//  SalesforceSDKCore
+//
```
bhariharan: Update copyright and move this file under the `SalesforceSDKCoreTests` folder.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
trooper2013: @bhariharan mainWindow is a wrapper (SFSDKContainer). setMainUIWindow is a convenience to set the UIWindow directly. They are not the same property.

bhariharan: Gotcha.

PR #2159, opened 2017-07-28
Pending requests should not be retried

bhariharan:
`shouldRetry` should be `NO` because this method is triggered only after successful session refresh.


bhariharan:
@khawkins @sgoldberg-sfdc

huminzhi:
## Danger has errored
```
 #  from DangerFile:5
 #  
 >  xcov.report(
 #     scheme: 'UnitTests',
```

</p>

PR #2165, opened 2017-08-09
Test querying null in smartstore

wmathurin:


wmathurin:
Just making sure it works as expected.

PR #2166, opened 2017-08-10
Update login failure to equate with Android

dborstelmann:
Login on Android does not automatically auth on failure.  This brings them together.  See https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/issues/33


bhariharan:
@wmathurin Do we need something similar in our `Cordova` plugin?

wmathurin:
@dborstelmann Thanks
@bhariharan Correct - I'll make the same change for hybrid (cordova based) apps.

PR #2167, opened 2017-08-10
Changing getAuthCredentials to behave like it does on Android

wmathurin:
Oauth cordova plugins's getAuthCredentials returns error "not authenticated" when not authenticated instead of triggering an authentication


wmathurin:
We might have to revisit [force.js](https://github.com/forcedotcom/SalesforceMobileSDK-Shared/blob/master/libs/force.js#L352)

bhariharan:
Could be because our template uses `force.login...` with `shouldAuthenticate = true` ;-)

huminzhi:
## Danger has errored
```
 #  from DangerFile:5
 #  
 >  xcov.report(
 #     scheme: 'UnitTests',
```

</p>

wmathurin:
Here is the related PR on the shared repo: https://github.com/forcedotcom/SalesforceMobileSDK-Shared/pull/449

PR #2170, opened 2017-08-11
Removing unused delegate

bhariharan:


bhariharan:
Forgot to commit some files.

bhariharan:
Good to go.

PR #2177, opened 2017-08-21
Init cursorCache to prevent getting nil cursor

booker0108:
Fixing the problem of inserting cursor to a nil object
No failure on inserting cursor to a nil object but lead to failure when trying to  manipulate cursor afterwards


wmathurin:
Why do we need this? Move cursor method calls cursor by id which creates the cursor cache map if needed.

booker0108:
I have a problem resulting in app crash with React Native if you don't init the cursorCache

1. smartstore.querySoup
2. smartstore.moveCursorToNextPage
 Then it crashed

huminzhi:
## Danger has errored
```
 #  from DangerFile:5
 #  
 >  xcov.report(
 #     scheme: 'UnitTests',
```

</p>

booker-laputa:
Will this be merged?

booker0108:
Will this pull request merge?

wmathurin:
Thanks @booker0108 !

PR #2179, opened 2017-08-22
Adding Login/Logout functions to the SFUserAccountManager

trooper2013:
In anticipation of the IDP authentication flow, this PR adds login/logout functions to the SFUserAccountManager. Here's whats done in this PR. 

1.  Created an SFSDKOAuthClient that is used by the SFUserAccountManager to handle all authentication functionality.   

2. SFSDKOAuthClient is meant to be an intermediary between User management and auth related functions. It wraps all request oriented state into a context object.  Allows for multiple auth requests to be handled in isolation (exceptions.. advanced auth). SFSDKOAuthClient adheres to the basic protocol and should be easily replaceable  or extended. 

T.B.D whether to deprecate SFAuthenticationManager or to wire the functionality through the SFUserAccountManager.


huminzhi:
## Danger has errored
[!] Invalid `DangerFile` file: No Junit file was found at test_output/report.junit
```
 #  from DangerFile:1
```

</p>

khawkins:
After a rough pass through the code, these are the questions I have out of the gate:

1. What's the value of moving authentication functions from one singleton (`SFAuthenticationManager`) to another (`SFUserAccountManager`)?
2. You say that "SFSDKOAuthClient is meant to be an intermediary between User management and auth related functions." Related to the question above, how is this different than what `SFAuthenticationManager` is already responsible for?

Looking through the code at first glance, I notice a couple of things:

1. `SFSDKOAuthClient` is managed as a property in the `SFUserAccountManager` singleton, effectively making it a singleton too.  This makes it hard to distinguish the value proposition over simply refactoring `SFAuthenticationManager` to do what you want to do. But see below for my additional thoughts on that, irrespective.
2. `SFSDKOAuthClient` is also tightly coupled to the `SFUserAccountManager` singleton, back-referencing it in many different places.  This seems like a recipe for state issues, and given that the idea is to decouple authentication from static state in the app, I'm not sure that we're achieving that goal.

I see the foundation to a successful effort of isolating authentication in a few key design features:

1. The base authentication process is _not_ a singleton, either explicitly or implicitly.  This allows any number of authentication processes to be in flight at once.
2. The process is self-contained.  I.e. you configure the parameters of an authentication request at initialization of the process/object, and that context is immutable from that point on.
3. The process has no _a priori_ parent or sibling dependencies.  E.g. no direct calls to `SFUserAccountManager`—it communicates solely through return values, callbacks, delegates, and notifications.
4. The responsibilities of the singleton (`SFUserAccountManager` or `SFAuthenticationManager` or whatever) are limited to managing the flow of allowable authentication processes in flight, given the business rules around standard auth vs. IDP auth, etc.

Anyway, those are my initial thoughts. Happy to discuss further.

sgoldberg-sfdc:
@khawkins, @bhariharan, and @trooper2013 Can we have a meeting to discuss what is driving these changes?

bhariharan:
@sgoldberg-sfdc Sure, but it's gotta be early in the morning or late in the evening (I'm working on IST this week).

trooper2013:
@khawkins  I agree with your assessment. I'm taking another pass at the changes to reflect some of your viewpoints.  @sgoldberg-sfdc The primary motivation for these changes are directly related to building out the IDP-flow functionality.  We can discuss these later today.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.h
```diff
@@ -0,0 +1,90 @@
...
+ value is determined by the SFDCOAuthClientIdPreference
+ configured via the settings bundle.
+ */
+@property (nonatomic, copy, nullable) NSString *oauthClientId;
+
+/** OAuth callback url to use for the OAuth login process.
+ Apps may customize this by setting this property before login.
+ By default this value is picked up from the main
+ bundle property SFDCOAuthRedirectUri
+ default: @"sfdc:///axm/detect/oauth/done")
```
bhariharan: The `default` specified here is wrong - should be our sample app's `callbackUrl`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.h
```diff
@@ -0,0 +1,361 @@
...
+/**
+ * An Auth window in which the auth flow will be presented
+ */
+- (SFSDKWindowContainer *)authWindow;
+//
+///**
+// * Retrieve cached context or create a new context
+// * @param credentials to use retrieve or create context
+// */
+//- (SFSDKOAuthClientContext *_Nullable)contextForCredentials:(SFOAuthCredentials *)credentials;
```
bhariharan: Remove commented out code.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
```diff
@@ -0,0 +1,791 @@
...
+
+@end
+
+@implementation SFSDKOAuthClient
+@synthesize authViewHandler = _authViewHandler;
+
+- (instancetype) init {
+    self = [super init];
+    if (self) {
+        //_currentRequestContexts = [SFSDKSafeMutableDictionary new];
```
bhariharan: Remove commented out code.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
```diff
@@ -0,0 +1,791 @@
...
+        SFAuthErrorHandler *currentHandler = (self.authErrorHandlerList.authHandlerArray)[i];
+        errorHandled = currentHandler.evalBlock(error, context.authInfo);
+        i++;
+    }
+
+    if (!errorHandled) {
+        // No error handlers could handle the error.  Pass through to the error blocks.
+        if (context.authInfo.authType == SFOAuthTypeUserAgent)
+            self.authViewHandler.authViewDismissBlock(self);
+        //[self execFailureBlocks];
```
bhariharan: Remove commented out code.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKMutableOAuthClientContext.h
```diff
@@ -0,0 +1,44 @@
...
+@interface SFSDKMutableOAuthClientContext : SFSDKOAuthClientContext
+
+@property (nonatomic, weak,readwrite,nullable) SFOAuthCredentials *credentials;
+@property (nonatomic, strong, readwrite,nullable) SFOAuthCoordinator *coordinator;
+@property (nonatomic, strong, readwrite,nullable) SFIdentityCoordinator *idCoordinator;
+@property (nonatomic, strong, readwrite,nullable) SFOAuthInfo *authInfo;
+@property (nonatomic, strong, readwrite,nullable) NSError *authError;
+@property (nonatomic, strong,readwrite,nullable) UIAlertController *statusAlert;
+
++ (instancetype _Nonnull)mutableContext:(SFSDKOAuthClientContext *_Nonnull)context;
```
khawkins: It would be more standard to implement `mutableCopy` on `SFSDKOAuthClientContext`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClientContext.h
```diff
@@ -0,0 +1,81 @@
...
+typedef void (^SFIdentityFailureCallbackBlock)(SFSDKOAuthClient *,NSError *);
+
+/** Object representing state of a current authentication context. Provides a means to isolate individual authentication requests
+ */
+@interface SFSDKOAuthClientContext : NSObject
+
+@property (nonatomic, weak,readonly) SFOAuthCredentials *credentials;
+@property (nonatomic, strong,readonly) UIAlertController *statusAlert;
+@property (nonatomic, strong, readonly) SFOAuthCoordinator *coordinator;
+@property (nonatomic, strong, readonly) SFIdentityCoordinator *idCoordinator;
```
khawkins: `statusAlert`, `coordinator`, and `idCoordinator` should not be part of the context.  These are operating classes, not data classes.  They should reasonably fall under the ownership of `SFSDKOAuthClient`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -104,111 +100,201 @@ - (void)dealloc {
...
+
+- (BOOL)loginWithCompletion:(SFUserAccountManagerSuccessCallbackBlock)completionBlock failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
+    SFOAuthCredentials *clientCredentials = [self retrieveClientCredentials];
+    return [self refreshCredentials:clientCredentials completion:completionBlock failure:failureBlock];
+}
+
+- (BOOL)refreshCredentials:(SFOAuthCredentials *)credentials completion:(SFUserAccountManagerSuccessCallbackBlock)completionBlock failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
+    
+    SFSDKOAuthClient *client = [SFSDKOAuthClient clientWithContextBlock:^(SFSDKOAuthClientConfig *config,SFSDKMutableOAuthClientContext *context) {
+        context.credentials = credentials;
```
khawkins: Given that this is the only item used in this builder pattern for `SFSDKOAuthClientContext`, I'd prefer that the constructor simply be amended to add this as a parameter.  E.g. this call would become:
```obj-c
SFSDKOAuthClient *client = [SFSDKOAuthClient clientCredentials:credentials configBlock:^(SFSDKOAuthClientConfig *config) {
    config.advancedAuthConfiguration = self.advancedAuthConfiguration;
    config.delegate = self;
    config.webViewDelegate = self;
    config.safariViewDelegate = self;
    config.successCallbackBlock = completionBlock;
    config.failureCallbackBlock = failureBlock;
}];
```
It would obviate the need to change `SFOAuthCoordinator` from its current paradigm of being configured up front with credentials—and indeed, `SFOAuthClient` should really just be a front-end to that class.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClientContext.m
```diff
@@ -0,0 +1,68 @@
...
+#import "SFSDKOAuthClient.h"
+#import "SFAuthErrorHandlerList.h"
+#import "SFAuthErrorHandler.h"
+#import "SFSDKOAuthViewHandler.h"
+
+@interface SFSDKOAuthClientContext()
+
+@property (nonatomic, assign) BOOL isAuthenticating;
+@property (nonatomic, weak,readwrite,nullable) SFOAuthCredentials *credentials;
+@property (nonatomic, weak,readwrite,nullable) id<SFSDKOAuthClientSafariViewDelegate> safariViewDelegate;
```
khawkins: There is a bunch of overlap between this private interface and the `SFSDKOAuthClientConfig` public interface.  Not clear what belongs where.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.h
```diff
@@ -0,0 +1,89 @@
...
+
+/**
+ The OAuth scopes associated with the app.
+ */
+@property (nonatomic, copy) NSSet<NSString*> *scopes;
+
+/**
+ @return The shared instance of this class.
+ */
++ (instancetype)sharedPreferences;
```
khawkins: Does this need to be a singleton?  It seems like the kind of thing that should instead just be owned by `SFUserAccountManager`, and referenced accordingly.

khawkins: Moreover, this is the kind of data that should be isolated to its context.  I.e. I would envision this being copied and passed into an `SFSDKOAuthClient` initialization method, so that it didn't compete with changes to the data _in situ_.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
```diff
@@ -0,0 +1,827 @@
...
++ (SFSDKOAuthClient *)webviewAuthInstance:(SFSDKOAuthClientConfig *)config {
+    if (self.clientProvider) {
+        return [self.clientProvider webviewAuthInstance:config];
+    }
+    return [[self alloc] initWithConfig:config];
+}
+
++ (SFSDKOAuthClient *)clientWithCredentials:(SFOAuthCredentials *)credentials configBlock:(void(^)(SFSDKOAuthClientConfig *))configBlock {
+
+    SFSDKOAuthClientConfig *config = [SFSDKOAuthClientConfig new];
```
khawkins: - `new` is kind of old—sorry for the pun. 😉  `alloc/init` is generally the convention these days.
- Se should be consistent, at any rate.  We're mixing both instantiation approaches in the same method.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClientContext.m
```diff
@@ -0,0 +1,61 @@
...
+ FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+#import "SFSDKOAuthClientContext.h"
+#import "SFIdentityData.h"
+#import "SFSDKOAuthClient.h"
```
khawkins: 1. This is circular.
2. It doesn't look like it's necessary?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -259,6 +345,17 @@ - (void)setAccountPersister:(id<SFUserAccountPersister>) persister {
     }
 }
 
+- (SFOAuthCredentials *)retrieveClientCredentials{
```
khawkins: Can we rename this?  It's misleading to say that it's retrieving credentials.  It's bootstrapping new credentials based on configuration.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -259,6 +345,17 @@ - (void)setAccountPersister:(id<SFUserAccountPersister>) persister {
...
+    NSString *identifier = self.oauthClientId;
+    SFOAuthCredentials *creds = [[SFOAuthCredentials alloc] initWithIdentifier:identifier clientId:self.oauthClientId encrypted:YES];
```
khawkins: The identifier passed to `SFOAuthCredentials` is supposed to be unique across the app instance.  It's not going to be here, as defined.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -259,6 +345,17 @@ - (void)setAccountPersister:(id<SFUserAccountPersister>) persister {
...
+    creds.redirectUri = self.oauthCompletionUrl;
+    creds.domain = self.loginHost;
+    creds.accessToken = nil;
+    creds.clientId = self.oauthClientId;
```
khawkins: This isn't necessary—it's set in the constructor.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -104,111 +99,202 @@ - (void)dealloc {
...
+    SFSDKOAuthClient *client = [SFSDKOAuthClient nativeBrowserAuthInstance:nil];
+    return [client handleURLAuthenticationResponse:appUrlResponse];
+}
+
+- (BOOL)loginWithCompletion:(SFUserAccountManagerSuccessCallbackBlock)completionBlock failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
+    SFOAuthCredentials *clientCredentials = [self retrieveClientCredentials];
+    return [self refreshCredentials:clientCredentials completion:completionBlock failure:failureBlock];
+}
+
+- (BOOL)refreshCredentials:(SFOAuthCredentials *)credentials completion:(SFUserAccountManagerSuccessCallbackBlock)completionBlock failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
```
khawkins: Can we rename this?  Given that it's ultimately called from every login method, the naming is confusing.  For a set of new credentials, it's not going to refresh the credentials.

Also, the analogous method in `SFAuthenticationManager` explicitly guarded against uninitialized credentials in this method, basically reserving it strictly for auth refresh scenarios.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
```diff
@@ -0,0 +1,827 @@
...
+static NSString * const kAlertDismissButtonKey = @"authAlertDismissButton";
+static NSString * const kAlertConnectionErrorFormatStringKey = @"authAlertConnectionErrorFormatString";
+static NSString * const kAlertVersionMismatchErrorKey = @"authAlertVersionMismatchError";
+
+static id<SFSDKOAuthClientProvider> _clientProvider = nil;
+
+
+@interface SFSDKOAuthClient()<SFOAuthCoordinatorDelegate,SFIdentityCoordinatorDelegate,SFSDKLoginHostDelegate,SFLoginViewControllerDelegate>
+
+@property (nonatomic, strong) SFSDKOAuthClientContext *context;
```
khawkins: This should be a `copy` property.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClientContext.h
```diff
@@ -0,0 +1,87 @@
...
+ */
+typedef void (^SFIdentitySuccessCallbackBlock)(SFSDKOAuthClient *client);
+/**
+ Callback block definition for Identity retrieval failure callback.
+ */
+typedef void (^SFIdentityFailureCallbackBlock)(SFSDKOAuthClient *,NSError *);
+
+/** Object representing state of a current authentication context. Provides a means to isolate individual authentication requests
+ */
+@interface SFSDKOAuthClientContext : NSObject
```
khawkins: This should support `NSCopying`, and its mutable subclass should implement `copyWithZone:` to return an immutable copy.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -104,111 +99,202 @@ - (void)dealloc {
...
+    return [self refreshCredentials:credentials
+                                  completion:completionBlock
+                                     failure:failureBlock];
+}
+
+- (void)logout {
+    [self logoutUser:[SFUserAccountManager sharedInstance].currentUser];
+}
+
+- (void)logoutUser:(SFUserAccount *)user {
```
khawkins: It looks like the guardrails on multiple logout calls were removed from this porting.  See [this PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1989).  This will likely result in regressions when these new interfaces are consumed.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.m
```diff
@@ -0,0 +1,159 @@
...
+    if (nil == host) {
+        [[NSUserDefaults msdkUserDefaults] removeObjectForKey:kSFUserAccountOAuthLoginHost];
+    } else {
+        [[NSUserDefaults msdkUserDefaults] setObject:host forKey:kSFUserAccountOAuthLoginHost];
+    }
+    
+    [[NSUserDefaults msdkUserDefaults] synchronize];
+    
+    // Only post the login host change notification if the host actually changed.
+    if ((oldLoginHost || host) && ![host isEqualToString:oldLoginHost]) {
```
bhariharan: As discussed offline, I don't think we should change the login host on the IDP app and/or post a notification for the same. We should simply use the login host value passed in from the SP to authenticate.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.h
```diff
@@ -0,0 +1,284 @@
...
+ Called when error occurs offline.
+ @param client The instance of SFSDKOAuthClient making the call.
+ */
+- (void)authClientContinueOfflineMode:(SFSDKOAuthClient *)client;
+
+/**
+ Called when a generic flow authentication is cancelled.
+ @param client The instance of SFSDKOAuthClient making the call.
+ */
+- (void)authClientDidChangeLoginHost:(SFSDKOAuthClient *)client loginHost:(NSString *)newLoginHost;
```
bhariharan: Do we need this? This is probably handled in the notification mechanism of the persistence layer.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.h
```diff
@@ -0,0 +1,284 @@
...
+
+/**
+ * Refresh credentials request.
+ */
+- (BOOL)refreshCredentials;
+
+/**
+ * Refresh credentials request.
+ */
+- (BOOL)refreshCredentials:(SFOAuthCredentials *)credentials;
```
bhariharan: Should we keep this? Doesn't the client have credentials built in?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
```diff
@@ -0,0 +1,864 @@
...
+    return self.idCoordinator.idData;
+}
+
+- (SFSDKOAuthViewHandler *)authViewHandler {
+
+    if (!self.config.authViewHandler) {
+        [readWriteLock lock];
+        __weak typeof(self) weakSelf = self;
+
+        if (self.config.advancedAuthConfiguration == SFOAuthAdvancedAuthConfigurationNone) {
```
bhariharan: This should be:
```
if (self.config.advancedAuthConfiguration == SFOAuthAdvancedAuthConfigurationRequire) {
<stuff in else block>
} else {
<stuff in if block>
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
```diff
@@ -0,0 +1,864 @@
...
+}
+
+
+#pragma mark - private methods
+- (void)revokeRefreshToken:(SFOAuthCredentials *)credentials
+{
+    if (credentials.refreshToken != nil) {
+         NSMutableString *host = [NSMutableString stringWithFormat:@"%@://", credentials.protocol];
+        [host appendString:credentials.domain];
+        [host appendString:@"/services/oauth2/revoke?token="];
```
bhariharan: Maybe extract this string to a constant.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.m
trooper2013: @bhariharan Good point. I will ensure that IDP flow handles the situation correctly. On the same note we have to ensure that the login screens for IDP APP  have the settings icon disabled in this context.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.h
trooper2013: @bhariharan This is required for the IDP client to notify its delegate of change of host that occurred. Should result in a re-login flow. The client doesn't directly make the change to the persistent store.

bhariharan:
LGTM! 👍

PR #2226, opened 2017-10-04
Bug in cookie setter

bhariharan:
We were setting `sid` as the domain instead of the actual domain itself.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKWebViewStateManager.m
```diff
@@ -36,7 +36,10 @@ @implementation SFSDKWebViewStateManager
 + (void)resetSessionWithNewAccessToken:(NSString *)accessToken isSecureProtocol:(BOOL)isSecure {
      //reset UIWebView related state if any
     [self removeUIWebViewCookies:@[SID_COOKIE] fromDomains:self.domains];
-    [self addSidCookieForDomain:SID_COOKIE withAccessToken:accessToken isSecureProtocol:isSecure];
+    NSAssert(self.domains != nil && [self.domains count] > 0, ERR_NO_DOMAIN_NAMES);
```
khawkins: This is probably not necessary, since `self.domains` is just a convenience to a method defined below, which hard-codes the domains we care about.

bhariharan:
@khawkins Done.

PR #2230, opened 2017-10-06
Expanding functionality for hybrid URLs and deferred authentication

khawkins:
Based on the 2x2 matrix we described:

| Deferred Login | Has Relative URL                       | Has Absolute URL                |
| -------------- | -------------------------------------- | ------------------------------- |
| No             | Require login and build abs URL        | Error since this is a misconfig |
| Yes            | Error since we can't build the abs URL | Load the abs URL w/o auth       |

This does sanity checking on the values, then updates the workflows mostly to support absolute URLs without authentication, which weren't working properly before.


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -216,6 +218,14 @@ - (void)viewDidLoad
     NSString *hybridViewUserAgentString = [self sfHybridViewUserAgentString];
     [SFSDKWebUtils configureUserAgent:hybridViewUserAgentString];
     self.baseUserAgent = hybridViewUserAgentString;
+    
+    // Run a basic validation of the app config inputs.
+    SFSDKAppConfigValidationResult *validationResult = [_hybridViewConfig validate];
```
khawkins: The one oddity in this approach is that, when authentication is set to `true`, auth happens before any of the hybrid view takes off.  The net result is that you can authenticate, then end up in a validation failure because e.g. you specified an absolute URL for your start page.

I'm not sure how to get around that without a more significant refactor.  One approach could be to manage the validation step in `SalesforceSDKManager` itself.

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          2 Errors
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Code coverage under minimum of 55%</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Needs testing on a Phone if change is non-trivial</td>
    </tr>
  </tbody>
</table>

### Tests: 

Classname | Name|
--- | ---|
SFSDKAuthRequestCommandTest | testSFSDKAuthRequestCommandWithParameters|
SFSDKAuthRequestCommandTest | testSFSDKAuthRequestCommandWithParameters|
SFSDKAuthRequestCommandTest | testSFSDKAuthRequestCommandWithParameters|
SFUserAccountManagerTests | testLogin|
SFUserAccountManagerTests | testLoginNotificationPosted|
SalesforceSDKManagerTests | testBrandedLoginPathInAuthManager|
SalesforceSDKManagerTests | testBrandedLoginPathInAuthManagerAndAuthorizeEndpoint|
SalesforceSDKManagerTests | testBrandedLoginPathInAuthManagerAndAuthorizeEndpoint|


## Current coverage for SalesforceAnalytics.framework is `43.82%`

---
## Current coverage for SalesforceSDKCore.framework is `48.85%`
Files changed | - | - 
--- | --- | ---
SFSDKAppConfigValidationResult.m | `0.00%` | :skull:
SFSDKAppConfig.m | `0.00%` | :skull:

---

</p>

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewConfig.m
```diff
@@ -119,6 +125,37 @@ - (void)setAttemptOfflineLoad:(BOOL)attemptOfflineLoad
...
+        return [[SFSDKAppConfigValidationResult alloc] initWithValidationSucceeded:NO message:@"No start page configured."];
+    }
+    
+    // Sanity check local URLs against absolute URL values.
+    if (self.isLocal && self.startPageIsAbsoluteUrl) {
+        return [[SFSDKAppConfigValidationResult alloc] initWithValidationSucceeded:NO message:@"Local start page should not be absolute URL."];
+    }
+    
+    // Start page makeup for remote apps beyond this point is subject to conditional configuration.
+    if (!self.isLocal) {
```
wmathurin: Same comment I did for Android.
Would the rules be more obvious if we wrote it this way
```obj-c
if (!self.isLocal && self.shouldAuthenticate && self.startPageIsAbsoluteUrl) {
    return [[SFSDKAppConfigValidationResult alloc] initWithValidationSucceeded:NO message:@"Cannot authenticate using an absolute URL for start page."];
}
if (!self.isLocal && !self.shouldAuthenticate && !self.startPageIsAbsoluteUrl) {
    return [[SFSDKAppConfigValidationResult alloc] initWithValidationSucceeded:NO message:@"Cannot configure relative URL start page with deferred authentication."];
}
```

khawkins: I'm fine with that.

khawkins:
@wmathurin @bhariharan Mirroring a couple of questions I had on Android:

1. Similar to the throw/catch question, it seems like the iOS `validate` side should follow a more standard iOS design pattern of returning a `BOOL`, and providing an `NSError **` reference parameter, obviating the need for a custom state object.
2. Thinking about it more, I think we _should_ move the configuration validation up in the process.  Indeed, `SalesforceSDKManager` already goes through some explicit validation of the configuration, after it's parsed.  We could add that basic validation to the base `SFSDKAppConfig` class, and then just call `validate` for any app config that's passed into `SalesforceSDKManager`, prior to launch.  In the iOS case, that gets rid of the issue about authenticating before validating in `SFHybridViewController`.

Thoughts?

wmathurin:
I agree with 1 (having a method that returns BOOL and take a NSError*) and 2 (move validation to SalesforceSDKManager).
Loosely related question: on Android we have a bootconfig.xml for native applications (and BootConfig.java can be initialized from it), but on iOS, we simply define the oauth key and callback url in the AppDelegate. Should we have a bootconfig.plist on iOS and move the setttings there?

bhariharan:
I agree with (1) as well. And I agree with Wolf's comment about `bootconfig.plist`. We've spoken about it forever, but never done it for want of time. `6.0` is the right vehicle for this change.

khawkins:
We have the conceptual plist, in the form of the ability to stage SalesforceSDKManager with an app config object.  Right now it's optional.  We could certainly standardize on it, and allow for a default configuration file, similar to what we do with hybrid.  I'll look at that approach.  Thanks for the feedback!  Stay tuned.

khawkins:
Can you guys take a look at my latest updates?  I'll annotate, but at a high level:

- Added a native bootconfig plist (you may want the name to change).
- Moved config validation into SalesforceSDKManager.
- Deprecated the old properties that handled configuration values.
- Moved to an `NSError` approach for config validation.

There will be more to do on making sure apps and templates, etc., get the new changes.  But I wanted to validate the approach first.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewConfig.h
```diff
@@ -25,8 +25,18 @@
...
+    SFSDKHybridAppConfigErrorCodeNoStartPage = 1066,
+    SFSDKHybridAppConfigErrorCodeLocalPageAbsoluteURL,
+    SFSDKHybridAppConfigErrorCodeAbsoluteURLNoAuth,
+    SFSDKHybridAppConfigErrorCodeRelativeURLAuth
+};
+
 NS_ASSUME_NONNULL_BEGIN
 
+// Default path to bootconfig.json on the filesystem.
+static NSString *const SFSDKDefaultHybridAppConfigFilePath = @"/www/bootconfig.json";
```
khawkins: Moved these values out to the header file for consumer visibility.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -238,22 +245,22 @@ - (BOOL)isLaunching
 
 - (NSString *)connectedAppId
 {
-    return [SFUserAccountManager sharedInstance].oauthClientId;
+    return self.appConfig.remoteAccessConsumerKey;
```
khawkins: These deprecate properties now point directly into the app config, for backward compatibility.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -350,9 +360,10 @@ - (BOOL)validateLaunchState:(NSError **)launchStateError
     BOOL validInputs = YES;
     NSMutableArray *launchStateErrorMessages = [NSMutableArray array];
     
-    // If an app config has been specified, set values from that first.
-    if (self.appConfig != nil) {
-        [self configureWithAppConfig];
+    // Attempt to load a default configuration, if one has not been configured.
+    if (_appConfig == nil) {
```
khawkins: This allows for the default configuration to be loaded, so apps don't have to specify that explicitly in code, if they have those file configurations.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -364,22 +375,12 @@ - (BOOL)validateLaunchState:(NSError **)launchStateError
...
-        [launchStateErrorMessages addObject:noCallbackUriError];
-        validInputs = NO;
-    }
-    if ([self.authScopes count] == 0) {
-        NSString *noAuthScopesError = @"No auth scopes set.  Cannot continue.";
-        [SFSDKCoreLogger e:[self class] format:noAuthScopesError];
-        [launchStateErrorMessages addObject:noAuthScopesError];
+    
+    NSError *appConfigError = nil;
+    BOOL appConfigValidated = [self.appConfig validate:&appConfigError];
```
khawkins: All app config (native and hybrid) happens as part of `launch`, now.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/Classes/AppDelegate.m
```diff
@@ -48,9 +48,6 @@ - (id)init
     self = [super init];
     if (self) {
         [SFAuthenticationManager sharedManager].advancedAuthConfiguration = SFOAuthAdvancedAuthConfigurationRequire;
-        [SalesforceSDKManager sharedManager].connectedAppId = RemoteAccessConsumerKey;
```
khawkins: Example code for RestAPIExplorer.  Explicit configuration in code no longer necessary.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/MobileSDKAppConfig.plist
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
khawkins: Forgot about the `bootconfig.plist` name.  Would you prefer I change it to that?  Picked this out of a hat. :)

shared/hybrid/AppDelegate+SalesforceHybridSDK.m
```diff
@@ -48,10 +48,8 @@ + (void)load
 
 - (AppDelegate *)sfsdk_swizzled_init
 {
-    SFHybridViewConfig *appConfig = [SFHybridViewConfig fromDefaultConfigFile];
```
khawkins: This all happens in `SalesforceSDKManager` now.  No need for explicit code.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKAppConfig.h
```diff
@@ -24,8 +24,18 @@
...
+
+typedef NS_ENUM(NSInteger, SFSDKAppConfigErrorCode) {
+    SFSDKAppConfigErrorCodeNoConsumerKey = 966,
+    SFSDKAppConfigErrorCodeNoRedirectURI,
+    SFSDKAppConfigErrorCodeNoOAuthScopes
+};
+
 NS_ASSUME_NONNULL_BEGIN
 
+static NSString *const SFSDKDefaultNativeAppConfigFilePath = @"/MobileSDKAppConfig.plist";
```
bhariharan: Could we call this `bootconfig.plist` for consistency?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -141,17 +141,17 @@ typedef void (^SFSnapshotViewControllerDismissalBlock)(UIViewController* snapsho
 /**
  The Connected App ID configured for this application.
  */
-@property (nonatomic, copy, nullable) NSString *connectedAppId;
+@property (nonatomic, copy, nullable) NSString *connectedAppId SFSDK_DEPRECATED(6.0, 7.0, "Use appConfig.remoteAccessConsumerKey to specify the consumer key.");
```
bhariharan: Could we remove these in `6.0`?

native/SampleApps/RestAPIExplorer/RestAPIExplorer/MobileSDKAppConfig.plist
bhariharan: Yes please :-)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
khawkins: We could, I suppose.  We never deprecated them though.  Happy to do whatever the team wants.

bhariharan: Makes sense.

khawkins:
@bhariharan That last batch is just cleanup of the various property areas in the rest of the SDK, and some of the test functionality that was affected.  And tests spotted a bug in my SDK manager implementation. :)

bhariharan:
👍

khawkins:
retest this please

khawkins:
test me

PR #2249, opened 2017-10-25
Improve fastlane PR unit test logic.  Remove bad check in danger.

brandonpage:
- Fail check if PR is against any branch except dev
- Remove Code Coverage check as PR's can be run against only a few schemes
- Test all schemes if code in SalesforceSDK library is changed


brandonpage:
test me

brandonpage:
test me

PR #2263, opened 2017-11-02
Updated user agent to include web user agent

bjpang:
This update is to append a web user agent to the default user agent making it more robust. Currently it uses the UIWebView because the WKWebView evaluates Javascript asynchronously, which would probably require us to use a semaphore and wait on the value. So for now we are using a hack-y method with UIWebView.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
@@ -893,7 +893,7 @@ static NSString* ailtnAppName = nil;
         // App type.
         NSString *appTypeStr = [self getAppTypeAsString];
         NSString *myUserAgent = [NSString stringWithFormat:
-                                 @"SalesforceMobileSDK/%@ %@/%@ (%@) %@/%@ %@%@ uid_%@ ftr_%@",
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -902,12 +903,38 @@ - (SFSDKUserAgentCreationBlock)defaultUserAgentString {
                                  appTypeStr,
                                  (qualifier != nil ? qualifier : @""),
                                  uid,
-                                 [[[SFSDKAppFeatureMarkers appFeatures].allObjects sortedArrayUsingSelector:@selector(caseInsensitiveCompare:)] componentsJoinedByString:@"."]
+                                 [[[SFSDKAppFeatureMarkers appFeatures].allObjects sortedArrayUsingSelector:@selector(caseInsensitiveCompare:)] componentsJoinedByString:@"."],
```
bhariharan: @bjpang You need to add another space separated `%@` for the new field in line 896 for this to work. Also remove the trailing comma after `webViewUserAgent`.

bhariharan: You can push a commit with the changes - you don't have to overwrite commit history.

bjpang: Ooh nice catch. I've updated it now.

PR #2268, opened 2017-11-08
Cleaning up frontDoor method, fixing use cases

khawkins:
Was running into path construction aberrations, and stayed to clean up some of the naming conventions as well.

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
@@ -404,6 +404,11 @@ SFSDK_USE_DEPRECATED_BEGIN
 
 - (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrlString returnUrlIsEncoded:(BOOL)isEncoded createAbsUrl:(BOOL)createAbsUrl
 {
+    // Special case: if returnUrlString itself is a frontdoor.jsp URL, parse its parameters and rebuild.
+    if ([returnUrlString containsString:@"frontdoor.jsp"]) {
+        return [self parseFrontDoorReturnUrlString:returnUrlString encoded:isEncoded];
+    }
+    
     SFOAuthCredentials *creds = [SFUserAccountManager sharedInstance].currentUser.credentials;
     NSURL *instUrl = creds.apiUrl;
     NSString *fullReturnUrlString = returnUrlString;
@@ -413,23 +418,34 @@ SFSDK_USE_DEPRECATED_BEGIN
      * other cases, because of differences between instance URL and community URL.
      */
     if (createAbsUrl && ![returnUrlString hasPrefix:@"http"]) {
-        fullReturnUrlString = [instUrl URLByAppendingPathComponent:returnUrlString].absoluteString;
+        NSURLComponents *retUrlComponents = [NSURLComponents componentsWithURL:instUrl resolvingAgainstBaseURL:NO];
+        retUrlComponents.path = [retUrlComponents.path stringByAppendingPathComponent:returnUrlString];
+        fullReturnUrlString = retUrlComponents.string;
     }
-    if ([returnUrlString containsString:@"frontdoor.jsp"]) {
-        NSRange r1 = [returnUrlString rangeOfString: isEncoded ? @"retURL%3D" : @"retURL="];
-        NSRange r2 = [returnUrlString rangeOfString: isEncoded ? @"%26display" : @"&display"];
-        NSRange range = NSMakeRange(r1.location + r1.length, r2.location - r1.location - r1.length);
-        NSString *newReturnUrlString = [returnUrlString substringWithRange: range];
-        if(isEncoded) newReturnUrlString = [newReturnUrlString stringByRemovingPercentEncoding];
-        [SFSDKHybridLogger d:[self class] format:[NSString stringWithFormat:@"%@", newReturnUrlString]];
-        return [self frontDoorUrlWithReturnUrl:newReturnUrlString returnUrlIsEncoded:YES createAbsUrl:NO];
-    }
-    NSString *encodedUrlString = (isEncoded ? fullReturnUrlString : [fullReturnUrlString stringByURLEncoding]);
+    
+    // Create frontDoor path based on credentials API URL.
+    NSURLComponents *frontDoorUrlComponents = [NSURLComponents componentsWithURL:instUrl resolvingAgainstBaseURL:NO];
+    frontDoorUrlComponents.path = [frontDoorUrlComponents.path stringByAppendingPathComponent:@"/secur/frontdoor.jsp"];
+    
+    // NB: We're not using NSURLComponents.queryItems here, because it unsufficiently encodes query params.
+    NSMutableString *frontDoorUrlString = [NSMutableString stringWithString:frontDoorUrlComponents.string];
+    NSString *encodedRetUrlValue = (isEncoded ? fullReturnUrlString : [fullReturnUrlString stringByURLEncoding]);
     NSString *encodedSidValue = [creds.accessToken stringByURLEncoding];
-    NSURL *frontDoorUrl = [instUrl URLByAppendingPathComponent:@"/secur/frontdoor.jsp"];
-    NSMutableString *frontDoorUrlString = [NSMutableString stringWithString:frontDoorUrl.absoluteString];
-    [frontDoorUrlString appendFormat:@"?sid=%@&retURL=%@&display=touch", encodedSidValue, encodedUrlString];
+    [frontDoorUrlString appendFormat:@"?sid=%@&retURL=%@&display=touch", encodedSidValue, encodedRetUrlValue];
+    
     return [NSURL URLWithString:frontDoorUrlString];
+}
+
+- (NSURL *)parseFrontDoorReturnUrlString:(NSString *)frontDoorUrlString encoded:(BOOL)encoded {
+    NSRange r1 = [frontDoorUrlString rangeOfString: encoded ? @"retURL%3D" : @"retURL="];
+    NSRange r2 = [frontDoorUrlString rangeOfString: encoded ? @"%26display" : @"&display"];
+    NSRange range = NSMakeRange(r1.location + r1.length, r2.location - r1.location - r1.length);
+    NSString *returnUrlString = [frontDoorUrlString substringWithRange: range];
+    if (encoded) {
+        returnUrlString = [returnUrlString stringByRemovingPercentEncoding];
+    }
+    [SFSDKHybridLogger d:[self class] format:@"%@ Extracted return URL string '%@' from original frontDoor URL '%@'", NSStringFromSelector(_cmd), returnUrlString, frontDoorUrlString];
+    return [self frontDoorUrlWithReturnUrl:returnUrlString returnUrlIsEncoded:YES createAbsUrl:NO];
 }
 
 - (NSString *)isLoginRedirectUrl:(NSURL *)url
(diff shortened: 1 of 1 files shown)


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -402,36 +402,34 @@ - (NSString *)sfHybridViewUserAgentString
     return userAgentString;
 }
 
-- (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrl returnUrlIsEncoded:(BOOL)isEncoded createAbsUrl:(BOOL)createAbsUrl
+- (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrlString returnUrlIsEncoded:(BOOL)isEncoded createAbsUrl:(BOOL)createAbsUrl
```
khawkins: We're using URLs and Strings interchangeably throughout this method, and it gets confusing in the naming conventions of the variables, knowing which one's a string and which one's a URL.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -402,36 +402,34 @@ - (NSString *)sfHybridViewUserAgentString
...
+    NSString *fullReturnUrlString = returnUrlString;
+    
     /*
      * We need to use the absolute URL in some cases and relative URL in some
      * other cases, because of differences between instance URL and community URL.
      */
-    if (createAbsUrl && ![returnUrl hasPrefix:@"http"]) {
-        fullReturnUrl = [NSString stringWithFormat:@"%@/%@", instUrl, returnUrl];
+    if (createAbsUrl && ![returnUrlString hasPrefix:@"http"]) {
+        fullReturnUrlString = [instUrl URLByAppendingPathComponent:returnUrlString].absoluteString;
```
khawkins: The former version was subject to blindly creating URLs like `http://example.com/root//subtree`.  I pulled back to a standard convention for appending paths.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -402,36 +402,34 @@ - (NSString *)sfHybridViewUserAgentString
...
+        NSString *newReturnUrlString = [returnUrlString substringWithRange: range];
+        if(isEncoded) newReturnUrlString = [newReturnUrlString stringByRemovingPercentEncoding];
+        [SFSDKHybridLogger d:[self class] format:[NSString stringWithFormat:@"%@", newReturnUrlString]];
+        return [self frontDoorUrlWithReturnUrl:newReturnUrlString returnUrlIsEncoded:YES createAbsUrl:NO];
     }
+    NSString *encodedUrlString = (isEncoded ? fullReturnUrlString : [fullReturnUrlString stringByURLEncoding]);
     NSString *encodedSidValue = [creds.accessToken stringByURLEncoding];
-    [frontDoorUrl appendFormat:@"secur/frontdoor.jsp?sid=%@&retURL=%@&display=touch", encodedSidValue, encodedUrl];
-    return [NSURL URLWithString:frontDoorUrl];
+    NSURL *frontDoorUrl = [instUrl URLByAppendingPathComponent:@"/secur/frontdoor.jsp"];
```
khawkins: Just another spot where I cleaned up the path creation to a standard methodology, and the surrounding functionality to support it.  The rest of the changes in this method were just about renaming variables to clearly state their data type.

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
    </tr>
  </tbody>
</table>

### Tests: 

Classname | Name|
--- | ---|
SmartSyncTestSuite | testFetchSObjects|
SmartSyncTestSuite | testSyncSObjectDetectConflictDelete|


## Current coverage for SalesforceAnalytics.framework is `42.94%`

---
## Current coverage for SalesforceHybridSDK.framework is `54.11%`
Files changed | - | - 
--- | --- | ---
SFHybridViewController.m | `23.57%` | :skull:

---
## Current coverage for SalesforceSDKCore.framework is `25.56%`

---
## Current coverage for SmartStore.framework is `51.90%`

---
## Current coverage for SmartSync.framework is `29.94%`

---

</p>

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
bhariharan: `URLByAppendingPathComponent` has some issues with base URLs that have more than one path component such as custom community URLs. We used to do this in the past and I changed it to string appending to fix those bugs.

bhariharan: Ideally, it would be nice if we could build up a string all the way through in this method.

khawkins: OMG that's crazy stupid, to the point of not even understanding how my change actually worked in my use case.  I created a test to go through various scenarios, and `URLByAppendingPathComponent` is nearly useless. 🙁

Okay, let me take a look at "manually" bullet proofing the scenario I was seeing.  I'll also look based on that and see if the structure of the builder functionality can be cleaned up a bit.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -402,36 +402,34 @@ - (NSString *)sfHybridViewUserAgentString
...
      */
-    if (createAbsUrl && ![returnUrl hasPrefix:@"http"]) {
-        fullReturnUrl = [NSString stringWithFormat:@"%@/%@", instUrl, returnUrl];
+    if (createAbsUrl && ![returnUrlString hasPrefix:@"http"]) {
+        fullReturnUrlString = [instUrl URLByAppendingPathComponent:returnUrlString].absoluteString;
     }
-    if([returnUrl containsString:@"frontdoor.jsp"]) {
-        NSRange r1 = [returnUrl rangeOfString: isEncoded ? @"retURL%3D" : @"retURL="];
-        NSRange r2 = [returnUrl rangeOfString: isEncoded ? @"%26display" : @"&display"];
+    if ([returnUrlString containsString:@"frontdoor.jsp"]) {
```
khawkins: In looking through some revamp of this method, is this block of functionality something we can revisit, or remove?  I don't really understand how the [original issue](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/1515) (fix [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/1519)) happens—and it seems like an edge case that should be handled in some other way, or is surfacing some other bug.  How do you end up with a return URL that itself is frontdoor.jsp?  That just shouldn't happen.

@bhariharan do you remember anything about this issue?

bhariharan: Unfortunately, I don't remember the exact use case. However, if I were to guess, I think it would have something to do with custom logout flows that are configurable in the community context. There are some use cases where this could potentially happen. There's probably no clean way to do this unless we get a first class native API that we can hit for session validity and other related functionality.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -402,36 +402,50 @@ - (NSString *)sfHybridViewUserAgentString
     return userAgentString;
 }
 
-- (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrl returnUrlIsEncoded:(BOOL)isEncoded createAbsUrl:(BOOL)createAbsUrl
+- (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrlString returnUrlIsEncoded:(BOOL)isEncoded createAbsUrl:(BOOL)createAbsUrl
 {
+    // Special case: if returnUrlString itself is a frontdoor.jsp URL, parse its parameters and rebuild.
+    if ([returnUrlString containsString:@"frontdoor.jsp"]) {
```
khawkins: Moved this to the top, and farmed it out to a helper method, since it's a self-contained code block that otherwise doesn't apply here.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -402,36 +402,50 @@ - (NSString *)sfHybridViewUserAgentString
...
-        if(isEncoded) newReturnUrl = [newReturnUrl stringByRemovingPercentEncoding];
-        [SFSDKHybridLogger d:[self class] format:[NSString stringWithFormat:@"%@", newReturnUrl]];
-        return [self frontDoorUrlWithReturnUrl: newReturnUrl returnUrlIsEncoded:TRUE createAbsUrl: FALSE];
-    }
-    NSString *encodedUrl = (isEncoded ? fullReturnUrl : [fullReturnUrl stringByURLEncoding]);
-    NSMutableString *frontDoorUrl = [NSMutableString stringWithString:instUrl];
-    if (![frontDoorUrl hasSuffix:@"/"]) {
-        [frontDoorUrl appendString:@"/"];
+    if (createAbsUrl && ![returnUrlString hasPrefix:@"http"]) {
+        NSURLComponents *retUrlComponents = [NSURLComponents componentsWithURL:instUrl resolvingAgainstBaseURL:NO];
```
khawkins: Ended up utilizing two principles:

1. `NSURLComponents` is (mostly) a good way to manipulate URLs.
2. `[NSString stringByAppendingPathComponent:]` is actually pretty intelligent about marrying path strings together—but you have to work exclusively on the path, because it doesn't understand (and thus messes with) e.g. `https://`

bhariharan: What happens here if `instUrl` itself contains a path component? For example, `instUrl = https://mycommunity.salesforce.com/community1` and `retUrl = <some URL>`. When we use `NSURLComponents` to break it up, would this still work? That was the reason we replaced it with `[NSString stringWithFormat:...` if I'm not mistaken.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -402,36 +402,50 @@ - (NSString *)sfHybridViewUserAgentString
...
+        retUrlComponents.path = [retUrlComponents.path stringByAppendingPathComponent:returnUrlString];
+        fullReturnUrlString = retUrlComponents.string;
     }
+    
+    // Create frontDoor path based on credentials API URL.
+    NSURLComponents *frontDoorUrlComponents = [NSURLComponents componentsWithURL:instUrl resolvingAgainstBaseURL:NO];
+    frontDoorUrlComponents.path = [frontDoorUrlComponents.path stringByAppendingPathComponent:@"/secur/frontdoor.jsp"];
+    
+    // NB: We're not using NSURLComponents.queryItems here, because it unsufficiently encodes query params.
```
khawkins: This is where `NSURLComponents` (and basically most of Cocoa's built-in URL encoding and decoding functionality) doesn't fare so well—so we depart for manual intervention here.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
khawkins: > What happens here if `instUrl` itself contains a path component?

`NSURLComponents` breaks a URL into all of its constituent parts, on import.  So in that case, `retUrlComponents.path` will be `/community1`.

> and `retUrl = <some URL>`

In this conditional, `retUrlString` is determined to be a path component (as opposed to an absolute URL).

And then as I mentioned, we just do manipulation on the paths themselves, then regurgitate the `NSURLComponents` object as a string at the end of the process.

bhariharan: Sounds good. 👍

PR #2293, opened 2017-11-30
Encrypt db if found unencrypted (5.3 bug fix)

wmathurin:


brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Warning">
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">Please provide a summary in the Pull Request description</td>
    </tr>
  </tbody>
</table>

## Current coverage for SalesforceAnalytics.framework is `43.55%`

---
## Current coverage for SalesforceSDKCore.framework is `15.63%`

---
## Current coverage for SmartStore.framework is `61.84%`
Files changed | - | - 
--- | --- | ---
SFSmartStoreDatabaseManager.m | `66.90%` | :warning:

---

</p>

libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
```diff
@@ -140,7 +140,25 @@ - (FMDatabase *)openStoreDatabaseWithName:(NSString *)storeName key:(NSString *)
...
+// This method checks for that situation and encrypt the database if needed
+- (void)fixFor530Bug:(NSString *)storeName key:(NSString *)key {
+    NSString *fullDbFilePath = [self fullDbFilePathForStoreName:storeName];
+    
+    __block BOOL needEncrypting = NO;
+    [[FMDatabaseQueue databaseQueueWithPath:fullDbFilePath] inDatabase:^(FMDatabase* db) {
+        needEncrypting = [[self class] verifyDatabaseAccess:db error:nil];
+    }];
+    
+    if (needEncrypting) {
```
khawkins: Can we additionally wrap this in an `#ifdef SQLITE_HAS_CODEC`?  If for whatever reason (e.g. testing) a developer didn't have encryption enabled, this would preclude him/her from working with an unencrypted database.

wmathurin: It's compile time flag, if it is not there - no encryption is possible.
If they don't want encryption, they should pass @"" for the key, and this method would not do anything.

khawkins: If the `SQLITE_HAS_CODEC` flag is not set, the encryption code gets called every time, and all errors out in various ways, because there's no guard against not executing encryption commands against DBs that can't support it.

As near as I can tell, stuff still works.  But it's not an ideal development paradigm, vs. a guard against even executing the code.

I don't understand the comment about compile time vs. runtime.  Whether or not SmartStore would encrypt data—or take encryption-based actions on data—has always been predicated on a compile-time constant.

khawkins: Cf. https://github.com/forcedotcom/fmdb/blob/9d594b9a0274f7aa5f2cf0802e3b2742276b9d37/src/fmdb/FMDatabase.m#L398

wmathurin: If `SQLITE_HAS_CODEC` is not defined, SQLCipher is basically disabled.
FMDatabase can be used with SQLite and SQLCipher.
Our code however expects SQLCipher.

If you don't want to encrypt your database while using SQLCipher, simply pass `@""` as the key.

We discovered an issue in 5.3.
Maybe you remember that FMDB not compiling because of setkey/rekey not being found when using XCode 9. We came up with a fix for cocoapod based apps which seemed to work (it compiled and ran fine) but it was actually the wrong header and SQLCipher was running without the ability to encrypt/decrypt (but not complaining about it).
We fixed our pod files (see https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/89/files) - but for 5.3 app upgrading to 6.0, we need this piece of code I just added to encrypt the database if it is not encrypted even though we have a key.

PR #2305, opened 2017-12-06
Binary download support for react native

wmathurin:
New flag returnBinary (default false) in network bridge sendRequest method.
When returnBinary is true the bridge will send back the base64 encoded body and the content type.

Related to: https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/pull/52


wmathurin:
Note: Testing by tweaking SmartSyncExplorerReactNative

PR #2358, opened 2018-01-12
parentChildren sync targets in JSON configs are not read

grevolution:
Fixed a bug where parentChildren sync targets defined in the <user/global>syncs.json file are not read properly and the SDK crashes because of nil target.


wmathurin:
Thanks @grevolution 
(I just checked the Android side, and it was not missed there - see [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/target/SyncDownTarget.java#L77)).

PR #2366, opened 2018-01-17
Clean ghosts should not remove records downloaded by the sync 

wmathurin:
iOS equivalent to https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1597

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -322,6 +322,7 @@ static NSMutableDictionary *syncMgrList = nil;
     SFSyncStateMergeMode mergeMode = sync.mergeMode;
     SFSyncDownTarget* target = (SFSyncDownTarget*) sync.target;
     long long maxTimeStamp = sync.maxTimeStamp;
+    NSNumber* syncId = [NSNumber numberWithInteger:sync.syncId];
 
     SFSyncDownTargetFetchErrorBlock failBlock = ^(NSError *error) {
         failSync(@"Server call for sync down failed", error);
@@ -354,7 +355,7 @@ static NSMutableDictionary *syncMgrList = nil;
             NSArray* recordsToSave = idsToSkip && idsToSkip.count > 0 ? [strongSelf  removeWithIds:records idsToSkip:idsToSkip idField:target.idFieldName] : records;
 
             // Save to smartstore.
-            [target saveRecordsToLocalStore:strongSelf  soupName:soupName records:recordsToSave];
+            [target saveRecordsToLocalStore:strongSelf  soupName:soupName records:recordsToSave syncId:syncId];
             countFetched += [records count];
             progress = 100*countFetched / totalSize;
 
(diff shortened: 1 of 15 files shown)


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -467,25 +468,25 @@ - (void) cleanResyncGhosts:(NSNumber*)syncId completionStatusBlock:(SFSyncSyncMa
...
-                                       attributes[@"syncTarget"] = NSStringFromClass([sync.target class]);
-                                       [SFSDKEventBuilderHelper createAndStoreEvent:@"cleanResyncGhosts" userAccount:nil className:NSStringFromClass([self class]) attributes:attributes];
-                                       completionStatusBlock(SFSyncStateStatusDone);
-                                   }];
-
+    [(SFSyncDownTarget *) sync.target cleanGhosts:self
+                                         soupName:soupName
+                                           syncId:[NSNumber numberWithInteger:sync.syncId]
+                                       errorBlock:^(NSError *e) {
+                                           [SFSDKSmartSyncLogger e:[self class] format:@"Failed to get list of remote IDs, %@", [e localizedDescription]];
```
bhariharan: Don't we have to use `weakSelf` inside the block here to prevent retain cycles?

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -467,25 +468,25 @@ - (void) cleanResyncGhosts:(NSNumber*)syncId completionStatusBlock:(SFSyncSyncMa
...
+                                           NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
+                                           attributes[@"syncId"] = [NSNumber numberWithInteger:sync.syncId];
+                                           attributes[@"syncTarget"] = NSStringFromClass([sync.target class]);
+                                           [SFSDKEventBuilderHelper createAndStoreEvent:@"cleanResyncGhosts" userAccount:nil className:NSStringFromClass([self class]) attributes:attributes];
```
bhariharan: Don't we have to use `weakSelf` inside the block here to prevent retain cycles?

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -467,25 +468,25 @@ - (void) cleanResyncGhosts:(NSNumber*)syncId completionStatusBlock:(SFSyncSyncMa
...
+                                           completionStatusBlock(SFSyncStateStatusFailed);
+                                       }
+                                    completeBlock:^(NSArray *localIds) {
+                                        NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
+                                        attributes[@"numRecords"] = [NSNumber numberWithInteger:localIds.count];
+                                        attributes[@"syncId"] = [NSNumber numberWithInteger:sync.syncId];
+                                        attributes[@"syncTarget"] = NSStringFromClass([sync.target class]);
+                                        [SFSDKEventBuilderHelper createAndStoreEvent:@"cleanResyncGhosts" userAccount:nil className:NSStringFromClass([self class]) attributes:attributes];
```
bhariharan: Don't we have to use `weakSelf` inside the block here to prevent retain cycles?

libs/SmartSync/SmartSync/Classes/Target/SFParentChildrenSyncDownTarget.m
```diff
@@ -159,20 +159,25 @@ - (NSString*) getQueryToRun:(long long)maxTimeStamp {
...
+                syncId:syncId
             errorBlock:errorBlock
          completeBlock:^(NSArray *localIdsArr) {
 
              // Taking care of ghost children
 
              // NB: ParentChildrenSyncDownTarget's getNonDirtyRecordIdsSql does a join between parent and children soups
              // We only want to look at the children soup, so using SoqlSyncDownTarget's getNonDirtyRecordIdsSql
-             NSMutableOrderedSet* localChildrenIds = [[self getIdsWithQuery:[super getNonDirtyRecordIdsSql:self.childrenInfo.soupName idField:self.childrenInfo.idFieldName] syncManager:syncManager] mutableCopy];
+             NSMutableOrderedSet *localChildrenIds = [[self getIdsWithQuery:[super getNonDirtyRecordIdsSql:self.childrenInfo.soupName
```
bhariharan: I think `self` here also will get retained since it's inside the block.

wmathurin: I wonder if we are not doing the strongself/weakself sometimes when we don't need to.

If self doesn't keep a reference to the block there is no chance of retain cycle.
When only using blocks inline, we shouldn't have to use weak self. (see https://stackoverflow.com/questions/13194613/is-it-necessary-to-use-weak-references-to-self-always-inside-blocks/13194818#13194818).

Also since, we have the `-Wimplicit-retain-self` flag, wouldn't the compiler warn us?

PR #2375, opened 2018-01-19
Bug fix  for sample

trooper2013:
Realized that the SmartSyncExplorer sample extensions code was failing in 6.0 while working with @jhneves for iOS extensions.


jhneves:
✌️

PR #2386, opened 2018-01-30
Turning of parallel builds in SwiftSDK

khawkins:
Since Carthage tries to build all framework targets in the workspace, parallel builds tend to sporadically error out on race conditions, as dependency project builds (CocoaLumberjack is the one I see the most) end up getting yanked out from under other upstream targets being built.


harry528tt:
cc: @guo-li Please check this fix for the msdk build jobs.

liguopku:
👍

PR #2390, opened 2018-02-01
Fixing linker errors with pod specs in some cases

bhariharan:
Transitive dependencies don't work with pod specs + `Xcode 9.2` while using `Swift Extensions` from those libraries.


bhariharan:
Example of a bug that you wouldn't see in the main repo but would if you attempt to run the `Swift` template (main repo doesn't use `Cocoapods`). Funnily enough, this appears to have been working fine until `Xcode 9.2` came out. I'm going to verify the other templates too.

wmathurin:
That's why I love xcode.

PR #2402, opened 2018-02-08
cleanResyncGhosts should call its callbacks after finishing

wmathurin:
Before 
The smartsync plugin methods for cleanResyncGhosts was returning right away and there was no way to know if/when it completed. 

New behavior 
cleanResyncGhosts calls its callbacks after finishing (success callback if successful and error callback otherwise).

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFSmartSyncPlugin/SFSmartSyncPlugin.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFSmartSyncPlugin/SFSmartSyncPlugin.m
@@ -187,16 +187,17 @@ NSString *const kSyncStoreNameArg = @"storeName";
 {
     NSDictionary *argsDict = [self getArgument:command.arguments atIndex:0];
     NSNumber* syncId = (NSNumber*) [argsDict nonNullObjectForKey:kSyncIdArg];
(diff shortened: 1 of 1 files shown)


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/Plugins/SFSmartSyncPlugin/SFSmartSyncPlugin.m
```diff
@@ -199,15 +185,22 @@ - (void) reSync:(CDVInvokedUrlCommand *)command
...
-    } command:command];
+    NSDictionary *argsDict = [self getArgument:command.arguments atIndex:0];
+    NSNumber* syncId = (NSNumber*) [argsDict nonNullObjectForKey:kSyncIdArg];
+    [SFSDKHybridLogger d:[self class] format:@"cleanResyncGhosts with sync id: %@", syncId];
+    id <CDVCommandDelegate> commandDelegate = self.commandDelegate;
+    NSString* callbackId = command.callbackId;
+    [[self getSyncManagerInst:argsDict] cleanResyncGhosts:syncId completionStatusBlock:^(SFSyncStateStatus syncStatus) {
+        CDVPluginResult* pluginResult;
+        if (syncStatus == SFSyncStateStatusDone) {
+            [SFSDKHybridLogger d:[self class] format:@"cleanResyncGhosts completed successfully"];
```
bhariharan: Do we need the `weak`/`strong` dance on `self` inside this block to avoid a retain cycle?

PR #2409, opened 2018-02-15
Made userId and orgId nonnull…

ssraza:
Since `userId` and `orgId` are being set in the initializer, they shouldn't be `nullable`

@trooper2013 @bhariharan


khawkins:
`isEqual:` needs a bit of revision in the implementation, to get rid of the `nil` comparisons.

ssraza:
Thanks for catching that @khawkins... I will update it shortly...

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerTests.m
```diff
@@ -147,22 +147,10 @@ - (void)tearDown {
 - (void)testAccountIdentityEquality {
     NSDictionary *accountIdentityMatrix = @{
                                             @"MatchGroup1": @[
```
ssraza: I updated this test case, since SFUserAccountIdentity initialization with `nil` should not be supported... Let me know what do you folks think...

PR #2411, opened 2018-02-15
Fixing SOSL response blocks

bhariharan:
Reported [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/2408).

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
@@ -1470,14 +1470,14 @@ static NSException *authException = nil;
 
     // An success block that we expected to succeed
     SFRestDictionaryResponseBlock dictSuccessBlock = ^(NSDictionary *d, NSURLResponse *rawResponse) {
-        [XCTAssertTrue([d isKindOfClass:[NSDictionary class]], @"Response should be a dictionary")];
+        XCTAssertTrue([d isKindOfClass:[NSDictionary class]], @"Response should be a dictionary");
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
```diff
@@ -1574,7 +1565,7 @@ - (void) testBlocks {
     self.currentExpectation = [self expectationWithDescription:@"performSOSLSearch-find {batman}"];
     [api performSOSLSearch:@"find {batman}"
                  failBlock:failWithUnexpectedFail
-             completeBlock:arraySuccessBlock];
+             completeBlock:dictSuccessBlock];
```
wmathurin: How was that test passing for the last year?

bhariharan: I'm not sure it was passing - our CI has been broken for a while now.

wmathurin: It was passing locally.
I think because we didn't send any message to (i.e. call any method on) the array object in `arraySuccessBlock`, nothing bad happened.
Could you add an assert inside `dictSuccessBlock` that checks that `[d isKindOfClass:[NSDictionary class]]` ?

bhariharan: Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2411/commits/118de98f83558e99e5e7c4d60a821acb1d6b7f1d).

PR #2415, opened 2018-02-20
Added branding settings for hostListVC

harry528tt:
@bhariharan @trooper2013


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostListViewController.m
```diff
@@ -167,6 +166,11 @@ - (void)viewWillAppear:(BOOL)animated {
     [self resizeContentForPopover];
     // style navigiation bar
     [super viewWillAppear:animated];
+    
+    SFSDKLoginViewControllerConfig *config = [SFUserAccountManager sharedInstance].loginViewControllerConfig;
```
bhariharan: @harry528tt Shouldn't this still be within the `if` block?

harry528tt: @bhariharan Actually I don't think it's necessary. The if condition at line 132 is for checking "rightBarButtonItem"; But the tintColor I set here is not only for that, but also for other navBarItems.

harry528tt:
retest this please

brandonpage:
@harry528tt Github pages take a while to index and actually display the results.  280 tests ran with 3 failures, but they look like flappers to me.

harry528tt:
@brandonpage thanks for the confirmation. @trooper2013 @bhariharan I moved the setting branding part from viewWillAppear to viewDidLoad.

PR #2419, opened 2018-02-21
Generating react tests bundle during build

wmathurin:
Packager is never needed to run the react tests.


libs/SalesforceReact/SalesforceReact.xcodeproj/project.pbxproj
```diff
@@ -957,12 +991,26 @@
...
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "$SRCROOT/node_modules/.bin/react-native bundle --platform ios --dev true --entry-file node_modules/react-native-force/test/alltests.js --bundle-output SalesforceReactTests/index.ios.bundle\n";
```
wmathurin: @bhariharan what would be an elegant way to do that on the Android side?

PR #2421, opened 2018-02-22
Add methods for updating accessibility of RSA Keys

sgoldberg-sfdc:
Notification extensions require access to the key pair when device is locked in order to support decrypting encrypted push notifications. These methods allow a client app to change the accessibility of a key pair.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.h
@@ -148,6 +148,20 @@ extern NSUInteger const kSFPBKDFDefaultSaltByteLength;
 + (nullable SecKeyRef)getRSAPrivateKeyRefWithName:(NSString *)keyName keyLength:(NSUInteger)length;
 
 /**
+ * Updates a private key's accessibility to given attribute
+ * @param keyName The name string used to generate the key.
+ * @param accessibleAttribute The accessibility attributed to use for keys
+ */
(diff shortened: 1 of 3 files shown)


sgoldberg-sfdc:
@bhariharan @trooper2013

sgoldberg-sfdc:
[SFKeyChainItemWrapper setAccessibleAttribute:] appears limited to Passwords stored in the keychain atm, and sets accessibility for all passwords. 

The goal here is to offer granular ability to set it specifically for one key/key-pair. In particular, the keys generated for use encrypting/decrypting push notifications.

khawkins:
IMO, to the extent that `SFKeychainItemWrapper` is limited, it should be expanded to accommodate the underlying keychain access need.  We've worked pretty hard to consolidate keychain management functionality into that class, where it used to be spread across numerous decentralized implementations.  I'd rather we didn't go back in that direction.

sgoldberg-sfdc:
@khawkins where do you draw the line between the keys and the keychain? The creation/and querying  of key pairs is currently handled in cryptoutils (even though they are stored in the keychain).

khawkins:
`SecItem*` calls should be managed in a single keychain manager class.  Keychain data access should be centralized.

sgoldberg-sfdc:
@bhariharan let me know if you require changes to this.

bhariharan:
I agree with @khawkins. I would prefer consolidating this functionality in `SFKeychainItemWrapper`.

bhariharan:
Discussed with @khawkins and @sgoldberg-sfdc offline. A new PR shall be submitted with a different approach.

PR #2423, opened 2018-02-22
Simplifying the Error handling for authentication

trooper2013:
A change to make handling of errors a bit easier for developers. Any SFUserAccountManager delegate has now the choice to let the SDK know if it has handled the error. If the Error has been handled then the SDK does not attempt to handle the error. Else it will use the ErrorManager and process the errors using the available ErrorHandlers. The ErrorManager is also made public just in case the error-handlers needs to be customized (in extreme cases). A bug related to displaying of alerts (broken from SafariViewController PR).


libs/SalesforceSDKCore/SalesforceSDKCore/SalesforceSDKCore.h
```diff
@@ -2,7 +2,7 @@
  SalesforceSDKCore.h
  SalesforceSDKCore
 
- Created by Bharath Hariharan on Fri Feb  9 09:53:54 PST 2018.
+ Created by Raj Rao on Thu Feb 22 15:09:12 PST 2018.
```
bhariharan: @trooper2013 Why didn't the new public header get added here? Could you wipe `DerivedData` and build the `Everything` scheme?

khawkins:
Looks like `TestUserAccountManagerDelegate` needs to be updated.

PR #2429, opened 2018-02-27
Add Nullable for SFPreference objectForKey

harry528tt:
@bhariharan @trooper2013 
Add nullable to this property to fix crash/ unit tests in S1.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFPreferences.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFPreferences.h
@@ -72,7 +72,7 @@ NS_ASSUME_NONNULL_BEGIN
 /** Returns the preferences object for the given key.
  @param key The key of the requested object.
  */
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFPreferences.h
```diff
@@ -72,7 +72,7 @@ NS_ASSUME_NONNULL_BEGIN
 /** Returns the preferences object for the given key.
  @param key The key of the requested object.
  */
-- (id)objectForKey:(NSString*)key;
+- (nullable id)objectForKey:(NSString*)key;
```
ssraza: @bhariharan @trooper2013 we believe this method should have a nullable return type. Since it's a dictionary under the hud, IMO it should be allowed to return `nil`. Specially with `removeObjectForKey` method, `nil` return can be a very common use case... Please let me know if you folks think otherwise...

PR #2434, opened 2018-02-27
Handle full page session refreshes

alsocalledchris:
Updated to better handle full page refreshes - remove session URL from redirect URL and URL decode absolute URL so don't get errors on redirect.

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
@@ -425,6 +425,15 @@ SFSDK_USE_DEPRECATED_BEGIN
     NSURL *instUrl = creds.apiUrl;
     NSString *fullReturnUrlString = returnUrlString;
     
+    // Catch if we have a absoute URL, if we do then url decode to avoid
+    // errors with 'Invalid page redirection'
+    if(createAbsUrl && [returnUrlString hasPrefix:@"http"]) {
+        NSString *decodedUrlString = [(NSString *)returnUrlString stringByReplacingOccurrencesOfString:@"+" withString:@" "];
+        decodedUrlString = [decodedUrlString stringByReplacingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+        NSURL *nsUrl = [NSURL URLWithString:decodedUrlString];
+        fullReturnUrlString = nsUrl.path;
+    }
+  
     /*
      * We need to use the absolute URL in some cases and relative URL in some
      * other cases, because of differences between instance URL and community URL.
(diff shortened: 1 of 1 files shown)


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -425,6 +425,15 @@ - (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrlString returnUrlIsEnco
     NSURL *instUrl = creds.apiUrl;
     NSString *fullReturnUrlString = returnUrlString;
     
+    // Catch if we have a absoute URL, if we do then url decode to avoid
+    // errors with 'Invalid page redirection'
+    if(createAbsUrl && [returnUrlString hasPrefix:@"http"]) {
+        NSString *decodedUrlString = [(NSString *)returnUrlString stringByReplacingOccurrencesOfString:@"+" withString:@" "];
```
bhariharan: I don't think you need the explicit cast here - `returnUrlString` is of type `NSString *`.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -425,6 +425,15 @@ - (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrlString returnUrlIsEnco
     NSURL *instUrl = creds.apiUrl;
     NSString *fullReturnUrlString = returnUrlString;
     
+    // Catch if we have a absoute URL, if we do then url decode to avoid
+    // errors with 'Invalid page redirection'
+    if(createAbsUrl && [returnUrlString hasPrefix:@"http"]) {
```
bhariharan: Shouldn't this be `https`? I don't think any pages within the `force.com` domain support `http`.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -591,8 +600,10 @@ - (void) webView:(WKWebView *) webView decidePolicyForNavigationAction:(WKNaviga
             [self refreshCredentials:[SFUserAccountManager sharedInstance].currentUser.credentials
              completion:^(SFOAuthInfo *authInfo, SFUserAccount *userAccount) {
                  [SFUserAccountManager sharedInstance].currentUser = userAccount;
+                 //Remove the session URL from the return URL as we have taken over the refresh process
+                 NSString *sessionRemovedRefreshUrl = [refreshUrl stringByReplacingOccurrencesOfString:@"%2Fvisualforce%2Fsession%3Furl%3D" withString:@""];
```
bhariharan: Where is this `visualforce` URL coming from?

brandonpage:
## Danger has errored
```
 >    xcov.report(
 #        scheme: 'UnitTests',
```

<p align="right" data-meta="generated_by_CI-Result-SalesforceHybridSDK">
</p>

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
chriswellingsfairsailcom: It was just the same as the line below - will catch both https and http

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
chriswellingsfairsailcom: Ah OK, thanks

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
chriswellingsfairsailcom: No idea - it is sometimes (mostly) there and others not (when not the refresh worked before this pull request's change) - so is causes the issue.

chriswellingsfairsailcom:
@bhariharan not sure how Android is working without this extra code - but it probably needs it too if you agree the above solution is correct.

bhariharan:
@alsocalledchris I think [this PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2438) should fix this.

alsocalledchris:
Many thanks @bhariharan, and thanks for picking up the other issues I have logged. When will those fixes make it into a release? :)

bhariharan:
I will post updates on the issues filed as and when I make progress.

PR #2438, opened 2018-03-01
Fixing session refresh bugs and UIWebView bugs in hybrid remote apps

bhariharan:
In some cases, the platform returns a `startURL` value with the additional prefix of `/visualforce/session?url=` (seems to be a new thing). This messes up our `frontdoor` computation during the token refresh flow.

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
@@ -62,6 +62,7 @@ static NSInteger  const kErrorCodeNoCredentials = 2;
 static NSString * const kErrorContextAppLoading = @"AppLoading";
 static NSString * const kErrorContextAuthExpiredSessionRefresh = @"AuthRefreshExpiredSession";
 static NSString * const kVFPingPageUrl = @"/apexpages/utils/ping.apexp";
+static NSString * const kVFSessionPrefix = @"%2Fvisualforce%2Fsession%3Furl%3D";
 
 // App feature constant.
 static NSString * const kSFAppFeatureUsesUIWebView = @"WV";
@@ -333,10 +334,8 @@ SFSDK_USE_DEPRECATED_BEGIN
             failureBlock(authInfo, error);
         }
     };
-
     if (![SFUserAccountManager sharedInstance].currentUser) {
-        [self loginWithCompletion:authCompletionBlock
-                                                             failure:authFailureBlock];
+        [self loginWithCompletion:authCompletionBlock failure:authFailureBlock];
     } else {
         [self refreshCredentials:[SFUserAccountManager sharedInstance].currentUser.credentials
                                                          completion:authCompletionBlock
@@ -416,15 +415,15 @@ SFSDK_USE_DEPRECATED_BEGIN
 
 - (NSURL *)frontDoorUrlWithReturnUrl:(NSString *)returnUrlString returnUrlIsEncoded:(BOOL)isEncoded createAbsUrl:(BOOL)createAbsUrl
 {
+
     // Special case: if returnUrlString itself is a frontdoor.jsp URL, parse its parameters and rebuild.
     if ([returnUrlString containsString:@"frontdoor.jsp"]) {
         return [self parseFrontDoorReturnUrlString:returnUrlString encoded:isEncoded];
     }
-    
     SFOAuthCredentials *creds = [SFUserAccountManager sharedInstance].currentUser.credentials;
     NSURL *instUrl = creds.apiUrl;
     NSString *fullReturnUrlString = returnUrlString;
-    
+
     /*
      * We need to use the absolute URL in some cases and relative URL in some
      * other cases, because of differences between instance URL and community URL.
@@ -434,17 +433,16 @@ SFSDK_USE_DEPRECATED_BEGIN
         retUrlComponents.path = [retUrlComponents.path stringByAppendingPathComponent:returnUrlString];
         fullReturnUrlString = retUrlComponents.string;
     }
-    
+
     // Create frontDoor path based on credentials API URL.
     NSURLComponents *frontDoorUrlComponents = [NSURLComponents componentsWithURL:instUrl resolvingAgainstBaseURL:NO];
     frontDoorUrlComponents.path = [frontDoorUrlComponents.path stringByAppendingPathComponent:@"/secur/frontdoor.jsp"];
-    
+
     // NB: We're not using NSURLComponents.queryItems here, because it unsufficiently encodes query params.
     NSMutableString *frontDoorUrlString = [NSMutableString stringWithString:frontDoorUrlComponents.string];
     NSString *encodedRetUrlValue = (isEncoded ? fullReturnUrlString : [fullReturnUrlString stringByURLEncoding]);
     NSString *encodedSidValue = [creds.accessToken stringByURLEncoding];
     [frontDoorUrlString appendFormat:@"?sid=%@&retURL=%@&display=touch", encodedSidValue, encodedRetUrlValue];
-    
     return [NSURL URLWithString:frontDoorUrlString];
 }
 
@@ -472,6 +470,9 @@ SFSDK_USE_DEPRECATED_BEGIN
         BOOL foundStartURL = (startUrlValue != nil);
         BOOL foundValidEcValue = ([ecValue isEqualToString:@"301"] || [ecValue isEqualToString:@"302"]);
         if (foundStartURL && foundValidEcValue) {
+            if ([startUrlValue containsString:kVFSessionPrefix]) {
+                startUrlValue = [startUrlValue stringByReplacingOccurrencesOfString:kVFSessionPrefix withString:@""];
+            }
             return startUrlValue;
         }
     }
@@ -591,6 +592,7 @@ SFSDK_USE_DEPRECATED_BEGIN
             [self refreshCredentials:[SFUserAccountManager sharedInstance].currentUser.credentials
              completion:^(SFOAuthInfo *authInfo, SFUserAccount *userAccount) {
                  [SFUserAccountManager sharedInstance].currentUser = userAccount;
+
                  // Reset the user agent back to Cordova.
                  [weakSelf authenticationCompletion:refreshUrl authInfo:authInfo];
              } failure:^(SFOAuthInfo *authInfo, NSError *error) {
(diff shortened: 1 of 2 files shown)


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -472,6 +470,9 @@ - (NSString *)isLoginRedirectUrl:(NSURL *)url
         BOOL foundStartURL = (startUrlValue != nil);
         BOOL foundValidEcValue = ([ecValue isEqualToString:@"301"] || [ecValue isEqualToString:@"302"]);
         if (foundStartURL && foundValidEcValue) {
+            if ([startUrlValue containsString:kVFSessionPrefix]) {
+                startUrlValue = [startUrlValue stringByReplacingOccurrencesOfString:kVFSessionPrefix withString:@""];
```
bhariharan: Basically we need to extract the `url` value from the embedded `url` parameter in this case.

khawkins: I wonder: is the right thing to do here to get rid of this prefix, or get rid of the frontDoor VF ping functionality?

bhariharan: We need the VF ping functionality to set cookies on those domains. Unless we start removing and setting cookies on the hybrid view ourselves. More info offline.

khawkins: I was just curious if the frontDoor redirect to that particular URL that's being removed, accomplished the same thing.  Seems like it might.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -668,7 +670,7 @@ - (BOOL) webView:(UIWebView *) webView shouldStartLoadWithRequest:(NSURLRequest
          * If the request is attempting to refresh an invalid session, take over
          * the refresh process via the OAuth refresh flow in the container.
          */
-        NSString *refreshUrl = [self isLoginRedirectUrl:webView.request.URL];
+        NSString *refreshUrl = [self isLoginRedirectUrl:request.URL];
```
bhariharan: In `UIWebView`, `request.URL` returns the correct value now, whereas `webView.request.URL` returns `nil`, since `webView.request` is `nil` (because it hasn't loaded the request yet).

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
khawkins: Can I request that we do this either directly or indirectly through [`parseFrontDoorReturnUrlString`](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2438/files#diff-bafe72208f1ff3afa7b2024a36312efeR449)? It’s basically the purpose of that method, albeit currently for another edge case.

khawkins: (Or basically start the amendment from that [special case check section](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2438/files#diff-bafe72208f1ff3afa7b2024a36312efeR420) in `frontDoorUrlWithReturnUrl:returnUrlIsEncoded:createAbsUrl:`)

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -416,35 +414,34 @@ - (NSString *)sfHybridViewUserAgentString
...
     NSURL *instUrl = creds.apiUrl;
     NSString *fullReturnUrlString = returnUrlString;
-    
+
     /*
      * We need to use the absolute URL in some cases and relative URL in some
      * other cases, because of differences between instance URL and community URL.
      */
-    if (createAbsUrl && ![returnUrlString hasPrefix:@"http"]) {
+    if (createAbsUrl && ![returnUrlString containsString:@"http"]) {
```
bhariharan: This was the real culprit (the difference between `hasPrefix` and `containsString`). Some URLs can be of the format `/visualforce/session?url=<blah>` and these should NOT be treated as relative URLs and prefixed by `instanceUrl`. We were doing that causing `fullReturnUrlString` a couple of lines below to become `null` and thereby `retUrl` to be `null`. However, if these URLs are passed along as-is, the page loads fine as expected post-refresh. cc: @khawkins.

khawkins: They should not be treated as relative URLs?  Why?  They are relative URLs.  Can you explain a little more?

bhariharan: It should be passed along as-is for `retUrl`, but it's being prepended by `instanceUrl` incorrectly.

khawkins: But then the actual issue is that this method is being called with the wrong `createAbsUrl` value.

As mentioned earlier, this seems like a special case that's very similar to [the initial call](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2438/files#diff-bafe72208f1ff3afa7b2024a36312efeR420) in this method.  I'd prefer to special-case it, rather than abandon the contract specified by the method.  You're going to get false negatives for things like `returnUrlString = @"/apex/httputils"` and the like.

khawkins: `parseFrontDoorReturnUrlString:encoded:` seems like exactly this case, with another URL format.  I.e. the actual return URL is embedded in the outer URL string.  The method parses that out, and runs it back through `frontDoorUrlWithReturnUrl:returnUrlIsEncoded:createAbsUrl:`.  It's not my favorite workflow (by a long stretch), but it's what makes the most sense in the current code base.

bhariharan: The behavior of something like `/apex/httputils` would be the same as before, even after this fix. In fact, nothing would change except session prefix requests already embedded with a full URL. I don't view this as a special case, it has always been this way (if the value of `startURL` contains `https://`, don't touch it). The regression was changing it to "if the value of `startURL` starts with `https://`).

khawkins: I got my edge case wrong, but the problem is still there.  An absolute URL string as a return URL vs. a relative one is totally based on a token that's insufficiently determinant. For cases of createAbsUrl:

- /apex/httputils: Return URL string = /apex/httputils
- /apex/mypage: Return URL string = https://na45.salesforce.com/apex/mypage.

It's totally arbitrary, and the first case fails its mandate to create an absolute URL.

khawkins: You could update it to something like:
```obj-c
if (createAbsUrl && !([returnUrlString containsString:@"http://"] || [returnUrlString containsString:@"https://"])) {
```
Which is better, but it's still a shot in the dark about knowing what's actually going on in the data.  It may or may not make a difference, in the end, but it certainly makes it a difficult method to maintain, because it's hard to understand what's going on.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
```diff
@@ -594,10 +592,9 @@ + (SFSDKOAuthClient *)idpAuthInstance:(SFSDKOAuthClientConfig *)config {
     } else {
         instance = [[SFSDKIDPAuthClient alloc] initWithConfig:config];
     }
-    SFSDKMutableOAuthClientContext *context = [[SFSDKMutableOAuthClientContext alloc] initWithAuthType:SFOAuthTypeAdvancedBrowser];
+    SFSDKMutableOAuthClientContext *context = [[SFSDKMutableOAuthClientContext alloc] initWithAuthType:SFOAuthTypeIDP];
```
bhariharan: This was the underlying cause of most of the hybrid remote refresh problems we were seeing specific to `iOS`, incorrectly assigning `OAuthClientContext`, most likely caused by a copy-paste error, and it was quite hard to hunt down. Let me explain:
- Going through a lot of async magical steps, we get to a point where `SFSDKOAuthClient` spins up an instance of itself with some context based on some input.
- There were some class methods for this (1 of each type) - `idpAuthInstance`, `nativeBrowserAuthInstance`, and `webviewAuthInstance`.
- Unfortunately, all these methods initialized themselves with the same auth type, i.e., `SFOAuthTypeAdvancedBrowser`, regardless of what the input was. :-(
- I fixed this by assigning the correct type to each.
- The next issue was there was no way to create a client with auth type `SFOAuthTypeRefresh`. I added a utility method for this and added a check in the creation block to use this auth type if a valid refresh token is present.
- The hybrid view controller relies on the auth type to assign the value of `createAbsUrl` (if it's of type `SFOAuthTypeRefresh`, it sets it to `NO`, which is what we want in this case). Since the type was incorrectly being set to `SFOAuthTypeAdvancedBrowser`, the downstream refresh flow was failing because this would lead to incorrect computation of `retURL`. I reverted the other special case since we don't need it anymore after fixing the underlying issue @khawkins.
cc: @wmathurin @trooper2013

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -711,33 +709,9 @@ - (BOOL) webView:(UIWebView *) webView shouldStartLoadWithRequest:(NSURLRequest
...
-        }
-        if (anyPluginsResponded) {
-            return shouldAllowRequest;
-        }
-        if ([url isFileURL]) {
-            return YES;
         } else {
             [[NSNotificationCenter defaultCenter] postNotification:[NSNotification notificationWithName:CDVPluginHandleOpenURLNotification object:url]];
         }
-        return NO;
```
bhariharan: `UIWebView` was broken for hybrid remote because of this piece of code. The default if everything falls through should `return YES` to load the page. I also removed the obsolete code above and lined things up with the `WKWebView` implementation. The `Cordova` library doesn't require this code to delegate explicitly anymore. Verified that everything's working as expected in `UIWebView` now, both first time load and session refresh flow.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKOAuthClient.m
khawkins: Nice sleuthing, @bhariharan !  I think @trooper2013 owes you a beer. 😉

PR #2439, opened 2018-03-06
Added Swift api documentation

trooper2013:
Added api documentation and the jazzy yaml file. 
FYI @rwhitley to generate docs do the following at the root of the sdk project.  Swift versions have to be specific for jazzy to generate docs. It maybe a bug in jazzy. 

```
jazzy --config ./jazzy.yaml  --swift-version 4.0.3
```


jazzy.yaml
```diff
@@ -0,0 +1,17 @@
+module: SalesforceSwiftSDK
```
bhariharan: Could we move this file under a top-level `doc` folder?

jazzy.yaml
```diff
@@ -0,0 +1,17 @@
...
+author: Salesforce
+author_url: https://github.com/forcedotcom
+github_url: https://github.com/forcedotcom/SalesforceMobileSDK-iOS
+exclude:
+  - "**/*.h"
+  - "**/*-ObjC.swift"
+  - "**/*Deprecations.swift"
+  - "**/*SwiftExtensions.swift"
+
+copyright: ''
```
bhariharan: Should we add our standard copyright here?

PR #2441, opened 2018-03-08
Fixing hybrid session refresh issues with passcode

bhariharan:
We were going through `SFOAuthCoordinator` and the entire user agent flow to perform a hybrid refresh action (both `hybrid_remote` and `hybrid_local`). This is the only piece of code that triggers the entire user agent flow, including spinning up a dedicated `SFSDKOauthClient`, for a simple refresh call, instead of using `SFRestAPI` or `SFOAuthSessionRefresher` to perform a lightweight refresh by hitting the token endpoint directly and getting a new access token. The problem with this is that if passcode is enabled and the passcode screen is on display due to timeout, the `OAuth` flow stops in the `SFOAuthCoordinator` and the callbacks in the plugin are never triggered. This is by design based on the assumption that the user agent flow would be triggered only for the user agent flow, not for the refresh flow. I changed this to make a lightweight API call using `SFRestAPI` to perform refresh and return the new credentials. The new design is now inline with what we have always done on `Android` - see [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SalesforceHybrid/src/com/salesforce/androidsdk/phonegap/ui/SalesforceDroidGapActivity.java#L315).

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
@@ -38,6 +38,7 @@
 #import <SalesforceSDKCore/SFSDKEventBuilderHelper.h>
 #import <SalesforceSDKCore/NSString+SFAdditions.h>
 #import <SalesforceSDKCore/SFSDKWebViewStateManager.h>
+#import <SalesforceSDKCore/SFRestAPI+Blocks.h>
 #import <Cordova/NSDictionary+CordovaPreferences.h>
 #import <Cordova/CDVUserAgentUtil.h>
 #import <objc/message.h>
@@ -300,6 +301,7 @@ SFSDK_USE_DEPRECATED_BEGIN
 
 - (void)authenticateWithCompletionBlock:(SFOAuthPluginAuthSuccessBlock)completionBlock failureBlock:(SFOAuthFlowFailureCallbackBlock)failureBlock
 {
+
     /*
      * Reconfigure user agent. Basically this ensures that Cordova whitelisting won't apply to the
      * WKWebView that hosts the login screen (important for SSO outside of Salesforce domains).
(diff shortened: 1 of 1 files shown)


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -859,11 +858,19 @@ - (void)loadVFPingPage
 - (void)refreshCredentials:(SFOAuthCredentials *)credentials
                  completion:(SFOAuthFlowSuccessCallbackBlock)completionBlock
                   failure:(SFOAuthFlowFailureCallbackBlock)failureBlock {
-    if ([SFUserAccountManager sharedInstance].useLegacyAuthenticationManager) {
-        [[SFAuthenticationManager sharedManager] refreshCredentials:credentials completion:completionBlock failure:failureBlock];
-    } else {
-        [[SFUserAccountManager sharedInstance] refreshCredentials:credentials completion:completionBlock failure:failureBlock];
```
khawkins: Is this something that needs to be fixed further up the chain?  Seems like `[SFUserAccountManager refreshCredentials:completion:failure:]` shouldn't be allowed to initiate a User Agent flow....

bhariharan: It doesn't trigger a user agent flow per se, it triggers a refresh flow. But the refresh flow is still using `SFOAuthCoordinator` and spinning up `SFSDKOAuthClient`, etc., instead of using `SFOAuthSessionRefresher`. I do want to refactor that code and make enhancements there, but it's too close to `6.1` for that. The goal of this exercise is to have the same design on `iOS` and `Android` as much as possible.

khawkins: Got it, thanks.  I don't see anything wrong with the specifics of your implementation.  Just was curious on the upstream question.

bhariharan: More meat for my `7.0 Enhancements` doc ;-)

PR #2443, opened 2018-03-09
Enforcing layout mode for auth Login view for iPhone devices

trooper2013:
The enforcement will work on iPhone device. Autorotation cannot be controlled for iPad devices. But the view should size accordingly now.  There seems to be a few issues with the approach taken for adjusting transparency on windows and its impact on rotations(views and keyboard). Toggling the hide attribute on UIWindow seems to resolve most of these issues, but requires more testing. Once the alternative is tested we can enable rotations for auth views as well. Merged all use use of topViewController code to single method in SFSDKWindowContainer.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -154,6 +154,14 @@ SFSDK_USE_DEPRECATED_END
     }
 }
 
+- (BOOL)shouldAutorotate {
+    return NO;
+}
+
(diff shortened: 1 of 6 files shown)


bhariharan:
Seems like a big change this close to `6.1`. Can we hold off on this until after `6.1` is released? What do you think @wmathurin?

trooper2013:
@wmathurin The statement  "seems to resolve most of these issues, but requires more testing" was related to switching to show/hide vs transparency. I haven't actually made those changes. The current PR resolves some rotation problems reported by s1.

bhariharan:
S1 has a workaround for now. I would prefer holding off on this until after `6.1`.

trooper2013:
Closing this PR for now. Will address in 6.2 with show/hide functionality for windows.

PR #2444, opened 2018-03-09
Unauthenticated analytics instrumentation support

khawkins:
Provides an analytics manager for any unauthenticated data capture.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager+Internal.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager+Internal.h
@@ -1,0 +1,51 @@
+/*
+ SFSDKSalesforceAnalyticsManager+Internal.h
+ SalesforceSDKCore
+ 
+ Created by Kevin Hawkins on 3/9/18.
+ 
+ Copyright (c) 2016-present, salesforce.com, inc. All rights reserved.
+ 
+ Redistribution and use of this software in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this list of conditions
+ and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice, this list of
+ conditions and the following disclaimer in the documentation and/or other materials provided
+ with the distribution.
+ * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+ endorse or promote products derived from this software without specific prior written
+ permission of salesforce.com, inc.
+ 
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+ IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#import <SalesforceSDKCore/SFSDKSalesforceAnalyticsManager.h>
+#import <SalesforceSDKCore/SFAuthenticationManager.h>
+
+@interface SFSDKAnalyticsTransformPublisherPair : NSObject
+
+@property (nonnull, nonatomic, readonly, strong) id<SFSDKTransform> transform;
+@property (nonnull, nonatomic, readonly, strong) id<SFSDKAnalyticsPublisher> publisher;
+
+- (nonnull instancetype)initWithTransform:(nonnull id<SFSDKTransform>)transform publisher:(nonnull id<SFSDKAnalyticsPublisher>)publisher;
+
+@end
+SFSDK_USE_DEPRECATED_BEGIN
+
+@interface SFSDKSalesforceAnalyticsManager () <SFAuthenticationManagerDelegate>
+
+SFSDK_USE_DEPRECATED_END
+@property (nonnull, nonatomic, readwrite, strong) SFSDKAnalyticsManager *analyticsManager;
+@property (nonnull, nonatomic, readwrite, strong) SFSDKEventStoreManager *eventStoreManager;
+@property (nullable, nonatomic, readwrite, strong) SFUserAccount *userAccount;
+@property (nonnull, nonatomic, readwrite, strong) NSMutableArray<SFSDKAnalyticsTransformPublisherPair *> *remotes;
+
+@end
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager+Internal.h
```diff
@@ -0,0 +1,51 @@
+/*
```
khawkins: I moved these interfaces out of the .m for testing purposes.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -37,47 +37,29 @@
...
-@end
-
 @implementation SFSDKSalesforceAnalyticsManager
 
-+ (instancetype) sharedInstanceWithUser:(SFUserAccount *) userAccount {
-    static dispatch_once_t pred;
-    dispatch_once(&pred, ^{
++ (void)initialize {
+    if (self == [SFSDKSalesforceAnalyticsManager class] && analyticsManagerList == nil) {
         analyticsManagerList = [[NSMutableDictionary alloc] init];
```
khawkins: Ended up moving this into `initialize`, because there were two singleton methods that each needed access to it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -131,16 +131,25 @@ - (instancetype) initWithUser:(SFUserAccount *) userAccount {
...
-        self.analyticsManager = [[SFSDKAnalyticsManager alloc] initWithStoreDirectory:rootStoreDir dataEncryptorBlock:dataEncryptorBlock dataDecryptorBlock:dataDecryptorBlock deviceAttributes:deviceAttributes];
-        self.eventStoreManager = self.analyticsManager.storeManager;
-        self.remotes = [[NSMutableArray alloc] init];
-        SFSDKAnalyticsTransformPublisherPair *tpp = [[SFSDKAnalyticsTransformPublisherPair alloc] initWithTransform:[[SFSDKAILTNTransform alloc] init] publisher:[[SFSDKAILTNPublisher alloc] init]];
-        [self.remotes addObject:tpp];
+        _analyticsManager = [[SFSDKAnalyticsManager alloc] initWithStoreDirectory:rootStoreDir dataEncryptorBlock:dataEncryptorBlock dataDecryptorBlock:dataDecryptorBlock deviceAttributes:deviceAttributes];
+        _eventStoreManager = self.analyticsManager.storeManager;
+        _remotes = [[NSMutableArray alloc] init];
+        
+        // There's no standard for unauthenticated instrumentation publishing, currently.  Consumers
```
khawkins: Comment pretty much covers it, but just a heads-up.  This will likely be different for different consumer scenarios.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -289,8 +298,8 @@ - (BOOL) readAnalyticsPolicy {
 
 - (void) publishOnAppBackground {
 
-    // Publishing should only happen for the current user, not for all users signed in.
-    if (![self.userAccount.accountIdentity isEqual:[SFUserAccountManager sharedInstance].currentUser.accountIdentity]) {
+    // Publishing should only happen for the current user, and any anonymous stats, not for all users signed in.
+    if (self.userAccount != nil && ![self.userAccount.accountIdentity isEqual:[SFUserAccountManager sharedInstance].currentUser.accountIdentity]) {
```
khawkins: Seemed like this would be the place to publish unauthenticated analytics data (presuming publish actions were defined).  But I could also see making publishing of unauthenticated log lines the responsibility of the consumer too.  Thoughts?

brandonpage:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">Please re-submit this PR to dev, we may have already fixed your issue.</td>
    </tr>
  </tbody>
</table>

<p align="right" data-meta="generated_by_PR-Check">
</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.h
```diff
@@ -61,6 +61,13 @@
  */
 + (void) removeSharedInstanceWithUser:(nonnull SFUserAccount *) userAccount;
 
+/**
+ * Returns an instance of this class associated with an unauthenticated context (no authenticated user account).
+ *
+ * @return Instance of this class.
+ */
++ (nonnull instancetype) sharedUnauthenticatedInstance;
```
bhariharan: Could we add a `removeSharedUnauthenticatedInstance` convenience method as well?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
bhariharan: I would leave the publishing of unauthenticated events the responsibility of the consumer at this time, given that we have no generic strategy to publish at the moment.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -112,16 +104,24 @@ + (void) removeSharedInstanceWithUser:(SFUserAccount *) userAccount {
...
+    return [self initWithUser:nil];
 }
 
 - (instancetype) initWithUser:(SFUserAccount *) userAccount {
     self = [super init];
     if (self) {
-        self.userAccount = userAccount;
+        _userAccount = userAccount;
         SFSDKDeviceAppAttributes *deviceAttributes = [[self class] getDeviceAppAttributes];
         NSString *rootStoreDir = [[SFDirectoryManager sharedManager] directoryForUser:userAccount type:NSDocumentDirectory components:@[ kEventStoresDirectory ]];
```
bhariharan: Shouldn't this be something along these lines:
```
NSString *rootStoreDir = nil;
if (_userAccount != nil) {
    NSString *rootStoreDir = [[SFDirectoryManager sharedManager] directoryForUser:userAccount type:NSDocumentDirectory components:@[ kEventStoresDirectory ]];
} else {
    NSString *rootStoreDir = [[SFDirectoryManager sharedManager] globalDirectoryOfType:NSDocumentDirectory components:@[ kEventStoresDirectory ]];
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.h
khawkins: I can do that.  I wasn't sure if there was value, given that it's just a toggle in memory—there are no constraints on creating / utilizing the object.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
khawkins: I looked at the code in `SFDirectoryManager`, and it defaults to a global directory, without a user.  But that said, I agree that it's better to be more explicit.  I'll change it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.h
khawkins: @bhariharan We talked offline about the value of removal.  Looking through the code for authenticated analytics managers, removing the manager singleton doesn't do anything other than remove it from memory.  The managers rely upon the authentication notifications to remove events from their stores.

Seems like that's probably a better way to go: explicit store clearing, vs. object manipulation.  What do you think?

khawkins: (These are ultimately fronted by the `[SFSDKAnalyticsManager reset]` method.)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
khawkins: That seems reasonable to me.  I'll update/revert.

bhariharan:
@khawkins I agree with you. I don't think we should bring in any other enhancements for the unauthenticated use case until we have a way to upload it. Other enhancements could be in a common internal platform layer for now.

PR #2445, opened 2018-03-10
Unregistering from notifications before starting actual logout

wmathurin:
If we unregister after sending out the will logout notification, then the SFRestAPI instance for the user will have been torn down already.
The unregister call will cause a new SFRestAPI to be built for the user. That user object credentials will be taken away during logout but the SFRestAPI will stay.
If you log back in as the same user, you will get that SFRestAPI for your network calls. Because it holds a user without credentials, the first network call will cause the app to go back to login!

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
@@ -1474,6 +1474,13 @@ static NSException *authException = nil;
         [self.currentExpectation fulfill];
     };
 
+    // An success block that we expected to succeed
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
```diff
@@ -1421,25 +1421,6 @@ - (void)testBlockUpdate {
                              [self.currentExpectation fulfill];
                          }];
     [self waitForExpectation];
-    self.currentExpectation = [self expectationWithDescription:@"performUpsertWithObjectType-upserting contact"];
-    fields[LAST_NAME] = lastName;
-    [api performUpsertWithObjectType:CONTACT
```
wmathurin: That upserts works with an admin user but gives a 400 "insufficient access rights on cross-reference id" error when running with a standard user.

PR #2460, opened 2018-03-19
Latest doc

bhariharan:


bhariharan:
Yes, I noticed that too and was just talking to @rwhitley about it. We need to update the `ignore` section of the doc XML for the `docReact` target.

PR #2465, opened 2018-03-20
Cleanup SFRestAPI instance on kSFNotificationUserDidLogout Notification

trooper2013:
Moved the cleanup  of SFRestAPI instances  to kSFNotificationUserDidLogout. Had to hang on to some key attributes for the logout notification. Removed some unneeded code in test.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
@@ -342,6 +342,12 @@ static NSString *const  kOptionsClientKey          = @"clientIdentifier";
     }
 
     [SFSDKCoreLogger d:[self class] format:@"Logging out user '%@'.", user.userName];
+    
+    //save for use with didLogout notification
+    NSString *userId = user.credentials.userId;
+    NSString *orgId = user.credentials.organizationId;
+    NSString *communityId = user.credentials.communityId;
+    
     NSDictionary *userInfo = @{ kSFNotificationUserInfoAccountKey : user };
     [[NSNotificationCenter defaultCenter]  postNotificationName:kSFNotificationUserWillLogout
                                                          object:self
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -361,6 +367,12 @@ - (void)postPushUnregistration:(SFUserAccount *)user {
         [SFSecurityLockout clearPasscodeState];
     }
     [SFSDKWebViewStateManager removeSession];
+    
+    //restore these id's inorder to enable post logout cleanup of components
+    user.credentials.userId = userId;
```
khawkins: What do we think about changing this interface in a future release?  This seems like a reasonable workaround in the short term—and aligned with the current interface—but in general, it feels like continuing to have an `SFUserAccount` at this point is a violation of the idea that an `SFUserAccount` is an authenticated identity.  @trooper2013 what are your thoughts?

khawkins: By changing the interface, I mean sending something that represents this data in the `kSFNotificationUserDidLogout` notification, but not an `SFUserAccount`.

brandonpage:
## Danger has errored
```
 >    xcov.report(
 #        scheme: 'UnitTests',
```

<p align="right" data-meta="generated_by_CI-Result-SalesforceSDKCore">
</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
trooper2013: @khawkins agreed. Having partly hydrated SFUserAccount at this point does lend itself to unexpected problems, and violates the definition of an SFUserAccount.  if and when we do provide data, I think it will be necessary to provide some basic supporting functions around it, as in what does one do with these id's?  A small change  for 7.0 perhaps?

khawkins: I think it's probably a small change—and 7.0 is probably the right place, because it's an interface change.  Conceptually, we wouldn't be changing the dynamics of what's happening here, since this user is already unauthenticated at this point.

PR #2470, opened 2018-03-23
Clean up SFPreference instance during logout.

trooper2013:
Added a new Notification when all users of a given org have logged out. Also added SFNotificationUserInfo as a way to encapsulate post logout data for this new post logout Notification.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
@@ -348,6 +348,10 @@ static NSString * const kGlobalScopingKey = @"-global-";
 }
 
 NSString *SFKeyForUserAndScope(SFUserAccount *user, SFUserAccountScope scope) {
+    return  SFKeyForUserIdAndScope(user.credentials.userId,user.credentials.organizationId,user.credentials.communityId,scope);
+}
+
+NSString *SFKeyForUserIdAndScope(NSString *userId,NSString *orgId, NSString *communityId, SFUserAccountScope scope) {
     NSString *key = nil;
     switch (scope) {
         case SFUserAccountScopeGlobal:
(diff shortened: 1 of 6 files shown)


brandonpage:
## Danger has errored
```
 >    xcov.report(
 #        scheme: 'UnitTests',
```

<p align="right" data-meta="generated_by_CI-Result-SalesforceSDKCore">
</p>

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -348,27 +348,31 @@ + (NSString *)loginStateDescriptionFromLoginState:(SFUserAccountLoginState)login
...
     NSString *key = nil;
     switch (scope) {
         case SFUserAccountScopeGlobal:
             key = kGlobalScopingKey;
             break;
             
         case SFUserAccountScopeOrg:
-            if (user.credentials.organizationId != nil) {
-                key = user.credentials.organizationId;
+            if (userId != nil) {
```
bhariharan: I'm not sure I understand why this changed. For org level scope, all we need is `orgId` - it should not require any knowledge of `userId`. In this case, `key` should be `orgId`, not `userId`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.h
```diff
@@ -247,6 +255,13 @@ FOUNDATION_EXTERN NSString * const kSFUserInfoAddlOptionsKey;
 
 @end
 
+/** User Information for post logout notifications.
+ */
+@interface SFNotificationUserInfo : NSObject
```
khawkins: Should we consider instead just using `SFUserAccountIdentity`?  CommunityID doesn't explicitly come into play for logout, and SFUserAccountIdentity otherwise actually has all of the information identifying a user.

trooper2013: @khawkins Good idea. I can weave SFUserAccountIdentity into the core of this change, but the approach I took was based on the assumption that, more than just these id's maybe required in the future. Nevertheless I could add SFUserAccountIdentity to the SFNotificationUserInfo object. what do you think?

khawkins: The future-proofing idea is a good one.  Yeah, maybe let's make the account identity a member of that object?

PR #2479, opened 2018-03-28
Use API permissions-agnostic endpoint to validate session

khawkins:
`requestForResources` is an API endpoint that's dependent on the user having API access, which is unsuitable for testing sessions that don't have API access.  Replaced with the [user info endpoint](https://help.salesforce.com/articleView?id=remoteaccess_using_userinfo_endpoint.htm).

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
@@ -864,7 +864,7 @@ SFSDK_USE_DEPRECATED_BEGIN
      * instead of going through the entire OAuth dance all over again.
      */
     SFOAuthInfo *authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeRefresh];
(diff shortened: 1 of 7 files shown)


khawkins:
I'll put together a PR for Android as well.

bhariharan:
If I'm not mistaken, userinfo endpoint should use loginURL as base URL, not instanceURL, since it's part of the OAuth services which is a part of the login pod.

khawkins:
Ah, you're right @bhariharan !  I'll rework.

bhariharan:
The idea behind using SFRestAPI was to get the free token refresh. Switching to loginURL will negate that, since SFRestAPI is built to work against instanceURL. Maybe we should leave it the way you've done it, after all, instanceURL in this case is a pass through to loginURL, so it should always work + we get to refresh as well.

khawkins:
I made changes to point to the login host, for the userinfo OAuth endpoint.  Let me know what you think.

khawkins:
I might change this up to follow the Android approach a little more closely.  It sticks out that this is the only `SFRestRequest` factory method that sets up all of its endpoint information beforehand, whereas the others build it in at request time.

PR #2480, opened 2018-04-01
Use of a safe mutable set for managing active requests 

trooper2013:
@sgoldberg-sfdc @bhariharan Please Review. Small change to add some guard rails around the management of active requests in SFRestAPI.  Added a SFSDKSafeMutableSet to the core. We can certainly reuse this class.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSafeMutableSet.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSafeMutableSet.m
@@ -1,0 +1,164 @@
+/*
+ Copyright (c) 2018-present, salesforce.com, inc. All rights reserved.
+ 
+ Redistribution and use of this software in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this list of conditions
+ and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice, this list of
+ conditions and the following disclaimer in the documentation and/or other materials provided
+ with the distribution.
+ * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+ endorse or promote products derived from this software without specific prior written
+ permission of salesforce.com, inc.
+ 
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+ IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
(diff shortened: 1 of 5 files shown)


brandonpage:
I don't know why CI reports failure and the link is a 404, all test passed and the job to publish the results was successful.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSafeMutableSet.m
```diff
@@ -0,0 +1,183 @@
...
+
+- (void)unionSet:(NSSet *)set {
+    dispatch_barrier_async(self.queue, ^{
+        [self.backingSet unionSet:set];
+    });
+}
+
+- (void)minusSet:(NSSet *)set {
+    dispatch_barrier_async(self.queue, ^{
+        [self.backingSet unionSet:set];
```
wmathurin: Bad copy/paste: should be `minusSet`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSafeMutableSet.m
```diff
@@ -0,0 +1,183 @@
...
+
+- (void)intersectSet:(NSSet *)set {
+    dispatch_barrier_async(self.queue, ^{
+        [self.backingSet intersectSet:set];
+    });
+}
+
+- (void)setSet:(NSSet *)set {
+    dispatch_barrier_async(self.queue, ^{
+        [self.backingSet intersectSet:set];
```
wmathurin: Bad copy/paste: should be `setSet`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -72,7 +72,7 @@ - (id)initWithUser:(SFUserAccount *)user {
     self = [super init];
     if (self) {
         self.user = user;
-        _activeRequests = [[NSMutableSet alloc] initWithCapacity:10];
+        _activeRequests = [SFSDKSafeMutableSet setWithCapacity:10];
```
wmathurin: Shouldn't we use an ordered set instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -376,7 +377,7 @@ -(void)flushPendingRequestQueue:(NSError *)error rawResponse:(NSURLResponse *)ra
 
 - (void)sendPendingRequests {
     @synchronized (self) {
-        NSSet *pendingRequests = [self.activeRequests copy];
+        NSSet *pendingRequests = [self.activeRequests asSet];
```
wmathurin: (Same question as above), shouldn't we send those in order?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
trooper2013: @wmathurin  Should we change the SFRestAPI to use an ordered set given the prior implementation did not care about the order?

wmathurin: I see it has been a set for years. I guess if app needs the requests to go out in a deterministic order, it would wait on the response of a request before firing the next one.

PR #2481, opened 2018-04-03
Fixed Error Handling for offline crashes & unhandled error situations

trooper2013:
Fixing Error handling (default case) i.e. when delegates do not handle the error situation. There were reports of crashes offline & in some cases infinite loops in the default offline  error handling. Delegates are always notified  first, so need to inform delegates of errors again from the error handlers.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKAuthErrorManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKAuthErrorManager.m
@@ -99,14 +99,21 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
     self.networkFailureAuthErrorHandler = [[SFAuthErrorHandler alloc]
                                            initWithName:kSFNetworkFailureAuthErrorHandler
                                            evalOptionsBlock:^BOOL(NSError *error, SFOAuthInfo *authInfo, NSDictionary *options) {
+                                               BOOL result = NO;
                                                if ([[weakSelf class] errorIsNetworkFailure:error]) {
-
-                                                   if (self.networkErrorHandlerBlock) {
-                                                       self.networkErrorHandlerBlock(error, authInfo, options);
-                                                       return YES;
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKAuthErrorManager.m
```diff
@@ -99,14 +99,21 @@ - (SFAuthErrorHandlerList *)populateDefaultAuthErrorHandlerList
     self.networkFailureAuthErrorHandler = [[SFAuthErrorHandler alloc]
                                            initWithName:kSFNetworkFailureAuthErrorHandler
                                            evalOptionsBlock:^BOOL(NSError *error, SFOAuthInfo *authInfo, NSDictionary *options) {
+                                               BOOL result = NO;
```
trooper2013: Moved this code into the ErrorManager. The block is invoked only if needed and the error is passed on to the next handler in other cases

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1178,44 +1178,44 @@ - (void)populateErrorHandlers
...
     
     self.errorManager.genericErrorHandlerBlock = ^(NSError *error, SFOAuthInfo *authInfo,NSDictionary *options) {
         __strong typeof (weakSelf) strongSelf = weakSelf;
-        SFSDKOAuthClient *client = [options objectForKey:kErroredClientKey];
-        [strongSelf disposeOAuthClient:client];
-        SFOAuthCredentials *credentials = [strongSelf newClientCredentials];
-        [strongSelf dismissAuthViewControllerIfPresent];
-        SFSDKOAuthClient *newClient = [strongSelf fetchOAuthClient:credentials completion:client.config.successCallbackBlock failure:client.config.failureCallbackBlock];
-        [newClient refreshCredentials];
+
```
trooper2013: Inlined the required handling along with a common error alert.

PR #2484, opened 2018-04-04
Config files for hybrid app go under www

wmathurin:
We had them in 2 build copy phases. Getting rid of the useless one.

### DIFF
--- a/hybrid/SampleApps/SmartSyncExplorerHybrid/SmartSyncExplorerHybrid.xcodeproj/project.pbxproj
+++ b/hybrid/SampleApps/SmartSyncExplorerHybrid/SmartSyncExplorerHybrid.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 49;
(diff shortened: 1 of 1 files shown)


rwhitley:
+1

hybrid/SampleApps/SmartSyncExplorerHybrid/SmartSyncExplorerHybrid.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 49;
```
brandonpage: 👍   But do we know how this happens?  A few months ago this broke the SmartStore unit tests (because 49 doesn't exist) and I had to make this change.  Is it just a bug in xcode?

wmathurin: That XCode did - I can revert it.

bhariharan:
👍

PR #2486, opened 2018-04-05
 Use IV for encryption in external storage

trooper2013:
@wmathurin and @bhariharan  please review. Here is what is done.

1. saveSoupEntryExternally simply uses an IV everytime a write occurs.
2. loadExternalSoupEntry attempts to load and decrypt the file using the given IV, if not it attempts to load it again with a null IV (backward compatibility), if it does decrypt successfully with a null IV, it rewrites the file with a IV (migrating up).
3. A simulated test for the above. i.e. saveSoupEntryExternally with null IV, read it & ensure that the migration occurred.

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
@@ -765,7 +765,8 @@ NSString *const EXPLAIN_ROWS = @"rows";
     SFSmartStoreEncryptionKeyBlock keyBlock = [SFSmartStore encryptionKeyBlock];
     if (keyBlock) {
         SFEncryptStream *encryptStream = [[SFEncryptStream alloc] initToFileAtPath:filePath append:NO];
-        [encryptStream setupWithKey:keyBlock().key andInitializationVector:nil];
+        SFEncryptionKey *encKey = keyBlock();
+        [encryptStream setupWithKey:encKey.key andInitializationVector:encKey.initializationVector];
         outputStream = encryptStream;
     } else {
         outputStream = [[NSOutputStream alloc] initToFileAtPath:filePath append:NO];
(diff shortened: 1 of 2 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -792,13 +793,22 @@ - (BOOL)saveSoupEntryExternally:(NSDictionary *)soupEntry
...
               soupTableName:(NSString *)soupTableName {
+    SFSmartStoreEncryptionKeyBlock keyBlock = [SFSmartStore encryptionKeyBlock];
+    SFEncryptionKey *encKey = nil;
+    if (keyBlock){
+       encKey = keyBlock();
+    }
+    return [self loadExternalSoupEntry:soupEntryId soupTableName:soupTableName withKey:encKey.key andIV:encKey.initializationVector];
+}
+
+- (id)loadExternalSoupEntry:(NSNumber *)soupEntryId
```
khawkins: Minor: Why not just send in the `SFEncryptionKey` as the parameter?

libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -810,14 +820,28 @@ - (id)loadExternalSoupEntry:(NSNumber *)soupEntryId
...
+            }
+            
+        } else {
+           NSString *errorMessage = [NSString stringWithFormat:@"Loading external soup from file failed! encrypted: %@, soupEntryId: %@, soupTableName: %@, filePath: '%@', error: %@.",
+                                      key ? @"YES" : @"NO",
+                                      soupEntryId,
+                                      soupTableName,
+                                      filePath,
+                                      error];
+            NSAssert(NO, errorMessage);
```
khawkins: I know this was here before, but this is not a good idea as a runtime exception generator, and it's exacerbated in this PR by further hiding the source of the issue—i.e. it doesn't fail fast.  The reason it's not good as a runtime check is because, if assertions are turned off, it's not going to fail _at all_.  Instead, it's going to drop through, and you're going to end up with unexpected behavior that's hard to track down.

`NSAssert` is good for sanity checks on arguments that would reasonably be caught during development cycles.  It's not good for user input, or data that's otherwise dynamically populated as a runtime experience.  If you truly need to fail the app in this scenario, you should raise an exception.

libs/SmartStore/SmartStore/Classes/SFSmartStore.m
trooper2013: @khawkins  Thats because I need to send in a nill IV at some point.  result = [self loadExternalSoupEntry:soupEntryId soupTableName:soupTableName withKey:key andIV:nil];

libs/SmartStore/SmartStore/Classes/SFSmartStore.m
trooper2013: @khawkins I will replace it with an exception.

libs/SmartStore/SmartStore/Classes/SFSmartStore.m
khawkins: I see now.  I missed the recursive call.

PR #2487, opened 2018-04-06
 Comment out test Codable's deserialization for nested dictionary.

trooper2013:
Changes in Swift 4.1 breaks deserialization of nested dictionaries.  Will adjust the code for Codable in the tests later to use a dynamic key value container.


khawkins:
To be fair, it cleans it up. `Any` is not _de facto_ `Decodable`.  It's the interface that needs changing.

PR #2496, opened 2018-04-16
Fix login host list nav bar color issue

harry528tt:
@trooper2013 @bhariharan @ssraza
We've seen an branding issue happened on the "Choose Connection" page. 
After looking into it, I found the issue can be solved by moving the setting up branding part from viewDidLoad() into viewWillAppear().

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostListViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostListViewController.m
@@ -139,20 +139,6 @@ static NSString * const SFDCLoginHostListCellIdentifier = @"SFDCLoginHostListCel
                                              target:nil
                                              action:nil];
     self.navigationItem.leftBarButtonItem = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemCancel target:self action:@selector(cancelLoginPicker:)];
-
-    SFSDKLoginViewControllerConfig *config = [SFUserAccountManager sharedInstance].loginViewControllerConfig;
-    
-    if (config.navBarColor) {
-        [self.navigationController.navigationBar setBarTintColor:config.navBarColor];
-    }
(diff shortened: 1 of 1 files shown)


bhariharan:
@harry528tt Are you able to reproduce this issue on our sample apps?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostListViewController.m
```diff
@@ -175,13 +161,29 @@ - (void)viewWillAppear:(BOOL)animated {
...
-    
+    [self setupBrandingForNavBar];
     [self.tableView reloadData];
     [self resizeContentForPopover];
     // style navigiation bar
     [super viewWillAppear:animated];
 }
 
+- (void)setupBrandingForNavBar {
+    [self.navigationController.navigationBar setTranslucent:false];
```
bhariharan: Why are we setting translucent to false?

trooper2013: @harry528tt I agree with @bhariharan.  Why cant this be done in your own subclass?

harry528tt: @bhariharan @trooper2013 I can move the setTranslucent to my own class. But for the remaining part, yes it's also happening on your sample app.

harry528tt:
@bhariharan @trooper2013 I took a look the "[self.navigationController.navigationBar setTranslucent:false];" again, I still think this should be added here instead of my own class. There are two main reasons, 1. It's also happening on the sample app,(Click gear button, the Choose Connection nav bar color got slightly lighter), but because of the current branding color for sample app's nav bar is Salesforce blue, it's very hard to notice. 2. We don't call this class in our code directly, we use SFLoginViewController, so we can't override the viewWillAppear for this class. Please let me know what you think.

PR #2499, opened 2018-04-19
Allow refresh credentials to use credentials.domain

trooper2013:
When credentials are specified use the credentials.domain if available in-order to setup the loginHost for refresh.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1277,7 +1277,7 @@ - (SFSDKOAuthClient *)fetchOAuthClient:(SFOAuthCredentials *)credentials cached:
         __weak typeof(self) weakSelf = self;
         client = [SFSDKOAuthClient clientWithCredentials:credentials configBlock:^(SFSDKOAuthClientConfig *config) {
             __strong typeof(self) strongSelf = weakSelf;
-            config.loginHost = strongSelf.loginHost;
+            config.loginHost = credentials.domain?:strongSelf.loginHost;
```
bhariharan: When and why would the 2 values differ?

bhariharan: What the code is doing here seems a little inconsistent. The rest of the values here are being set by the configured values in `strongSelf`, none of the other values are being set from the `credentials` passed in. @trooper2013, @ssraza and I had a quick meeting and we have an alternate solution in place.

trooper2013:
Need to revisit later.

PR #2502, opened 2018-04-23
 Add condition to ensure next active window isn't currently hidden

haoguan:
We ran into an issue where foregrounding the application incorrectly set the next key window to a `_UIInteractiveHighlightEffectWindow` presumably created by the OS and only present in iOS 11.1 builds. This window happens to be the last window in the list and passes the alpha and keyboard if conditions. To resolve this, I added an extra condition to ensure the next active window is not currently set to be hidden. 

I also took the time to do a bit of refactoring by reusing the `findActiveWindow` method.

@bhariharan @trooper2013


bhariharan:
@trooper2013 is working on fixing some other issues with windowing. We might want to hold off on this until that goes in - that might address this issue as well. Stay tuned!

PR #2508, opened 2018-04-27
Fix encryptUsingRSAforData plaintext length

sgoldberg-sfdc:
-  strlen assumes the data is null terminated. This was leading to extraneous characters sometimes appearing at end of decrypted text.


khawkins:
Can we add or update a test for this?  The previous test should have otherwise caught this disparity.

PR #2509, opened 2018-04-27
Validate plain text is not assumed to be null terminated

sgoldberg-sfdc:
The previous encryptUsingRSAForData assumed the input data was null terminated, as it used strlen to determine length of input data. 

This was randomly picking up extraneous data depending what was in memory beyond the bytes of the NSData plain text.


sgoldberg-sfdc:
@khawkins

PR #2521, opened 2018-05-11
Add support to return non-mutable dictionary from SFSDKSafeMutableDictionary

sgoldberg-sfdc:


sgoldberg-sfdc:
@bhariharan @trooper2013 @khawkins

PR #2523, opened 2018-05-17
 Window implementation to handle rotations and restore use of alpha

trooper2013:
Another stab at resolving issues related to multi-windowing approaches in the SDK. Multiple visible UI windows tend to have some really bad side effects with rotations (keyboard and views) and  status bar. We previously resorted to using the hidden property, unfortunately using hidden property on the UIWindow leads to  really bad flicker issue ( black screen ).  Reverted back to using alpha with a slightly different strategy.

A debugging  of UIKIT revealed the following facts. 

1.  All UIWindows are rotated when rotation occurs. 

2. All preference calls are delegated to the window's rootviewcontroller if present. 

3. Multiple windows with different behaviors will lead to weird UI experience.  For instance a visible window  may be locked to portrait mode, but during rotation the status bar will still continue to rotate because another window may allow rotations. It will also lead to  keyboard window being in the wrong orientation.

Strategy used.

- Stash(nullify) the rootviewcontroller when the window is presented and unstash(restore) when dismissed. Extended UIWindow (SFSDKUIWindow) to handle the stash and unstash.
- Windows are created lazily and the references are removed when  the windows are dismissed.

NOTE: 
For Applications that care about rotations, should use the SFSDKUIWindow instead of the UIWindow and also remove and re-attach their views on viewDidAppear  as implemented in the SFSDKNavigationController.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -99,6 +99,16 @@ - (void)viewDidLayoutSubviews {
 
 - (void)viewWillAppear:(BOOL)animated {
     [super viewWillAppear:animated];
+    UIWindow *currentWindow = (UIWindow *) self.view.window;
```
wmathurin: Should we move that block of code to some helper method (maybe in the SFSDKUIWindow class) so that one would just need a one line call?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -32,11 +32,81 @@
...
+
+- (instancetype)initWithFrame:(CGRect)frame andName:(NSString *)windowName {
+    self = [super initWithFrame:frame];
+    if (self) {
+        _windowName = windowName;
+    }
+    return self;
+}
+
+- (void)stashRootViewController {
```
wmathurin: Could you put the comment from the PR in this class to explain what is going on and why we are doing it this way?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
trooper2013: @wmathurin Its a good suggestion,  should we just add it to a base UIViewController 
 SDK instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowContainer.m
```diff
@@ -49,49 +56,50 @@ - (instancetype)initWithWindow:(UIWindow *)window name:(NSString *) windowName {
 
 - (UIWindow *)window {
     if (_window == nil) {
-        _window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
-         _window.hidden = YES;
-        _window.rootViewController = _viewController;
+        _window = [[SFSDKUIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds  andName:_windowName];
```
trooper2013: Window is now lazily loaded. No need to create these upfront.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
wmathurin: Are you thinking of swizzling in a new implementation of viewWillAppear for UIViewController that would refresh the views or to simply add a helper method to the class via a category?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
trooper2013: This stash function is called when the window is not the key responder anymore. When another window is make key & visible, UIWindow:resignKeyWindow will be invoked by UIKIT. This is when we stash(copy then nullify the reference) for the rootviewcontroller.

wmathurin: What I meant earlier, is could you add the comments in the source code so that our future selfs can make sense of the code without having to do a git blame etc.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -32,11 +32,81 @@
...
+    if (self.rootViewController) {
+        _stashedController = self.rootViewController;
+        super.rootViewController = nil;
+    }
+}
+
+- (void)unstashRootViewController {
```
trooper2013: This unstash function is called when the window is made key window. UIKIT invokes the UIWindow:becomeKeyWindow. This is when we unstash (set) the rootviewcontroller.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -32,11 +32,81 @@
...
+    }
+}
+@end
+
 @interface SFSDKWindowManager()<SFSDKWindowContainerDelegate>
 
 @property (nonatomic, strong) NSHashTable *delegates;
 @property (nonatomic, strong,readonly) NSMapTable<NSString *,SFSDKWindowContainer *> * _Nonnull namedWindows;
-@property (nonatomic, strong,readonly) NSMapTable<UIWindow *,SFSDKWindowContainer *> * _Nonnull reverseLookupTable;
+@property (nonatomic,weak) SFSDKWindowContainer *lastActiveWindow;
```
trooper2013: Moved snapshot handling logic to the WindowManager. Capture the last active window when the snapshot is being present and restore when its dismissed.

libs/SmartStore/SmartStore/Classes/SFSmartStoreInspectorViewController.m
```diff
@@ -456,11 +451,14 @@ - (void)layoutSubviews
     [self.resultGrid reloadData];
 }
 
-- (BOOL)shouldAutorotateToInterfaceOrientation:(UIInterfaceOrientation)interfaceOrientation
-{
+- (BOOL)shouldAutorotate {
```
trooper2013: removed use of old deprecated apis.

libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
```diff
@@ -160,7 +160,7 @@ - (void)fixFor530Bug:(NSString *)storeName key:(NSString *)key {
     }
 }
 
-- (FMDatabaseQueue *)openStoreQueueWithName:(NSString *)storeName key:(NSString *)key error:(NSError **)error {
+- (FMDatabaseQueue *)openStoreQueueWithName:(NSString *)storeName key:(NSString *)key error:(NSError * __autoreleasing *)error {
```
trooper2013: took care of warning.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
trooper2013: I was thinking its easier for devs to call super instead if class is extended.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowContainer.m
```diff
@@ -49,49 +56,50 @@ - (instancetype)initWithWindow:(UIWindow *)window name:(NSString *) windowName {
...
+        //_window.hidden = NO;
```
bhariharan: Remove commented out code.

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/AppDelegate.m
```diff
@@ -129,7 +130,7 @@ - (void)setUserLoginStatus :(BOOL) loggedIn {
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
 {
-    self.window = [[UIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
+    self.window = [[SFSDKUIWindow alloc] initWithFrame:[UIScreen mainScreen].bounds];
```
bhariharan: Do we need these changes in the template apps too?

trooper2013: @bhariharan IMO having it here is a good for us to test rotations and should also serve as an example for developers. I have added notes here.

bhariharan:
@trooper2013 It looks like there are conflicts in `libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKWindowManagerTests.m` that need to be resolved.

PR #2524, opened 2018-05-18
Fix intermittent crash on logout when accessing SFUserAccount.apiURL in Swift

sgoldberg-sfdc:
-  This property may actually return a nil due to instanceURL being nullable.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
@@ -79,7 +79,7 @@ typedef NS_ENUM(NSUInteger, SFUserAccountLoginState) {
  on the server-side. This URL takes into account the
  current community if available (see `communityId`).
  */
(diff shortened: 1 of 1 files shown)


jamesward:
@sgoldberg-sfdc I've invited you to this org, visit https://github.com/forcedotcom to accept the invite.  Then go to https://cla.salesforce.com/status/forcedotcom/SalesforceMobileSDK-iOS/pull/2524 to revalidate the PR.

khawkins:
This reflects the underlying nullability of the properties it fronts, so 👍 to that.  That said, a nil `apiUrl` at the user account level suggests an improperly configured set of credentials.  User accounts are implicitly authenticated, and authenticated credentials should not have a nil `apiUrl`.

Somewhere along the line, there's a reconciliation that should happen between `SFOAuthCredentials` and `SFUserAccount`.  But that may be an exercise for another time.

bhariharan:
Spoke with @sgoldberg-sfdc and it looks like we have code paths that set these properties to `nil`.

PR #2529, opened 2018-05-24
Fixed compilation issue with Swift tests

trooper2013:


trooper2013:
@wmathurin I suppose its just stricter code checking for optionals in Xcode 9.3.1 (swift 4.1)

PR #2534, opened 2018-05-29
Fixing provisioning of test targets and import of SFSDK_DEPRECATED

bhariharan:


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncCacheManager.h
```diff
@@ -22,6 +22,8 @@
  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#import <SalesforceSDKCore/SalesforceSDKConstants.h>
```
bhariharan: These imports are required for `SFSDK_DEPRECATED`. It wasn't complaining on the other targets because it's imported in the `PCH` file.

libs/SalesforceReact/SalesforceReact.xcodeproj/project.pbxproj
```diff
@@ -313,6 +315,20 @@
 		};
 /* End PBXContainerItemProxy section */
 
+/* Begin PBXCopyFilesBuildPhase section */
```
bhariharan: The `SalesforceReactTestApp` should embed the `SalesforceReact` framework binary like the other test targets.

PR #2551, opened 2018-06-13
Updating CocoaLumberjack version

nuvolarmobileteam:
Currently the project uses a CocoaLumberjack version that produces some warnings at build time. Those issues have been fixed in newer versions of CocoaLumberjack library.
I suggest updating the version to one that does not produce warnings.
screen shot 2018-06-13 at 13 45 22

### DIFF
--- a/SalesforceAnalytics.podspec
+++ b/SalesforceAnalytics.podspec
@@ -19,7 +19,7 @@ Pod::Spec.new do |s|
 
   s.subspec 'SalesforceAnalytics' do |sdkanalytics|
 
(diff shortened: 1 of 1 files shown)


bhariharan:
@nuvolarmobileteam It's not as simple as bumping up the version in the `podspec` within our workspace. We have made some custom changes to `CocoaLumberjack` to accommodate binary generation for our hybrid apps and some other use cases. The upgrade will need to take those into account as well, i.e. we will need to upgrade our fork. We will consider this in a future release. For now, you can bump it up in your app's workspace in the `podspec`.

Issue #2544, opened 2018-06-08
Switch user method fails to switch the user  when the id’s of users are same (Case Insensitive)


nzsrini:
Version of Mobile SDK Used: 6.1.0

Issue found in Native App or Hybrid App: Native iOS app
OS Version: 11.3.0
Device: iPad

Assume we have userid's from salesforce which 15 digit number( Case senstive not same).
005300000040EVc
005300000040EvC

(We have this case scenario).
005300000040EVc: UserA :user_id
005300000040EvC: UserB:user_id

As the users are from the same organization. The SDK below code returns true for the above condition even though both users are different.

if ([self.currentUser.accountIdentity isEqual:newCurrentUser.accountIdentity]) {
[SFSDKCoreLogger w:[self class] format:@"%@ new user identity is the same as the current user. No action taken.", NSStringFromSelector(_cmd)];

SFUserAccountIdentity.m : compares ca

BOOL userIdsEqual = [objectToCompare.userId isEqualToEntityId:self.userId];
BOOL orgIdsEqual = [objectToCompare.orgId isEqualToEntityId:self.orgId];
return userIdsEqual && orgIdsEqual;

==== isEqultoEntitiyId isn't comparing case sensitive so it is returning it as same user
code in NSString + SFAdditions.mp
if ([self caseInsensitiveCompare:entityId] == NSOrderedSame) return YES; /

Please can you look into it .

nzsrini:
any update on this Please, we need to fix the issue pretty urgently

trooper2013:
@nzsrini the 15 digit User Id's are case sensitive and cannot be compared as you suggest. You should look into why the case of the identifier has changed.

nzsrini:
@trooper2013  Hi Rao,

What we are saying is the SDK class (SFUserAccountManager.m) --switchToUser is comparing with 15 digit userid rather than 18 digit and not switching the USer, And logging the below message.
    [SFSDKCoreLogger w:[self class] format:@"%@ new user identity is the same as the current user.  No action taken.", NSStringFromSelector(_cmd)];
====== Function=========================
(void)switchToUser:(SFUserAccount *)newCurrentUser {
    if ([self.currentUser.accountIdentity isEqual:newCurrentUser.accountIdentity]) {
        [SFSDKCoreLogger w:[self class] format:@"%@ new user identity is the same as the current user.  No action taken.", NSStringFromSelector(_cmd)];
=====

Please can you look into it.

[...]

PR #2556, opened 2018-06-16
Fixing isEqualToEntityId 

wmathurin:
Added a test which fails without the fix

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSString+SFAdditions.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSString+SFAdditions.m
@@ -213,10 +213,12 @@ static inline BOOL IsValidEntityId(NSString *string) {
 }
 
 - (BOOL)isEqualToEntityId:(NSString*)entityId {
-    if ([self caseInsensitiveCompare:entityId] == NSOrderedSame) return YES; // for entityId like `me`
-    
-    if (![entityId length]) return NO;
-    if (!IsValidEntityId(self) || !IsValidEntityId(entityId)) return NO;
+    if (!IsValidEntityId(self) || !IsValidEntityId(entityId)) {
+        if ([self caseInsensitiveCompare:entityId] == NSOrderedSame)
+            return YES; // for entityId like `me`
+        else
+            return NO;
+    }
     
     NSString *id18self = ([self length] == SFEntityIdLength18) ? self : [self entityId18];
     NSString *id18other = ([entityId length] == SFEntityIdLength18) ? entityId : [entityId entityId18];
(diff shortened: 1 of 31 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSString+SFAdditions.m
```diff
@@ -213,10 +213,12 @@ - (nullable NSString*)entityId18 {
 }
 
 - (BOOL)isEqualToEntityId:(NSString*)entityId {
-    if ([self caseInsensitiveCompare:entityId] == NSOrderedSame) return YES; // for entityId like `me`
```
wmathurin: We only want to do the case insensitive compare for the "me" case - in other words when not dealing with valid entity ids.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSString+SFAdditions.m
```diff
@@ -213,10 +213,12 @@ - (nullable NSString*)entityId18 {
...
-    
-    if (![entityId length]) return NO;
```
wmathurin: This one is weird, empty string would not be equal to itself.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSString+SFAdditions.m
```diff
@@ -213,10 +213,12 @@ - (nullable NSString*)entityId18 {
...
-    if (!IsValidEntityId(self) || !IsValidEntityId(entityId)) return NO;
+    if (!IsValidEntityId(self) || !IsValidEntityId(entityId)) {
```
wmathurin: If one or both are invalid ids, compare without looking at case (for the "me" - i guess we allow "me" and "Me" ??).

libs/SalesforceSDKCore/SalesforceSDKCoreTests/NSString+SFAdditionsTests.m
```diff
@@ -0,0 +1,82 @@
...
+
+    XCTAssertTrue([USER_B_ID_15 isEqualToEntityId:USER_B_ID_15]);
+    XCTAssertTrue([USER_B_ID_15 isEqualToEntityId:USER_B_ID_18]);
+    XCTAssertTrue([USER_B_ID_18 isEqualToEntityId:USER_B_ID_15]);
+    XCTAssertTrue([USER_B_ID_18 isEqualToEntityId:USER_B_ID_18]);
+
+    XCTAssertFalse([USER_A_ID_15 isEqualToEntityId:@""]);
+    XCTAssertFalse([USER_A_ID_15 isEqualToEntityId:ME]);
+    XCTAssertFalse([USER_A_ID_15 isEqualToEntityId:OTHER_ME]);
+    XCTAssertFalse([USER_A_ID_15 isEqualToEntityId:USER_B_ID_15]);
```
wmathurin: This line fails with the old code.

wmathurin:
A customer actually had two users on one device with identical ids when ignoring the case. See https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/2544

wmathurin:
@bhariharan I agree with you. But given the code (and comment), the callers of this method expects it to handle fake ids like "me". For 6.2, we should keep it backwards compatible.

PR #2558, opened 2018-06-18
Only requests requiring authentication are replayed after a refresh

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -380,11 +380,13 @@ -(void)flushPendingRequestQueue:(NSError *)error rawResponse:(NSURLResponse *)ra
     }
 }
 
-- (void)sendPendingRequests {
+- (void)resendActiveRequestsRequiringAuthentication {
```
wmathurin: Renamed method to better reflect what it now does

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -440,7 +442,9 @@ - (SFRestRequest *)requestForUserInfo {
 
 - (SFRestRequest *)requestForVersions {
     NSString *path = @"/";
-    return [SFRestRequest requestWithMethod:SFRestMethodGET path:path queryParams:nil];
+    SFRestRequest* request = [SFRestRequest requestWithMethod:SFRestMethodGET path:path queryParams:nil];
```
wmathurin: This API doesn't need an auth token. We use it in the new test below - but we needed SFRestAPI to know what authentication was not required.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFNativeRestRequestListener.h
```diff
@@ -29,6 +29,7 @@
 @interface SFNativeRestRequestListener : SFSDKTestRequestListener <SFRestDelegate>
 
 @property (nonatomic, strong) SFRestRequest *request;
+@property (nonatomic, assign) NSTimeInterval sleepDuringLoad;
```
wmathurin: To slow down the consumption of the response. Without it, the request will get out of the activeRequests array before the refresh completes in the new test below.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
```diff
@@ -1299,6 +1299,62 @@ -(void)testInvalidAccessToken_MultipleRequests {
...
 }
 
+// - set an invalid access token (simulate expired)
+// - make multiple simultaneous requests (some not requiring authentication)
+// - requests not requiring authentication should succeed and only be sent once
+// - other requests will fail in some arbitrary order
+// - ensure that a new access token is retrieved using refresh token
+// - ensure that all requests eventually succeed
+//
+-(void)testInvalidAccessToken_UnAuthAndAuthRequests {
```
wmathurin: And here is the new test.
NB: this test fails without the fix above.

wmathurin:
The change is trivial, but getting a test to demonstrate the issue took a while.
If you comment out the fix, the new test will fail.

PR #2561, opened 2018-06-19
customRegistrationKeys for push notification

wfanforever:


bhariharan:
That works for me. @wfanforever Do you want to send an updated PR?

PR #2563, opened 2018-06-20
Using Mobile SDK user agent for login and oauth requests

wmathurin:
iOS used to only set the Mobile SDK user agent in SFRestAPI and in SFHybridViewController.
On Android all network requests already use the Mobile SDK user agent.
With this PR, iOS will now behave like Android and use the Mobile SDK user agents for all network requests.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -355,7 +355,7 @@ static NSString * const kSFECParameter = @"ec";
         _view.autoresizingMask = UIViewAutoresizingFlexibleHeight | UIViewAutoresizingFlexibleWidth;
          _view.clipsToBounds = YES;
         _view.translatesAutoresizingMaskIntoConstraints = NO;
-        
+        _view.customUserAgent = [SalesforceSDKManager sharedManager].userAgentString(@"");
         _view.UIDelegate = self;
     }
     return _view;
@@ -368,9 +368,6 @@ static NSString * const kSFECParameter = @"ec";
 {
     self.authenticating = NO;
     self.advancedAuthState = SFOAuthAdvancedAuthStateNotStarted;
-    if (info.authType == SFOAuthTypeUserAgent) {
-        [self resetWebUserAgent];
-    }
     if ([self.delegate respondsToSelector:@selector(oauthCoordinator:didFailWithError:authInfo:)]) {
         [self.delegate oauthCoordinator:self didFailWithError:error authInfo:info];
     } else if ([self.delegate respondsToSelector:@selector(oauthCoordinator:didFailWithError:)]) {
@@ -386,9 +383,6 @@ static NSString * const kSFECParameter = @"ec";
 {
     self.authenticating = NO;
     self.advancedAuthState = SFOAuthAdvancedAuthStateNotStarted;
-    if (authInfo.authType == SFOAuthTypeUserAgent) {
-        [self resetWebUserAgent];
-    }
     if ([self.delegate respondsToSelector:@selector(oauthCoordinatorDidAuthenticate:authInfo:)]) {
         [self.delegate oauthCoordinatorDidAuthenticate:self authInfo:authInfo];
     } else if ([self.delegate respondsToSelector:@selector(oauthCoordinatorDidAuthenticate:)]) {
@@ -472,8 +466,6 @@ static NSString * const kSFECParameter = @"ec";
         return;
     }
     
-    [self configureWebUserAgent];
-
     self.initialRequestLoaded = NO;
     
     // notify delegate will be begin authentication in our (web) vew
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -887,32 +879,6 @@ - (void) updateCredentials:(NSDictionary *) params {
 
 }
 
-- (void)configureWebUserAgent
```
wmathurin: These methods don't actually set or reset the user agent on the web view. They simply change the user defaults.
It's enough when using cordova (because it looks for the user agent in the user defaults and sets it on its web view) but it doesn't do anything otherwise.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -355,7 +355,7 @@ - (WKWebView *)view {
         _view.autoresizingMask = UIViewAutoresizingFlexibleHeight | UIViewAutoresizingFlexibleWidth;
          _view.clipsToBounds = YES;
         _view.translatesAutoresizingMaskIntoConstraints = NO;
-        
+        _view.customUserAgent = [SalesforceSDKManager sharedManager].userAgentString(@"");
```
wmathurin: Setting the user agent for the login webview.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -192,11 +189,7 @@ + (NSString *)userAgentString {
 }
 
 + (NSString *)userAgentString:(NSString*)qualifier {
-    NSString *returnString = @"";
-    if ([SalesforceSDKManager sharedManager].userAgentString != NULL) {
-        returnString = [SalesforceSDKManager sharedManager].userAgentString(qualifier);
-    }
-    return returnString;
+    return [SalesforceSDKManager sharedManager].userAgentString(qualifier);
```
wmathurin: [SalesforceSDKManager sharedManager].userAgentString cannot be null so no need to check for NULL.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -170,11 +170,6 @@ - (NSURLRequest *)prepareRequestForSend:(SFUserAccount *)user {
             [self.request setValue:bearer forHTTPHeaderField:@"Authorization"];
         }
 
-        // Sets Mobile SDK user agent on REST API requests, if it hasn't been set already elsewhere.
```
wmathurin: No longer setting user agent from SFRestRequest, instead doing it in SFNetwork.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -67,7 +68,12 @@ - (instancetype)initWithBackgroundSession {
     return self;
 }
 
-- (NSURLSessionDataTask *)sendRequest:(NSURLRequest *)urlRequest dataResponseBlock:(SFDataResponseBlock)dataResponseBlock {
+- (NSURLSessionDataTask *)sendRequest:(NSMutableURLRequest *)urlRequest dataResponseBlock:(SFDataResponseBlock)dataResponseBlock {
+    // Sets Mobile SDK user agent if it hasn't been set already elsewhere.
+    if (![urlRequest.allHTTPHeaderFields.allKeys containsObject:@"User-Agent"]) {
```
wmathurin: Setting user agent in SFNetwork.

wmathurin: That way it applies to oauth requests as well.

wmathurin:
Tested with Charles (with apps before and after the change).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -887,32 +879,6 @@ - (void) updateCredentials:(NSDictionary *) params {
...
-        [[NSUserDefaults msdkUserDefaults] registerDefaults:userAgentDict];
-    }
-}
-
-- (void)resetWebUserAgent
-{
-    if (self.userAgentForAuth != nil) {
-        // If the current web user agent has not changed from the one we set, reset it.  Otherwise, assume it's
-        // already been altered out of band, and we shouldn't touch it.
-        NSString *currentWebUserAgent = [[NSUserDefaults msdkUserDefaults] objectForKey:kOAuthUserAgentUserDefaultsKey];
```
bhariharan: 👍 This was the legacy way to do it for `UIWebView` if I'm not mistaken.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
bhariharan: Would this be set beforehand for requests originating from `SmartSync`? Where would the `qualifier` come in for `SmartSync` requests?

wmathurin:
@bhariharan SmartSync sets the header before calling sendRequest (see [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/SmartSync/SmartSync/Classes/Util/SFSmartSyncNetworkUtils.m#L36).
Therefore the user agent will already be set by the time it gets to SFNetwork:setRequest - and it won't be overwritten thanks to `if (![urlRequest.allHTTPHeaderFields.allKeys containsObject:@"User-Agent"])`.
(I also tested that with Charles).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
sgoldberg-sfdc: Thats correct!

PR #2564, opened 2018-06-20
Dev info dialog works over login screen

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -403,9 +403,9 @@ - (void)setIsDevSupportEnabled:(BOOL)isDevSupportEnabled {
 
 - (void)showDevSupportDialog
 {
-    SFSDKWindowContainer * mainWindow = [SFSDKWindowManager sharedManager].mainWindow;
-    if ([self isDevSupportEnabled] && mainWindow.isEnabled) {
-        UIViewController * topViewController = mainWindow.topViewController;
+    SFSDKWindowContainer * activeWindow = [SFSDKWindowManager sharedManager].activeWindow;
```
wmathurin: Needed to use activeWindow instead of mainWindow.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -487,10 +488,14 @@ - (NSArray*) getDevSupportInfos
     return devInfos;
 }
 
-- (NSString*) usersToString:(NSArray*)userAccounts {
+- (NSString*) userToString:(SFUserAccount*)user {
```
wmathurin: Otherwise it crashes when trying to put a nil in the devInfos array (in the case we have not logged in yet).

libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
```diff
@@ -76,6 +76,8 @@ + (SFSmartStoreDatabaseManager *)sharedManager
 + (SFSmartStoreDatabaseManager *)sharedManagerForUser:(SFUserAccount *)user
 {
     @synchronized (self) {
+        if (user == nil) return nil;
```
wmathurin: We shouldn't be able to create database manager for nil.

libs/SmartStore/SmartStore/Classes/SmartStoreSDKManager.m
```diff
@@ -97,7 +97,7 @@ -(NSArray*) getDevActions:(UIViewController *)presentedViewController
 
 - (NSArray*) getDevSupportInfos
 {
-    SFSmartStore *store = [SFSmartStore sharedStoreWithName:kDefaultSmartStoreName];
+    SFSmartStore *store = [SFSmartStore sharedGlobalStoreWithName:kDefaultSmartStoreName];
```
wmathurin: We are only getting a handle to the store to get the compile options.
Since we might not be logged in yet, we should get a global store rather than a user store.

PR #2567, opened 2018-06-21
Fix crash for non JSON PN registration responses

sgoldberg-sfdc:
This addresses a crash when the server returns a non JSON response but a 200 Status OK.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
@@ -202,9 +202,13 @@ static NSString * const kSFAppFeaturePushNotifications = @"PN";
         [SFSDKAppFeatureMarkers registerAppFeature:kSFAppFeaturePushNotifications];
         NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse*) rawResponse;
         NSInteger statusCode = httpResponse.statusCode;
+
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
```diff
@@ -202,10 +202,14 @@ - (BOOL)registerSalesforceNotificationsWithCompletionBlock:(void (^)(void))compl
         [SFSDKAppFeatureMarkers registerAppFeature:kSFAppFeaturePushNotifications];
         NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse*) rawResponse;
         NSInteger statusCode = httpResponse.statusCode;
+
         if (statusCode < 200 || statusCode >= 300) {
             [SFSDKCoreLogger e:[strongSelf class] format:@"Registration for notifications with Salesforce failed with status %ld", statusCode];
             [SFSDKCoreLogger e:[strongSelf class] format:@"Response:%@", response];
             [strongSelf postPushNotificationRegistration:failBlock];
+        } else if (![response isKindOfClass:[NSDictionary class]]) {
+            [SFSDKCoreLogger e:[strongSelf class] format:@"Registration for notifications with Salesforce failed due to unexpected response: %@", response];
```
bhariharan: When would this occur? Doesn't the server API contract guarantee that JSON would be returned for `200 OK`?

PR #2568, opened 2018-06-26
What's New and Upgrade Steps for 6.2

bhariharan:


bhariharan:
@rwhitley

readme.md
```diff
@@ -35,25 +35,21 @@ If you have problems building any of the projects, take a look at the online [FA
...
-**SmartSync Enhancements**
-- We have addressed an issue in Mobile SDK 6.1 where `cleanResyncGhosts` was operating incorrectly.
-
-**SalesforceSwiftSDK Enhancements**
-- We have added new promise APIs based on `PromiseKit` for existing `SmartStore` and `SmartSync` APIs.
-
-**Test Framework Enhancements**
-- We have added an entire suite of tests for `SalesforceReact` in Mobile SDK 6.1.
+**SmartSync Data Framework Enhancements**
+- Starting with Mobile SDK 6.2, the `SmartSync` data framework now returns detailed information about the error, if an error occurs during a sync operation.
```
wmathurin: Should we say the error is stored in the record that failed to sync up or is it too much detail for the readme

bhariharan: Might be better suited for the dev guide. I'm okay either way.

wmathurin: Maybe something like
Starting with Mobile SDK 6.2, the SmartSync data framework saves the error returned by the server on records that fail to sync up.

PR #2578, opened 2018-07-03
 Upgraded to PromiseKit 6.0

trooper2013:
Changed some of the required initializers and handling of wrapped values, as suggested in PromiseKit 6.x.

### DIFF
--- a/libs/SalesforceSwiftSDK/SalesforceSwiftSDK/Extensions/SFRestAPIExtensions.swift
+++ b/libs/SalesforceSwiftSDK/SalesforceSwiftSDK/Extensions/SFRestAPIExtensions.swift
@@ -150,7 +150,7 @@ extension SFRestAPI {
          - Returns: SFRestRequest wrapped in a promise.
          */
         public func versions() -> Promise<SFRestRequest> {
-            return  Promise(.pending) {  resolver in
+            return  Promise {  resolver in
                 resolver.fulfill(self.api!.requestForVersions())
             }
         }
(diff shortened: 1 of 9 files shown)


libs/SalesforceSwiftSDK/SalesforceSwiftSDK/Extensions/SFRestAPIExtensions.swift
```diff
@@ -150,7 +150,7 @@ extension SFRestAPI {
          - Returns: SFRestRequest wrapped in a promise.
          */
         public func versions() -> Promise<SFRestRequest> {
```
wmathurin: (I'm probably a bit late for that comment) 
But what's the value of having promises for synchronous APIs.
Why is 
```swift
restApi.Promises.resources()
          .then { (request) in
             restApi.send(request)
          }
```
Better than
```swift
restApi.send(restApi.requestForResources())
```

libs/SalesforceSwiftSDK/SalesforceSwiftSDK/Extensions/SFSmartStoreExtensions.swift
```diff
@@ -190,7 +190,7 @@ extension SFSmartStore {
          - Returns: SFSoupSpec wrapped in a promise.
```
wmathurin: Same question for SFSmartStoreExtensions.swift.
All smartstore operations are synchronous so why wrap them in promises?

libs/SalesforceSwiftSDK/SalesforceSwiftSDK/Extensions/SFRestAPIExtensions.swift
trooper2013: @wmathurin Just a matter of  convenience, it allows you to start a promise chain from any of these factory methods. As for smartstore it also does offer a fluent api.

PR #2581, opened 2018-07-05
Removing OCMock

trooper2013:
Refactored the few tests to work without use of OCMock. Removed all references to OCMock.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
@@ -3313,7 +3313,7 @@
 				GCC_WARN_UNDECLARED_SELECTOR = YES;
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -3313,7 +3223,7 @@
 				GCC_WARN_UNDECLARED_SELECTOR = YES;
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				GCC_WARN_UNUSED_FUNCTION = YES;
-				HEADER_SEARCH_PATHS = "$(SRCROOT)/../../external/ocmock/Source";
+				HEADER_SEARCH_PATHS = "";
```
wmathurin: Do we need any value at all here?

bhariharan: Removing it from the UI creates an empty string (reason 9803 why I love `Xcode`). In order to remove the entry, we need to hand-edit it using a text editor and remove the line.

trooper2013: @wmathurin Thats the only value that was present here. Was able to compile everything, so I guess not.

PR #2586, opened 2018-07-10
cleanResyncGhosts returns numRecords 

wmathurin:
Change done across the board:
* obj-c
* swift
* cordova plugin
* react native module


wmathurin:
Merging - I don't think the CI failures are legit.

PR #2591, opened 2018-07-12
Add some loging around process pool getting, setting

khawkins:
We've been trying to track down some tricky issues with web state reset, and I'm thinking logging process pool changes might help that investigation.  For your consideration.


bhariharan:
Looks like the failing test has nothing to do with this PR.

PR #2592, opened 2018-07-12
Cleanup CI

brandonpage:
- Paralyze nightly builds for stability and readability.
- Update to Xcode 9.4/iOS 11.4 - thought I had done this already.
- Fix test reporting
- Fix test not running on rebuild
- yaml/fastlane cleanup

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -255,10 +255,11 @@ workflows:
       - test-SalesforceSwiftSDK
 
   # Cron is UTC timezone, which is 7 hours ahead of PST.
+  # Run tests at 1am.
   nightly-test-ios11:
     triggers:
       - schedule:
-          cron: "50 20 * * 2,3,4,5,6"
+          cron: "05 23 * * 2,3,4,5,6"
           filters:
             branches:
               only:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -399,16 +254,53 @@ workflows:
...
+    triggers:
+      - schedule:
+          cron: "00 08 * * 2,3,4,5,6"
+          filters:
+            branches:
+              only:
+                - dev
+    jobs:
+      - test-SalesforceAnalytics:
+            context: nightly-test
```
bhariharan: Shouldn't this be `nightly-test-ios11` (same as line 259 above)?

brandonpage: Nope, 259 is the workflow name. This context name should be nightly-test, it's actually the same one you already added for Android the other day. Since I have fastlane default to latest iOS version/device no extra work is needed there.

Then for the iOS 10 tests we will make a second one, nightly-test-ios10, that has the same nightly env var and an env var to specify iOS 10.

bhariharan: Cool.

PR #2596, opened 2018-07-19
 Decoupling authentication from launch in the SalesforceSDKManager

trooper2013:
Deprecated launch related functions. Added 2 net new Api(s) to the SDKManager. Changed code in the AppDelegate's for samples. Now the only thing required is to invoke the initializeSDK for the Manager. Since the process of init merely set's up configurations  for the sdk in a synchronous fashion there is no need for setting up  launch  or authentication related blocks. The developer has to invoke the SFUserAccountManager login api(s) on-demand. 

Bootstrap the SDK using the following.
```
[SalesforceSDKManager initializeSDK];  
[SalesforceSDKManager initializeSDKWithClass:SomeClass.class];

or using any  one of relevant subclasses

[SmartSyncSDKManager initializeSDK];
[SalesforceHybridSDKManager initializeSDK];
```

For Logout & User Switch Notifications 
```
[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleUserDidLogout:) name:kSFNotificationUserDidLogout object:nil];
        
[[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleUserDidSwitch:) name:kSFNotificationUserDidSwitch  object:nil];
```


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SalesforceHybridSDKManager.m
```diff
@@ -41,6 +41,10 @@ - (NSString*) pathForUserSyncsConfig;
 @implementation SalesforceHybridSDKManager
 @synthesize appConfig = _appConfig;
 
++ (void)initializeSDK {
```
bhariharan: Should we just call this `init`? The `SDK` part is redundant.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -251,7 +251,17 @@ typedef void (^SFSnapshotViewControllerDismissalBlock)(UIViewController* snapsho
  actions.
  @return YES if the launch successfully kicks off, NO if launch is already running.
  */
-- (BOOL)launch;
+- (BOOL)launch SFSDK_DEPRECATED(7.0, 8.0, "SalesforceSDKManager launch has been deprecated. Use initializeSDK instead of launch");
+
+/**
+ Initializes the SDK.
+ */
++ (void)initializeSDK;
```
bhariharan: Should we just call this `init`? The `SDK` part is redundant.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -251,7 +251,17 @@ typedef void (^SFSnapshotViewControllerDismissalBlock)(UIViewController* snapsho
...
+
+/**
+ Initializes the SDK.  Class instance to be used to instantiate the sdkManager.
+ */
++ (void)initializeSDKWithClass:(Class)className;
```
bhariharan: `initWithClass`?

libs/SmartStore/SmartStore/Classes/SmartStoreSDKManager.m
```diff
@@ -41,6 +41,10 @@ -(instancetype)init {
     return self;
 }
 
++ (void)initializeSDK {
```
bhariharan: Should we just call this `init`? The `SDK` part is redundant.

libs/SmartSync/SmartSync/Classes/Manager/SmartSyncSDKManager.m
```diff
@@ -28,6 +28,10 @@
 
 @implementation SmartSyncSDKManager
 
++ (void)initializeSDK {
```
bhariharan: Should we just call this `init`? The `SDK` part is redundant.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SalesforceHybridSDKManager.m
wmathurin: In objective-c, init is the instance level initializer (constructor) and initialize the class level initializer. So I don't think we should call it init. Having the name based on initialize makes sense.

bhariharan: Ah yes. I completely forgot. That makes sense.

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/AppDelegate.m
```diff
@@ -66,14 +66,12 @@ - (id)init
...
-        [SmartSyncSDKManager sharedManager].appConfig.remoteAccessConsumerKey = config.remoteAccessConsumerKey;
-        [SmartSyncSDKManager sharedManager].appConfig.oauthRedirectURI = config.oauthRedirectURI;
-        [SmartSyncSDKManager sharedManager].appConfig.oauthScopes = [NSSet setWithArray:config.oauthScopes];
-        __weak typeof(self) weakSelf = self;
-        [[SmartSyncSDKManager sharedManager] addDelegate:self];
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleUserDidLogout:) name:kSFNotificationUserDidLogout object:nil];
         
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleUserDidSwitch:) name:kSFNotificationUserDidSwitch  object:nil];
+        [SmartSyncSDKManager initializeSDK];
+        [[SmartSyncSDKManager sharedManager] addDelegate:self];
```
wmathurin: Why do we still need to be a delegate of the SDK manager? Couldn't we do everything with notification?

native/SampleApps/RestAPIExplorer/RestAPIExplorer/AppDelegate.swift
```diff
@@ -139,6 +129,20 @@ class AppDelegate : UIResponder, UIApplicationDelegate
     }
     
     // MARK: - Private methods
+    func loginIfRequired() -> Void  {
+        if SFUserAccountManager.sharedInstance().currentUser != nil  {
+            self.setupRootViewController()
+        }
+        
+        _ = SFUserAccountManager.sharedInstance().Promises.login()
```
wmathurin: Should that be in the else block?

native/SampleApps/RestAPIExplorer/RestAPIExplorer/AppDelegate.swift
```diff
@@ -139,6 +129,20 @@ class AppDelegate : UIResponder, UIApplicationDelegate
     }
     
     // MARK: - Private methods
+    func loginIfRequired() -> Void  {
```
wmathurin: There is still a lot of boilerplate code (loginIfRequired, all the handle* methods). Could we provide that through a helper class and only require a one liner in the AppDelegate?

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/AppDelegate.m
trooper2013: @wmathurin the delegate was already  part of  the sample to clear keyboard on backgrounding

native/SampleApps/RestAPIExplorer/RestAPIExplorer/AppDelegate.swift
trooper2013: Good catch. It should have been.

native/SampleApps/SmartSyncExplorer/SmartSyncExplorer/Classes/AppDelegate.m
wmathurin: Ah ok.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/AppDelegate.swift
wmathurin: The only code that are app specific is the setup of the root view controller, and the bit with the initial view controller. Could we move everything to a helper (loginIfRequired, handle logout and user switching)? Of course some apps might need more control and could look at the code of the helper as a guide, but all the sample apps / template apps would have much simple AppDelegate classes.

PR #2598, opened 2018-07-19
Add support for passing in absolute URL

bhariharan:
Should address [this issue](https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/issues/83).


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -123,41 +123,51 @@ - (void)setCustomRequestBodyStream:(NSInputStream* (^)(void))bodyInputStreamBloc
 
 - (NSURLRequest *)prepareRequestForSend:(SFUserAccount *)user {
     if (user) {
```
qingqingliu: @bhariharan ,  in case of full URL in path, we would still want to support adding query parameters to the path, right? like the line from https://github.com/bhariharan/SalesforceMobileSDK-iOS/blob/d511432813afd4d7b400eb85c204052e8359218b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m#L166

PR #2610, opened 2018-07-31
Fixing compilation and removing deprecation markers

bhariharan:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -251,6 +251,7 @@ typedef void (^SFSnapshotViewControllerDismissalBlock)(UIViewController* snapsho
  actions.
  @return YES if the launch successfully kicks off, NO if launch is already running.
  */
+- (BOOL)launch;
```
bhariharan: @trooper2013 FYI....`launch` was accidentally removed. Adding it back.

PR #2613, opened 2018-08-03
Make SFSDKSafeMutableSet public

sgoldberg-sfdc:


bhariharan:
@sgoldberg-sfdc Could you build the `Everything` scheme? This will update the `podspec` file and the umbrella header to add the new public header.

PR #2616, opened 2018-08-08
Fix hybrid remote session refresh case for SAML/SSO login flows

bhariharan:
If the org uses a SAML/SSO provider for login, it is possible to customize the SAML/SSO login URL to drop all arguments from the URL. Besides, the login redirect is beyond the control of the Salesforce IDP at this point and hence, the login redirect when a session timeout occurs will not contain `ec=301` or `startURL`. This leaves the user staring at a login page within the hybrid `WebView` that does nothing. Luckily, we can fetch the login URL in this case from `.well-known` and the container needs to do 2 things to solve this problem.
- We need to parse the SSO URLs from `.well-known` (which we already fetch) and check if these URLs are hit as part of the redirects that indicate session timeout.
- If this occurs, load the `startURL`.
This PR addresses these issues.

cc: @khawkins 

The equivalent `Android` PR is [here](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1763).

### DIFF
--- a/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
+++ b/libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
@@ -32,6 +32,7 @@
 #import <SalesforceSDKCore/NSURL+SFAdditions.h>
 #import <SalesforceSDKCore/SFUserAccountManager.h>
 #import <SalesforceSDKCore/SFAuthErrorHandlerList.h>
+#import <SAlesforceSDKCore/SFSDKAuthConfigUtil.h>
 #import <SalesforceSDKCore/SFSDKWebUtils.h>
 #import <SalesforceSDKCore/SFSDKResourceUtils.h>
 #import <SalesforceSDKCore/SFSDKEventBuilderHelper.h>
@@ -92,6 +93,8 @@ static NSString * const kSFAppFeatureUsesUIWebView = @"WV";
  * UIWebView for processing the error page, in the event of a fatal error during bootstrap.
  */
 @property (nonatomic, strong) UIWebView *errorPageUIWebView;
+
+@property (nonatomic, strong) SFOAuthOrgAuthConfiguration *authConfig;
 
 /**
  * Whether or not the input URL is one of the reserved URLs in the login flow, for consideration
@@ -184,10 +187,19 @@ static NSString * const kSFAppFeatureUsesUIWebView = @"WV";
         _hybridViewConfig = (viewConfig == nil ? [SFHybridViewConfig fromDefaultConfigFile] : viewConfig);
         NSAssert(_hybridViewConfig != nil, @"_hybridViewConfig was not properly initialized. See output log for errors.");
         self.startPage = _hybridViewConfig.startPage;
-        
+
         // Setup global stores and syncs defined in static configs
         [[SalesforceHybridSDKManager sharedManager] setupGlobalStoreFromDefaultConfig];
         [[SalesforceHybridSDKManager sharedManager] setupGlobalSyncsFromDefaultConfig];
+        __weak typeof(self) weakSelf = self;
+        [SFSDKAuthConfigUtil getMyDomainAuthConfig:^(SFOAuthOrgAuthConfiguration *authConfig, NSError *error) {
+            __strong typeof(weakSelf) strongSelf = weakSelf;
+            dispatch_async(dispatch_get_main_queue(), ^{
+                if (!error) {
+                    strongSelf.authConfig = authConfig;
+                }
+            });
+        } oauthCredentials:[SFUserAccountManager sharedInstance].currentUser.credentials];
     }
     return self;
 }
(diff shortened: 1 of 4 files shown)


libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -459,22 +471,50 @@ - (NSURL *)parseFrontDoorReturnUrlString:(NSString *)frontDoorUrlString encoded:
 
 - (NSString *)isLoginRedirectUrl:(NSURL *)url
 {
-    if (url == nil || [url absoluteString] == nil || [[url absoluteString] length] == 0) {
+    if (url == nil || url.absoluteString == nil || url.absoluteString.length == 0) {
```
bhariharan: Standardizing on dot notation.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -184,10 +187,19 @@ - (id) initWithConfig:(SFHybridViewConfig *) viewConfig useUIWebView:(BOOL) useU
...
+
         // Setup global stores and syncs defined in static configs
         [[SalesforceHybridSDKManager sharedManager] setupGlobalStoreFromDefaultConfig];
         [[SalesforceHybridSDKManager sharedManager] setupGlobalSyncsFromDefaultConfig];
+        __weak typeof(self) weakSelf = self;
+        [SFSDKAuthConfigUtil getMyDomainAuthConfig:^(SFOAuthOrgAuthConfiguration *authConfig, NSError *error) {
+            __strong typeof(weakSelf) strongSelf = weakSelf;
+            dispatch_async(dispatch_get_main_queue(), ^{
+                if (!error) {
+                    strongSelf.authConfig = authConfig;
```
bhariharan: Fetching org config once as part of view controller initialization and saving it.

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -459,22 +471,50 @@ - (NSURL *)parseFrontDoorReturnUrlString:(NSString *)frontDoorUrlString encoded:
...
+            NSString *ecValue = [url valueForParameterName:@"ec"];
+            BOOL foundStartURL = (startUrlValue != nil);
+            BOOL foundValidEcValue = ([ecValue isEqualToString:@"301"] || [ecValue isEqualToString:@"302"]);
+            if (foundValidEcValue) {
+                if (foundStartURL) {
+                    return startUrlValue;
+                } else {
+                    return self.startPage;
+                }
+            } else if ([self isSamlLoginRedirect:url.absoluteString]) {
```
bhariharan: Checking if a SAML login redirect is occurring.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.h
```diff
@@ -36,17 +34,20 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, readonly) BOOL useNativeBrowserForAuth;
 
+/**
+ List of configured SSO URLs.
+ */
+@property (nonatomic, strong, readonly, nullable) NSArray<NSString *> *ssoUrls;
```
bhariharan: Exposing a new property and fixing nullability specifiers.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.m
```diff
@@ -49,6 +51,20 @@ - (BOOL)useNativeBrowserForAuth {
     return [self.authConfigDict[kAuthConfigMobileSDKKey][kAuthConfigUseNativeBrowserKey] boolValue];
 }
 
+- (NSArray<NSString *> *)ssoUrls {
```
bhariharan: Extracting SSO URLs.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKAuthConfigUtilTests.m
```diff
@@ -69,6 +70,35 @@ - (void)testBrowserBasedLoginEnabled {
     [self waitForExpectationsWithTimeout:20 handler:nil];
 }
 
+- (void)testGetSSOUrls {
```
bhariharan: Tests for positive and negative cases.

bhariharan:
End-to-end test complete. Works as expected (same as Android).

bhariharan:
Any feedback?

libs/SalesforceHybridSDK/SalesforceHybridSDK/Classes/SFHybridViewController.m
```diff
@@ -184,10 +187,19 @@ - (id) initWithConfig:(SFHybridViewConfig *) viewConfig useUIWebView:(BOOL) useU
         _hybridViewConfig = (viewConfig == nil ? [SFHybridViewConfig fromDefaultConfigFile] : viewConfig);
         NSAssert(_hybridViewConfig != nil, @"_hybridViewConfig was not properly initialized. See output log for errors.");
         self.startPage = _hybridViewConfig.startPage;
-        
+
         // Setup global stores and syncs defined in static configs
         [[SalesforceHybridSDKManager sharedManager] setupGlobalStoreFromDefaultConfig];
         [[SalesforceHybridSDKManager sharedManager] setupGlobalSyncsFromDefaultConfig];
+        __weak typeof(self) weakSelf = self;
+        [SFSDKAuthConfigUtil getMyDomainAuthConfig:^(SFOAuthOrgAuthConfiguration *authConfig, NSError *error) {
```
khawkins: (Sorry, late to the party on this review.)

In general, `init` methods are not a great place to take actions—particularly network activities.  And even more particularly, asynchronous network activities—init methods should be synchronous.

This is really looking for a different design pattern.  Some suggestions:

1. Given that this information is general, and likely to become germane to more parts of the SDK footprint than the hybrid view, shouldn't this happen as part of the SDK bootstrapping process?
2. If this is going to stay in the hybrid view controller, it really needs to move somewhere that's part of an existing async flow.  I would suggest `viewDidLoad`.

khawkins: Another point, probably in favor of the first approach: this information feels pretty tightly coupled to the authentication of a user.  Shouldn't that process be the process that owns the retrieval and management of this configuration?

khawkins: A further point that puts me 99% in favor of the first approach: we're [already retrieving this data](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m#L230) as part of authentication, which is making this an additional, redundant network call.  We should simply persist the retrieved data as part of e.g. the user account, and reference it here.

bhariharan: Thanks for the feedback, @khawkins. Yes - we are already retrieving the data. I'll look into a persistence solution.

PR #2617, opened 2018-08-15
Fixing corner case with auth timeout

bhariharan:
The URL that comes back from the server is not always encoded (in some communities it's not encoded). This PR computes whether it's encoded or not dynamically to handle this situation and prevent a downstream crash while attempting to build `frontdoorUrl`.


bhariharan:
Merging for now because `Communities` need this hotfix.

PR #2621, opened 2018-08-21
Using SFAuthenticationSession for advanced auth flows

wmathurin:
I don't think the work is quite done but I wanted to get some feedback.

Basically, the new class is super easy to use. I took out the code that is no longer used at all, but I feel like we could go further e.g. SFSDKOAuthClientSafariViewDelegate could be further simplified. Also, maybe more renaming should take place: should we use "safari" or "browser" in the method or property names.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -435,10 +435,6 @@ static NSString * const kSFECParameter = @"ec";
     [SFSDKAppFeatureMarkers registerAppFeature:kSFAppFeatureSafariBrowserForLogin];
     __weak typeof(self) weakSelf = self;
     _authSession = [[SFAuthenticationSession alloc] initWithURL:nativeBrowserUrl callbackURLScheme:nil completionHandler:^(NSURL * _Nullable callbackURL, NSError * _Nullable error) {
-
-        NSLog(@"url-->%@", callbackURL);
-        NSLog(@"error-->%@", error);
-        
         if (error) {
             [weakSelf.delegate oauthCoordinatorDidCancelBrowserAuthentication:self];
         }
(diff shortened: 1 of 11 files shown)


bhariharan:
LGTM! I agree - `browser` might be better suited than `Safari` going forward.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -431,10 +433,22 @@ - (void)continueNativeBrowserFlow {
...
-    svc.delegate = self;
-    self.advancedAuthState = SFOAuthAdvancedAuthStateBrowserRequestInitiated;
-    [self.delegate oauthCoordinator:self didBeginAuthenticationWithSafariViewController:svc];
+    __weak typeof(self) weakSelf = self;
+    _authSession = [[SFAuthenticationSession alloc] initWithURL:nativeBrowserUrl callbackURLScheme:nil completionHandler:^(NSURL * _Nullable callbackURL, NSError * _Nullable error) {
+        if (error) {
+            [weakSelf.delegate oauthCoordinatorDidCancelBrowserAuthentication:self];
+        }
+        else {
+            [[SFUserAccountManager sharedInstance] handleAdvancedAuthenticationResponse:callbackURL options:nil];
```
trooper2013: @wmathurin Can we do away with the dependency on SFUserAccountManager? Is it possible to perhaps invoke a completion block instead or  we could fire a notification?

wmathurin: SFUserAccountManager's handleAdvancedAuthenticationResponse calls SFSDKURLHandlerManager's processRequest which calls SFSDKAdvancedAuthURLHandler's processRequest which calls SFUserAccountManager's handleNativeAuthResponse which goes to SFSDKOAuthClient's handleURLAuthenticationResponse which goes back to SFOAuthCoordinator's handleAdvancedAuthenticationResponse !

So I guess I could call the SFOAuthCoordinator method directly.

wmathurin: Done [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2621/commits/343477a4e708bd7b298140c131bba7542b04edf8)

wmathurin:
Getting rid of alert during advanced auth flow with [this](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2621/commits/e72490a168d5a679e2d28d9606567a0260cc3ac3) since SFAuthenticationSession is also showing an alert:
screen shot 2018-08-22 at 4 14 55 pm

PR #2622, opened 2018-08-22
Modernizing Obj-c api(s) for swift

trooper2013:
Summary of Changes. 

Most of these changes directly impact swift apps. This is the first PR that applies the changes to SDK core. A few more will follow after this one.

1.  Renamed api(s) using NS_SWIFT_NAME.
- There is no need for our Prefixes SF or SFSDK for public apis. 
_Some exceptions 
SalesForceSDKManager is renamed to SalesforceSDK
SFRestAI is renamed to RestClient_
- All SFRestAPI factory methods have been renamed to start with build..()
- All SFRestAPI perform methods have been renamed to start with execute..()
-  Changed all annotations on parameters to reflect onSuccess and onFailure for apis that have completionBlocks. It  adds consistency for naming across the sdk.

2. Grouped and name-spaced some of the global constants into enums NS_TYPED_ENUM. Moved some extern definitions that were sitting in the wrong header to more appropriate location.

3. The SalesforceSwiftSDK has been refactored to reflect all the above changes. We will move this out later. It helps to validate the shape of the changed api(s).


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -31,14 +31,14 @@
...
-typedef UIViewController * __nullable (^SFSnapshotViewControllerCreationBlock)(void);
+typedef UIViewController * __nullable (^SFSnapshotViewControllerCreationBlock)(void)  NS_SWIFT_NAME(SnapshotViewCreationBlock);
 
 typedef NS_ENUM(NSUInteger, SFAppType) {
     kSFAppTypeNative,
     kSFAppTypeHybrid,
     kSFAppTypeReactNative,
     kSFAppTypeNativeSwift
-};
+} NS_SWIFT_NAME(AppType);
```
wmathurin: Just to be sure, the swift code will look like:
```swift
AppType.Native
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.h
```diff
@@ -30,6 +30,50 @@
 #import <Foundation/Foundation.h>
 #include "SFOAuthCoordinator.h"
 NS_ASSUME_NONNULL_BEGIN
+
+typedef NSString * const AuthPreferencesNotification NS_TYPED_ENUM NS_SWIFT_NAME(AuthPreferences.Notification);
+typedef NSString * const AuthPreferencesConstants NS_TYPED_ENUM NS_SWIFT_NAME(AuthPreferences.Constants);
+/**
+ Key identifying login host
+ */
+FOUNDATION_EXTERN AuthPreferencesConstants kSFUserAccountOAuthLoginHost;
```
wmathurin: What are the swift names here? Could you maybe add comments so that a developer looking at the header file doesn't have to figure out what typedef + FOUNDATION_EXTERN does?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
trooper2013: @wmathurin It will be AppType.native.

PR #2628, opened 2018-08-28
Removing wypopovercontroller

wmathurin:
wypopovercontroller was used by old RestAPIExplorer


wmathurin:
I stand corrected. It's used by SmartSyncExplorer (see ContactListViewController).

PR #2649, opened 2018-09-20
Remove Hybrid and React

brandonpage:


external/wypopovercontroller/.gitignore
```diff
@@ -0,0 +1,18 @@
+# Xcode
```
bhariharan: Why are these files committed? Shouldn't `external/wypopovercontroller` be deleted? If not, it should still be consumed as submodule.

brandonpage: I had deleted `external/wypopovercontroller` but SmartSync wouldn't compile without it.  I added it back to avoid modifying the code this late, but I can take a closer look.

bhariharan: Looks like the submodule got removed and the files got added back incorrectly.

brandonpage: Yeah, sorry I was going to write that in comments but I just got pulled into a call.  wypopovercontroller should have been added back as a submodule, that is a mistake.  Shared also got added back as files to avoid creating another PR to create a branch over there.  That is the correct way to do it, but I don't know if its worth it if this branch just get deleted when we release 7.0 proper.

.circleci/config.yml
```diff
@@ -50,7 +50,7 @@ aliases:
 defaults: &defaults
   working_directory: ~/SalesforceMobileSDK-iOS
   macos:
-    xcode: "9.4.1"
+    xcode: "10.0.0"
```
brandonpage: 🎉

install.sh
```diff
@@ -47,12 +47,6 @@ git submodule init
...
 
-# Get react native
-pushd "libs/SalesforceReact"
-rm -rf node_modules
-npm install
-popd
-
 
 # Remove the old Xcode templates, if they still exist.
 hybrid_template_dir="${HOME}/Library/Developer/Xcode/Templates/Project Templates/Application/Hybrid Force.com App.xctemplate"
```
brandonpage: This is the only remaining mention of hybrid, but it didn't seem like it should be removed.

external/wypopovercontroller/.gitignore
bhariharan: Let's do the submodule bit correctly. I'm not sure we want zombie files just hanging around with no source.

jamesward:
@da-real-bpage You've been invited to the org: https://github.com/forcedotcom

Once accepted, refresh the CLA bot: https://cla.salesforce.com/status/forcedotcom/SalesforceMobileSDK-iOS/pull/2649

install.sh
bhariharan: It can be removed. It's almost 5 yrs old.

PR #2652, opened 2018-09-27
Add error message to SyncState for failed sync down

michchan:
Set message from error to current sync state
Return sync state to react native instead of "Sync failed" for sync failure

Please also refer to: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1786

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -253,6 +253,8 @@ static NSMutableDictionary *syncMgrList = nil;
 
     SyncFailBlock failSync = ^(NSString* failureMessage, NSError* error) {
         __strong typeof(weakSelf) strongSelf = weakSelf;
+        //Set error message to sync state
+        [sync setErrorJSON: (NSString*) [[error.userInfo valueForKey:@"error"] objectAtIndex:0]];
         [SFSDKSmartSyncLogger e:[strongSelf class] format:@"runSync failed:%@ cause:%@ error%@", sync, failureMessage, error];
         updateSync(kSFSyncStateStatusFailed, kSyncManagerUnchanged, kSyncManagerUnchanged, kSyncManagerUnchanged);
     };
--- a/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
+++ b/libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
@@ -254,8 +254,7 @@ static NSMutableDictionary *syncMgrList = nil;
     SyncFailBlock failSync = ^(NSString* failureMessage, NSError* error) {
         __strong typeof(weakSelf) strongSelf = weakSelf;
         //Set error message to sync state
-//        [sync setErrorJSON:(NSString*) [[error.userInfo valueForKey:@"error"] objectAtIndex:0]];
-        [sync setErrorJSON:(NSString*) error.userInfo];
+        [sync setErrorJSON: error.userInfo];
         [SFSDKSmartSyncLogger e:[strongSelf class] format:@"runSync failed:%@ cause:%@ error%@", sync, failureMessage, error];
         updateSync(kSFSyncStateStatusFailed, kSyncManagerUnchanged, kSyncManagerUnchanged, kSyncManagerUnchanged);
     };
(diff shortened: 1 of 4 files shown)


michchan:
The pull request failed but it actually works,
and I have looked into the testing codes it seems not related to my changes.
Does anyone know how to resolve it?

Failing tests:
`-[MetadataManagerTests testCleanCache]
-[MetadataManagerTests testCommonMRUObjectsFromServer]
-[MetadataManagerTests testGlobalMRUObjectsFromServer]
-[MetadataManagerTests testRemoveMRUCache]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[ParentChildrenSyncTests testReSyncWithUpdatedChildren]
-[SyncManagerTests testSyncUpWithLocallyDeletedRecordsWithoutOverwrite]
-[SyncManagerTests testSyncUpWithLocallyDeletedRecordsWithoutOverwrite]
-[SyncManagerTests testSyncUpWithLocallyUpdatedRecordsWithoutOverwrite]
-[SyncManagerTests testSyncUpWithLocallyUpdatedRecordsWithoutOverwrite]
-[SyncManagerTests testSyncUpWithLocallyUpdatedRecordsWithoutOverwrite]
-[SyncManagerTests testSyncUpWithLocallyUpdatedRecordsWithoutOverwrite]`

libs/SmartSync/SmartSync/Classes/Util/SFSyncState.m
```diff
@@ -340,4 +348,16 @@ -(id)copyWithZone:(NSZone *)zone
     return clone;
 }
 
+#pragma mark - Setter and getter of property errorJSON
```
wmathurin: Why do we need a setter and getter?

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
```diff
@@ -253,6 +253,8 @@ - (void) runSync:(SFSyncState*) sync updateBlock:(SFSyncSyncManagerUpdateBlock)u
 
     SyncFailBlock failSync = ^(NSString* failureMessage, NSError* error) {
         __strong typeof(weakSelf) strongSelf = weakSelf;
+        //Set error message to sync state
+        [sync setErrorJSON: (NSString*) [[error.userInfo valueForKey:@"error"] objectAtIndex:0]];
```
wmathurin: See my comment here: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1793#discussion_r221398663

The sync error is not necessarily the error returned by the service.
During sync up, if a record doesn't save on the server, we keep going and only network errors or other unexpected failures will cause the sync to fail.

libs/SmartSync/SmartSync/Classes/Util/SFSyncState.h
```diff
@@ -138,6 +142,11 @@ extern NSString * const kSFSyncStateMergeModeLeaveIfChanged;
 + (SFSyncStateMergeMode) mergeModeFromString:(NSString*)mergeMode;
 + (NSString*) mergeModeToString:(SFSyncStateMergeMode)mergeMode;
 
+/** Setter and getter of property errorJSON
+ */
+- (void) setErrorJSON:(NSString * _Nonnull)errorJSON;
```
wmathurin: Do we need a getter/setter?

michchan: Resolved in commit: 3ebdacfc3fd423b87501bc0da3b856778c0b173c
But the setter is still needed for setting error into the sync object and the setter interface is still required by compiler.

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.m
michchan: Resolved in commit ba068115bbd4cc32a66d8e5f7fd213ba3632e977 & d16984db460575863a9a68d6d0d9a4602ad33b17.
I tried to keep to error.userInfo instead of converting it to string ([error.userInfo description]) so that it returns as JSON on the JS side.
Just wonder how to do if we want to align the error returned in iOS with Android.
Say it also returns JSON instead of string in Android.

michchan: Resolved in d1f5324ae905e05629c06d5e552f9727b3021f06

PR #2641, opened 2018-09-11
Change MD5_DIGEST_LENGTH to CC_MD5_DIGEST_LENGTH

sgoldberg-sfdc:
-  MD5_DIGEST_LENGTH is no longer defined in SDK 12
-  MD5_DIGEST_LENGTH was always an alias for CC_MD5_DIGEST_LENGTH

PR #2662, opened 2018-10-12
Update NSData+SFAdditions.m

jolivr:
Defined MD5_DIGEST_LENGTH to fix the undeclared identifier error in the md5 method


bhariharan:
[This PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2641) has the right fix. It's already fixed in the `dev` branch.

PR #2665, opened 2018-10-17
Avoiding wasteful deserialization/serialization when querying smartstore

wmathurin:
- smartstore query from hybrid or react native application no longer do any deserialization or serialization in native land / and there is no javascript code change required for hybrid app, however react.force.smartstore.js is changed (see https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/pull/89)
- native app only deserialize once the full result is built (*)
- SFStoreCursor refactored to be more like Android's StoreCursor (for instance the query doesn't have to be run the first time by the caller).

(*) probably not faster but that way the code is cleaner (query is calling queryAsString)

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
@@ -800,61 +800,107 @@ NSString *const EXPLAIN_ROWS = @"rows";
 }
 
 - (id)loadExternalSoupEntry:(NSNumber *)soupEntryId
-              soupTableName:(NSString *)soupTableName {
-    SFSmartStoreEncryptionKeyBlock keyBlock = [SFSmartStore encryptionKeyBlock];
-    SFEncryptionKey *encKey = nil;
-    if (keyBlock){
-       encKey = keyBlock();
-    }
-    return [self loadExternalSoupEntry:soupEntryId soupTableName:soupTableName withKey:encKey.key andIV:encKey.initializationVector];
+              soupTableName:(NSString *)soupTableName
+{
+    return [SFJsonUtils objectFromJSONString:[self loadExternalSoupEntryAsString:soupEntryId soupTableName:soupTableName]];
 }
 
-- (id)loadExternalSoupEntry:(NSNumber *)soupEntryId
-              soupTableName:(NSString *)soupTableName withKey:(NSData *)key andIV:(NSData *)initializationVector {
+- (NSString*)loadExternalSoupEntryAsString:(NSNumber *)soupEntryId
+                             soupTableName:(NSString *)soupTableName
+{
     NSString *filePath = [self externalStorageSoupFilePath:soupEntryId
                                              soupTableName:soupTableName];
+    
+    SFSmartStoreEncryptionKeyBlock keyBlock = [SFSmartStore encryptionKeyBlock];
+    NSData* key;
+    NSData* initializationVector;
+    if (keyBlock) {
+        SFEncryptionKey *encKey = keyBlock();
+        key = encKey.key;
+        initializationVector = encKey.initializationVector;
+    }
+    
+    NSString* entryAsString = [self readFromEncryptedFile:filePath key:key iv:initializationVector];
+    
+    // Before 6.2, we were using nill IV when encrypting.
+    // Starting in 6.2, we are using a non-nil IV when encrypting.
+    // If it doesn't look like proper json, it means the entry was encrypted with pre 6.2 SDK.
+    // It needs to be stored back with a non-nil IV encryption.
+    if (![entryAsString hasPrefix:@"{"]) {
+        NSError* error = nil;
+        NSDictionary* entry = [NSJSONSerialization JSONObjectWithData:[entryAsString dataUsingEncoding:kCFStringEncodingUTF8] options:NSJSONReadingAllowFragments error:&error];
+        
+        if(!entry && error) {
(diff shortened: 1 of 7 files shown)


libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
```diff
@@ -866,38 +866,6 @@ - (void)testQuerySpecPageSize
     XCTAssertEqual(querySpecPageSize, expectedPageSize, @"Page size value should reflect input value.");
 }
 
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wnonnull"
-
-- (void)testCursorTotalPages
```
wmathurin: totalEntries is no longer passed as a parameter to the SFStoreCursor init method.
Instead SFStoreCursor's runs the query to figure out the number of entries. So to keep this test, we would have to actually populate the db and pass a query that returns the desired number of entries. Seems a lot just to test the following lines of code:
```objective-c
float totalPagesFloat = totalEntries / (float) querySpec.pageSize;
float totalPagesFloat = totalEntries / (float)querySpec.pageSize;	        
NSUInteger totalPages = ceilf(totalPagesFloat);
int totalPages = ceilf(totalPagesFloat);	
if (0 == totalEntries)
    totalPages = 0;
```

libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
```diff
@@ -866,38 +866,6 @@ - (void)testQuerySpecPageSize
     XCTAssertEqual(querySpecPageSize, expectedPageSize, @"Page size value should reflect input value.");
 }
 
-#pragma clang diagnostic push
```
wmathurin: Do we still need those pragma's?

libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -645,8 +646,10 @@ - (void)inDatabase:(void (^)(FMDatabase *db))block error:(NSError* __autoreleasi
             if (error != nil) {
                 *error = [self errorForException:exception];
             }
+            success = FALSE;
```
bhariharan:
```suggestion
            success = NO;
```

bhariharan: Trying new `GitHub` feature ;-)

libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
bhariharan: I don't think so. They're ignoring warnings based on nullability markers (or lack thereof). We should have pretty good coverage of nullability markers now.

libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
```diff
@@ -866,38 +866,6 @@ - (void)testQuerySpecPageSize
...
-                              kQuerySpecParamIndexPath: @"a",
-                              kQuerySpecParamPageSize: @(unevenDividePageSize)};
-    querySpec = [[SFQuerySpec alloc] initWithDictionary:allQuery  withSoupName:kTestSoupName];
-    cursor = [[SFStoreCursor alloc] initWithStore:nil querySpec:querySpec totalEntries:totalEntries firstPageEntries:nil];
-    XCTAssertEqual([cursor.totalEntries unsignedIntValue], totalEntries, @"Wrong value for totalEntries");
-    cursorTotalPages = [cursor.totalPages intValue];
-    XCTAssertEqual(cursorTotalPages, expectedPageSize, @"%lu entries across a page size of %lu should make %lu total pages.", (unsigned long)totalEntries, (unsigned long)unevenDividePageSize, (unsigned long)expectedPageSize);
-}
-
 #pragma clang diagnostic pop
```
bhariharan:
```suggestion
#pragma clang diagnostic pop
```
This can be removed too.

libs/SmartStore/SmartStore/Classes/SFStoreCursor.m
```diff
@@ -42,112 +39,57 @@ @interface SFStoreCursor ()
...
+    [resultBuilder appendString:@"{"];
+    [resultBuilder appendFormat:@"\"%@\":\"%@\", ", @"cursorId", self.cursorId];
+    [resultBuilder appendFormat:@"\"%@\":%@, ", @"currentPageIndex", self.currentPageIndex ?: @0];
+    [resultBuilder appendFormat:@"\"%@\":%@, ", @"pageSize", self.pageSize ?: @0];
+    [resultBuilder appendFormat:@"\"%@\":%@, ", @"totalPages", self.totalPages ?: @0];
+    [resultBuilder appendFormat:@"\"%@\":%@, ", @"totalEntries", self.totalEntries ?: @0];
+    [resultBuilder appendFormat:@"\"%@\":", @"currentPageOrderedEntries"];
+    [store queryAsString:resultBuilder querySpec:self.querySpec pageIndex:[self.currentPageIndex integerValue] error:error];
+    [resultBuilder appendString:@"}"];
+    return resultBuilder;
```
radhap: @wmathurin do you think I should add JSON validation check here as well ?

wmathurin: I think it would be overkill if we already checked that the query-as-string method returned something that is a properly json-ized object.

PR #2673, opened 2018-10-26
Swift Naming change for public api(s)

trooper2013:
Swift Naming change for public api(s).


libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.h
```diff
@@ -81,8 +89,14 @@ NS_SWIFT_NAME(SyncManager)
  * @param user The user associated with the store.
  * @param storeName The name of the store associated with the given user.
  */
-+ (void)removeSharedInstanceForUser:(SFUserAccount*)user storeName:(nullable NSString*)storeName NS_SWIFT_NAME(removeSharedInstance(user:storeName:));
++ (void)removeSharedInstanceForUser:(SFUserAccount*)user storeName:(nullable NSString*)storeName NS_SWIFT_UNAVAILABLE("");
```
bhariharan: Should we just remove this method from `Obj C` as well?

libs/SmartSync/SmartSync/Classes/Manager/SFSmartSyncSyncManager.h
```diff
@@ -59,7 +59,15 @@ NS_SWIFT_NAME(SyncManager)
  * @param user The user associated with the store.
  * @param storeName The name of the SmartStore associated with the user.
  */
-+ (instancetype)sharedInstanceForUser:(SFUserAccount*)user storeName:(nullable NSString *)storeName NS_SWIFT_NAME(sharedInstance(user:storeName:));
++ (instancetype)sharedInstanceForUser:(SFUserAccount*)user storeName:(nullable NSString *)storeName  NS_SWIFT_UNAVAILABLE("");
```
bhariharan: Should we just remove this method from `Obj C` as well?

libs/SmartStore/SmartStore/Classes/SFSmartStore.h
```diff
@@ -360,7 +360,7 @@ NS_SWIFT_NAME(SmartStore)
  
  @return The array of updated entries in the soup.
  */
-- (NSArray<NSDictionary*>*)upsertEntries:(NSArray<NSDictionary*>*)entries toSoup:(NSString*)soupName NS_SWIFT_NAME(upsert(entries:soupName:));
+- (NSArray<NSDictionary*>*)upsertEntries:(NSArray<NSDictionary*>*)entries toSoup:(NSString*)soupName NS_SWIFT_NAME(upsert(entries:forSoupNamed:));
```
wmathurin: Should we make this one unavailable in swift and add an upsertEntries that take a `NSError**` (and therefore that throws in Swift)?
(The upsert with external id path takes a `NSError** `/ throws).

PR #2680, opened 2018-11-07
Use of os_log as default Logger for the SDK

trooper2013:
This PR  makes changes the underlying logger implementation. It removes the dependency on CocoaLumberjack. The SFSDKLogger which has been an implementation of  CocoaLumberjack will still be made available in another repo. Apps using SFSDKLogger will be required to add an additional dependency to the project.  

- The logger code is moved in to SalesforceSDKCommon. A few other classes are moved here from within core. e.g. all the SFDKSafe...Classes
- SalesforceAnalytics is dependent on the SalesforceSDKCommon.  
- The main SDK logger now is SFLogger which has an identical interface as that of SFSDKLogger. 
- By default all sdk logs will be directed to os_log. 
- In-order to change the behavior you could add an object that implements the SFLogging protocol. 
- The LogLevels to be used will be SFLogLevel instead of DDLogLevel. The SFLogLevels map directly to the os_log log levels (default, debug, error, info  & fault). The mapping from these to relevant levels should be done by the object implementing SFLogging protocol.
- The SDK's own implementation of the SFLogging protocol is done in the SFDefaultLogger. 
-  The SDK can be configured to use the relative SFLogging implementation as shown below.
```
 [SFLogger setInstanceClass:[TestLoggingImpl class]];
```


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEventBuilder.m
```diff
@@ -69,7 +69,7 @@ - (SFSDKInstrumentationEvent *) buildEvent {
         errorMessage = @"Mandatory field 'page' not set!";
     }
     if (errorMessage) {
-        [SFSDKAnalyticsLogger w:[self class] format:@"WARNING: Building event failed! REASON: %@", errorMessage];
+        [SFSDKAnalyticsLogger log:[self class] format:@"WARNING: Building event failed! REASON: %@", errorMessage];
```
wmathurin: Shouldn't be `warn` (or whatever the method is called to log at log level warning)?

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
```diff
@@ -103,64 +118,64 @@
...
  * @param cls Class.
  * @param format The format message, and optional arguments to expand in the format.
  * @param ... The arguments to the message format string.
  */
-- (void)i:(nonnull Class)cls format:(nonnull NSString *)format, ...;
+- (void)d:(nonnull Class)cls format:(nonnull NSString *)format, ...;
 
 /**
- * Logs a verbose log line.
+ * Logs a default log line.
```
wmathurin: Should be warning instead of default

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
```diff
@@ -103,64 +118,64 @@
...
- * @param message Log message.
+ * @param format The format message, and optional arguments to expand in the format.
+ * @param ... The arguments to the message format string.
  */
-- (void)v:(nonnull Class)cls message:(nonnull NSString *)message;
+- (void)w:(nonnull Class)cls format:(nonnull NSString *)format, ...;
 
 /**
- * Logs a verbose log line.
+ * Logs a default log line.
```
wmathurin: Should be warning instead of default

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
```diff
@@ -27,64 +25,81 @@
...
  */
 
-#define DD_LEGACY_MACROS 0
-#import <CocoaLumberjack/DDLog.h>
-#import "SFSDKFileLogger.h"
+#import <Foundation/Foundation.h>
+#import <OS/log.h>
+
+typedef NS_ENUM(NSUInteger, SFLogLevel){
+    SFLogLevelDefault =  OS_LOG_TYPE_DEFAULT,
```
wmathurin: There is no warning level?

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
```diff
@@ -169,135 +184,145 @@
...
  * @param cls Class.
  * @param message Log message.
  */
-+ (void)w:(nonnull Class)cls message:(nonnull NSString *)message;
++ (void)d:(nonnull Class)cls message:(nonnull NSString *)message;
+
 
 /**
- * Logs an info log line.
+ * Logs a default log line.
```
wmathurin: Shouldn't it be warning?

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
```diff
@@ -169,135 +184,145 @@
...
  * @param cls Class.
  * @param format The format message, and optional arguments to expand in the format.
  * @param ... The arguments to the message format string.
  */
-+ (void)i:(nonnull Class)cls format:(nonnull NSString *)format, ...;
++ (void)w:(nonnull Class)cls format:(nonnull NSString *)format, ...;
 
 /**
- * Logs an info log line.
+ * Logs a default log line.
```
wmathurin: Shouldn't it be warning?

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
wmathurin: Should we have a SFLogLevelWarning even if we end up mapping it to OS_LOG_TYPE_DEFAULT?

libs/SalesforceSDKCommon/SalesforceSDKCommon.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,961 @@
...
+					"DEBUG=1",
+					"$(inherited)",
+				);
+				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
+				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
+				GCC_WARN_UNDECLARED_SELECTOR = YES;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				GCC_WARN_UNUSED_FUNCTION = YES;
+				GCC_WARN_UNUSED_VARIABLE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
```
bhariharan: I think this should be `11.0`.

libs/SalesforceSDKCommon/SalesforceSDKCommon.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,961 @@
...
+				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				GCC_C_LANGUAGE_STANDARD = gnu11;
+				GCC_NO_COMMON_BLOCKS = YES;
+				GCC_WARN_64_TO_32_BIT_CONVERSION = YES;
+				GCC_WARN_ABOUT_RETURN_TYPE = YES_ERROR;
+				GCC_WARN_UNDECLARED_SELECTOR = YES;
+				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
+				GCC_WARN_UNUSED_FUNCTION = YES;
+				GCC_WARN_UNUSED_VARIABLE = YES;
+				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
```
bhariharan: I think this should be `11.0`.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
```diff
@@ -169,135 +184,145 @@
  * @param format The format message, and optional arguments to expand in the format.
  * @param ... The arguments to the message format string.
  */
-- (void)d:(nonnull Class)cls format:(nonnull NSString *)format, ...;
+- (void)f:(nonnull Class)cls format:(nonnull NSString *)format, ...;
```
bhariharan: Comment should say "fault" log line.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
trooper2013: @wmathurin There isn't an equivalent of warn. Warn is mapped to default level. From what I read, the default level  seems to be appropriate for warnings.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
trooper2013: @wmathurin Thats correct there isn't a warning equivalent. The Default sounds like the best choice for warning.

PR #2691, opened 2018-11-15
Fix to avoid crash on failed request

wmathurin:
Simple way to reproduce: revoke token and then try a request (that requires being authenticated) 
=> delegate are notified of failure before logout starts
=> in RestAPIExplorer updateUI is called with response = nil and error = nil
=> updateUI calls JSONSerialization.data(withJSONObject:response as Any, ...) which crashes with nil

Now
only trying to parse response if not nil
passing error through so it can be displayed


bhariharan:
> We should refactor SFRestAPI's replayRequestIfRequired.
Agreed. It's a massive method that does a lot of stuff.

PR #2697, opened 2018-11-20
Moving SFJsonUtils to SalesforceSDKCommon and using it of NSJSONSeria…

wmathurin:
…lization

In some places, we were calling NSJSONSerialization methods directly without checking if the data was nil - which will cause a crash.


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEvent.m
```diff
@@ -85,9 +86,8 @@ - (instancetype) initWithJson:(NSData *) jsonRepresentation {
     self = [super init];
     if (self && jsonRepresentation) {
         NSError *error;
-        NSDictionary *dict = [NSJSONSerialization JSONObjectWithData:jsonRepresentation
```
wmathurin: This line has crashed in prod and is the reason for this PR.

PR #2732, opened 2018-12-15
Adding missing style.css

wmathurin:
@rwhitley noticed style.css were missing

It looks better - but it's still a lot of empty docs


rwhitley:
👍

PR #2735, opened 2018-12-18
Set current login host during user switch

trooper2013:
@sgoldberg-sfdc  @wmathurin Based on discussions, here is the PR to master for the user switching issue.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
@@ -1068,6 +1068,8 @@ static NSString *const  kOptionsClientKey          = @"clientIdentifier";
                 [self willChangeValueForKey:@"currentUser"];
                 _currentUser = user;
                 [self setCurrentUserIdentity:user.accountIdentity];
+                if (user.credentials.domain)
+                    self.loginHost = user.credentials.domain;
                 [self didChangeValueForKey:@"currentUser"];
                 userChanged = YES;
             } else {
@@ -1267,7 +1269,7 @@ static NSString *const  kOptionsClientKey          = @"clientIdentifier";
         __weak typeof(self) weakSelf = self;
         client = [SFSDKOAuthClient clientWithCredentials:credentials configBlock:^(SFSDKOAuthClientConfig *config) {
             __strong typeof(self) strongSelf = weakSelf;
-            config.loginHost = strongSelf.loginHost;
+            config.loginHost = credentials.domain ?: strongSelf.loginHost;
             config.brandLoginPath = strongSelf.brandLoginPath;
             config.additionalTokenRefreshParams = strongSelf.additionalTokenRefreshParams;
             config.additionalOAuthParameterKeys = strongSelf.additionalOAuthParameterKeys;
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerTests.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerTests.m
@@ -88,6 +88,7 @@ static NSString * const kOrgIdFormatString = @"00D000000000062EA%lu";
 @property (nonatomic, strong) SFUserAccountManager *uam;
 @property (nonatomic, strong) SFSDKAuthViewHandler *authViewHandler;
 @property (nonatomic, strong) SFSDKLoginViewControllerConfig *config;
+@property (nonatomic, strong) NSString *origLoginHost;
 
 - (SFUserAccount *)createNewUserWithIndex:(NSUInteger)index;
 - (NSArray *)createAndVerifyUserAccounts:(NSUInteger)numAccounts;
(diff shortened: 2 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1267,7 +1269,7 @@ - (SFSDKOAuthClient *)fetchOAuthClient:(SFOAuthCredentials *)credentials cached:
         __weak typeof(self) weakSelf = self;
         client = [SFSDKOAuthClient clientWithCredentials:credentials configBlock:^(SFSDKOAuthClientConfig *config) {
             __strong typeof(self) strongSelf = weakSelf;
-            config.loginHost = strongSelf.loginHost;
+            config.loginHost = credentials.domain ?: strongSelf.loginHost;
```
khawkins: This is food for thought down the road, not necessarily this PR.  It strikes me that creating `SFOAuthCredentials` instances to have some of their required fields populated non-atomically is fundamentally problematic, and something of an anti-pattern.  We do a number of checks like this in credentials objects, when we shouldn't have to: `SFOAuthCredentials` should have a number of fields that are required for it to be a valid data object, which in turn would imply requiring non-null values in initializers. `SFOAuthCredentials` should not exist without a `protocol`, `domain`, `identifier`, `clientId`, and `redirectUri`.

Something to think about for the future. I suppose you could debate whether you take that even further and say that a valid SFOAuthCredentials object implies a user that has authenticated at some point, so further values like `instanceUrl` become required too.  But regardless of that, the fields above should never not be a part of the credentials object—you can't do anything credentials-related with that object, without them.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1068,6 +1068,8 @@ - (void)setCurrentUser:(SFUserAccount*)user {
                 [self willChangeValueForKey:@"currentUser"];
                 _currentUser = user;
                 [self setCurrentUserIdentity:user.accountIdentity];
+                if (user.credentials.domain)
```
khawkins: I hope this isn't solving a problem, and is more of a nice to have.  `SFUserAccountManager.loginHost` should not be driving actions for any existing user, and if it is, such a dependency should be rectified/eliminated.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
trooper2013: @khawkins Thanks for the feedback. You've made some good arguments. In addition we should be thinking of the mutability of this object and its implications. Will add it to future improvements.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
khawkins: I should add that the other change in this PR looks like it may be rectifying exactly such a case?  I don't see any other obvious areas where `loginHost` would be used against this suggestion, so that's encouraging.

sgoldberg-sfdc: Is this setting the loginHost for the userAccountManager to the current users domain?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
khawkins: @sgoldberg-sfdc That's essentially it, though changing the implementation of `SFOAuthCredentials` could end up changing the shape of a lot of the current code that utilizes it—so who can be sure what this particular branch of code would look like, post-refactor.  But yes: if you say that credentials always have the appropriate bootstrapping data elements—and as @trooper2013 says, potential considerations on their immutability—in theory you should not need this check, as you'll always have the login host value that's relevant to this code, within the credentials object.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
khawkins: Yes.  As part of setting `currentUser`—which *should* entail a guarantee of a valid `SFUserAccount` *and* its underlying credentials—this code would change `loginHost` to reflect the current user's login host.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
sgoldberg-sfdc: in terms of mutability, wouldn't mutability of this property be necessary to support org splits/instance refresh?

khawkins: Depends on whether you considered the current credentials valid in the event that such data needed to change.  You may forego mutability in favor of cloning with updated data elements. It just depends on your feeling for the sanctity of the data.

Given that login host information should only change rarely, if ever, for a given user, I'd consider its mutability to be more of a liability than an asset, in general practice.

PR #2736, opened 2018-12-18
Set current login host during  user switch

trooper2013:
Cherry picked changes from master to dev


bhariharan:
@trooper2013 We should merge `master` into `dev` to avoid orphaned branches.

PR #2737, opened 2018-12-18
Regenerated doc

wmathurin:
Re-ran generate_docs.sh on a machine with
1) XCode 9.2
2) Only XCode 9.2
3) Mac OS 10.13.1 (High Sierra)
4) AppleDoc built/installed from sources

Not sure if 2), 3) or 4) was the key to success


bhariharan:
👍

wmathurin:
Was able to regenerate doc by doing just 4) (installing apple doc from sources)  - didn't need to xcode-select xcode 9.2, or run on mac os 13.1.

PR #2744, opened 2019-01-07
Updating readme.md per Apple's review

rwhitley:
- “Building from the command line has been tested using ant 1.8. Older versions might work, but we recommend using the latest version of ant.” — This doesn’t have any dependencies on ant, does it?
- The FAQ that the README links to appears to be very out of date. Is it worth linking to it anymore?
- As with the general documentation, in the What’s New in 7.0 section:
-- “Apple Inc.” is typically rendered without a comma.
-- “Mobile SDK no longer supports the Cocoa Lumberjack logging framework”: How about “no longer requires” instead of “no longer supports”. (Also, in the documentation, it appears that SFSDKLogger is deprecated? If so, it would be worth calling out here.)
-- “iOS base version” should be “iOS base SDK version”, to reduce confusion.
-- PromiseKit is included in both “Swift API Updates” and “Deprecation”. Perhaps it should just be under “Deprecation”, so that the “Swift API Updates” section is not leading with the removal of something.
-- Instead of the item starting with “PromiseKit APIs are no longer available…”, how about: “PromiseKit is no longer a dependency of the Mobile SDK. The Swift APIs have been designed around closures and/or delegates for asynchronous calls.”
- “The SDK also provides Objective-C wrappers for the Salesforce REST API…” Should this be “Swift and Objective-C”, or maybe something like “provides wrappers for the Salesforce REST API that can be used from both Swift and Objective-C”?

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -26,21 +26,18 @@ After cloning the repo:
 
 This script pulls the submodule dependencies from GitHub, to finalize setup of the workspace.  You can then work with the Mobile SDK by opening `SalesforceMobileSDK.xcworkspace` from Xcode.
 
-The Salesforce Mobile SDK for iOS requires iOS 11.0 or greater.  The install.sh script checks for this, and aborts if the configured SDK version is incorrect.  Building from the command line has been tested using ant 1.8.  Older versions might work, but we recommend using the latest version of ant.
-
-If you have problems building any of the projects, take a look at the online [FAQ](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/wiki/FAQ) for troubleshooting tips.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -63,16 +61,16 @@ through My Domain settings.
...
 - iOS minimum version (deployment target): iOS 11
-- iOS base version: iOS 12
+- iOS base SDK version: iOS 12
 - Xcode minimum version: 10
   
 **Deprecation**
-- `PromiseKit` APIs are no longer supported. Instead, use standard delegates or blocks to handle asynchronous callbacks.
-
+- `PromiseKit` is no longer a dependency of Mobile SDK. Instead, you can use standard delegates or blocks to handle asynchronous calls.
+- `SFSDKLogger` is now deprecated in Mobile SDK apps. Use `os_log()` instead.
```
brandonpage: This is a great catch.

rwhitley: I agree.

PR #2745, opened 2019-01-08
Fix special character handling bug in SmartStore

wmathurin:
Not all special characters were properly handled following this change: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2665

### DIFF
--- a/libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
+++ b/libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
@@ -583,6 +583,62 @@
 }
 
 /**
+ * Test query against soup with special characters when soup has string index
+ */
+-(void) testQueryDataWithSpecialCharactersWithStringIndex
+{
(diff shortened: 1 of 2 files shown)


libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
```diff
@@ -582,6 +582,62 @@ -(void) trySmartQuery:(NSString*)indexType
...
+
+/**
+ * Test query against soup with special characters when soup has json1 index
+ */
+-(void) testQueryDataWithSpecialCharactersWithJSON1Index
+{
+    [self tryQueryDataWithSpecialCharacters:kSoupIndexTypeJSON1];
+}
+
+-(void) tryQueryDataWithSpecialCharacters:(NSString*)indexType
```
wmathurin: NB: This test fails without the change in SFSmartStore.m

PR #2752, opened 2019-01-22
Upgrading to SQLCipher 4.0.1

wmathurin:
Taking the path of least resistance, opening every database with 3.x settings (except of kdf iter, where we are still using 2.x value for performance reason).

We should evaluate if 4.x settings have performance benefits (e.g. the larger page size). 
If they do, we will have to either:
- have code to try 4.x settings first then fall back on 3.x settings if that fails (to open older databases),
- or we should have code (or an API) for migrating older databases to 4.x.

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
@@ -182,13 +182,28 @@ static NSString * const kSFSmartStoreVerifyReadDbErrorDesc = @"Could not read fr
 + (FMDatabase*) setKeyForDb:(FMDatabase*) db key:(NSString *)key error:(NSError **)error {
     [db setLogsErrors:YES];
     [db setCrashOnErrors:NO];
-    FMDatabase* unlockedDb = [self unlockDatabase:db key:key kdfIter:4000];
+    FMDatabase* unlockedDb = [self unlockDatabase:db key:key pragmas:@[@"PRAGMA cipher_default_kdf_iter = 4000"]];
     
+    if (!unlockedDb) {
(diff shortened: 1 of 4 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
```diff
@@ -182,13 +182,9 @@ + (FMDatabase *)openDatabaseWithPath:(NSString *)dbPath key:(NSString *)key erro
 + (FMDatabase*) setKeyForDb:(FMDatabase*) db key:(NSString *)key error:(NSError **)error {
     [db setLogsErrors:YES];
     [db setCrashOnErrors:NO];
-    FMDatabase* unlockedDb = [self unlockDatabase:db key:key kdfIter:4000];
     
-    if (!unlockedDb) {
-        // You must be have created your database with SDK 3.2 or 3.1 and cocoapods
-        // And therefore you are using sqlcipher 3.1 with 64000 iterations
-        unlockedDb = [self unlockDatabase:db key:key kdfIter:64000];
```
wmathurin: When we started using SQLCipher 3.x (in SDK 3.1), we didn't put in the code to keep using kdf iter of 4000 right away. Apps using the compiled library should be okay because we compiled with that default value - but apps using cocoapods ended up using the new kdf iter of 64000.
We fixed that a few months later. 
That code was there to gracefully handle opening those databases.
It's been 4 years. It's unlikely there are running apps using a database created in that short window. So dropping the code to support that case.

bhariharan: Ah yes, I remember this. The difference was based on the way it was being consumed, i.e. Cocoapods had a different kdf iter.

PR #2754, opened 2019-01-22
SFRestAPI Upgrades, Ability to invoke endpoints in an unauthenticated context

trooper2013:
@bhariharan @wmathurin please review. This PR Does the following. 

Note: All Core,SmartStore & SmartSync tests are passing. Im still testing all sample apps to ensure that the result is as expected.

1. Fixes issues related to community files api(s). [Read about the required changes](https://developer.salesforce.com/docs/atlas.en-us.chatterapi.meta/chatterapi/connect_resources_files_rendition.htm).

2. Adds a sharedGlobalInstance of SFRestAPI for use in unauthenticated context.

3. Allows ability to invoke custom REST API endpoints.

```
 SFRestRequest *request = [SFRestRequest customUrlRequestWithMethod:SFRestMethodGET baseURL:@"https://api.github.com" path:@"/orgs/forcedotcom/repos" queryParams:nil];
 request.requiresAuthentication = NO;  
 [[SFRestAPI sharedGlobalInstance] sendRESTRequest:request failBlock:^(NSError *  e, 
      NSURLResponse * rawResponse) {
       
  } completeBlock:^(id  resp, NSURLResponse *  rawResponse) {
           NSDictionary *response = resp;
           // response should have forcedotcom repos      
  }];
```
4. Allows the setting of service type in Request. This enables the OS to prioritize requests.  
[Read](https://developer.apple.com/documentation/foundation/nsurlrequestnetworkservicetype?language=objc).

5. Added Unit tests for most of the changes.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Files.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Files.m
@@ -125,7 +125,7 @@
 }
 
 - (NSString *)communitiesUrlPathIfRequired {
-    if (!self.user.credentials.communityId) {
+    if (!self.user.communityId) {
         return @"";
     }
     return [NSString stringWithFormat:@"/communities/%@", self.user.communityId];
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Files.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Files.m
@@ -27,7 +27,7 @@
 #import <SalesforceSDKCommon/SFJsonUtils.h>
 #import "SFRestAPI+Files.h"
 #import "SFRestRequest+Internal.h"
-
+#import "SFOAuthCredentials.h"
 #define ME @"me"
 #define PAGE @"page"
 #define VERSION @"versionNumber"
@@ -40,28 +40,28 @@
 @implementation SFRestAPI (Files)
 
 - (SFRestRequest *) requestForOwnedFilesList:(NSString *)userId page:(NSUInteger)page {
-    NSString *path = [NSString stringWithFormat:@"/%@/connect/files/users/%@", self.apiVersion, (userId == nil ? ME : userId)];
+    NSString *path = [NSString stringWithFormat:@"/%@/connect%@/files/users/%@", self.apiVersion,[self communitiesUrlPathIfRequired],  (userId == nil ? ME : userId)];
     NSMutableDictionary *params = [NSMutableDictionary dictionary];
     if (page) params[PAGE] = @(page);
     return [SFRestRequest requestWithMethod:SFRestMethodGET path:path queryParams:params];
 }
 
 - (SFRestRequest *) requestForFilesInUsersGroups:(NSString *)userId page:(NSUInteger)page {
-    NSString *path = [NSString stringWithFormat:@"/%@/connect/files/users/%@/filter/groups", self.apiVersion, (userId == nil ? ME : userId)];
+    NSString *path = [NSString stringWithFormat:@"/%@/connect%@/files/users/%@/filter/groups", self.apiVersion, [self communitiesUrlPathIfRequired], (userId == nil ? ME : userId)];
     NSMutableDictionary *params = [NSMutableDictionary dictionary];
     if (page) params[PAGE] = @(page);
     return [SFRestRequest requestWithMethod:SFRestMethodGET path:path queryParams:params];
 }
 
 - (SFRestRequest *) requestForFilesSharedWithUser:(NSString *)userId page:(NSUInteger)page {
-    NSString *path = [NSString stringWithFormat:@"/%@/connect/files/users/%@/filter/sharedwithme", self.apiVersion, (userId == nil ? ME : userId)];
+    NSString *path = [NSString stringWithFormat:@"/%@/connect%@/files/users/%@/filter/sharedwithme", self.apiVersion, [self communitiesUrlPathIfRequired],(userId == nil ? ME : userId)];
     NSMutableDictionary *params = [NSMutableDictionary dictionary];
     if (page) params[PAGE] = @(page);
     return [SFRestRequest requestWithMethod:SFRestMethodGET path:path queryParams:params];
 }
 
 - (SFRestRequest *) requestForFileDetails:(NSString *)sfdcId forVersion:(NSString *)version {
-    NSString *path = [NSString stringWithFormat:@"/%@/connect/files/%@", self.apiVersion, sfdcId];
+    NSString *path = [NSString stringWithFormat:@"/%@/connect%@/files/%@", self.apiVersion,[self communitiesUrlPathIfRequired], sfdcId];
     NSMutableDictionary *params = [NSMutableDictionary dictionary];
     if (version) params[VERSION] = version;
     return [SFRestRequest requestWithMethod:SFRestMethodGET path:path queryParams:params];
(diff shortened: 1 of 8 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.h
```diff
@@ -38,6 +38,17 @@ typedef NS_ENUM(NSInteger, SFRestMethod) {
     SFRestMethodPATCH
 } NS_SWIFT_NAME(RestRequest.Method);
 
+/**
+ * HTTP methods for requests.
+ */
+typedef NS_ENUM(NSInteger, SFSDKNetworkServiceType) {
```
bhariharan: Would it make sense for this to live in `SFNetwork` instead since it's more of a network property than API client related?

trooper2013: This is part of NSURLRequest in foundation. Hence it made more sense to add it our SFRestRequest. i.e. easier to lookup and use for a developer.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -272,7 +289,7 @@ - (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)del
                 [strongSelf notifyDelegateOfResponse:delegate request:request data:dataForDelegate rawResponse:response];
             }
             // 401 (and sometimes 403) indicates refresh is required.
-            else if (request.shouldRefreshOn403 ? (statusCode == 401 || statusCode == 403) : (statusCode == 401)) {
+            else if (request.requiresAuthentication && request.shouldRefreshOn403 ? (statusCode == 401 || statusCode == 403) : (statusCode == 401)) {
```
wmathurin: Or maybe we should make sure shouldRetry is no if requiresAuthenticate is false.
By changing 
```objective-c
- (void)send:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate {
    [self send:request delegate:delegate shouldRetry:YES];
}
```
to
```objective-c
- (void)send:(SFRestRequest *)request delegate:(id<SFRestDelegate>)delegate {
    if (request.requiresAuthentication) {
        [self send:request delegate:delegate shouldRetry:YES];
    }
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Files.m
```diff
@@ -40,41 +40,41 @@
 @implementation SFRestAPI (Files)
 
 - (SFRestRequest *) requestForOwnedFilesList:(NSString *)userId page:(NSUInteger)page {
-    NSString *path = [NSString stringWithFormat:@"/%@/connect/files/users/%@", self.apiVersion, (userId == nil ? ME : userId)];
+    NSString *path = [NSString stringWithFormat:@"/%@/connect%@/files/users/%@", self.apiVersion,[self communitiesUrlPathIfRequired],  (userId == nil ? ME : userId)];
```
wmathurin: Ideally the community bug fix should be in a separate PR since it's completely unrelated to the support for unauthenticated calls.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.h
bhariharan: Sounds good.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -39,7 +39,7 @@ - (id)initWithMethod:(SFRestMethod)method serviceHostType:(SFSDKRestServiceHostT
         self.baseURL = baseURL;
         self.path = path;
         self.queryParams = [queryParams mutableCopy];
-        self.endpoint = kSFDefaultRestEndpoint;
+        self.endpoint = (hostType == SFSDKRestServiceHostTypeCustom)?@"":kSFDefaultRestEndpoint;
```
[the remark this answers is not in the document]
bhariharan: `baseUrl` doesn't always have the information of whether an API should go against `loginUrl` or `instanceUrl`. I think we still need `hostType` for those special `loginUrl` URLs like the `userInfo` endpoint. I forget the exact specifics.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -122,84 +126,83 @@ - (void)setCustomRequestBodyStream:(NSInputStream* (^)(void))bodyInputStreamBloc
...
 
 - (NSURLRequest *)prepareRequestForSend:(SFUserAccount *)user {
-    if (user) {
-
-        /*
-         * If an absolute URL is passed in, use it as-is. If a relative URL is passed in,
-         * parse it and put the pieces together to construct the full URL.
-         */
-        NSMutableString *fullUrl = nil;
-        if ([[self.path lowercaseString] hasPrefix:@"https://"]) {
```
bhariharan: @trooper2013 We still need this `if` condition as I mentioned in our conversation offline. There are a few different use cases for this. Some are covered by your new method that takes a custom endpoint (the native use cases), but removing this will break existing clients that rely on this behavior (we have some users internally). Also, it will remove the ability to accomplish this from our `Cordova` and `ReactNative` apps. We need to bring this `if` condition back, at least for now.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
```diff
@@ -122,84 +126,83 @@ - (void)setCustomRequestBodyStream:(NSInputStream* (^)(void))bodyInputStreamBloc
...
+
+    //Set the service host type
+    NSURLRequestNetworkServiceType serviceType = [self urlRequestServiceType:self.networkServiceType];
+    [self.request setNetworkServiceType:serviceType];
+    
+    // Sets HTTP method on the request.
+    [self.request setHTTPMethod:[SFRestRequest httpMethodFromSFRestMethod:self.method]];
+
+    // Sets OAuth Bearer token header on the request (if not already present).
+    if (self.requiresAuthentication && ![self.request.allHTTPHeaderFields.allKeys containsObject:@"Authorization"]) {
```
bhariharan: This should be:
```
if (user && self.requiresAuthentication && ![self.request.allHTTPHeaderFields.allKeys containsObject:@"Authorization"])
```
This guards against someone who has an unauthenticated instance (`user == nil`), but has `self.requiresAuthentication == YES`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.m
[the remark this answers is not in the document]
wmathurin: Looking at `prepareRequestForSend`, it looks like if path starts with https then it is used the full url. In that case, base url, end point and host type don't matter. 

Maybe the method `customUrlRequestWithMethod` should just take one parameter: the full url (which would be fed into the path field of the SFRestRequest).

Also the comment for the method is misleading since it doesn't help the case where you want to hit a custom end point on salesforce (e.g. an api under `/services/apexrest`). For that case, a useful factory method would be one taking end point and path.

So in summary, useful new factory methods would be:
```objective-c
+ (instancetype)customUrlRequestWithMethod:(SFRestMethod)method fullURL:(NSString *)fullURL queryParams:(nullable NSDictionary<NSString*, id> *)queryParams;

+ (instancetype)customEndPointRequestWithMethod:(SFRestMethod)method endPoint:(NSString *)endPoint path:(NSString *)path queryParams:(nullable NSDictionary<NSString*, id> *)queryParams;
```

PR #2765, opened 2019-01-31
use atomic for communities

liguopku:
@brandonpage @trooper2013 @sgoldberg-sfdc 

We are seeing a crash on hockeyapp: https://rink.hockeyapp.net/manage/apps/31036/app_versions/179/crash_reasons/259010550?type=crashes

After some investigation, we think it is a read/write race condition when we try to change user.communities in a callback in main and also read user.communities in another background thread.
screen shot 2019-01-31 at 12 03 53 pm

Thus change user.communities from nonatomic to atomic to make sure it is threadsafe.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
@@ -115,7 +115,7 @@ NS_SWIFT_NAME(UserAccount)
 
 /** The list of communities (as SFCommunityData item)
  */
-@property (nonatomic, copy, nullable) NSArray<SFCommunityData *> *communities;
+@property (atomic, copy, nullable) NSArray<SFCommunityData *> *communities;
 
 /** Returns YES if the user has an access token and, presumably,
  a valid session.


liguopku:
@trooper2013 In our locally debugging, we found that there is a possible race condition happens in 
1. we are changing communities in main thread at https://git.soma.salesforce.com/iOS/Salesforce1/blob/release-18.1/Chatter/src/Chatter/Chatter/Classes/Common/ViewControllers/Containers/Switcher/SFSwitcherView.m#L175
2. we are reading communities in another background thread at https://git.soma.salesforce.com/iOS/Salesforce1/blob/release-18.1/Chatter/src/Chatter/Chatter/Classes/Common/ViewControllers/Containers/Switcher/SFSwitcherView.m#L312

by the time we try to execute `enumerateObjectsUsingBlock` in `populateItems ` method, user. communities has been changed and previous communities will point to a bad memory. thus it will cause a crash in the following report
screen shot 2019-01-31 at 1 09 45 pm
screen shot 2019-01-31 at 1 09 32 pm

the issues is hard to reproduce thus we could not test out before/after the fix to verify the change. 

for the other nonatomic property, it won't run it the same cases if they are readonly
for nonreadonly property, there is a risk they run into similar case but it will depends on how ofter we are updating their values...

Using a SFSDKSafeMutableSet as a collection of all the properties can help fix it here but requires much more work to refactor the current class. SFSDKSafeMutableSet also need to support NSCoding?

trooper2013:
@liguopku We should have a discussion offline. I don't think that this fix is complete. Are there other alternatives e.g. your code could check the status of the user also take a copy of the collection maybe? In the grand scheme of things we may need to address the mutability/immutability of the SFUserAccount.

liguopku:
sure thanks! let me reach out to you offline to discuss.

PR #2774, opened 2019-02-08
Thread Safety for  User Account

trooper2013:
This PR attempts to make read-write(s) to UserAccount Thread Safe.

1. SFUserAccount thread safety is implemented using dispatch barrier.
2. Made SFUserAccountIdentity and SFOAuthCredential properties readOnly for public apis.
3. Some SFUserAccountProperties are backed by credentials and identity, the rest are Deprecated.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials+Internal.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials+Internal.h
@@ -42,18 +42,33 @@ extern NSException * SFOAuthInvalidIdentifierException(void);
 
 @interface SFOAuthCredentials ()
 
-@property (nonatomic, strong) NSDictionary * additionalOAuthFields;
 
-// Facilitates NSCopying implementation.
-@property (nonatomic, readwrite) NSDictionary *legacyIdentityInformation;
+@property (nonatomic, readwrite, nullable) NSString *protocol;
+@property (nonatomic, readwrite, nullable) NSString *domain;
+@property (nonatomic,readwrite) NSString *identifier;
+@property (nonatomic, readwrite, nullable) NSString *clientId;
+@property (nonatomic, readwrite, nullable) NSString *redirectUri;
+@property (nonatomic, readwrite, nullable) NSString *jwt;
+@property (nonatomic, readwrite, nullable) NSString *refreshToken;
+@property (nonatomic, readwrite, nullable) NSString *accessToken;
+@property (nonatomic, readwrite, nullable) NSString *organizationId;
+@property (nonatomic, readwrite, nullable) NSURL *instanceUrl;
+@property (nonatomic, readwrite, nullable) NSString *communityId;
+@property (nonatomic, readwrite, nullable) NSURL *communityUrl;
+@property (nonatomic, readwrite, nullable) NSDate *issuedAt;
+@property (nonatomic, readwrite, nullable) NSURL *identityUrl;
+@property (nonatomic, readwrite, nullable) NSURL *apiUrl;
+@property (nonatomic, readwrite, nullable) NSString *userId;
 
-/** Holds the attributes that have changed during auth flow.
- */
-@property (nonatomic) NSMutableDictionary * credentialsChangeSet;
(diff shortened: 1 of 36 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -97,198 +96,273 @@ - (instancetype)initWithCredentials:(SFOAuthCredentials*) credentials {
...
+
+- (NSArray<SFCommunityData *> *)communities {
+    __block NSArray<SFCommunityData *> *communitiesLocal = nil;
+    dispatch_sync(_syncQueue, ^{
+        communitiesLocal = self->_communities;
+    });
+    return communitiesLocal;
+}
+
+- (void)setCommunities:(NSArray<SFCommunityData *> *)communities {
```
bhariharan: For `8.0`, this should also be removed (we spoke about it earlier). This property that backs `SFCommunityData` should be moved to an extension in the app code. It has no place in the `Mobile SDK`, since we don't populate it, nor do we use it. It satisfies an app specific use case.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -97,198 +96,273 @@ - (instancetype)initWithCredentials:(SFOAuthCredentials*) credentials {
...
+            if ([fm fileExistsAtPath:photoPath]) {
+                if (![fm removeItemAtPath:photoPath error:&error]) {
+                    [SFSDKCoreLogger e:[self class] format:@"Unable to remove previous photo from disk: %@", error];
+                }
             }
+            NSData *data = UIImagePNGRepresentation(photo);
+            if (![data writeToFile:photoPath options:NSDataWritingAtomic error:&error]) {
+                [SFSDKCoreLogger e:[self class] format:@"Unable to write photo to disk: %@", error];
+            }
+            [self willChangeValueForKey:@"photo"];
```
bhariharan: Should this pattern of `willChange` and `didChange` be changed to a notification pattern like the rest of the SDK?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials+Internal.h
```diff
@@ -42,18 +42,33 @@ extern NSException * SFOAuthInvalidIdentifierException(void);
...
+@property (nonatomic, readwrite, nullable) NSURL *apiUrl;
+@property (nonatomic, readwrite, nullable) NSString *userId;
 
-/** Holds the attributes that have changed during auth flow.
- */
-@property (nonatomic) NSMutableDictionary * credentialsChangeSet;
+@property (nonatomic, readwrite, strong, nullable) NSDictionary * additionalOAuthFields;
+@property (nonatomic, readwrite, nullable) NSString *challengeString;
+@property (nonatomic, readwrite, nullable) NSString *authCode;
+@property (nonatomic, readwrite, nullable) NSDictionary *legacyIdentityInformation;
```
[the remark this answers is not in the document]
trooper2013: @bhariharan  I think we should remove it completely. Theres some more remnants of this code. I will remove those too. This is where it came from  PR #661

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
trooper2013: @bhariharan We should then deprecate communityUrlWithId, communityWithId too. What do u think?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
trooper2013: We can certainly  this legacy code to NSNotification.

sirisak1985:
good app set up

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
bhariharan: Sounds good.

PR #2778, opened 2019-02-13
Updating framework version

bhariharan:
Looks like we missed one file.


wmathurin:
Updated `setversion.sh` here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2781

PR #2783, opened 2019-02-19
BatchingSyncUpTarget for iOS

wmathurin:
1) Repurposed `AdvancedSyncUpTarget` to support batching (https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/d01d667c451d2154d31ec489f70aeedcf97f3be8)
2) Sync manager changes were pretty mininmal (https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/155226bfe3c606f0d3ef70db5b55e38e8f192dac)
2) For tests, moved existing sync up tests to a new class `SyncUpTargetTests` and made sure, there was only one place in the whole class where the sync up target was actually built. Added a subclass `BatchingSyncUpTargetTests` which creates `SFBatchingSyncUpTarget`'s instead of `SFSyncUpTarget`'s. So all the existing sync up tests are now also run with a batching sync up target. Using batch size 2 (see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/334a80a60ff8b6280e275d9d1edacd5a328be824) to test going to the server more than once during a sync (since existing tests deal with 3-5 dirty records).


wmathurin:
@trooper2013 Maybe we could have helping code (builders) in SalesforceSDKCore.

PR #2791, opened 2019-02-27
Using constructors instead of factory methods in SFBatchingSyncUpTarget

wmathurin:
More in line with super classes.

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/SFBatchingSyncUpTarget.h
+++ b/libs/SmartSync/SmartSync/Classes/SFBatchingSyncUpTarget.h
@@ -34,12 +34,15 @@ NS_SWIFT_NAME(BatchingSyncUpTarget)
  */
 @interface SFBatchingSyncUpTarget : SFSyncUpTarget <SFAdvancedSyncUpTarget>
 
-/** Factory methods
+/** Constructor
  */
+- (instancetype)initWithCreateFieldlist:(nullable NSArray *)createFieldlist
+                        updateFieldlist:(nullable NSArray *)updateFieldlist
+                           maxBatchSize:(nullable NSNumber *)maxBatchSize;
+ 
 
-+ (instancetype)newSyncTargetWithCreateFieldlist:(nullable NSArray *)createFieldlist
-                                 updateFieldlist:(nullable NSArray *)updateFieldList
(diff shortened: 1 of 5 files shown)


libs/SmartSync/SmartSync/Classes/SFBatchingSyncUpTarget.h
```diff
@@ -34,12 +34,15 @@ NS_SWIFT_NAME(BatchingSyncUpTarget)
  */
 @interface SFBatchingSyncUpTarget : SFSyncUpTarget <SFAdvancedSyncUpTarget>
 
-/** Factory methods
+/** Constructor
  */
+- (instancetype)initWithCreateFieldlist:(nullable NSArray *)createFieldlist
```
trooper2013: @wmathurin Could we  specify the type in the list  like below?
``` (instancetype)initWithCreateFieldlist:(nullable NSArray<NSString *> *)createFieldlist  updateFieldlist:(nullable NSArray<NSString *> *)updateFieldlist maxBatchSize:(nullable NSNumber *)maxBatchSize; ```

Comment (bhariharan) [reply libs/SmartSync/SmartSync/Classes/SFBatchingSyncUpTarget.h]: Does this need `NS_SWIFT_NAME` magic too?

Comment (wmathurin) [reply libs/SmartSync/SmartSync/Classes/SFBatchingSyncUpTarget.h]: It's on line 31.

Comment (wmathurin) [reply libs/SmartSync/SmartSync/Classes/SFBatchingSyncUpTarget.h]: Here is the generated swift header
```swift
open class BatchingSyncUpTarget : SyncUpTarget, AdvancedSyncUpTarget {

    
    /** Constructor
     */
    public init(createFieldlist: [Any]?, updateFieldlist: [Any]?, maxBatchSize: NSNumber?)

    
    
    /** Factory method
     */
    open class func build(dict: [AnyHashable : Any]?) -> Self
}
```

wmathurin:
After last commit, here is the generated Swift 4.2 header for SFBatchingSyncUpTarget
```swift
open class BatchingSyncUpTarget : SyncUpTarget, AdvancedSyncUpTarget {

    public init(createFieldlist: [String]?, updateFieldlist: [String]?, maxBatchSize: NSNumber?)

    open class func build(dict: [AnyHashable : Any]?) -> Self

}
```

PR #2794, opened 2019-02-28
Restrict passcode flow during refresh.

trooper2013:
Trigger of passcode flow during refresh can lead to various problem .e.g if refresh is being done in background.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
@@ -1351,9 +1351,6 @@ static NSString *const  kOptionsClientKey          = @"clientIdentifier";
                                           lockoutTime:(client.idData.mobileAppScreenLockTimeout * 60)
                                      biometricAllowed:biometricUnlockAvailable];
     } else {
-        [SFSecurityLockout  updateInactivityConfigurationIfRequired:client.idData.mobileAppPinLength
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1338,17 +1338,24 @@ - (void)retrievedIdentityData:(SFSDKOAuthClient *)client
...
+        [SFSecurityLockout setLockScreenFailureCallbackBlock:^{
+            [weakSelf handleFailure:client.context.authError client:client notifyDelegates:YES];
+        }];
+        // Check to see if a passcode needs to be created or updated, based on passcode policy data from the
+        // identity service.
+        [SFSecurityLockout setInactivityConfiguration:client.idData.mobileAppPinLength
+                                          lockoutTime:(client.idData.mobileAppScreenLockTimeout * 60)
+                                     biometricAllowed:biometricUnlockAvailable];
+    } else {
+        [SFSecurityLockout  updateInactivityConfigurationIfRequired:client.idData.mobileAppPinLength
```
bhariharan: Why is this being updated during refresh flow? We should leave it alone.

bhariharan: We have always maintained the stance that passcode changes occur ONLY In the following scenarios:
1. Add new user.
2. Logout user.
For refresh scenarios, we should not update the policy. It adds a lot of unnecessary complexity.

PR #2796, opened 2019-03-01
Fix passcode flows

brandonpage:
- Stop app from forcing passcode verification on switch to new user.
- Fix issue where lockout time could increase.
- Fix issue of user not being able to login if initial passcode creation is interrupted.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecurityLockout.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecurityLockout.m
@@ -219,16 +219,6 @@ static BOOL _showPasscode = YES;
     return newLockoutTime !=0 || [self usersHavePasscodePolicy];
 }
 
-+ (BOOL)needsPasscodePolicyChange:(NSUInteger)newLockoutTime passcodeLength:(NSUInteger)newPasscodeLength {
-    BOOL result = NO;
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecurityLockout.m
```diff
@@ -219,16 +219,6 @@ + (BOOL)needsPasscodeFlow:(NSUInteger)newLockoutTime {
...
 }
 
-+ (BOOL)needsPasscodePolicyChange:(NSUInteger)newLockoutTime passcodeLength:(NSUInteger)newPasscodeLength {
-    BOOL result = NO;
-    if (securityLockoutTime == 0 && newLockoutTime > 0) {
-        result = YES;
-    }else if (newLockoutTime !=0 && newLockoutTime < securityLockoutTime){
-         result = YES;
-    }
-    return result || (newPasscodeLength > [self passcodeLength]);
```
brandonpage: This line is dangerous due to the or.  Since later on we just check != instead of <, it was possible to increase the lockout time if the same policy also increased the passcode length.

trooper2013: @brandonpage Let's discuss this offline.

PR #2802, opened 2019-03-05
Remove all related instances of analyticsmanager and sfrestapi when user logs out

sgoldberg-sfdc:
-  For community users we were only removing the currently logged in instance
-  This change will remove the singleton instances for the user for all communities


bhariharan:
@trooper2013 Let's discuss enhancements offline. I have a proposal.

PR #2805, opened 2019-03-07
Always pass auth info with browser auth type when cancelling from browser flow

paultiarks:
When using a host configured for advanced browser auth, the user cancelling the process caused a notification to be posted with an SFOAuthInfo object with the wrong authentication type. Since only browser auth flows used this code path, it was suggested by @rao-r to build a new SFOAuthInfo with the browser auth type and send that.

Any suggestions on where to fix the root issue to avoid having to use this approach would be appreciated.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
@@ -513,8 +513,9 @@ static NSString *const  kOptionsClientKey          = @"clientIdentifier";
 
 - (BOOL)authClientDidCancelBrowserFlow:(SFSDKOAuthClient *)client {
     BOOL result = NO;
+    SFOAuthInfo *authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeAdvancedBrowser];
     NSDictionary *userInfo = @{ kSFNotificationUserInfoCredentialsKey: client.credentials,
-                                kSFNotificationUserInfoAuthTypeKey: client.context.authInfo };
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -513,8 +513,9 @@ - (void)authClient:(SFSDKOAuthClient *)client willBeginBrowserAuthentication:(SF
 
 - (BOOL)authClientDidCancelBrowserFlow:(SFSDKOAuthClient *)client {
     BOOL result = NO;
+    SFOAuthInfo *authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeAdvancedBrowser];
     NSDictionary *userInfo = @{ kSFNotificationUserInfoCredentialsKey: client.credentials,
-                                kSFNotificationUserInfoAuthTypeKey: client.context.authInfo };
+                                kSFNotificationUserInfoAuthTypeKey: authInfo };
```
bhariharan: Same question as the other PR -> Why doesn't `client.context.authInfo` have the correct `authType` here? @trooper2013 @paultiarks

trooper2013: @bhariharan There is no harm in doing this here. We changed advanced auth to be configured thru a domain/org setting only. The Auth Client is configured to start with UserAgent and kicked off.

bhariharan: Gotcha. So the `client` changed dynamically.

PR #2806, opened 2019-03-08
Using Secure Enclave to encrypt key store 

wmathurin:
Added new helper methods in `SFCryptoUtils` for Elliptic Curve key pair generation / encryption / decryption (with and without Secure Enclave).

Added a new class `SFSecureEncryptionKey` which takes advantage of the new Elliptic Curve helper methods (and takes advantage of Secure Enclave when it is available - but works also without it). 
It is secured because there is no way to get the bits of keys (when secure enclave is available).
It can be used in all places that use `SFEncryptionKey` except in SmartStore and in files (because those expect to get the bits of the keys).

Changed `SFKeyStoreManager` to use `SFSecureEncryptionKey` by default as the encryption key (inside the `SFKeyStoreKey` used to encrypt its `SFKeyStore`).
Also changed `SFKeyStoreManager` to automatically switch its `SFKeyStore` to a `SFSecureEncryptionKey` if it is found to still be using a `SFEncryptionKey`.
(The key store contains all the keys used to encrypt sensitive data like the OAuth tokens, SmartStore, files, etc).

Added tests (including for upgrade).


libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSecurityTests.m
```diff
@@ -196,6 +199,33 @@ - (void)testUpgradeTo60PasscodeEnabled
     [mgr removeKeyWithLabel:keyLabel];
 }
 
+- (void)testUpgradeTo71
```
wmathurin: Upgrade test

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFKeyStoreManager.m
```diff
@@ -118,13 +119,28 @@ - (void)initializeKeyStores
         self.generatedKeyStore.keyStoreKey = [self createDefaultKey];
     }
     else {
+        // Pre SDK 7.1 SFGeneratedKeyStore were encrypted with SFEncryptionKey
+        // Starting in SDK 7.1, we use SFSecureEncryptionKey instead
+        // Switch to SFSecureEncryptionKey if needed
+        [self switchToSecureKeyIfNeeded:self.generatedKeyStore];
```
wmathurin: That's the upgrade code.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -60,7 +60,14 @@
 		4F06AF921C49A18E00F70798 /* SFPreferencesTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F06AF6D1C49A16A00F70798 /* SFPreferencesTests.m */; };
 		4F06AF931C49A18E00F70798 /* SFSecurityLockoutTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F06AF6F1C49A16A00F70798 /* SFSecurityLockoutTests.m */; };
 		4F06AF941C49A18E00F70798 /* SFTestSDKManagerFlow.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F06AF711C49A16A00F70798 /* SFTestSDKManagerFlow.m */; };
+		4F16F4C1222A1C53008529FB /* SalesforceAnalytics.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = B716A398218F5E37009D407F /* SalesforceAnalytics.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
bhariharan: These entries should be removed. `Embed Frameworks` should only apply at the app level. I don't think we should embed one framework within another.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -818,6 +825,8 @@
 			dstPath = "";
 			dstSubfolderSpec = 10;
 			files = (
+				4F16F4C2222A1C5A008529FB /* SalesforceSDKCommon.framework in Embed Frameworks */,
```
bhariharan: Remove this - aren't they already linked?

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -3011,6 +3032,7 @@
 				CLANG_WARN_ENUM_CONVERSION = YES;
 				CLANG_WARN_INT_CONVERSION = YES;
 				CLANG_WARN_OBJC_ROOT_CLASS = YES_ERROR;
+				ENABLE_BITCODE = NO;
```
bhariharan: Why no `Bitcode`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecureEncryptionKey.m
```diff
@@ -0,0 +1,193 @@
...
+
+@end
+
+@implementation SFSecureEncryptionKey
+
+#pragma mark - Factory methods and constructor
+
++ (instancetype) createKey:(NSString*)label
+{
+    [SFSDKCryptoUtils deleteECKeyPairWithName:label]; // Delete existing key if any
```
bhariharan: Should this delete and recreate if it already exists, or just leave it alone and return the existing instance, i.e. call `retrieveKey` instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecureEncryptionKey.m
```diff
@@ -0,0 +1,193 @@
...
+        }
+        // Creation failed
+        else {
+            return nil;
+        }
+    }
+    return self;
+}
+
+- (void) dealloc {
```
bhariharan: Is this file not ARC-enabled either?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.m
```diff
@@ -27,6 +27,8 @@
 #import <CommonCrypto/CommonCrypto.h>
 #import "NSData+SFAdditions.h"
 #import <Security/Security.h>
+#include "TargetConditionals.h"
```
bhariharan: Is this file -> `https://opensource.apple.com/source/CarbonHeaders/CarbonHeaders-18.1/TargetConditionals.h`? Should it be `#import`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecureEncryptionKey.m
wmathurin: The compiler does not automatically manage the lifetimes of Core Foundation objects; you must call CFRetain and CFRelease.
https://developer.apple.com/library/archive/releasenotes/ObjectiveC/RN-TransitioningToARC/Introduction/Introduction.html#//apple_ref/doc/uid/TP40011226-CH1-SW1

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.m
```diff
@@ -499,4 +629,44 @@ + (size_t)encodeLength:(unsigned char *)buf length:(size_t)length {
...
+        // Handle the error. . .
+        NSError *error = [NSError errorWithDomain:NSOSStatusErrorDomain code:status userInfo:nil];
+        [SFSDKCoreLogger e:[self class] format:@"Error getting EC SecKeyRef with tag %@. Error code: %@", keyTagString, error.localizedDescription];
+        return nil;
+    }
+    
+    return keyRef;
+}
+
++ (void)deleteKeyByTag:(NSString*)keyTagString {
```
trooper2013: @wmathurin We should either return an error or a boolean to allow calling code to handle  the failure.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.m
```diff
@@ -230,6 +236,130 @@ + (nullable NSData*)decryptUsingRSAforData:(NSData *)data withKeyRef:(SecKeyRef)
...
+        CFRelease(publicKey);
+    }
+    if (privateKey != nil) {
+        CFRelease(privateKey);
+    }
+}
+
++ (void)deleteECKeyPairWithName:(NSString *)keyName
+{
+    [SFSDKCryptoUtils deleteKeyByTag:[NSString stringWithFormat:@"%@.%@", kSFECPublicKeyTagPrefix, keyName]];
```
trooper2013: What would happen in the case that there is a failure to delete the private key and not the public or vice versa?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.m
```diff
@@ -230,6 +236,130 @@ + (nullable NSData*)decryptUsingRSAforData:(NSData *)data withKeyRef:(SecKeyRef)
...
+{
+#if TARGET_OS_SIMULATOR
+    return NO;
+#else
+    LAContext *context = [[LAContext alloc] init];
+    return [context canEvaluatePolicy:LAPolicyDeviceOwnerAuthenticationWithBiometrics error:nil];
+#endif
+}
+
++ (void)createECKeyPairWithName:(NSString *)keyName accessibleAttribute:(CFTypeRef)accessibleAttribute useSecureEnclave:(BOOL)useSecureEnclave
```
trooper2013: @wmathurin  Shouldn't we should bubble up errors?

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
wmathurin: Compiler complains when building tests to run on device:
```
Showing All Messages
ld: -bundle and -bitcode_bundle (Xcode setting ENABLE_BITCODE=YES) cannot be used together
```
So either ENABLE_BITCODE has to be set to NO or the Match-O type for SalesforceSDKCoreTests target has to be changed to something else than Bundle.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.m
wmathurin: It's unlikely since they have the same access controls.
When creating/retrieving a key, we see if we can get both otherwise we consider that the key doesn't exist - we then either create a new key or return nil (depending on whether you do a create or a retrieve). See the getKeyRefs method.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
wmathurin: Without them, it only runs on simulator, and crashes on device.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
wmathurin: Actually disabling bitcode seems to be the only way to run it on device.
> ld: ‘/Applications/Xcode10.2.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/Library/Frameworks/XCTest.framework/XCTest’ does not contain bitcode. You must rebuild it with bitcode enabled (Xcode setting ENABLE_BITCODE), obtain an updated library from the vendor, or disable bitcode for this target. file ‘/Applications/Xcode10.2.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/Library/Frameworks/XCTest.framework/XCTest’ for architecture arm64

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecureEncryptionKey.m
wmathurin: We can go either way. I don't have a preference.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
bhariharan: I didn't realize this was for the test app. That makes sense.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecureEncryptionKey.m
bhariharan: I would probably go with leaving it alone and returning the existing instance, but I'll leave it up to you.

PR #2820, opened 2019-04-03
Supports sharing of SmartStore using plain text header for encrypted sqlite files.

trooper2013:
This PR does the following

- Allows for encrypted sqlite files to be shared between apps. It does so by enabling the ability for sqlcipher to store the headers for the sqlite file in plain text. Usually any app holding a write lock to a file is jetsam'd  by the jetsam monitor when the app is suspended. This rule applies to all files but sqlite files. SQLCipher encrypts the sqlite files and jetsam monitor is not able to detect the sqlite file causing the app to be jetsam'd 

- Switches the journaling mode for the database. 

- Stores the salt in keychain, since the salt is not stored with the database anymore. The salt  is applied subsequently using pragma. 

- Switches the DB to this shared mode only if app groups have been enabled. 

Note: Various upgrade scenarios have to be tested.  Will push out another PR for handling that situation.

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStore+Internal.h
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStore+Internal.h
@@ -211,6 +211,11 @@ typedef NS_ENUM(NSUInteger, SFSmartStoreFtsExtension) {
 + (NSString *)encKey;
 
 /**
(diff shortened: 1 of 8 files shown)


trooper2013:
@wmathurin I will push tests and upgrade scenarios in another PR. So going back from salt in keychain to salt in header will require us to patch the sqlite file.

wmathurin:
Tests failed on CI (compilation errors in tests classes).

PR #2821, opened 2019-04-03
Fixing my domain config fetching for unauthenticated users

bhariharan:
The whole point of fetching domain config from `.well-known` is that it can be used to determine the type of authentication, i.e. should not be tied to a user account. The existing API encourages callers to pass in `user.credentials`, which breaks in the unauthenticated use case because our code constructs a request to `null://null`, which fails. All we need is `loginDomain` to make the request to. This lines up the implementation with how it is on `Android`. I mentioned this briefly to @trooper2013 earlier this week. I'll have a related PR for the hybrid repo too.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKAuthConfigUtil.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKAuthConfigUtil.h
@@ -28,6 +28,7 @@
 
 #import "SFOAuthOrgAuthConfiguration.h"
 #import "SFOAuthCredentials.h"
+#import "SalesforceSDKConstants.h"
 
 @interface SFSDKAuthConfigUtil : NSObject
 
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKAuthConfigUtil.h
```diff
@@ -28,11 +28,14 @@
...
+#import "SalesforceSDKConstants.h"
 
 @interface SFSDKAuthConfigUtil : NSObject
 
 typedef void (^ _Nonnull MyDomainAuthConfigBlock)(SFOAuthOrgAuthConfiguration * _Nullable authConfig, NSError * _Nullable error);
 
-+ (void)getMyDomainAuthConfig:(nonnull MyDomainAuthConfigBlock)authConfigBlock oauthCredentials:(nonnull SFOAuthCredentials *)oauthCredentials;
++ (void)getMyDomainAuthConfig:(nonnull MyDomainAuthConfigBlock)authConfigBlock loginDomain:(nonnull NSString *)loginDomain;
+
++ (void)getMyDomainAuthConfig:(nonnull MyDomainAuthConfigBlock)authConfigBlock oauthCredentials:(nonnull SFOAuthCredentials *)oauthCredentials SFSDK_DEPRECATED(7.1, 8.0, "Use getMyDomainConfig:authConfigBlock loginDomain:loginDomain instead.");
```
bhariharan: Deprecating old API - we don't really need a `credentials` object.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKAuthConfigUtil.h
```diff
@@ -28,11 +28,14 @@
...
 #import "SFOAuthOrgAuthConfiguration.h"
 #import "SFOAuthCredentials.h"
+#import "SalesforceSDKConstants.h"
 
 @interface SFSDKAuthConfigUtil : NSObject
 
 typedef void (^ _Nonnull MyDomainAuthConfigBlock)(SFOAuthOrgAuthConfiguration * _Nullable authConfig, NSError * _Nullable error);
 
-+ (void)getMyDomainAuthConfig:(nonnull MyDomainAuthConfigBlock)authConfigBlock oauthCredentials:(nonnull SFOAuthCredentials *)oauthCredentials;
++ (void)getMyDomainAuthConfig:(nonnull MyDomainAuthConfigBlock)authConfigBlock loginDomain:(nonnull NSString *)loginDomain;
```
bhariharan: New API takes in `loginDomain`.

PR #2824, opened 2019-04-05
Add lifetime to properties to remove warnings

artcodex:
The missing lifetime properties in this pr cause warnings which when you turn on warning as errors cause errors in the code. 

Seems reasonable to have lifetime attributes here since the default assign doesn't make that much sense.

@trooper2013 @wmathurin


trooper2013:
@artcodex Which version of sdk andf xcode are you using? We don't see any warnings.

artcodex:
@trooper2013 xcode 10.1 and this is on 7.0 sdk

bhariharan:
@artcodex Try the latest from `dev` on `Xcode 10.2`. We don't see any warnings there.

PR #2825, opened 2019-04-05
Reading all bytes then converting to string when reading back file

wmathurin:
If you convert each buffer's worth of bytes to a string you might end up corrupting the string (because a multi bytes character could have been split at the buffer boundary.


sgoldberg-sfdc:
Makes perfect sense. Great work whomever found this.

PR #2827, opened 2019-04-05
Fixing weak access setting for iOS private Key

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecureEncryptionKey.m
```diff
@@ -77,7 +77,7 @@ - (instancetype) initWithLabel:(NSString*)label autoCreate:(BOOL)autoCreate
         // Does not exist and should be auto created
         BOOL useEnclave = [SFSDKCryptoUtils isSecureEnclaveAvailable];
         [SFSDKCryptoUtils createECKeyPairWithName:label
-                              accessibleAttribute:kSecAttrAccessibleAlways
+                              accessibleAttribute:kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
```
trooper2013: @wmathurin why not kSecAttrAccessibleWhenUnlockedThisDeviceOnly?

wmathurin: That would break background operations.

wmathurin: After first unlock, the app is not yet running. A notification might come in, as a result the push notification extension in the Salesforce App will not be able to show a preview => it should default to a generic message (we will validate when integrating).

PR #2846, opened 2019-05-02
Fix setting of userId

trooper2013:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountIdentity.m
```diff
@@ -48,7 +48,7 @@ - (id)initWithUserId:(NSString *)userId orgId:(NSString *)orgId
 {
     self = [super init];
     if (self) {
-        _userId = userId;
+        self.userId = userId;
         _orgId = orgId;
```
bhariharan: Should this be `self.orgId`?

trooper2013: no override for the setter should not matter but added for consistency anyway.

PR #2854, opened 2019-05-07
Fixed crash on loginPage, returning as NSNull object

ssraza:
@bhariharan @trooper2013

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.m
@@ -70,7 +70,7 @@ static NSString * const kAuthConfigLoginPageUrlKey     = @"LoginPageUrl";
 - (NSString *)loginPageUrl {
     NSString *loginPageUrl = nil;
     NSDictionary *loginPage = self.authConfigDict[kAuthConfigLoginPageKey];
-    if (loginPage) {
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.m
```diff
@@ -70,7 +70,7 @@ - (BOOL)useNativeBrowserForAuth {
 - (NSString *)loginPageUrl {
     NSString *loginPageUrl = nil;
     NSDictionary *loginPage = self.authConfigDict[kAuthConfigLoginPageKey];
-    if (loginPage) {
+    if (![loginPage isKindOfClass:[NSNull class]] && [loginPage count] > 0) {
         loginPageUrl = loginPage[kAuthConfigLoginPageUrlKey];
```
trooper2013: @ssraza We should guard the retrieval of kAuthConfigLoginPageUrlKey  too.

ssraza: why we want to `kAuthConfigLoginPageUrlKey` 🤔. `loginPageUrl` is a nullable property right? will it not gonna return `nit` if value is not there?

trooper2013: You are right that's fine.

PR #2858, opened 2019-05-09
loadConfigFromFile no longer logs an error if file doesn't exist

wmathurin:
It returns the error to the caller - caller that cares should log an error

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSDKStoreConfig.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSDKStoreConfig.m
@@ -44,7 +44,7 @@ static NSString *const kStoreConfigIndexes = @"indexes";
 - (nullable id)initWithResourceAtPath:(NSString *)path {
     self = [super init];
     if (self) {
-        NSDictionary *config = [SFSDKResourceUtils loadConfigFromFile:path];
+        NSDictionary *config = [SFSDKResourceUtils loadConfigFromFile:path error:nil];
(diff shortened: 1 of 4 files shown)


libs/SmartStore/SmartStore/Classes/SFSDKStoreConfig.m
```diff
@@ -44,7 +44,7 @@ @implementation SFSDKStoreConfig
 - (nullable id)initWithResourceAtPath:(NSString *)path {
     self = [super init];
     if (self) {
-        NSDictionary *config = [SFSDKResourceUtils loadConfigFromFile:path];
+        NSDictionary *config = [SFSDKResourceUtils loadConfigFromFile:path error:nil];
         self.soupsConfig = config == nil ? nil : config[kStoreConfigSoups];
```
wmathurin: Code does not always expect a config - so if no config found, we don't need to log any error.

wmathurin:
@bhariharan 
> Are we okay on semver? ;-)

=> Adressed with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2858/commits/6208b3f9a6bcdcb5d57b3b86379a06bcd4d03ce2

bhariharan:
Thanks, @wmathurin. Maybe we should mark the old one deprecated and remove it in `8.0`?

wmathurin:
@bhariharan Doing it here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2860

Issue #2884, opened 2019-06-01
Calling cleanResyncGhosts causes subseqent SyncAlreadyRunningError


TerryBrett:
1. Version of Mobile SDK Used: **7.1.0**
2. Issue found in Native App or Hybrid App: **Native**
3. OS Version: **iOS 12.2**
4. Device: **iPad**

5. Steps to reproduce:
- Call SyncManager.reSync(named: "MySyncDown")
- On completion, call SyncManager.cleanResyncGhosts(forName: "MySyncDown")
- On completion, call SyncManager.reSync(named: "MySyncDown") (***error**)

6. Actual behavior:
First reSync and cleanResyncGhosts succeeds.
Second reSync fails with error

7. Expected Behavior:
8. Error Log:
SFSmartSyncSyncManager SyncAlreadyRunningError: sync 1 is still running

**NOTE:**
It appears that the SFSyncTask init() method calls 
        [self.syncManager addToActiveSyncs:self];
and that in it's updateSync() method it calls
        [self.syncManager removeFromActiveSyncs:self];
which is the only two calls that change the active syncs list.

However, the SFCleanSyncGhostsTask overrides updateSync() and the removeFromActiveSyncs() call is never made.

wmathurin:
Thanks for reporting the issue. The fix is on this PR: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2887

PR #2886, opened 2019-06-02
Ensure task is removed from active syncs

TerryBrett:
Ensure the task has been removed from the active syncs list. See https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/2884

The task is added to the active syncs list in the init(), and usually removed in the updateSync() method, however the SFCleanSyncGhostsTask overrides updateSync() causing the task to be left in the list, and the subsequent attempt to do a sync down fails.


wmathurin:
Thanks for your submission. Using this https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2887 to fix the problem instead.

PR #2887, opened 2019-06-03
Clean resync ghost tasks was not removing itself from active syncs up…

wmathurin:
…on completion

Fixing https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/2884
Enhanced existing test to validate change


wmathurin:
NB: problem does not occur on Android thanks to https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/master/libs/SmartSync/src/com/salesforce/androidsdk/smartsync/manager/SyncTask.java#L97

PR #2889, opened 2019-06-05
 Added new Async API to switch to new user

trooper2013:
This PR adds the following.

1. A async api for switchToNewUser. Ensures that current user is not cleared and is set only after the login is completed. Errors out if called without a currentUser. 

2. Deprecated older Sync api. 

Note: Tests are WIP.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.h
@@ -495,6 +495,12 @@ NS_SWIFT_NAME(UserAccountManager)
 - (void)switchToNewUser NS_SWIFT_NAME(switchToNewUserAccount());
 
 /**
+ Switches to a new user. Sets the current user only if the login succeeds. Completion block is
+ invoked if the login flow completes, or if any errors are encountered during the flow.
+ */
+- (void)switchToNewUserWithCompletion:(void (^)(NSError *, SFUserAccount *))completion NS_SWIFT_NAME(switchToNewUserAccount(_:));
+
+/**
  Switches away from the current user, to the given user account.
  @param userAccount The user to switch to.
  */
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.h
@@ -492,7 +492,7 @@ NS_SWIFT_NAME(UserAccountManager)
 /**
  Switches away from the current user, to a new user context.
  */
-- (void)switchToNewUser NS_SWIFT_NAME(switchToNewUserAccount());
+- (void)switchToNewUser NS_SWIFT_NAME(switchToNewUserAccount()) SFSDK_DEPRECATED(7.2, 8.0, "Use switchToNewUserWithCompletion instead.");
 
 /**
  Switches to a new user. Sets the current user only if the login succeeds. Completion block is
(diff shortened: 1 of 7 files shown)


wmathurin:
We should probably merge this after 7.1.2 is released.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.h
```diff
@@ -492,7 +492,13 @@ NS_SWIFT_NAME(UserAccountManager)
 /**
  Switches away from the current user, to a new user context.
  */
-- (void)switchToNewUser NS_SWIFT_NAME(switchToNewUserAccount());
+- (void)switchToNewUser NS_SWIFT_NAME(switchToNewUserAccount()) SFSDK_DEPRECATED(7.2, 8.0, "Use switchToNewUserWithCompletion instead.");
```
wmathurin: NB: For Objective-C consumers, it will just be in warning if they use the deprecated API in 7.2, but for Swift consumers, it's an API change.
(We didn't similar things with some of the 7.1 SmartSync changes).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1483,6 +1485,28 @@ - (void)switchToNewUser {
     [self switchToUser:nil];
 }
 
+- (void)switchToNewUserWithCompletion:(void (^)(NSError *error, SFUserAccount * currentAccount))completion {
+    if (!self.currentUser) {
```
bhariharan: Not sure I understand this. Why can't we switch to a new user without `currentUser` being set? This is basically the deferred authentication scenario, where there's a playground mode, followed by an API call to this API. At this point, `self.currentUser` will be `nil`.

trooper2013: @bhariharan  The check is to ensure that there is a logged in user at the least i.e. the switch is occurring from the context of  a logged in user.  Why would this call be made otherwise vs. using loginWithCompletion.

bhariharan: Well, I was thinking they'd be the same. Basically deprecate `loginWithCompletion` in this class and rename it to `switchToNewUser`. That makes the APIs sound the same - `switchToUser` and `switchToNewUser`. Also, `switchToUser` would fall back on `switchToNewUser` if no user is passed in. Makes sense?

trooper2013: @bhariharan I don't think we should. To start with login api has the advantage of being concise in what the api means and does, as opposed to "switch to a new User". Secondly the login api **does not set the current user** or technically speaking  "switch to the user".  A developer could set the current user to any of the logged in users. Third reason is that the login api and logout together symmetry.

We could rename the api to loginAndSwitchUser instead? A more radical approach would be to consider the addition of a  user to the UserAccountManager as a login? So then the developer would always add a user  account ->  set the current user or switch user. Perhaps we should disallow the setting the currentUser and resort to switchUser at all times.  We should chat about it offline.

bhariharan: @trooper2013 The whole point of this API is to avoid having the app set `currentUser`, or expecting the app developer to handle that case. Basically, the end result of `loginWithCompletion` and `switchToNewUser` should be the same - we should set `currentUser` to that newly logged in user automatically, and then pass along control to the app by calling their completion blocks. I can't think of many use cases where an app would want to login with a new user and then not switch to that user. Login should automatically land the user on that newly logged in user's state.

I'm okay with keeping both `loginWithCompletion` and `switchToNewUser` the way they are - let's make sure that all the methods make setting of `currentUser` internal to the SDK.

PR #2893, opened 2019-06-11
Adds a NSNotification for Refresh flow

trooper2013:
Fires kSFNotificationUserDidRefreshToken when the refresh flow completes successfully.

e.g.  usage
``` 
NotificationCenter.default.addObserver(self, selector:  #selector(refreshFlowCompleted(_:)), name: UserAccountManager.didRefreshToken, object: nil)

@objc func refreshFlowCompleted(_ notification: Notification) {
         if let account = notification.userInfo?[UserAccountManager.userInfoAccountKey] as?  UserAccount {
           ... 
           print("Refresh Flow Completed %@",account)
        }
    }

```


wmathurin:
We are cutting 7.1.2 tomorrow - so we could keep it in dev.

bhariharan:
@wmathurin Could we get this in `7.1.2`? We need this for parity with Android on the `SecuritySDK`. I didn't realize it was missing. I assume `7.1.2` will go from `master`?

PR #2904, opened 2019-06-14
Name for sync is in the 'name' property

wmathurin:


libs/SmartSync/SmartSync/Classes/Util/SFSyncState.h
```diff
@@ -89,7 +89,7 @@ NS_SWIFT_NAME(SyncState)
 @interface SFSyncState : NSObject <NSCopying>
 
 @property (nonatomic, readonly) NSInteger syncId;
-@property (nonatomic, strong, readonly) NSString* syncName;
+@property (nonatomic, strong, readonly) NSString* name;
```
wmathurin: In the .m we have @property (nonatomic, strong, readwrite) NSString* name - so syncName is never set to anything.

PR #2907, opened 2019-06-18
Set's the default presentation mode to full screen for SDK's Auth Controllers

trooper2013:
In iOS 13 the default presentation mode mode for view controllers has changed and  is not  full screen, due to which view controllers can be swiped to be removed from the screen or moved. This fix work is backward compatible with iOS 12 and prior.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
@@ -705,7 +705,7 @@ static NSString * const kSFMobileSDKNativeSwiftDesignator = @"NativeSwift";
     // If we make the snapshot window the active window now, that's where the SFAuthenticationSession's view controller will end up
     // Then when the application is foregrounded and the snapshot window is dismissed, we will lose the SFAuthenticationSession
     SFSDKWindowContainer* activeWindow = [SFSDKWindowManager sharedManager].activeWindow;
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -705,7 +705,7 @@ - (void)handleAppWillResignActive:(NSNotification *)notification
     // If we make the snapshot window the active window now, that's where the SFAuthenticationSession's view controller will end up
     // Then when the application is foregrounded and the snapshot window is dismissed, we will lose the SFAuthenticationSession
     SFSDKWindowContainer* activeWindow = [SFSDKWindowManager sharedManager].activeWindow;
-    if (([activeWindow isAuthWindow]  && ![activeWindow.topViewController isKindOfClass:[SFLoginViewController class]]) ||  [activeWindow isPasscodeWindow]) {
```
wmathurin: Why drop the `&& ![activeWindow.topViewController isKindOfClass:[SFLoginViewController class]]`?

It came with:
https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/af1152d8d387191fa07c2891f0a3762d85e918d9

trooper2013: @wmathurin was noticing some problems with snapshots over the authwindow. Removing this check disallows the the snapshot  from being rendered over auth window.

wmathurin: That check was to only disabling snapshot over SFAuthenticationSession - so you are saying it's also causing issues with regular login.

PR #2909, opened 2019-06-19
Caching in SmartStore now uses NSCache with a count limit

wmathurin:
For SmartSql -> SQL, we use a new class SFSmartSqlCache (which internally also uses NSCache)


libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -205,11 +209,16 @@ - (id) initWithName:(NSString*)name user:(SFUserAccount *)user isGlobal:(BOOL)is
...
+        _soupNameToTableName.countLimit = CACHES_COUNT_LIMIT;
+        
+        _attrSpecBySoup = [[NSCache alloc] init];
+        _attrSpecBySoup.countLimit = CACHES_COUNT_LIMIT;
         
-        _smartSqlToSql = [[NSMutableDictionary alloc] init];
+        _indexSpecsBySoup = [[NSCache alloc] init];
+        _indexSpecsBySoup.countLimit = CACHES_COUNT_LIMIT;
+        
+        _smartSqlToSql = [[SFSmartSqlCache alloc] initWithCountLimit:CACHES_COUNT_LIMIT];
```
trooper2013: @wmathurin shouldn't the number of objects  held in cache be SFSmartSqlCache  be higher as compared to the other caches?

trooper2013: @wmathurin Should we provide a way to clear the caches? What would happen if the soup features changed?

wmathurin: I don't want to make clearing the caches a public api.
I need to check the code to see if altering a soup causes the cache to be cleared or not (it might not).

wmathurin: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2909/commits/2d55ca969ded6e5948c14ac9cfa585134c543f45 takes care of clearing caches when soup is changed.

PR #2912, opened 2019-06-19
Prevent AILTN publisher background task from ending prematurely.

brandonpage:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -243,6 +245,7 @@ - (void) publishEvents:(NSArray<SFSDKInstrumentationEvent *> *) events {
                 }
                 publishCompleteBlock = nil;
             }
+            [self cleanupBackgroundTask];
```
trooper2013: @brandonpage Shouldn't this  clean up occur in the else block above it?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -313,14 +316,10 @@ - (void) publishOnAppBackground {
         return;
     }
     dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        __block UIBackgroundTaskIdentifier task;
         task = [[SFApplicationHelper sharedApplication] beginBackgroundTaskWithName:NSStringFromClass([self class]) expirationHandler:^{
-            [[SFApplicationHelper sharedApplication] endBackgroundTask:task];
-            task = UIBackgroundTaskInvalid;
+            [self cleanupBackgroundTask];
```
trooper2013: Is the retain intentional?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
brandonpage: Don't we want to end the background task regardless of success/failure? If the operation fails the events aren't deleted and will be sent (with new ones) the next time the app is backgrounded.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
trooper2013: @brandonpage on a second thought this is ok 👍

PR #2919, opened 2019-06-25
Fix not calling the completion block on the metadata fetch failure

dtrofimov:
The completion block of `SFMetadataSyncManager` `-fetchFromServer:completionBlock:` was never called if the request failed.


wmathurin:
Thanks for your contribution.

PR #2925, opened 2019-07-02
Adding pragma to upgrade db to sqlcipher 4.x

wmathurin:
Now using 4.x settings
Added method + test to check runtime settings (also shown in dev menu)
Tested upgrade with GlobalStoreTester:
- tested 7.0 to 7.2 upgrade: goes from sqlcipher 3.4.2 to 4.2
- tested 7.1 to 7.2 upgrade: goes from sqlcipher 4.0 to 4.2
- tested dowgrades also - they fail as expected  (7.2 -> 7.1 because of db, 7.2 -> 7.0 because of encryption keys)

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
@@ -199,15 +199,14 @@ static NSString * const kSFSmartStoreVerifyReadDbErrorDesc = @"Could not read fr
 
 + (FMDatabase*) unlockDatabase:(FMDatabase*)db key:(NSString*)key salt:(NSString *)salt {
     if ([db open]) {
-        // Using sqlcipher 3.x default settings
-        // => should open 3.x databases without any migration
-        [[db executeQuery:@"PRAGMA cipher_default_compatibility = 3"] close];
         // Using sqlcipher 2.x kdf iter because 3.x default (64000) and 4.x default (256000) are too slow
-        // => should open 2.x databases without any migration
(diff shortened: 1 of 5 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
```diff
@@ -199,15 +199,14 @@ + (FMDatabase*) setKeyForDb:(FMDatabase*) db key:(NSString *)key salt:(NSString
...
-        [[db executeQuery:@"PRAGMA cipher_default_compatibility = 3"] close];
         // Using sqlcipher 2.x kdf iter because 3.x default (64000) and 4.x default (256000) are too slow
-        // => should open 2.x databases without any migration
         [[db executeQuery:@"PRAGMA cipher_default_kdf_iter = 4000"] close];
        
         if (key)
            [db setKey:key];
+
+        // Migrating and upgrading an existing database in place (preserving data and schema) if necessary
+        [[db executeQuery:@"PRAGMA cipher_migrate"] close];
```
trooper2013: @wmathurin Has this to be run everytime a DB is opened?

wmathurin: To avoid running it for newly created database, I would have to change the signature of a bunch of methods to indicate the db was just created.
To avoid running for an existing database that has the right settings, I would have to check the settings first etc.

Actually looking at SQLCipher sources, it does nothing if it is already the right version. See [here](https://github.com/sqlcipher/sqlcipher/blob/v4.2.0/src/crypto_impl.c#L1353).

PR #2927, opened 2019-07-03
Update appDisplayName property

brandonpage:
Instead of creating a new property I decided to update the existing `appDisplayName` property so it can be used as the SDK's source of truth name for the app for any UI.  Currently it is only used for IDP and Biometric, but this should be the property used for any UI (or external facing messages) going forward.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.m
@@ -217,7 +217,18 @@ NSString * const kOAuthAppName = @"oauth_app_name";
 - (NSString *)appDisplayName
 {
     NSUserDefaults *defs = [NSUserDefaults msdkUserDefaults];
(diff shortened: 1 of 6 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.m
```diff
@@ -217,7 +217,18 @@ - (void)setIsIdentityProvider:(BOOL)isIdentityProvider
 - (NSString *)appDisplayName
 {
     NSUserDefaults *defs = [NSUserDefaults msdkUserDefaults];
-    return [defs stringForKey:kOAuthAppName];
+    NSString *appName = [defs stringForKey:kOAuthAppName];
+    
+    if (appName) {
+        return appName;
+    } else {
+        NSString *bundleDispalyName = [[NSBundle mainBundle] objectForInfoDictionaryKey:@"CFBundleDisplayName"];
```
bhariharan: @brandonpage Typo on the variable - should be `bundleDisplayName` (it is `bundleDispalyName`).

Issue #2853, opened 2019-05-07
'Use Custom Domain' screen crash when invalid domain entered


rilchyshyn:
We are receiving the same crash report from various clients (see [crashlytics-report.txt](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/files/3153027/crashlytics-report.txt)).

To reproduce it, just try to enter _wrong_ custom domain value (in 'Use Custom Domain' screen) **twice** (see [how-to-reproduce](https://user-images.githubusercontent.com/48205812/57307210-d2a2fb00-70ec-11e9-830b-b1672a1500e4.gif)).

The issue caused on calling 
`- [SFUserAccountManager notifyUserCancelledOrDismissedAuth:andAuthInfo:]` 
with `info` parameter equals nil (see [try1](https://user-images.githubusercontent.com/48205812/57375858-338f0980-71a7-11e9-82a8-6051183c3717.png), [try2](https://user-images.githubusercontent.com/48205812/57375883-46094300-71a7-11e9-9772-28164b677433.png)).

We use pod `'SalesforceSDKCore','7.0.0'` from source https://github.com/forcedotcom/SalesforceMobileSDK-iOS-Specs.git. But it also reproducible on version 6.2.0.

trooper2013:
@rilchyshyn We will take a look. Meanwhile you can customize the handling of this situation by doing the following in your app delegate. 

```
UserAccountManager.shared.add(self)

extension AppDelegate : UserAccountManagerDelegate {
    func userAccountManager(accountManager userAccountManager: UserAccountManager, didFailAuthenticationWith error: Error, info: AuthInfo) -> Bool {
        // return true if error is handled. if false the SDK will use a default handler.
    }
}
```

rilchyshyn:
Thanks for the response! 
We will think about the pros and cons of implementing custom behavior instead of wait for the bug fix.
Anyway, this workaround could prevent the described crash and is very helpful.

wmathurin:
Not able to reproduce the issue.
![BadDomain](https://user-images.githubusercontent.com/1012459/59235065-77a96b80-8ba4-11e9-86f5-68aa954149c9.gif)

wmathurin:
How do you start the login flow?
Do you have the following in your `AppDelegate.swift`?
```swift
[...]

PR #2928, opened 2019-07-03
Add nil check for authInfo

bbirman:
For #2853

After the first invalid attempt auth info is set to nil and then would refresh but the second invalid attempt is happening before the refresh so auth info is still nil causing the crash. I initially looked into catching it earlier in the cycle but thought it would be more consistent to let it go through and not include the auth info if it didn't exist, let me know if another point would be better

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
@@ -1589,11 +1589,16 @@ static int const kSFSDKUserAccountManagerErrorCode = 100;
 
 - (void)notifyUserCancelledOrDismissedAuth:(SFOAuthCredentials *)credentials andAuthInfo:(SFOAuthInfo *)info
  {
-    NSDictionary *userInfo = @{ kSFNotificationUserInfoCredentialsKey:credentials,
-                                kSFNotificationUserInfoAuthTypeKey: info };
+    NSMutableDictionary *userInfo = [NSMutableDictionary new];
+    userInfo[kSFNotificationUserInfoCredentialsKey] = credentials;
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1589,11 +1589,16 @@ - (void)notifyUserChange:(NSString *)notificationName withUser:(SFUserAccount *)
 
 - (void)notifyUserCancelledOrDismissedAuth:(SFOAuthCredentials *)credentials andAuthInfo:(SFOAuthInfo *)info
  {
-    NSDictionary *userInfo = @{ kSFNotificationUserInfoCredentialsKey:credentials,
-                                kSFNotificationUserInfoAuthTypeKey: info };
+    NSMutableDictionary *userInfo = [NSMutableDictionary new];
+    userInfo[kSFNotificationUserInfoCredentialsKey] = credentials;
+    if (info) {
```
wmathurin: But why is the info null on the second cancel?

bbirman: It's set to nil [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/c7c22f1d883f036a4beb81173b0347f7c4ae2b45/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m#L361) after the first cancel call and then would be populated again after the refresh but the second cancel can happen before the refresh finishes

wmathurin: 👍

PR #2936, opened 2019-07-16
Merge pull request #831 from rpatnayakuni/jsonWork

wmathurin:
W-6310157 - Partial JSon validation work into 7.1.2


wmathurin:
Code does not deserialize json during queries. However externally stored soups could have cut-off entries if app is killed during a write. Querying from such a soup in an hybrid app could cause top level error in hybrid app (callback will never be invoked because the javascript injected by cordova in the webview will have a syntax error).
NB: The check is not the long term solution.

wmathurin:
Merging - this is a cherry pick already reviewed multi times on the fork.

PR #2937, opened 2019-07-16
saveSoupEntryExternally now writes to a tmp file then rename the file…

wmathurin:
… to make write closer to atomic operations

Also added debug logging to the method


wmathurin:
PR moved here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2938

PR #2938, opened 2019-07-17
saveSoupEntryExternally now writes to a tmp file then rename the file…

wmathurin:
… to make write closer to atomic operations

Also added debug logging to the method

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
@@ -828,14 +828,14 @@ NSUInteger CACHES_COUNT_LIMIT = 1024;
                     soupEntryId:(NSNumber *)soupEntryId
                   soupTableName:(NSString *)soupTableName {
     
-    // Helper to log debug messages
-    void (^logDebug)(NSString* step) = ^(NSString* step) {
-        NSString *debugMessage = [NSString stringWithFormat:@"%@ (%@): soupEntryId: %@, soupTableName: %@",
-                                  NSStringFromSelector(_cmd),
-                                  step,
-                                  soupEntryId,
-                                  soupTableName];
-        [SFSDKSmartStoreLogger d:[self class] format:debugMessage];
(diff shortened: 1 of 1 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -827,41 +827,82 @@ - (BOOL)createExternalStorageDirectory:(NSString *)soupTableName {
...
                   soupTableName:(NSString *)soupTableName {
+    
+    // Helper to log debug messages
+    void (^logDebug)(NSString* step) = ^(NSString* step) {
+        NSString *debugMessage = [NSString stringWithFormat:@"%@ (%@): soupEntryId: %@, soupTableName: %@",
+                                  NSStringFromSelector(_cmd),
+                                  step,
+                                  soupEntryId,
+                                  soupTableName];
+        [SFSDKSmartStoreLogger d:[self class] format:debugMessage];
```
radhap: do we want this to be info, so we can get this info in release builds as well? rather than just debug

libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -827,41 +827,82 @@ - (BOOL)createExternalStorageDirectory:(NSString *)soupTableName {
...
+        if (success) {
+            success = [[NSFileManager defaultManager] moveItemAtPath:tmpFilePath toPath:filePath error:&error];
+        }
+    }
+    logDebug(@"4/4 Done renaming tmp file");
+    
+    if (success) {
+        return YES;
+    } else {
+        NSString *errorMessage = [NSString stringWithFormat:@"Saving external soup to file failed! encrypted: %@, soupEntryId: %@, soupTableName: %@, tmpFilePath: '%@', filePath: '%@', error: %@.",
```
khawkins: Minor: can we just simplify this to `if (!success)`, and return `success` at the end?

PR #2951, opened 2019-07-25
Added SFSDKOAuth2 class for token endpoint calls.

trooper2013:
This is the first PR  for rewiring the refresh flow. It does the following. 

1.  Added a SFSDKOAuth2 class that has the ability to make endpoint api calls. Wraps request & response in appropriate classes.  

-  The response object embodies the response NVPairs or the error.  
-  The SFSDKOAuth2 is an implementation of  SFSDKOAuth protocol &  SFSDKOAuthSessionManaging. Allows for extension of the client and managing off the URLSession that is used by the class.

2. Moves some code  & constants from the SFOAuthCordinator to the SFSDKOAuth2 
& SFSDKAuthConstants. 

3. Changes a few classes inorder to now refer to the methods & constants in the  SFSDKOAuth2 class.

e.g. usage 
```
        let oauthClient = SFSDKOAuth2()
        oauthClient.accessToken(forRefresh: request) { (tokenResponse) in
             if tokenResonse.hasError {
                    // tokenResponse.error should have the details of error 
                   // handle error return
             }
             // response should have all the values from endpoint.
        }

```

NOTE: The next follow up PR will rewire the components to use the SFSDKOAuth2 class. I have currently added tests for token refresh. I have not tried the code flow.  I will test that functionality in the follow up PR.  I think we can also move other methods to this class e.g. token revocation & other util methods.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
```diff
@@ -0,0 +1,116 @@
...
+@property (nonatomic, copy, nullable) NSString *clientID;
+@property (nonatomic, copy, nullable) NSString *approvalCode;
+@property (nonatomic, copy, nullable) NSString *codeVerifier;
+@property (nonatomic, assign) NSTimeInterval timeout;
+@property (nonatomic, strong, nullable) NSURL *serverURL;
+@property (nonatomic, strong, nullable) NSDictionary * additionalTokenRefreshParams;
+@property (nonatomic, strong, nullable) NSArray<NSString *> *additionalOAuthParameterKeys;
+@end
+
+@interface SFSDKOAuthTokenEndpointResponse : NSObject
```
bhariharan: Is this a redefined interface from the one above? Can we combine them, since most properties are `nullable` anyway?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
```diff
@@ -0,0 +1,116 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SFSDKOAuthTokenEndpointErrorResponse : NSObject
+@property  (nonatomic, readonly) NSString *tokenEndpointErrorCode;
+@property  (nonatomic, readonly) NSString *tokenEndpointErrorDescription;
+@property  (nonatomic, readonly) NSError *error;
+@end
+
+
+@interface SFSDKOAuthTokenEndpointRequest : NSObject
```
bhariharan: I think some of these properties should be `nonnull`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
```diff
@@ -0,0 +1,416 @@
...
+- (NSDictionary *)asDictionary {
+    return self.values;
+}
+@end
+
+@implementation SFSDKOAuth2
+
+- (void)accessTokenForApprovalCode:(SFSDKOAuthTokenEndpointRequest *)endpointReq completion:(void (^)(SFSDKOAuthTokenEndpointResponse *)) completionBlock {
+    
+    NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:endpointReq.serverURL cachePolicy:NSURLRequestReloadIgnoringCacheData timeoutInterval:endpointReq.timeout];
```
bhariharan: Should we use `SFNetwork` here? Or does it have other implications such as auto-refresh?

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKAuthUtilTests.swift
```diff
@@ -0,0 +1,142 @@
...
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import XCTest
+import Foundation
+@testable import SalesforceSDKCore
+
+class SFSDKAuthUtilTests: XCTestCase {
```
bhariharan: First `Swift` test. Yay!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
trooper2013: @bhariharan I need to refine the nullability for  the  properties.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
bhariharan: My bad - I didn't read the `Request` and `Response` difference ;-) Ignore this comment.

PR #2961, opened 2019-08-03
Adding the ability to set API versions at the REST request layer

bhariharan:
Currently, a consuming app can only set the API version at the REST API layer, which sets it for all requests coming through. API version should be set at the REST request layer - this gives a consuming app the ability to specify which version to use in a more granular manner. This PR is a first stab at allowing for this. This feature already exists on `Android`.

Outstanding Work:

* Replace deprecated method usage in `SmartSync` and test targets with the newer factory methods (will be in the next PR).
* Get rid of deprecated methods (will be in `Mobile SDK 8.0` - @bbirman this one's for you).


sgoldberg-sfdc:
👍 Thanks for this Bharath

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -3250,6 +3250,7 @@
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				SDKROOT = iphoneos;
 				SUPPORTED_PLATFORMS = "iphonesimulator iphoneos";
+				SWIFT_COMPILATION_MODE = wholemodule;
```
bbirman: What does this do?

bhariharan: Good question - that was the result of clicking `Update to Xcode recommended settings` in `Xcode 10.3`.

trooper2013:
@bhariharan Weren't you going to make the api version Nullable?  i.e. the default will be used if nil.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Files.h
```diff
@@ -37,99 +37,209 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+/**
+ * Build a request that will delete the specified file share.
+ *
+ * @param shareId The Id of the file share record (aka ContentDocumentLink).
+ * @param apiVersion API version.
+ * @return A new SFRestRequest that can be used to delete this share.
  */
-- (SFRestRequest *) requestForDeleteFileShare:(NSString *)shareId;
+- (SFRestRequest *)requestForDeleteFileShare:(NSString *)shareId apiVersion:(NSString *)apiVersion;
```
trooper2013: Should we make the version Nullable in all calls?

bhariharan:
@trooper2013 I'll do that in the next PR. I'm planning on moving the constants to a class and plugging them in.

bhariharan:
@trooper2013 Merging this for now - will add more in subsequent PRs, so that it's easier to review.

PR #2963, opened 2019-08-05
Adding fallback and nullability markers for API version

bhariharan:
This PR:
- Adds nullability markers for `apiVersion`.
- Falls back on the default API version, if `apiVersion` is not supplied.


bhariharan:
@trooper2013 Here's the fallback to default API version.

Issue #2964, opened 2019-08-06
Resync fails if using 12-hour clock


jlcvp:
Please fill out the following details:

1. Version of Mobile SDK Used: tested from 6.1 to 7.0
2. Issue found in Native App or Hybrid App: Hybrid - ios build. 
3. OS Version: MacOS Mojave. 
4. Device: iPad Pro 2018. 
5. Steps to reproduce: Run a resync with iOS clock format set to 12-hour, language English(UK) and region Ireland.
6. Actual behavior: resync fails because the where clause of `LastModifiedDate > ...` is malformed because the hour part of datetime is only 1 digit-long, despite of the format being set to `yyyy-MM-dd'T'HH:mm:ss.SSSZ` by `SFSmartSyncObjectUtils` constructor. [SFSmartSyncObjectUtils.m#L35](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/9c51f4c89424878ed1acb6405f9d5e2e8784f5fa/libs/SmartSync/SmartSync/Classes/Util/SFSmartSyncObjectUtils.m#L35)
7. Expected Behavior: resync occurs normally
8. Error Log: 
```
Object Name: PriceBookEntry \nWhere LastModifiedDate > 2019-07-16T9:35:21.000-0300 \n^\nERROR at Row:1:Column:268\nline 1:268 no viable alternative at character '9'
```

wmathurin:
Fixed with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2965 and https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2971

PR #2965, opened 2019-08-06
Enforcing sync isoformatter locale for date to be ISO8601 compliant

jlcvp:
This commit addresses forcedotcom/SalesforceMobileSDK-iOS#2964

### DIFF
--- a/libs/SmartSync/SmartSync/Classes/Util/SFSmartSyncObjectUtils.m
+++ b/libs/SmartSync/SmartSync/Classes/Util/SFSmartSyncObjectUtils.m
@@ -32,6 +32,8 @@ __strong static NSDateFormatter *isoDateFormatter;
 + (void) initialize {
     utcDateFormatter = [NSDateFormatter new];
     isoDateFormatter = [NSDateFormatter new];
+    NSLocale *enUSPOSIXLocale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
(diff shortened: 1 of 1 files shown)


libs/SmartSync/SmartSync/Classes/Util/SFSmartSyncObjectUtils.m
```diff
@@ -32,6 +32,8 @@ @implementation SFSmartSyncObjectUtils
 + (void) initialize {
     utcDateFormatter = [NSDateFormatter new];
     isoDateFormatter = [NSDateFormatter new];
+    NSLocale *enUSPOSIXLocale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
+    [isoDateFormatter setLocale:enUSPOSIXLocale];
```
wmathurin: I would have thought that setting the date format would be sufficient but you are right, we also need to set the locale. Also you need to set the time zone to UTC. Could you add:
```objective-c
[isoDateFormatter setTimeZone:[NSTimeZone timeZoneForSecondsFromGMT:0]];
```
For more info see: https://developer.apple.com/library/archive/qa/qa1480/_index.html


For more information see: https://developer.apple.com/library/archive/qa/qa1480/_index.html
Could you also

wmathurin: I will do the `setTimeZone` in a separate PR. Thanks for your contribution.

PR #2966, opened 2019-08-07
Refactors components to use SFSDKOAuth2 Refresh call

trooper2013:
Part 2 of the refactoring the refresh flow. All refresh calls are now made through the SFSDKOAuth2 class. These calls are simply token endpoint calls. These calls should be lighter as compared to the prior method of spinning up the coordinators.


libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKAuthUtilTests.swift
```diff
@@ -73,13 +73,14 @@ class SFSDKAuthUtilTests: XCTestCase {
...
-        XCTAssertNotNil(endpointResponse!.accessToken)
-        XCTAssertNil(endpointResponse!.refreshToken)
-        XCTAssertNotNil(endpointResponse!.scopes)
-        XCTAssertNotNil(endpointResponse!.instanceUrl)
-        XCTAssertNotNil(endpointResponse!.signature)
-        XCTAssertNotNil(endpointResponse!.issuedAt)
+        let response = try! require(endpointResponse)
+        XCTAssertFalse(response.hasError)
+        XCTAssertNotNil(response.accessToken)
+        XCTAssertTrue(response.refreshToken==nil || response.refreshToken!.count < 1)
```
brandonpage: 👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -312,5 +313,53 @@ - (BOOL)hasPropertyValueChangedForKey:(NSString *) key {
     return [_credentialsChangeSet objectForKey:key]!=nil;
 }
 
+- (NSURL *)overrideDomainIfNeeded{
+    
+    NSString *refreshDomain = self.communityId ? self.communityUrl.absoluteString : self.domain;
```
bhariharan: This seems incorrect. Refresh domain should always go against the configured login host, i.e. login URL. This is because the token endpoint is a part of the login pool and is agnostic of instance URL and/or community URL. This is especially important when pods are taken down and/or migrated to a different instance server.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/TestSetupUtils.m
```diff
@@ -108,11 +108,13 @@ + (void)synchronousAuthRefresh
      completion:^(SFOAuthInfo *authInfo, SFUserAccount *userAccount) {
          authListener.returnStatus = kTestRequestStatusDidLoad;
          user = userAccount;
+         
      } failure:^(SFOAuthInfo *authInfo, NSError *error) {
          authListener.lastError = error;
          authListener.returnStatus = kTestRequestStatusDidFail;
      }];
     [authListener waitForCompletion];
+    [SFUserAccountManager sharedInstance].currentUser = user;
```
bhariharan: Remove this. It's deprecated.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
```diff
@@ -87,7 +87,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) BOOL hasError;
 @property (nonatomic, readonly, nullable) SFSDKOAuthTokenEndpointErrorResponse *error;
 @property (nonatomic, readonly) NSString *accessToken;
-@property (nonatomic, readonly) NSString *refreshToken;
+@property (nonatomic, readonly, nullable) NSString *refreshToken;
```
bhariharan: Refresh token should be `nonnull`.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKAuthUtilTests.swift
bhariharan: Fix spacing here before and after `==`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
trooper2013: @bhariharan The code has been remained the same since 5.0.  https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/9024d4ccc40b770e1d8cdb1fddd9b489b30b6392/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m#L647

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
trooper2013: @bhariharan Im reusing the token endpoint call.  The refreshToken is not  returned during the refresh flow so it has to be optional right?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
bhariharan: Right - we can take a look separately after this PR.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
bhariharan: Right - but it would still have the old value, right? If an app retrieves this `property`, it should always have a value, right?

trooper2013: @bhariharan Sounds good to me.

PR #2971, opened 2019-08-15
ISO8601 compliance (part 2)

wmathurin:
When working with fixed format dates, such as RFC 3339, you set the dateFormat property to specify a format string.
For most fixed formats, you should also set the locale property to a POSIX locale ("en_US_POSIX"), and set the timeZone property to UTC.
Source: https://developer.apple.com/documentation/foundation/nsdateformatter

This PR is part 2 of https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2965


wmathurin:
Note: all SmartSync tests passing locally

az-oolloow:
@wmathurin which release will this be part of?

PR #2975, opened 2019-08-21
Ability to share URL sessions

bbirman:
The org hit its request limit, so I still need to rerun the tests when it resets

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
@@ -30,15 +30,15 @@
 #import <Foundation/Foundation.h>
 #import "SalesforceSDKConstants.h"
 
+SFSDK_DEPRECATED(7.3, 8.0, "Will be removed")
 NS_SWIFT_NAME(NetworkManaging)
 @protocol SFNetworkSessionManaging
-- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sessionWithIdentifier:sessionConfiguration instead");
-- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sessionWithIdentifier:sessionConfiguration instead");
-- (nonnull NSURLSession *)sessionWithIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nonnull NSURLSessionConfiguration *)configuration;
+- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sharedInstanceWithIdentifier:sessionConfiguration instead");
+- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sharedInstanceWithIdentifier:sessionConfiguration instead");
 @end
 
-extern NSString * __nonnull const kSFNetworkEphemeralSessionIdentifier NS_SWIFT_NAME(SFNetworkEphemeralSessionIdentifier);
-extern NSString * __nonnull const kSFNetworkBackgroundSessionIdentifier NS_SWIFT_NAME(SFNetworkBackgroundSessionIdentifier);
+extern NSString * __nonnull const kSFNetworkEphemeralInstanceIdentifier NS_SWIFT_NAME(SFNetworkEphemeralInstanceIdentifier);
+extern NSString * __nonnull const kSFNetworkBackgroundInstanceIdentifier NS_SWIFT_NAME(SFNetworkBackgroundInstanceIdentifier);
 
 NS_SWIFT_NAME(Network)
 @interface SFNetwork : NSObject
@@ -46,44 +46,57 @@ NS_SWIFT_NAME(Network)
 typedef void (^SFDataResponseBlock) (NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) NS_SWIFT_NAME(DataResponseBlock);
 
 @property (nonatomic, readonly, strong, nonnull) NSURLSession *activeSession;
-@property (class, readonly, nonnull) NSDictionary *sharedSessions;
 
 /**
  * Returns an instance of this class with the default ephemeral session configuration.
  *
  * @return Instance of this class.
  */
-+ (nonnull instancetype)defaultEphemeralNetwork;
++ (nonnull instancetype)sharedEphemeralInstance;
 
 /**
  * Returns an instance of this class with the default background session configuration.
  *
  * @return Instance of this class.
  */
-+ (nonnull instancetype)defaultBackgroundNetwork;
++ (nonnull instancetype)sharedBackgroundInstance;
+
+/**
+ * Returns instance of this class for the given identifier with the default ephemeral session configuration.
+ *
+ * @return Instance of this class.
+ */
++ (nonnull instancetype)sharedEphemeralInstanceWithIdentifier:(nonnull NSString *)identifier;
+
+/**
+ * Returns instance of this class for the given identifier with the default background session configuration.
+ *
+ * @return Instance of this class.
+ */
++ (nonnull instancetype)sharedBackgroundInstanceWithIdentifier:(nonnull NSString *)identifier;
 
 /**
  * Returns an instance of this class with the given session configuration.
  *
- * @param identifier Identifier for the session.
- * @param sessionConfiguration  Configuration to use for the session. Defaults to the ephemeral configuration.
+ * @param identifier Identifier for the instance
+ * @param sessionConfiguration Configuration to use for the session.
  * @return Instance of this class.
  */
-+ (nonnull instancetype)networkWithSessionIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nullable NSURLSessionConfiguration *)sessionConfiguration;
++ (nonnull instancetype)sharedInstanceWithIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nonnull NSURLSessionConfiguration *)sessionConfiguration;
 
 /**
  * Initializes this class with an ephemeral session configuration.
  *
  * @return Instance of this class.
  */
-- (nonnull instancetype)initWithEphemeralSession SFSDK_DEPRECATED(7.3, 8.0, "Use defaultEphemeralNetwork instead");
+- (nonnull instancetype)initWithEphemeralSession SFSDK_DEPRECATED(7.3, 8.0, "Use sharedEphemeralInstance instead");
 
 /**
  * Initializes this class with a background session configuration.
  *
  * @return Instance of this class.
  */
-- (nonnull instancetype)initWithBackgroundSession SFSDK_DEPRECATED(7.3, 8.0, "Use defaultBackgroundNetwork instead");
+- (nonnull instancetype)initWithBackgroundSession SFSDK_DEPRECATED(7.3, 8.0, "Use sharedBackgroundInstance instead");
 
 /**
  * Sends a REST request and calls the appropriate completion block.
@@ -105,7 +118,7 @@ typedef void (^SFDataResponseBlock) (NSData * _Nullable data, NSURLResponse * _N
  * Sets a session configuration to be used for network requests in Mobile SDK.
  *
  * @param sessionConfig Session configuration to be used.
- * @param identifier Identifier for the session to use this config.
+ * @param identifier Identifier for the instance to use this config.
  */
 + (void)setSessionConfiguration:(nonnull NSURLSessionConfiguration *)sessionConfig identifier:(nonnull NSString *)identifier;
 
@@ -114,34 +127,40 @@ typedef void (^SFDataResponseBlock) (NSData * _Nullable data, NSURLResponse * _N
  *
  * @param manager Object that implements the SFNetworkSessionManaging protocol.
  */
-+ (void)setSessionManager:(nonnull id<SFNetworkSessionManaging>)manager;
++ (void)setSessionManager:(nonnull id<SFNetworkSessionManaging>)manager SFSDK_DEPRECATED(7.3, 8.0, "Will be removed");
 
 /**
- * Removes the default ephemeral session from `sharedSessions`.
+ * Removes shared instance for the default ephemeral identifier.
  */
-+ (void)removeSharedEphemeralSession;
++ (void)removeSharedEphemeralInstance;
 
 /**
- * Removes the default background session from `sharedSessions`.
+ * Removes shared instance for the default background identifier.
  */
-+ (void)removeSharedBackgroundSession;
++ (void)removeSharedBackgroundInstance;
 
 /**
- * Removes shared session for given identifier.
+ * Removes shared instance for given identifier.
  *
  * @param identifier Identifier for the session.
  */
-+ (void)removeSharedSessionForIdentifier:(nullable NSString *)identifier;
++ (void)removeSharedInstanceForIdentifier:(nullable NSString *)identifier;
 
 /**
- * Removes all shared sessions.
+ * Removes all shared instances.
  */
-+ (void)removeAllSharedSessions;
++ (void)removeAllSharedInstances;
 
 /**
- * Generates a unique session identifier
+ * Returns list of identifiers for all shared instances.
+ * @return Array of identifiers.
  */
-+ (nonnull NSString *)uniqueSessionIdentifier;
++ (nullable NSArray *)sharedInstanceIdentifiers;
+
+/**
+ * Generates a unique instance identifier.
+ */
++ (nonnull NSString *)uniqueInstanceIdentifier;
 
 
 @end
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
@@ -32,9 +32,13 @@
 
 NS_SWIFT_NAME(NetworkManaging)
 @protocol SFNetworkSessionManaging
-- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig;
-- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig;
+- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sessionWithConfigurationIdentifier:sessionConfiguration instead");
+- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sessionWithConfigurationIdentifier:sessionConfiguration instead");
+- (nonnull NSURLSession *)sessionWithConfigurationIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nullable NSURLSessionConfiguration *)configuration useSharedSession:(BOOL)useSharedSession;
 @end
+
+extern NSString * __nonnull const kSFNetworkEphemeralSessionIdentifier NS_SWIFT_NAME(SFNetworkEphemeralSessionIdentifier);
+extern NSString * __nonnull const kSFNetworkBackgroundSessionIdentifier NS_SWIFT_NAME(SFNetworkBackgroundSessionIdentifier);
 
 NS_SWIFT_NAME(Network)
 @interface SFNetwork : NSObject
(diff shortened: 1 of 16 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -41,6 +45,42 @@ @implementation SFNetwork
...
+                    sessionConfiguration = [NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:identifier];
+                } else {
+                    // Case 1: Identifier is `kSFNetworkEphemeralSessionIdentifier`
+                    // Case 2: Custom identifier provided but session configuration wasn't, default to ephemeral as well
+                    sessionConfiguration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
+                }
+            }
+
+            if (useSharedSession) {
+                session = [NSURLSession sessionWithConfiguration:sessionConfiguration delegate:self delegateQueue:nil];
```
wmathurin: Why is the delegate set here but not in the else case?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
```diff
@@ -32,31 +32,46 @@
 
 NS_SWIFT_NAME(NetworkManaging)
 @protocol SFNetworkSessionManaging
-- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig;
-- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig;
+- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sessionWithConfigurationIdentifier:sessionConfiguration:useSharedSession instead");
+- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sessionWithConfigurationIdentifier:sessionConfiguration:useSharedSession instead");
+- (nonnull NSURLSession *)sessionWithConfigurationIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nullable NSURLSessionConfiguration *)configuration useSharedSession:(BOOL)useSharedSession;
```
wmathurin: Instead of having special identifiers that imply certain config, could we keep the old factory methods, and maybe add the useSharedSession:(BOOL)useSharedSession to them?
They would all feed into the new method `(nonnull NSURLSession *)sessionWithConfigurationIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nullable NSURLSessionConfiguration *)configuration useSharedSession:(BOOL)useSharedSession` but at least callers would not need to know about `kSFNetworkEphemeralSessionIdentifier` and `kSFNetworkBackgroundSessionIdentifier`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
bbirman: The delegate method `(void)URLSession:(NSURLSession *)session didBecomeInvalidWithError:(nullable NSError *)error` is only used to remove items from the shared session dictionary. I can change it so that the delegate it used for both cases and have a check in the method instead if that's preferable (that'd also be good if we want to do anything with the error)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
bbirman: For that route, should the identifiers still be external for use with `removeSharedSessionForConfigurationIdentifier`? Or should there be convenience wrappers for that too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
wmathurin: I see. If you set the delegate for the non-shared case, how would you know that the NSURLSession is shared or not? Is it possible to create two NSURLSession with the same identifier, one shared and one not shared?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
wmathurin: Hmm, I guess.
By the way, what happens if useSharedSession is false, and you pass an identifier used before?

bbirman: For ephemeral sessions it would just spin up new ones and there wouldn't be a conflict because the identifier wouldn't be used beyond determining the config type... For background sessions there'd be a problem with `[NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:identifier]`, I'm thinking that for the false case we could append the hash like it already is in the existing background init method (thanks for catching!)

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFNetworkTests.m
```diff
@@ -0,0 +1,130 @@
+//
```
bhariharan: Use our copyright.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
bhariharan: I agree with Wolf's suggestion. If possible, could we avoid callers having to pass in an identifier?
Some thoughts/suggestions:
- have an extra `useSharedSession` argument that a caller could use to explicitly request a private session by passing in `NO`. If `NO` is passed in, we randomly generate a hash to append to the name (already there), and store it in a `Dictionary` in memory.
- keep the existing API signature (without deprecating), but the implementation would simply call the overloaded method, passing in `useSharedSession` as `YES`.
- add an API to return all available identifier names along with the type of session they are, i.e. `background` or `ephemeral`.
- add an API to remove a session from memory that takes an identifier as an argument.
- add an API to remove all sessions in memory.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
```diff
@@ -32,31 +32,46 @@
...
 @end
 
+extern NSString * __nonnull const kSFNetworkEphemeralSessionIdentifier NS_SWIFT_NAME(SFNetworkEphemeralSessionIdentifier);
```
bhariharan: I don't think we need to `extern` these identifiers, especially if we add an API to fetch all available sessions.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -320,6 +320,14 @@ - (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)del
     }
 }
 
+- (SFNetwork *)networkForRequest:(SFRestRequest *)request {
```
bhariharan: What happens if token refresh is thrown in the mix here? For example, let's assume we have 2 requests, A & B. A has been set up to use ephemeral session and B has been set up to use background session. Both of them are enqueued and 1 of them triggers the refresh flow.
- Which session is used for the refresh flow?
- Once refresh is done (assuming it's successful), when the requests are replayed (queue is flushed), do they retain their original session instances?

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFNetworkTests.m
```diff
@@ -0,0 +1,130 @@
...
+
+#import <XCTest/XCTest.h>
+#import "SalesforceSDKCore/SFNetwork.h"
+
+@interface SFNetworkTests : XCTestCase <SFNetworkSessionManaging>
+
+@end
+
+#pragma clang diagnostic push
```
bhariharan: Is this to hide protocol warnings?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
trooper2013: @bhariharan All auth related calls should be using a dataTask &. non background sessions. SessionRefresher should be using its own(auth) related session. Once completion block is invoked the pendingRequests should get fired using the SFRestApi's shared session Right @bbirman?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -320,6 +320,14 @@ - (void)enqueueRequest:(SFRestRequest *)request delegate:(id<SFRestDelegate>)del
...
+    if (request.shouldRunInBackground) {
```
trooper2013: @bbirman We should also handle this scenario  which  is not common but likely to happen i.e  RestClient will be used to invoke rest api(s) from a different host altogether.i.e If  request.serviceHostType is SFSDKRestServiceHostTypeCustom we shouldn't reuse the same URLSession.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.h
```diff
@@ -221,14 +221,18 @@ NS_SWIFT_NAME(RestRequest)
...
 /**
  * Used to specify if the SDK should attempt to refresh tokens on HTTP 403. If YES, the SDK will
  * attempt to refresh on HTTP 403. If NO, refresh will not be attempted.
  */
 @property (nonatomic, assign) BOOL shouldRefreshOn403;
 
+/**
+ * Specifies if the request should run in a background URL session. NO by default.
+ */
+@property (nonatomic, assign) BOOL shouldRunInBackground;
```
trooper2013: @bbirman Should we simply use networkServiceType instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
trooper2013: I think the api is getting a bit complicated to understand. What do you guys think about having  [SFNetwork defaultNetwork] && [SFNetwork defaultBackgroundNetwork] instead for shared mode for salesforce platform? The RestApi will use the above 2 methods for all cases except for external hosts.   Our Auth can create and dispose its own new sessions. Everything else can be custom.

bbirman: @trooper2013 👍 @wmathurin and I were talking about getting rid of the useSharedSession flag yesterday too

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
bbirman: Yup! The sessions for auth and the pending requests would be separate-- @bhariharan let me know if I'm not understanding a piece

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFNetworkTests.m
bbirman: Yup, it's complaining about not implementing the two deprecated methods but putting them in gives a deprecated warning anyway

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
bhariharan: That can still use the shared session, right?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
```diff
@@ -32,30 +32,58 @@
...
+extern NSString * __nonnull const kSFNetworkBackgroundSessionIdentifier NS_SWIFT_NAME(SFNetworkBackgroundSessionIdentifier);
+
 NS_SWIFT_NAME(Network)
 @interface SFNetwork : NSObject
 
 typedef void (^SFDataResponseBlock) (NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) NS_SWIFT_NAME(DataResponseBlock);
 
 @property (nonatomic, readonly, strong, nonnull) NSURLSession *activeSession;
+@property (class, readonly, nonnull) NSDictionary *sharedSessions;
```
bhariharan: Keeping this as a `property` has a lot of problems - it doesn't play well with the quasi-singleton approach and the class methods. You need an instance of the class to read and write to this dictionary. Instead, you should probably use a `static` dictionary to store the shared sessions. For example, look at `SFSDKSalesforceAnalyticsManager` or any of our other quasi-singleton classes that follow the same design. So, this would be something like:
```
static NSMutableDictionary *sharedSessions = nil;
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -41,6 +45,41 @@ @implementation SFNetwork
 static NSURLSessionConfiguration *kSFSessionConfig;
 __weak static id<SFNetworkSessionManaging> kSFNetworkManager;
 
++ (instancetype)defaultEphemeralNetwork {
+    NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
```
bhariharan: This method should simply call the other new class method (`networkWithSessionIdentifier`), passing in the default identifier and a default session configuration.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -41,6 +45,41 @@ @implementation SFNetwork
...
+    return [[self alloc] initWithSessionIdentifier:kSFNetworkEphemeralSessionIdentifier sessionConfiguration:sessionConfiguration];
+}
+
++ (instancetype)defaultBackgroundNetwork {
```
bhariharan: This method should simply call the other new class method (`networkWithSessionIdentifier`), passing in the default identifier and a default session configuration.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -41,6 +45,41 @@ @implementation SFNetwork
...
+    NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:kSFNetworkBackgroundSessionIdentifier];
+    return [[self alloc] initWithSessionIdentifier:kSFNetworkBackgroundSessionIdentifier sessionConfiguration:sessionConfiguration];
+}
+
++ (instancetype)networkWithSessionIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nullable NSURLSessionConfiguration *)sessionConfiguration {
+    return [[self alloc] initWithSessionIdentifier:identifier sessionConfiguration:sessionConfiguration];
```
bhariharan: I don't think we should `alloc` a new instance of this class every time it's called. Instead, you want to look up the `static` dictionary to figure out if an instance exists with the session identifier passed in. If so, return it. If not, then we `alloc` and put that new instance in the `static` dictionary. Here is a similar example from `SFSDKSalesforceAnalyticsManager` (it would look similar to this):
```
+ (instancetype) sharedInstanceWithUser:(SFUserAccount *) userAccount {
    @synchronized ([SFSDKSalesforceAnalyticsManager class]) {
        if (!userAccount) {
            userAccount = [SFUserAccountManager sharedInstance].currentUser;
        }
        if (!userAccount) {
            return nil;
        }
        NSString *key = SFKeyForUserAndScope(userAccount, SFUserAccountScopeCommunity);
        if (!key) {
            return nil;
        }
        id analyticsMgr = analyticsManagerList[key];
        if (!analyticsMgr) {
            if (userAccount.loginState != SFUserAccountLoginStateLoggedIn) {
                [SFSDKCoreLogger w:[self class] format:@"%@ A user account must be in the  SFUserAccountLoginStateLoggedIn state in order to create a SFSDKSalesforceAnalyticsManager instance for a user.", NSStringFromSelector(_cmd)];
                return nil;
            }
            analyticsMgr = [[self alloc] initWithUser:userAccount];
            analyticsManagerList[key] = analyticsMgr;
        }
        return analyticsMgr;
    }
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -41,6 +45,41 @@ @implementation SFNetwork
...
+}
+
+- (instancetype)initWithSessionIdentifier:(nonnull NSString *)identifier sessionConfiguration:(nullable NSURLSessionConfiguration *)sessionConfiguration {
+    self = [super init];
+    if (self) {
```
bhariharan: `init` would be very simple. See comments above about structure.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -97,4 +141,53 @@ + (void)setSessionManager:(id<SFNetworkSessionManaging>)manager {
     kSFNetworkManager = manager;
 }
 
++ (NSDictionary *)sharedSessions {
```
bhariharan: Create a class method for this instead of a `property`, once this becomes a `static` dictionary.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -97,4 +141,53 @@ + (void)setSessionManager:(id<SFNetworkSessionManaging>)manager {
...
+    return [[self mutableSharedSessions] dictionary];
+}
+
++ (void)removeSharedEphemeralSession {
+    [[self class] removeSharedSessionForIdentifier:kSFNetworkEphemeralSessionIdentifier];
```
bhariharan: Can be `[SFNetwork removeSharedSessionForIdentifier:kSFNetworkEphemeralSessionIdentifier];`. Remove usage of `self` from this class method.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -97,4 +141,53 @@ + (void)setSessionManager:(id<SFNetworkSessionManaging>)manager {
...
+}
+
++ (void)removeSharedBackgroundSession {
+    [[self class] removeSharedSessionForIdentifier:kSFNetworkBackgroundSessionIdentifier];
```
bhariharan: Can be `[SFNetwork ...`. Remove usage of `self` from this class method.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -97,4 +141,53 @@ + (void)setSessionManager:(id<SFNetworkSessionManaging>)manager {
...
+}
+
++ (void)removeSharedSessionForIdentifier:(nullable NSString *)identifier {
+    if (identifier) {
+        [[self mutableSharedSessions] removeObject:identifier];
```
bhariharan: Remove usage of `self` from this class method. This should access the `static` dictionary directly.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -97,4 +141,53 @@ + (void)setSessionManager:(id<SFNetworkSessionManaging>)manager {
...
+    }
+}
+
++ (void)removeAllSharedSessions {
+    [[self mutableSharedSessions] removeAllObjects];
```
bhariharan: Remove usage of `self` from this class method. This should access the `static` dictionary directly.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
```diff
@@ -30,32 +30,73 @@
...
+SFSDK_DEPRECATED(7.3, 8.0, "Will be removed")
 NS_SWIFT_NAME(NetworkManaging)
 @protocol SFNetworkSessionManaging
-- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig;
-- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig;
+- (nonnull NSURLSession *)ephemeralSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sharedInstanceWithIdentifier:sessionConfiguration instead");
+- (nonnull NSURLSession *)backgroundSession:(nonnull NSURLSessionConfiguration *)sessionConfig SFSDK_DEPRECATED(7.3, 8.0, "Use sharedInstanceWithIdentifier:sessionConfiguration instead");
 @end
 
+extern NSString * __nonnull const kSFNetworkEphemeralInstanceIdentifier NS_SWIFT_NAME(SFNetworkEphemeralInstanceIdentifier);
```
bhariharan: Minor comment: `Swift` name should probably be `NetworkEphemeralInstanceIdentifier` (no prefixes in `Swift`, right @trooper2013?).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h
```diff
@@ -30,32 +30,73 @@
...
+extern NSString * __nonnull const kSFNetworkBackgroundInstanceIdentifier NS_SWIFT_NAME(SFNetworkBackgroundInstanceIdentifier);
```
bhariharan: Minor comment: `Swift` name should probably be `NetworkBackgroundInstanceIdentifier` (no prefixes in `Swift`, right @trooper2013?).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -41,6 +46,48 @@ @implementation SFNetwork
...
+    return [SFNetwork sharedInstanceWithIdentifier:kSFNetworkEphemeralInstanceIdentifier sessionConfiguration:sessionConfiguration];
+}
+
++ (instancetype)sharedEphemeralInstanceWithIdentifier:(NSString *)identifier {
+    NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
+    return [SFNetwork sharedInstanceWithIdentifier:identifier sessionConfiguration:sessionConfiguration];
+}
+
++ (instancetype)sharedBackgroundInstance {
+    NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration backgroundSessionConfigurationWithIdentifier:kSFNetworkBackgroundInstanceIdentifier];
```
bhariharan: The implementation of this method can be:
```
return [SFNetwork sharedBackgroundInstanceWithIdentifier:kSFNetworkBackgroundInstanceIdentifier];
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -41,6 +46,48 @@ @implementation SFNetwork
 static NSURLSessionConfiguration *kSFSessionConfig;
 __weak static id<SFNetworkSessionManaging> kSFNetworkManager;
 
++ (instancetype)sharedEphemeralInstance {
+    NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
```
bhariharan: The implementation of this method can be:
```
return [SFNetwork sharedEphemeralInstanceWithIdentifier:kSFNetworkEphemeralInstanceIdentifier];
```

PR #2978, opened 2019-08-24
Bug fix for corner case where custom login host is updated with app version updates

bhariharan:
If a custom host is configured though the app's `plist`, and that custom host value is changed with a new version of the app, our code would hold on to the old custom host value. The app would have to be uninstalled and reinstalled to recover. This is because we use `NSUserDefaults` to store the host selection value and a different persistent store for all hosts. This PR addresses that bug by ensuring they are in sync. It also fixes some log levels. Tested using `SmartSyncExplorer`.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
@@ -64,11 +64,13 @@ static NSString * const SFSDKLoginHostNameKey = @"SalesforceLoginHostNameKey";
     if (self) {
         self.loginHostList = [NSMutableArray array];
         SFManagedPreferences *managedPreferences = [SFManagedPreferences sharedPreferences];
+        SFSDKLoginHost *production = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_PRODUCTION"] host:@"login.salesforce.com" deletable:NO];
+        SFSDKLoginHost *sandbox = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_SANDBOX"] host:@"test.salesforce.com" deletable:NO];
 
         // Add the Production and Sandbox login hosts, unless an MDM policy explicitly forbids this.
         if (!(managedPreferences.hasManagedPreferences && managedPreferences.onlyShowAuthorizedHosts)) {
-            [self.loginHostList addObject:[SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_PRODUCTION"] host:@"login.salesforce.com" deletable:NO]];
-            [self.loginHostList addObject:[SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_SANDBOX"] host:@"test.salesforce.com" deletable:NO]];
+            [self.loginHostList addObject:production];
+            [self.loginHostList addObject:sandbox];
         }
 
         // Load from managed preferences (e.g. MDM).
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
```diff
@@ -78,35 +80,42 @@ - (id)init {
...
 
-            // Add the login host from info.plist only if it is not already added.
-            if(![self loginHostForHostAddress:customHost]) {
-                [self.loginHostList addObject:[SFSDKLoginHost hostWithName:customHost host:customHost deletable:NO]];
+            /*
+             * Add the login host from info.plist if it doesn't exist already.
+             * This also handles the case where the custom host configured
+             * was changed between version updates of the application.
+             */
+            if (![self loginHostForHostAddress:customHost]) {
```
bhariharan: If the custom host has changed, we should remove the old entries and put in the new custom host value. Since the key is the old host name which we no longer have, we simply remove all of them and add `Production`, `Sandbox`, and the new custom host back in. This does NOT mess with the MDM use case, and applies only to the custom host set through `plist`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.m
```diff
@@ -93,13 +87,13 @@ - (NSString *)loginHost {
         [defaults synchronize];
         return host;
     }
-    
-    // Fetch from the standard defaults or bundle.
+
+    // Fetch from the standard defaults or bundle and ensues that login host still exists.
     NSString *loginHost = [defaults stringForKey:kSFUserAccountOAuthLoginHost];
-    if ([loginHost length] > 0) {
+    if (loginHost.length > 0 && [[SFSDKLoginHostStorage sharedInstance] loginHostForHostAddress:loginHost] != nil) {
```
bhariharan: This is the other important fix. Here, we check if `NSUserDefaults` has a value stored that's not persisted in our server list, i.e. a zombied server host. If so, we replace it with the new value (this is handled in the code below).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKAuthConfigUtil.m
```diff
@@ -40,14 +40,14 @@ + (void)getMyDomainAuthConfig:(MyDomainAuthConfigBlock)authConfigBlock oauthCred
 
 + (void)getMyDomainAuthConfig:(MyDomainAuthConfigBlock)authConfigBlock loginDomain:(NSString *)loginDomain {
     NSString *orgConfigUrl = [NSString stringWithFormat:@"https://%@%@", loginDomain, kSFOAuthEndPointAuthConfiguration];
-    [SFSDKCoreLogger d:[self class] format:@"%@ Advanced authentication configured. Retrieving auth configuration from %@", NSStringFromSelector(_cmd), orgConfigUrl];
+    [SFSDKCoreLogger i:[self class] format:@"%@ Advanced authentication configured. Retrieving auth configuration from %@", NSStringFromSelector(_cmd), orgConfigUrl];
```
bhariharan: Fixed some log levels in this file to better showcase them. We need these logs in release builds too.

wmathurin:
There are a couple of test failures that seem to have been caused by the change?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthPreferences.m
```diff
@@ -59,32 +58,34 @@
...
 
-- (void)setLoginHost:(NSString*)host {
+- (void)setLoginHost:(NSString *)host {
     NSString *oldLoginHost = [self loginHost];
-    
     if (nil == host) {
         [[NSUserDefaults msdkUserDefaults] removeObjectForKey:kSFUserAccountOAuthLoginHost];
     } else {
+
+        // Persists the login host if it doesn't exist already.
```
bhariharan: @wmathurin This is the issue the test failure exposed. We weren't persisting in some cases. Also, this is the only code change from before (I don't know why `GitHub` is showing the merged changes on this PR). All tests are passing now.

PR #2982, opened 2019-08-31
Fixing bug in saveSoupEntryExternally causing externally stored files to be cut-off in some cases

wmathurin:
When we upsert soup entries that are stored in external files, we use `NSJSONSerialization:writeJSONObject` to write the JSON to a file.
That method can fail if with certain unicode characters. See [corruption tester](https://github.com/wmathurin/CorruptionTester).
`saveSoupEntryExternally` should exit with an error, causing the `upsert` to revert.
Unfoturnately, we were looking at the return value `NSJSONSerialization:writeJSONObject` to decide if an error occurred.
But this method returns the number of bytes written so it could return a non zero value even though it fails to write the entire JSON.
As a result, the file ends up being be cut-off, which causes issues at query time.

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
@@ -865,10 +865,15 @@ NSUInteger CACHES_COUNT_LIMIT = 1024;
     log(@"1/4 Starting to write to tmp file");
     [outputStream open];
     NSError *error = nil;
-    BOOL success = [NSJSONSerialization writeJSONObject:soupEntry
-                                               toStream:outputStream
-                                                options:0
-                                                  error:&error];
+    
+    // NSJSONSerialization:writeJSONObject returns the number of bytes written
+    // So NSJSONSerialization:writeJSONObject can return a value > 0 while there is an error
+    [NSJSONSerialization writeJSONObject:soupEntry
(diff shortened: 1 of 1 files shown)


wmathurin:
Apple documentation states:
> The number of bytes written to the stream, or 0 if an error occurs.
I was able to see it not be the case - but I am having trouble seeing it again.

sgoldberg-sfdc:
If this is a valid case, then this is a better measure of success, but do we want to log the error?

wmathurin:
To get a return value > 0 (which would be interpreted as success by the code before this PR) with an error, simply have lone (unpaired) UTF-16 surrogate (e.g. U+D800) in the string.
Such a string is still valid according to JSON grammar, but not handled properly by a number of JSON parser (including NSJSONSerialization). See http://seriot.ch/parsing_json.php

It's easy to produce such strings with [CorruptionTester](https://github.com/wmathurin/CorruptionTester). I'm not sure they can happen "in the wild".
If they did though, they would cause JSON files to be cut-off.

wmathurin:
@sgoldberg-sfdc see logging improvements in https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2982/commits/acb72d8fba2ffcb269f7addd3dd3b5e76fc700c4

PR #2985, opened 2019-09-04
Ability to set the photo with a method and callback to ensure the asy…

jeanbovet:
…nchronous behavior is visible to consumer

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
@@ -264,6 +264,10 @@ static NSString * const kGlobalScopingKey = @"-global-";
 }
 
 - (void)setPhoto:(UIImage *)photo {
+    [self setPhoto:photo completion:nil];
+}
+
+- (void)setPhoto:(UIImage*)photo completion:(void (^ __nullable)(void))completion {
     dispatch_barrier_async(_syncQueue, ^{
         NSError *error = nil;
         NSString *photoPath = [self photoPathInternal];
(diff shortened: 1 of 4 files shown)


jeanbovet:
Closing to rebase on dev

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -264,6 +264,10 @@ - (UIImage*)photo {
 }
 
 - (void)setPhoto:(UIImage *)photo {
+    [self setPhoto:photo completion:nil];
+}
+
+- (void)setPhoto:(UIImage*)photo completion:(void (^ __nullable)(void))completion {
```
trooper2013: @jeanbovet Since we are adding a completion block, we should also care about the success/failure. What do you think?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -274,13 +278,18 @@ - (void)setPhoto:(UIImage *)photo {
...
-            NSData *data = UIImagePNGRepresentation(photo);
-            if (![data writeToFile:photoPath options:NSDataWritingAtomic error:&error]) {
-                [SFSDKCoreLogger e:[self class] format:@"Unable to write photo to disk: %@", error];
+            if (photo) {
+                NSData *data = UIImagePNGRepresentation(photo);
+                if (![data writeToFile:photoPath options:NSDataWritingAtomic error:&error]) {
+                    [SFSDKCoreLogger e:[self class] format:@"Unable to write photo to disk: %@", error];
+                }
             }
             [self willChangeValueForKey:@"photo"];
```
trooper2013: We should move the willChange and didChange  inside the `if then` block?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
jeanbovet: That's a great question! I did consider it but backtracked because I wasn't sure how you wanted to treat error. For example, in the current implementation, either error in deleting or storing the image will still set the user: should I change that as well to not assign the photo if storing the photo fails?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
jeanbovet: Which if then block are you thinking?

PR #2986, opened 2019-09-04
Ability to set the photo with a method and callback to ensure the asy…

jeanbovet:
…nchronous behavior is visible to consumer

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
@@ -100,6 +100,8 @@ NS_SWIFT_NAME(UserAccount)
 @property (nonatomic, copy) NSString *userName SFSDK_DEPRECATED(7.1, 8.0, "Use SFUserAccount.idData properties instead");
 
 /** The user's photo. Usually store a thumbnail of the user.
+ Note: setting the photo is asynchronous. Use the method
+ below to get a callback when the photo is set.
  */
 @property (nonatomic, strong, nullable) UIImage *photo;
 
@@ -167,6 +169,12 @@ NS_SWIFT_NAME(UserAccount)
  @return The object for a particular key
  */
 - (nullable id)customDataObjectForKey:(id)key;
+
+/** Sets the user's photo.
+ @param photo The user photo, usually the thumbnail of the user.
+ @param completion Optional callback block invoked when the photo has been set
+ */
+- (void)setPhoto:(UIImage*_Nullable)photo completion:(void (^ __nullable)(void))completion;
 
 /** Function that returns a key that uniquely identifies this user account for the
  given scope. Note that if you use SFUserAccountScopeGlobal,
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
@@ -172,9 +172,9 @@ NS_SWIFT_NAME(UserAccount)
 
 /** Sets the user's photo.
  @param photo The user photo, usually the thumbnail of the user.
- @param completion Optional callback block invoked when the photo has been set
+ @param completion Optional callback block invoked when the photo has been set. If not set, an error is returned.
  */
-- (void)setPhoto:(UIImage*_Nullable)photo completion:(void (^ __nullable)(void))completion;
+- (void)setPhoto:(UIImage*_Nullable)photo completion:(void (^ __nullable)(NSError* _Nullable))completion;
 
 /** Function that returns a key that uniquely identifies this user account for the
  given scope. Note that if you use SFUserAccountScopeGlobal,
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -274,13 +278,18 @@ - (void)setPhoto:(UIImage *)photo {
...
-            NSData *data = UIImagePNGRepresentation(photo);
-            if (![data writeToFile:photoPath options:NSDataWritingAtomic error:&error]) {
-                [SFSDKCoreLogger e:[self class] format:@"Unable to write photo to disk: %@", error];
+            if (photo) {
+                NSData *data = UIImagePNGRepresentation(photo);
+                if (![data writeToFile:photoPath options:NSDataWritingAtomic error:&error]) {
+                    [SFSDKCoreLogger e:[self class] format:@"Unable to write photo to disk: %@", error];
+                }
             }
             [self willChangeValueForKey:@"photo"];
```
trooper2013: We should move the willChange and didChange inside the if then block?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
```diff
@@ -168,6 +170,12 @@ NS_SWIFT_NAME(UserAccount)
  */
 - (nullable id)customDataObjectForKey:(id)key;
 
+/** Sets the user's photo.
+ @param photo The user photo, usually the thumbnail of the user.
+ @param completion Optional callback block invoked when the photo has been set
+ */
+- (void)setPhoto:(UIImage*_Nullable)photo completion:(void (^ __nullable)(void))completion;
```
trooper2013: @jeanbovet Since we are adding a completion block, we should also care about the success/failure. What do you think?

jeanbovet: That's a great question! I did consider it but backtracked because I wasn't sure how you wanted to treat error. For example, in the current implementation, either error in deleting or storing the image will still set the user: should I change that as well to not assign the photo if storing the photo fails?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
jeanbovet: Which if then block are you thinking?

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountPhotoTests.m
```diff
@@ -0,0 +1,85 @@
+/*
+ Copyright (c) 2017-present, salesforce.com, inc. All rights reserved.
```
bhariharan: `2019-present`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
bhariharan:
```
[self willChangeValueForKey:@"photo"];
self->_photo = photo;
[self didChangeValueForKey:@"photo"];
```
This code should be inside the `if (photo) {` block, shouldn't it?

trooper2013: @jeanbovet ideally we should be doing this only if the value was set i.e. if there isn't an error.`[data writeToFile:photoPath options:NSDataWritingAtomic error:&error]` Since you are refactoring this piece of code it would be  good to fix this. In the past it's possible that a nil is sent, the value is not set but we still notify the observers.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
trooper2013: @jeanbovet Since the error will be bubbled up to the app, it could choose to ignore or handle it if needed.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
jeanbovet: Sure, it makes sense to set it only if there was no error. However, we still need to honor the fact that the user can set a nil photo, right?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.h
jeanbovet: Updated, let me know what you think

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
jeanbovet: Updated, let me know what you think

PR #2990, opened 2019-09-06
Getting webview useragent from WKWebView

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -962,23 +964,22 @@ - (SFSDKUserAgentCreationBlock)defaultUserAgentString {
...
-                                 webViewUserAgent
+                                 self.webViewUserAgent == nil ? @"" : self.webViewUserAgent
                                  ];
         return myUserAgent;
     };
 }
 
-- (NSString *)getUIWebViewUserAgent {
-    static NSString *webViewUserAgent = nil;
+- (void)computeWebViewUserAgent {
```
trooper2013: @wmathurin Does the actual base agent String change drastically or is it the same?

wmathurin: The web view user agent is appended at the end of the Mobile SDK generated user agent. Are you asking if UI web view and WK web view produce different user agent?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -962,23 +964,22 @@ - (SFSDKUserAgentCreationBlock)defaultUserAgentString {
...
     static dispatch_once_t onceToken;
+    self.webView = [[WKWebView alloc] initWithFrame:CGRectZero];
     dispatch_once_on_main_thread(&onceToken, ^{
-        // Grabs the current user agent. This is very hackish but WKWebView, which we
-        // want to transition too currently evaluates Javscript asynchronously (11/2/17)
-        UIWebView *webView = [[UIWebView alloc] initWithFrame:CGRectZero];
-        webViewUserAgent = [webView stringByEvaluatingJavaScriptFromString:@"navigator.userAgent"];
+        [self.webView loadHTMLString:@"<html></html>" baseURL:nil];
+        [self.webView evaluateJavaScript:@"navigator.userAgent" completionHandler:^(id __nullable userAgent, NSError * __nullable error) {
```
bhariharan: Do we need the `strong-weak` dance for `self` within this completion block?

bhariharan: Also, does `webView` have to be a `property`? Can't it be a local variable within this method? We need it only once anyway.

wmathurin: If I use a local variable, the evaluateJavaScript completes with the error code 3: "The WKWebView was invalidated". Can't find a way around it except for doing [this](https://gist.github.com/Koze/cfda5d2af12f6215424e#file-webviewgetuseragent2-m-L19).

sgoldberg-sfdc: evaluateJavascript is async as opposed to stringByEvaluatingJavaScriptFromString. It is likely when the webview goes out of scope, the webcontent process is killed, hence the error.

PR #2994, opened 2019-09-11
Suppress deprecated warning

bbirman:
Goes along with #2975, it'll go away in 8.0


bhariharan:
@bbirman I don't think we want to suppress the deprecation warning. It'll make it harder to delete it in `8.0` (we may not discover it). I'd recommend leaving it as-is, and removing it from `dev` after `7.3` goes out.

bhariharan:
It's okay either way, I guess. As long as we don't forget to remove it.

PR #2997, opened 2019-09-13
Set color directly instead of creating image from color

manushitalwar:


wmathurin:
Reverting here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3002/commits/a150e3f29b1c586fa84f5e5246bbc0ef89e025a3

We will look into adding dark mode support in 8.0.

PR #3015, opened 2019-09-20
Conditional compilation (otherwise it won't build with XCode 10)

wmathurin:


wmathurin:
iOS 12.4 simulators are only available with XCode 10.3.
XCode 11.0 has 12.2 simulators.

Without that change we cannot build the SDK on XCode 10.
Alternatively we could make the deployment target 12.2 and limits ourselves to supporting XCode 11.

bhariharan:
So, the options are:
- Add conditional compilation and support `Xcode 10+` and `Xcode 11+`.
- No conditional compilation, `iOS 12.2` minimum, and require `Xcode 11+`.

If that's the case, I like option (2) better, i.e. make `Xcode 11` required for `Mobile SDK 8.0`. We've always done that in the past anyway (require the latest version of `Xcode`). I don't see why we should support `Xcode 10`.

wmathurin:
Circle CI Xcode 11 image (https://circle-macos-docs.s3.amazonaws.com/image-manifest/v1136/index.html) does not have 12.4 simulators. However locally my Xcode 11 does.
We are still at GM seed 2, maybe that will get fixed with the GA.
In the mean time, this PR should allow the SDK to build on XCode 10 and allow tests to be run against iOS 12.

PR #3008, opened 2019-09-17
Changing deployment target from 11.0 to 12.4

wmathurin:
Also fixed resulting deprecation warnings

PR #3017, opened 2019-09-21
Fix issue happening with iOS 12 as target

wmathurin:
Archivers defaults to secure coding.
Since the objects we archive don't follow that protocol, we need to pass requiresSecureCoding = NO when archiving/unarchiving.
I missed that one place when I did https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3008.

There was one test failing because of it (testMultipleAccounts).
More importantly, on restart, one had to login again!


wmathurin:
Here is the diff without white spaces changes https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3017/files?w=1

PR #3018, opened 2019-09-22
Make projects in workspace Xcode 11.0 compatible

bhariharan:
A checkbox in `Xcode`.

### DIFF
--- a/libs/MobileSync/MobileSync.xcodeproj/project.pbxproj
+++ b/libs/MobileSync/MobileSync.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 51;
+	objectVersion = 52;
 	objects = {
 
 /* Begin PBXAggregateTarget section */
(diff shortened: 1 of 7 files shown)


bhariharan:
I don't think so - this is just the project file format (which hasn't changed). Could you run it on `Xcode 10.3` locally and check? I overwrote it with `Xcode 11` :-(

wmathurin:
XCode 10.3 won't open the projects anymore.

Screen Shot 2019-09-23 at 11 02 53 AM

bhariharan:
@wmathurin Do we care? In the past, we've always required the latest version of `Xcode` with the major version. Besides, you need it for `iOS 13` anyway. Keeps the test matrix easier.

wmathurin:
Waiting to see if Xcode 11 Circle CI will add 12.4 simulators. Otherwise, we need to go back and lower deployment target to 12.2 everywhere.

bhariharan:
Closing this for now, until `Xcode 11` adds a `iOS 12.4` simulator.

PR #3023, opened 2019-09-25
False negative "json parse error" were being fired when running a bad…

wmathurin:
… query (e.g. querying a non-existent soup)

Added tests also.

### DIFF
--- a/libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
+++ b/libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
@@ -47,8 +47,8 @@
 #define kTestSmartStoreName  @"testSmartStore"
 #define kTestSoupName        @"testSoup"
 
-@interface SFSmartStoreTests ()
-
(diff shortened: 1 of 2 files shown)


libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
```diff
@@ -1391,6 +1391,104 @@ - (void)testReadMultiByteCharacterAroundBufferBoundary {
...
+        XCTAssertNotNil(error, @"Error expected");
+        XCTAssertNil(serializedCursor, @"Serialized cursor should be nil");
+    }
+}
+
+/**
+Test running an invalid query with SFStoreCursor while JSON check are on
+Make sure we do NOT get a notification for a JSON parsing error
+*/
+- (void)testInvalidQueryWithStoreCursorWithRawJsonCheckOn {
```
wmathurin: NB: this test fails without the change to SFStoreCursor.m from this PR.

PR #3024, opened 2019-09-25
Added Rest Api Swift Extension (WIP)

trooper2013:
This PR is the first of the few PR's to follow for adding Swift Extensions to the mobile SDK. Refines some of the Blocks API(s) and introduces an alternative API with a single completionBlock and ResultType.

Todo: Add some RestClient Swift tests.

PR #3026, opened 2019-09-26
Adds Swift Extensions (with and without Combine) for RestClient 

trooper2013:
Sorry @wmathurin  & @bhariharan The merge for the previous PR was troublesome. Ive incorporated some of your comments from #3024 in this PR.  Replaying my changes was much easier.

1. Added RestAPI Extensions. (Basic Overlays)
2. Added a Combine extension (iOS 13+)
3. Refined some of the Blocks API(s)
4. Reflected changes in RestApiExplorer.


libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -1935,6 +1946,7 @@
 					};
 					CE4CE26A1C0E449C009F6029 = {
 						CreatedOnToolsVersion = 7.1;
+						LastSwiftMigration = 1100;
```
bhariharan: @wmathurin Do you think this will cause problems with `Xcode 10.x`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+        self.data = data
+        self.urlResponse = urlResponse
+    }
+    
+    /// Parse the response as a Json Dictionary.
+    public func asJson() throws -> Any {
+        do {
+            let jsonData = try JSONSerialization.jsonObject(with: data, options: [])
+            return jsonData
+        } catch let error {
```
bhariharan: Just like `Java try-catch-throw` (reminds me of how this was a no-no in `Obj C` land except in special cases ;-)).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+        let request = self.request(forSObjectTree: objectType, objectTrees: objectTrees, apiVersion: self.apiVersion)
+        self.send(request: request, completionBlock)
+        return request
+    }
+    
+    /// Owned Files request
+    /// - Parameter userId: User identifier.
+    /// - Parameter page: Page number.
+    /// - Parameter completionBlock: The completion block to invoke.
+    func ownedFiles(_ userId: String?, page: Int, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) -> RestRequest {
```
bhariharan: I think `userId` should be required, not optional.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+       let request = self.request(forOwnedFilesList: userId, page: UInt(page), apiVersion: self.apiVersion)
+       self.send(request: request, completionBlock)
+       return request
+    }
+    
+    /// Get files in users groups
+    /// - Parameter userId: User identifier.
+    /// - Parameter page: Page number.
+    /// - Parameter completionBlock: The completion block to invoke.
+    func filesInUsersGroups(_ userId: String?, page: Int, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) -> RestRequest {
```
bhariharan: Same here. I think `userId` should be required, not optional.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+        let request = self.requestForFiles(inUsersGroups: userId, page: UInt(page), apiVersion: self.apiVersion)
+        self.send(request: request, completionBlock)
+        return request
+    }
+    
+    /// Files shared with user.
+    /// - Parameter userId: User Identifier
+    /// - Parameter page: Page number
+    /// - Parameter completionBlock: The completion block to invoke.
+    func filesSharedWithUser(_ userId: String?, page: Int, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) -> RestRequest {
```
bhariharan: Same as above.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+       return request
+    }
+    
+    /// Upload a File.
+    /// - Parameter data: Data contents.
+    /// - Parameter name: Name for the file
+    /// - Parameter description: Description for file.
+    /// - Parameter mimeType: The mime type.
+    /// - Parameter completionBlock: The completion block to invoke.
+    func uploadFileShare(_ data: Data, name: String, description: String, mimeType: String, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) -> RestRequest {
```
bhariharan: `description` might be optional here, IIRC.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+    }
+    
+    /// Upload a photo.
+    /// - Parameter data: Data contents.
+    /// - Parameter name: Name for the file
+    /// - Parameter description: Description for file.
+    /// - Parameter mimeType: The mime type.
+    /// - Parameter userId: Identifier for user.
+    /// - Parameter completionBlock: The completion block to invoke.
+    func uploadProfilePhoto(_ data: Data, name: String, description: String, mimeType: String, userId: String, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) -> RestRequest {
```
bhariharan: Same here - `description` might be optional.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+        let request = self.request(forProfilePhotoUpload: data, fileName: name, mimeType: mimeType, userId: userId, apiVersion: self.apiVersion)
+       self.send(request: request, completionBlock)
+       
+       return request
+    }
+ 
+}
+
+@available(iOS 13.0, tvOS 13.0, watchOS 6.0, *)
```
bhariharan: Do we need `tvOS` and `watchOS` declarations here? We don't support either of those.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+        let request = self.compositeRequest(requests, refIds: referenceIds, allOrNone: allOrNone,apiVersion: self.apiVersion)
+        self.send(request: request, completionBlock)
+        return request
+    }
+    
+    /// Execute a batch request
+    /// - Parameter requests: Array of RestRequewst
+    /// - Parameter haltOnError: Boolean to indicate whether to stop or continue when error occurs in any request.
+    /// - Parameter completionBlock: The completion block to invoke.
+    func batch(_ requests:[RestRequest], haltOnError: Bool, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) -> RestRequest {
```
wmathurin: We could do better than a generic RestResponse.
The shape of the response is well known. See https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/responses_composite_batch.htm.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+        self.send(request: request, completionBlock)
+        return request
+    }
+     
+    /// Execute a Composite request
+    /// - Parameter requests: Array of RestRequewst
+    /// - Parameter referenceIds: String array of reference identifiers.
+    /// - Parameter allOrNone: Boolean to indicate whether to set allOrNone.
+    /// - Parameter completionBlock: The completion block to invoke.
+    func composite(_ requests:[RestRequest], referenceIds: [String], allOrNone: Bool, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) -> RestRequest {
```
wmathurin: We could do better than a generic RestResponse.
The response body contains an array with the responses for each sub request. For more information see: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/responses_composite.htm

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+        }
+    }
+}
+
+extension RestClient {
+    
+    /// Execute a request.
+    /// - Parameter request: RestRequest object
+    /// - Parameter completionBlock: The completion block to invoke.
+    public func send(request: RestRequest, _ completionBlock: @escaping (Result<RestResponse,RestClientError>) -> () ) {
```
wmathurin: I wonder if we should just provide this one (and the iOS13 publisher one) and let them use the RestRequest factory methods.

wmathurin: Maybe the other methods (query/queryAll/search/describe etc etc) are only useful if they return a better typed response. When we know the request, we know the shape of the response.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
wmathurin: Maybe.
Swift 5 works starting in 10.2. So we could use that (or 10.3).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
wmathurin: Throw in swift is more performant than in objective-c.

> Error handling in Swift resembles exception handling in other languages, with the use of the try, catch and throw keywords. Unlike exception handling in many languages—including Objective-C—error handling in Swift does not involve unwinding the call stack, a process that can be computationally expensive. As such, the performance characteristics of a throw statement are comparable to those of a return statement.

Source: https://docs.swift.org/swift-book/LanguageGuide/ErrorHandling.html

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
trooper2013: @bhariharan its merely reflecting what the [obj-c api](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/ffa6ca5c31008ca41756eff2a357c45660bdd94d/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI%2BFiles.h#L42) does. Should we change the objc-api and impl? The obj-c api replaces userId with "ME" in the url path.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
trooper2013: Same reasons as above.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
trooper2013: In the Obj-c api the description is not optional. Should we change the impl?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
trooper2013: I think we should keep the restrictions in place for watchOS.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
trooper2013: @wmathurin  I agree if we decide to internalize the other apis. i.e. these apis should not  be available in swift. Otherwise it will just add to the confusion (combine, resulttype &  older api shape)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -0,0 +1,432 @@
...
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+import Combine
+/// Errors that can be thrown while using RestClient
+public enum RestClientError: Error {
+    case ApiResponseIsEmpty
+    case ApiInvocationFailed(underlyingError: Error, urlResponse: URLResponse?)
```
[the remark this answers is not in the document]
trooper2013: @bhariharan It is optional because the SFRestAPI failureBlock has modeled it as a nullable.

PR #3030, opened 2019-10-02
Using 12.2 for our deployment target (instead of 12.4) 

wmathurin:
because 12.4 simulators are not available in Xcode 11 circle CI image


.circleci/config.yml
```diff
@@ -8,15 +8,6 @@ executors:
       BASH_ENV: ~/.bashrc
       FASTLANE_SKIP_UPDATE_CHECK: "true"
       CHRUBY_VER: 2.6.3
-  mac-xcode10:
-    working_directory: ~/SalesforceMobileSDK-iOS
-    macos:
-      xcode: 10.3.0
```
brandonpage: For the `mac` executer above, should we move to `11.1.0`? 
https://discuss.circleci.com/t/xcode-11-1-gm-image-released/32668

That would require changing iOS `13.0` to `13.1` also.

wmathurin: Maybe in a separate PR, I have not had a chance to make sure everything works fine with 13.1.

PR #3037, opened 2019-10-07
Adds Batch Request/Response objects for SFRestAPI

trooper2013:
Similar change as we made for the Composite Request. Reflects the  changes in the RestClient for swift.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFSDKBatchRequest.m
```diff
@@ -0,0 +1,146 @@
...
+    }
+    return self;
+}
+
+-(NSArray<SFRestRequest *> *)batchRequests {
+    return [NSArray arrayWithArray:_requests];
+}
+
+
+-(void)addRequest:(SFRestRequest *)subRequest {
```
wmathurin: IMHO, we only need the addRequest on the builder object. The BatchRequest should have a constructor with all the necessary parameters - that the builder’s build method would invoke.

PR #3039, opened 2019-10-07
Move task identifier to local variable to avoid reentrancy issues leading to BTNF

sgoldberg-sfdc:
-  In 7.1 the Background Task Identifier was moved from a local variable to a global
-  This created a situation wherein if the publishOnAppBackground is called a second time before the first background task is ended, we overwrite the task id, and the original background task is never properly ended
-  Apple will then terminate the app due to a Background Task Not Finished.


sgoldberg-sfdc:
The UIBackgroundTaskInvalid is at least deterministic. The old public behavior which did not specify a value used whatever was in the value of the global at the time. Either random data, UIBackgroundTaskInvalid, or a valid task identifier that was unintended. 

Let's discuss other approaches though.

PR #3040, opened 2019-10-07
Move task identifier to local variable to avoid reentrancy issues leading to BTNF

sgoldberg-sfdc:
- In 7.1 the Background Task Identifier was moved from a local variable to a global
- This created a situation wherein if the publishOnAppBackground is called a second time before the first background task is ended, we overwrite the task id, and the original background task is never properly ended
- Apple will then terminate the app due to a Background Task Not Finished.
- This approach, moving it to an atomic property will allow us to gate calls to publishOnBackground so only one instance is active at a time.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
@@ -51,8 +51,6 @@ static NSString * const kSFAppFeatureAiltnEnabled = @"AI";
 
 static NSMutableDictionary *analyticsManagerList = nil;
 
-UIBackgroundTaskIdentifier task;
-
 @implementation SFSDKSalesforceAnalyticsManager
 
 + (void)initialize {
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKSalesforceAnalyticsManager.m
```diff
@@ -315,13 +313,16 @@ - (void) publishOnAppBackground {
...
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-        __block typeof(self) weakSelf = self;
-        task = [[SFApplicationHelper sharedApplication] beginBackgroundTaskWithName:NSStringFromClass([self class]) expirationHandler:^{
-            [weakSelf cleanupBackgroundTask];
-        }];
-        [self publishAllEvents];
-    });
+    // Avoid re-entrance if task is active
+    if (self.task == UIBackgroundTaskInvalid) {
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
wmathurin: There is a small window for a race condition here.

sgoldberg-sfdc: Yeah, I saw that as well but seems much less likely that the current state. We can revise in subsequent PRs but need to CP something in 220. 

I thought maybe we can just convert this to use a serial queue but the publish event is async in itself, so the execution of the block would terminate allowing another block in before we end the task on a successful publish. Same issue with a critical section. Have to unravel the multiple asynchronous behaviors here to have a truly bullet proof approach.

wmathurin: Sounds good.

PR #3042, opened 2019-10-09
Supports use of scene delegate in app

trooper2013:
Uses scenes to retrieve the key window.


trooper2013:
> LGTM.
> Will you be updating the iOS 13 templates to use SceneDelegate ?

Yes I will update templates today.

PR #3048, opened 2019-10-16
Native smartstore query are back to building the result NSArray incrementally

wmathurin:
Both query and queryAsString now calls runQuery which expects either a NSMutableArray or NSMutableString and getDataFromRow is back.

This work was already done on Android, see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/1850


bhariharan:
👍

PR #3050, opened 2019-10-17
Encrypt user photo

bbirman:
Also removed the old photo setter

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
@@ -43,6 +43,7 @@ static const char * kSyncQueue = "com.salesforce.mobilesdk.sfuseraccount.syncque
 /** Key that identifies the global scope
  */
 static NSString * const kGlobalScopingKey = @"-global-";
+static NSString * const kUserAccountPhotoEncryptionKeyLabel = @"com.salesforce.userAccount.photos.encryptionKey";
 
 @interface SFUserAccount ()
 {
@@ -141,7 +142,7 @@ static NSString * const kGlobalScopingKey = @"-global-";
     return [userPhotoPath stringByAppendingPathComponent:[SFDirectoryManager safeStringForDiskRepresentation:_credentials.userId]];
 }
 
-- (UIImage*)photo {
+- (UIImage *)photo {
     if (nil == _photo) {
         __weak __typeof(self) weakSelf = self;
         dispatch_sync(_syncQueue, ^{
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -141,25 +142,33 @@ - (NSString *)photoPathInternal:(NSError**)error {
     return [userPhotoPath stringByAppendingPathComponent:[SFDirectoryManager safeStringForDiskRepresentation:_credentials.userId]];
 }
 
-- (UIImage*)photo {
+- (UIImage *)photo {
     if (nil == _photo) {
```
bbirman: @bhariharan I realized this has a check so it only goes to disk when nil so I didn't move it to a fetch method like we talked about before

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager+Internal.h
```diff
@@ -32,6 +32,13 @@
 
 @class SFSDKAuthPreferences;
 
+extern NSString * _Nonnull const kSFSDKUserAccountManagerErrorDomain;
+
+typedef NS_ENUM(NSUInteger, SFSDKUserAccountManagerErrorCode) {
```
bbirman: @trooper2013  this is where I'm thinking the other SFUserAccountManagerErrorDomain codes could move to when it's ready, the others were 10001-10004, so could drop them in or renumber them all later

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -377,6 +385,42 @@ + (NSString *)loginStateDescriptionFromLoginState:(SFUserAccountLoginState)login
...
+        if (error) {
+            [SFSDKCoreLogger e:[self class] format:@"Unable to write photo to disk: %@", *error];
+        }
+        return NO;
+    }
+
+    return YES;
+}
+
+- (UIImage *)decryptPhoto:(NSString *)photoPath {
```
wmathurin: autoCreate could be NO here - if the key does not exist and is created at this point, there is no chance it could decrypt the file.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccount.m
```diff
@@ -141,25 +142,33 @@ - (NSString *)photoPathInternal:(NSError**)error {
...
             __strong typeof(weakSelf) strongSelf = weakSelf;
             NSString *photoPath = [strongSelf photoPathInternal:nil];
             NSFileManager *manager = [[NSFileManager alloc] init];
             if ([manager fileExistsAtPath:photoPath]) {
-                strongSelf->_photo = [[UIImage alloc] initWithContentsOfFile:photoPath];
+                UIImage *decryptedPhoto = [self decryptPhoto:photoPath];
+                if (decryptedPhoto) {
+                    strongSelf->_photo = decryptedPhoto;
+                } else {
+                    // Check for upgrade scenario
```
bhariharan: Maybe add a `TODO:` here to remove the upgrade scenario in `Mobile SDK 9.0`.

PR #3052, opened 2019-10-17
Update host page presentation style

TLiu2121999:


TLiu2121999:
this change is for keeping the old behavior. So no need for tests updates. 
on iOS13 the default vc presentation style changed to "pop over", we want to keep the old style for now.

TLiu2121999:
@wmathurin could you merge this one please?

PR #3055, opened 2019-10-18
Swift extensions for SmartStore - work in progress

wmathurin:
We have decent swift interfaces already (through NS_SWIFT_NAME) and all SmartStore operations are synchronous. So not a lot to do here.

However I simplified two operations:
1) soup registration: simply takes a name and an array of paths. No need for SoupSpec or IndexSpec (we want to phase out external storage and non-json indexing).

2) query: simply takes smartsql and return Result<> or Future<> (on iOS 13). No need for QuerySpec. Also no need for page size or page index since that can be expressed inside the smartsql.

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/Extensions/SmartStore.swift
+++ b/libs/SmartStore/SmartStore/Classes/Extensions/SmartStore.swift
@@ -1,0 +1,89 @@
+import Foundation
+
+/*
+ SmartStire.swift
+ SalesforceSDKCore
+ 
+ Copyright (c) 2019-present, salesforce.com, inc. All rights reserved.
+ 
+ Redistribution and use of this software in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this list of conditions
+ and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice, this list of
+ conditions and the following disclaimer in the documentation and/or other materials provided
+ with the distribution.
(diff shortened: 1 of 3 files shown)


libs/SmartStore/SmartStore/Classes/Extensions/SmartStore.swift
```diff
@@ -0,0 +1,105 @@
+import Foundation
+
+/*
+ SmartStire.swift
```
bhariharan: Typo: Should be `SmartStore.swift`.

libs/SmartStore/SmartStore/Classes/Extensions/SmartStore.swift
```diff
@@ -0,0 +1,105 @@
...
+ SmartStore extension
+ */
+extension SmartStore {
+    /**
+     Register a soup and returns a Result<>
+     @param withName: name of the soups
+     @param withIndexPaths: array of paths to index
+     @return a Result<Bool, SmartStoreError>
+     */
+    func registerSoup(withName soupName: String, withIndexPaths indexPaths: [String]) -> Result<Bool, SmartStoreError> {
```
wmathurin: @trooper2013 @bhariharan Do we want to start introducing methods that return a Result<> instead of throwing an error? I did it for this one, but there are many more we could do it for.

wmathurin:
Merging now. We can refine in subsequent PRs.

libs/SmartStore/SmartStore/Classes/Extensions/SmartStore.swift
bhariharan: Sure, that sounds good to me!

Issue #3060, opened 2019-10-22
Crash UIApplicationInvalidInterfaceOrientation in -[UIViewController setInterfaceOrientation:]


MichaelWTA:
Similar to old ticket that was abandoned:
https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/2723

We only support landscape in our app by setting `Supported interface orientations (iPad)` in our `info.plist` file.

1. Version of Mobile SDK Used: 7.3.0
2. Issue found in Native App or Hybrid App: Native
3. OS Version: iOS 12.4.1
4. Device: iPad Air 2
5. Steps to reproduce: Unknown. Seems to occur randomly on SF login screen. One report reads as follows:
> opened the app (horizontally), SF SSO tried to go vertically, I turned the iPad vertically and then horizontaly, app crashed (no message)
6. Actual behavior: Crash with no message
7. Expected Behavior: Don't crash or try to present vertical login screen
8. Error Log:

> Last Exception BackTrace: UIApplicationInvalidInterfaceOrientation: Supported orientations has no common orientation with the application, and [UIAlertController shouldAutorotate] is returning YES

> 0	CoreFoundation	0x000000020061527c	__exceptionPreprocess
> 1	libobjc.A.dylib	0x00000001ff7ef9f4	objc_exception_throw
> 2	CoreFoundation	0x000000020051f4ac	+[NSException raise:format:]
> 3	UIKitCore	0x000000022cd0210c	-[UIViewController __supportedInterfaceOrientations]
> 	4	UIKitCore	0x000000022cd02668	-[UIViewController __withSupportedInterfaceOrientation:apply:]
> 5	UIKitCore	0x000000022cd02e8c	-[UIViewController setInterfaceOrientation:]
> 6	UIKitCore	0x000000022cceda88	-[UIViewController viewDidMoveToWindow:shouldAppearOrDisappear:]
> 7	UIKitCore	0x000000022d71e340	-[UIView(Internal) _didMoveFromWindow:toWindow:]
> 8	UIKitCore	0x000000022d711b20	__45-[UIView(Hierarchy) _postMovedFromSuperview:]_block_invoke
> 9	Foundation	0x0000000200f8faf0	-[NSISEngine withBehaviors:performModifications:]
> 10	UIKitCore	0x000000022d711a04	-[UIView(Hierarchy) _postMovedFromSuperview:]
[...]

PR #2503, opened 2018-04-24
Solves window related status bar & rotation issues

trooper2013:
@bhariharan based on our offline discussions about issues related to multiple windows, I'm finally  changing the window manager implementation to move away from the current strategy of adjusting transparency of windows. This fix involves moving away from the current approach of adjusting transparency of windows, to a model where the windows will show and hide instead.   Side effect  - window transitions may not be as smooth.

PR #3063, opened 2019-10-23
Avoid returning restrictive supportedInterfaceOrientation that causes crashes.

MichaelWTA:
This will avoid crashes when the parent application doesn't support the given orientation.
Should solve this issue:
https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/3060

And is related to this common issue described on StackOverflow:
https://stackoverflow.com/questions/32169216/ios-error-supported-orientations-has-no-common-orientation-with-the-application

Let me know whether I should be concerned with the failing tests, or if this doesn't fit the desired code style.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -247,7 +247,7 @@
 }
 
 - (UIInterfaceOrientationMask)supportedInterfaceOrientations {
-    return UIInterfaceOrientationMaskPortrait;
+    return UIInterfaceOrientationMaskAll;
 }
 
 - (void)layoutWebView {
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/AppLockView/SFSDKAppLockViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/AppLockView/SFSDKAppLockViewController.m
@@ -74,7 +74,7 @@
 
 - (UIInterfaceOrientationMask)supportedInterfaceOrientations
 {
-    return UIInterfaceOrientationMaskPortrait;
+    return UIInterfaceOrientationMaskAll;
 }
 
 - (void)setupNavBar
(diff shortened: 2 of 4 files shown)


trooper2013:
@MichaelWTA Could you provide a description of why this fix is required. We do not officially support all orientations. You can extend the SFLoginViewController to change the behavior.

MichaelWTA:
@trooper2013 could you point me towards some documentation stating that you don't support all orientations? 

Do you support landscape orientations? If so, this is necessary to avoid crashing the apps that use your SDK.

bhariharan:
@MichaelWTA As @trooper2013 mentioned, this is by design. Any UI screens within the SDK, such as the login view controller are designed to work only in portrait mode.

MichaelWTA:
I understand.
It would be great if this was documented clearly somewhere so that users of your SDK don't end up with random crashes in their apps.

MichaelWTA:
Hey, just pinging to see if there's any response to this.

Specifically the questions of: 
> Is the portrait only functionality documented anywhere, or is it just meant to be discovered? This is dangerous since this causes crashes on landscape only apps.

and 

> Is there a reason for only supporting portrait? It seems like it'd be pretty simple to also support landscape, especially since the proposed solution is to just subclass the view and make it support landscape.

wmathurin:
1) I don't think it's documented.

2) Portrait only came with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2503. But following the subsequent https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2523, rotation should work fine. 

We will look into it.

MichaelWTA:
#2 doesn't seem to be the case, the issue still exists and the given PR didn't solve it. The issue is wherever the code reads 
```
-(UIInterfaceOrientationMask)supportedInterfaceOrientations {
    return UIInterfaceOrientationMaskPortrait;
}
```
which is still does in a number of places.

MichaelWTA:
@wmathurin any updates or progress on this?

wmathurin:
We are looking into it for our next release (8.0) scheduled for the end of the year.

bhariharan:
We will evaluate this in a future release.

PR #3066, opened 2019-10-26
Migrate encrypted push notification handling

bbirman:
- Migrated encrypted push notification code from the internal shared repo. Didn’t make any logic changes, updated naming like SFS -> SFSDK
- Added manager flag to use encrypted push notifications

I kept the same header visibility as the originals but I couldn’t find any usage of PushNotificationFieldsConstants or PushNotificationError in other apps so I’m not sure if we should keep those two public or not— let me know if they should be changed


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationEncryption.m
```diff
@@ -0,0 +1,40 @@
...
+#import <SalesforceSDKCore/SFSDKCryptoUtils.h>
+#import "SFSDKPushNotificationEncryption.h"
+#import "SFSDKPushNotificationEncryptionConstants.h"
+
+@implementation SFSDKPushNotificationEncryption
+
++ (NSString *)getRSAPublicKey {
+    NSString *rsaPublicKey = [SFSDKCryptoUtils getRSAPublicKeyStringWithName:kPNEncryptionKeyName keyLength:kPNEncryptionKeyLength];
+    if (rsaPublicKey == nil) {
+        [SFSDKCryptoUtils createRSAKeyPairWithName:kPNEncryptionKeyName keyLength:kPNEncryptionKeyLength accessibleAttribute:kSecAttrAccessibleAfterFirstUnlock];
```
bbirman: Changed this from deprecated `kSecAttrAccessibleAlways`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.h
```diff
@@ -45,6 +45,10 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nullable, nonatomic, strong) NSDictionary* customPushRegistrationBody;
 
+/** Use this flag to indicate if the notifications should be encrypted. NO by default.
```
bhariharan: Add a comment here that this flag should always be set to `NO` for `Apex push notifications`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
```diff
@@ -39,18 +40,6 @@
 #pragma clang diagnostic push
```
bhariharan: Do we still need these pragmas that ignore warnings? Can we remove them now?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
```diff
@@ -184,6 +151,15 @@ - (BOOL)registerSalesforceNotificationsWithCompletionBlock:(void (^)(void))compl
         [bodyDict addEntriesFromDictionary: _customPushRegistrationBody];
     }
     
+    if (self.encryptionEnabled) {
+        NSString *rsaPublicKey = [SFSDKPushNotificationEncryption getRSAPublicKey];
+        if (!rsaPublicKey) {
+            [SFSDKCoreLogger e:[self class] format:@"Cannot register for notifications with Salesforce: no RSA key for encrypted notifications"];
+            return NO;
```
bhariharan: This is interesting. If we can't get a key, should we fallback on unencrypted push? 🤔

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationDecryption.h
```diff
@@ -0,0 +1,36 @@
...
+*/
+
+#import <Foundation/Foundation.h>
+#import <UserNotifications/UserNotifications.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SFSDKPushNotificationDecryption : NSObject
+
++ (BOOL)decryptNotificationContent:(nonnull UNMutableNotificationContent *)notificationContent error:(NSError * _Nullable * _Nullable)error;
```
bhariharan: How would the notification extension in the app get the decrypted content?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationEncryption.h
```diff
@@ -0,0 +1,35 @@
...
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#import <Foundation/Foundation.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface SFSDKPushNotificationEncryption : NSObject
```
bhariharan: Could we remove this from the public header footprint? It's only required for registration, which is completely internal to `Mobile SDK` at this point. Apps should never need to call this.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationEncryption.m
```diff
@@ -0,0 +1,40 @@
...
+WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#import <SalesforceSDKCore/SFSDKCryptoUtils.h>
+#import "SFSDKPushNotificationEncryption.h"
+#import "SFSDKPushNotificationEncryptionConstants.h"
+
+@implementation SFSDKPushNotificationEncryption
+
++ (NSString *)getRSAPublicKey {
```
bhariharan: I suggest moving this method inside the class that performs the registration. Make it a private method inside that class. Apps don't need to call this since the SDK handles registration.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationEncryptionConstants.h
```diff
@@ -0,0 +1,31 @@
...
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
+
+#ifndef SFSDKPushNotificationEncryptionConstants_h
+#define SFSDKPushNotificationEncryptionConstants_h
+
+static NSString * const kPNEncryptionKeyName = @"com.salesforce.mobilesdk.notificationKey";
```
bhariharan: Remove from public header footprint.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
wmathurin: Under what circumstances can getRSAPublicKey fail?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationEncryption.m
wmathurin: Should the app decide the accessible attribute to use? In other words, should `getRSAPublicKey` take a `accessibleAttribute` parameter?

wmathurin: It will probably be flagged by security if we use `kSecAttrAccessibleAlways`. See https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/2827

bhariharan: I don't know about that. Devs might unknowingly change this to some more restrictive value and PNs will stop working suddenly. We see GitHub issues being raised for keychain issues due to incorrect values quite frequently. We might want to keep this internal to the SDK.

wmathurin: Then I would use `kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly`.

bhariharan: Agreed. Sounds good 👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
bbirman: - I was on the fence about it, my thought was that if an app required encrypted push we shouldn't fall over to something less secure without them knowing
- I don't know what could cause failure, just that the methods this eventually calls into for creation (`SecKeyCreateRandomKey`, `SecKeyCopyPublicKey`) and retrieval are nullable

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationEncryptionConstants.h
bbirman: This header isn't public, but I could move it to `SFPushNotificationManager` along with the `getRSAPublicKey` method if you want to get rid of the file

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationDecryption.h
bbirman: It modifies the given content, I also just realized this is missing a doc comment, will add!

PR #3076, opened 2019-11-01
Enables Adv. Auth for iOS13 & 12, adv auth nav bug & rest client property access

trooper2013:
We do not need the preprocessor checks for min version of iOS.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
@@ -38,11 +38,9 @@ public enum RestClientError: Error {
 }
 
 public struct RestResponse {
-    private static let emptyJsonDictionaryArrayResponse = [String: Any]()
-    private static let emptyJsonArrayResponse = [[String: Any]]()
     private static let emptyStringResponse = ""
     private (set) var data: Data
-    private (set) var urlResponse: URLResponse
+    public private (set) var urlResponse: URLResponse
     
     /// Initializes the RestResponse with a Data object and URLResponse
     /// - Parameter data: Response as raw Data
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthRootController.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthRootController.h
@@ -31,10 +31,6 @@
  */
 
 #import "SFSDKRootController.h"
-
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= 130000
+#import <AuthenticationServices/AuthenticationServices.h>
 @interface SFSDKAuthRootController : SFSDKRootController <ASWebAuthenticationPresentationContextProviding>
-#else
-@interface SFSDKAuthRootController : SFSDKRootController
-#endif
 @end
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthRootController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthRootController.m
@@ -26,17 +26,9 @@ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH 
 */
 
 #import "SFSDKAuthRootController.h"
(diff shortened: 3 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthRootController.h
```diff
@@ -31,10 +31,6 @@
  */
 
 #import "SFSDKRootController.h"
-
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= 130000
+#import <AuthenticationServices/AuthenticationServices.h>
 @interface SFSDKAuthRootController : SFSDKRootController <ASWebAuthenticationPresentationContextProviding>
```
wmathurin: Does this build on iOS 12?
Before the if def, I had a cast in the @available link but no reference to iOS 13 classes otherwise.

trooper2013: Yes I was able to build it for for iOS 12.

wmathurin: And no runtime issue?

trooper2013: Compiles and runs in both.

wmathurin: How come the compiler does not complain? ASWebAuthenticationPresentationContextProviding is only available on iOS 13.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -38,11 +38,9 @@ public enum RestClientError: Error {
 }
 
 public struct RestResponse {
-    private static let emptyJsonDictionaryArrayResponse = [String: Any]()
-    private static let emptyJsonArrayResponse = [[String: Any]]()
     private static let emptyStringResponse = ""
     private (set) var data: Data
```
qingqingliu: @trooper2013 , we need to the same for the data property above to expose the data with public private(set)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthRootController.m
```diff
@@ -26,17 +26,9 @@
...
 
 #import "SFSDKAuthRootController.h"
-#import <AuthenticationServices/AuthenticationServices.h>
 
 @implementation SFSDKAuthRootController
-
-/**
-  Implementating  ASWebAuthenticationPresentationContextProviding so that ASWebAuthenticationSession can be started from the window
- */
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= 130000
```
qingqingliu: @trooper2013 , we are using deployment target to check? not the actual OS version? Is that actually the proper way to to do this?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
```diff
@@ -1861,14 +1861,12 @@ - (void)presentLoginView:(SFSDKAuthViewHolder *)viewHandler {
             }];
         }
         else {
-#if __IPHONE_OS_VERSION_MIN_ALLOWED >= 130000
```
qingqingliu: @trooper2013, thanks for making this change. I am not sure why if using __IPHONE_OS_VERSION_MIN_ALLOWED >= 130000 to do conditional logic based on deployment target is the right thing to do in general, should we do an inventory to see if we should remove from other places?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
trooper2013: there's an api already for that. asData()

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m
trooper2013: @qingqingliu These are the only special cases.

PR #3079, opened 2019-11-06
Leveraging new batch and composite classes

wmathurin:


wmathurin:
With the last few commits, all MobileSync tests are passing again. Let's see if they do on CI...

wmathurin:
CI passed !

PR #3083, opened 2019-11-07
Set swift version in podspec.

brandonpage:


wmathurin:
Should we revert the change that added a swift version to the pure objective c template?

brandonpage:
@bhariharan @wmathurin Even with this change you will get a compiler error if you do not specify the swift version in the project files for core, store and sync.

wmathurin:
What about the change to the template app's project file: https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/230/commits/d09121fe0bf46295099d01de4d93e5d9cf14c117

Is it still needed?

PR #3093, opened 2019-11-21
Sync up can now provided a external id field name

wmathurin:
Locally created records providing an external id are upserted during sync up.

* SyncUpTarget and BatchSyncUpTarget both support the new field
* Added test for SyncUpTarget and BatchSyncUpTarget
* Using the new field in one of the example config and modified config test accordingly

NB:
* ParentChildrenSyncUpTarget does not support it (yet)

### DIFF
--- a/libs/MobileSync/MobileSync/Classes/Target/SFBatchSyncUpTarget.m
+++ b/libs/MobileSync/MobileSync/Classes/Target/SFBatchSyncUpTarget.m
@@ -4,7 +4,7 @@
  Redistribution and use of this software in source and binary forms, with or without modification,
  are permitted provided that the following conditions are met:
  * Redistributions of source code must retain the above copyright notice, this list of conditions
- and the following disclaimer.
+and the following disclaimer.
  * Redistributions in binary form must reproduce the above copyright notice, this list of
  conditions and the following disclaimer in the documentation and/or other materials provided
  with the distribution.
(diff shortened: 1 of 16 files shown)


nzsrini:
Thanks for this field.
Logic fails 
when the application doesn't receive the response from Salesforce when the network is unavailable but the record is already created at S.F end.

The application tries to update when the network is available again using external field then application receive a response with statuscode= 200 (Sucess).
///
In parseIdsFromResponses method, code is checking only status code==201.(Created)
 when we receive status 200 parseIdsFromResponses is returning null.

When it returns null it is not updated to local store.

====================================
+ (NSDictionary<NSString*, NSString*> *)parseIdsFromResponses:(NSArray<SFSDKCompositeResponse*>*)responses {
    NSMutableDictionary *refIdToId = [NSMutableDictionary new];
    for (SFSDKCompositeSubResponse* response in responses) {
        if (response.httpStatusCode == 201 ) {
            NSString *serverId = response.body[kCreatedId];
            refIdToId[response.referenceId] = serverId;
        }
    }
    return refIdToId;
}
-======
--- From Salesforce --
Example JSON response
In API version 46.0 and later, the HTTP status code is 200 (OK) and the successful response is:
{
    "id" : "001D000000Kv3g5IAB",
    "success" : true,
    "errors" : [ ],
    "created": false
}

PR #3096, opened 2019-12-03
Use named pasteboard when external paste disabled

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -742,6 +745,7 @@ - (void)handleAuthCompleted:(NSNotification *)notification
     // Will set up the passcode timer for auth that occurs out of band from SDK Manager launch.
     [SFSecurityLockout setupTimer];
     [SFSecurityLockout startActivityMonitoring];
+    [self swizzlePasteboardIfNeeded];
```
trooper2013: @bbirman Isn't this linked only to Debug Mode?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -775,6 +779,10 @@ - (void)handlePostLogout
     [SFSecurityLockout cancelPasscodeScreen];
     [SFSecurityLockout stopActivityMonitoring];
     [SFSecurityLockout removeTimer];
+    if (pasteboardSwizzled) {
+        // Reset to original implementations
+        [self swizzlePasteboard];
```
trooper2013: @bbirman Why do we need to swizzle here?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -789,6 +797,7 @@ - (void)handleUserDidSwitch:(SFUserAccount *)fromUser toUser:(SFUserAccount *)to
 {
     [SFSecurityLockout setupTimer];
     [SFSecurityLockout startActivityMonitoring];
+    [self swizzlePasteboardIfNeeded];
```
trooper2013: Same question. Disabling cut&paste is a managed preference for app. So why do we need to check here again?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -870,6 +878,22 @@ - (void)clearClipboard
...
 
+- (void)swizzlePasteboardIfNeeded {
+    BOOL externalPasteDisabled = [SFManagedPreferences sharedPreferences].shouldDisableExternalPasteDefinedByConnectedApp;
+    if ((externalPasteDisabled && !pasteboardSwizzled) || (!externalPasteDisabled && pasteboardSwizzled)) {
+        [self swizzlePasteboard];
+    }
+}
+
+- (void)swizzlePasteboard {
+    method_exchangeImplementations(class_getClassMethod([UIPasteboard class], @selector(generalPasteboard)), class_getInstanceMethod([SalesforceSDKManager class], @selector(sdkPasteboard)));
```
trooper2013: This swizzle should be run only once in the app. I think we should swizzle for sure and decide which pasteboard to use in the swizzled method. i.e. we should check the managed preference within the swizzled method. What do u think?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
bbirman: It's configured on the connected app so it could change between users on different orgs, this would check if it needs to be changed to the named pasteboard or to the general one depending on the current state and the setting

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
bbirman: So we'd keep the general pasteboard's implementation, swizzle with a new pasteboard method and then that method would either return the sdk's named pasteboard or the original general pasteboard?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -142,6 +139,12 @@ + (void)initialize {
         // For dev support
         method_exchangeImplementations(class_getInstanceMethod([UIWindow class], @selector(motionEnded:withEvent:)), class_getInstanceMethod([UIWindow class], @selector(sfsdk_motionEnded:withEvent:)));
 
+        // Pasteboard
```
trooper2013: @bbirman why are we doing this here in the sfsdk_motionEnded?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -139,6 +139,12 @@ + (void)initialize {
...
```
bcriscuolo: @bbirman I believe that swizzling guidance is to call during `+load`. Summary here: https://nshipster.com/method-swizzling/ (while not the definitive answer, nshipster has been a good summary reference).

`+initialize` _should_ work within the context of the SalesforceSDKManager (as it is not a category) but it feels like moving to recommended patterns may be better, esp at this core level.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -185,6 +191,14 @@ + (instancetype)sharedManager {
     return sdkManager;
 }
 
++ (UIPasteboard *)sdkPasteboard {
```
bcriscuolo: 👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
bcriscuolo: @bbirman any thoughts on the above?

bbirman: It’s updated in #3099

Issue #3111, opened 2019-12-17
Crash in Unit Tests on sdkPasteboard


AnthonyMillerSF:
1. Version of Mobile SDK Used: 8.0.0
2. Issue found in Native App or Hybrid App: Native App
3. OS Version: 13.2
4. Device: iPhone 8 (Simulator)
5. Steps to reproduce:
  - Unit tests with an app target
  - In a test create a view controller with a `UISearchBar`
  -  Display view controller:
```swift
let window = UIWindow.init(frame: UIScreen.main.bounds)
window.rootViewController = controller
window.makeKeyAndVisible()
RunLoop.main.run(until: Date())
```

6. Actual behavior:
Unit tests crash with EXC_BAD_INSTRUCTION during a `dispatch_once` in `[SFUserAccountManager sharedInstance]`
7. Expected Behavior:
Should not crash
8. Error Log:
```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x00007fff51a0b170 in _dispatch_once_wait.cold.1 ()
#1	0x00007fff519def44 in _dispatch_once_wait ()
#2	0x0000000104c73d89 in _dispatch_once [inlined] at /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator13.2.sdk/usr/include/dispatch/once.h:84
#3	0x0000000104c73d6e in +[SFUserAccountManager sharedInstance] at /Users/anthony.miller/Documents/repos/REDACTED/Pods/SalesforceSDKCore/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFUserAccountManager.m:153
#4	0x0000000104c2ff4c in +[SFSDKEventBuilderHelper createAndStoreEvent:userAccount:className:attributes:] at /Users/anthony.miller/Documents/repos/REDACTED/Pods/SalesforceSDKCore/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKEventBuilderHelper.m:43
#5	0x0000000104bdf120 in -[SFManagedPreferences storeAnalyticsEvent] at /Users/anthony.miller/Documents/repos/REDACTED/Pods/SalesforceSDKCore/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m:165
[...]

PR #3112, opened 2019-12-17
Fix deadlock on [SFUserAccountManager sharedInstance]

AnthonyMDev:
When the `[SFUserAccountManager sharedInstance]` is created and sends the `SFUserAccountManagerDidFinishUserInitNotification`, if any observer attempts to access the `[SFUserAccountManager sharedInstance]` when the notification is posted, a deadlock is created.

This is because the `dispatch_once` block for posting the notification is still being executed when the `dispatch_once` token is accessed again during the call to access the `[SFUserAccountManager sharedInstance]`.

More information on why this is a problem can be found here:
https://stackoverflow.com/questions/19176219/why-am-i-getting-deadlock-with-dispatch-once

This PR fixes this bug, identified in https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/3111


AnthonyMDev:
Not sure why these tests are failing. :/ I can't get the tests to run locally since I don't have test_credentials. Not sure how to get a proper test credential.

AnthonyMillerSF:
I am a Salesforce employee and I have joined the org on GitHub. So CLA should be clear.

PR #3113, opened 2019-12-27
Handle invalid values for instrumentation event json.

jwutke-sf:
The SDK can crash if the consumer passes in any invalid keys or values for `NSJSONSerialization`. The Apple docs here suggest checking before calling `dataWithJSONObject`. 
https://developer.apple.com/documentation/foundation/nsjsonserialization/1413636-datawithjsonobject?language=objc#discussion
> If obj will not produce valid JSON, an exception is thrown. This exception is thrown prior to parsing and represents a programming error, not an internal error. You should check whether the input will produce valid JSON before calling this method by using isValidJSONObject:.

This change will ignore events that are invalid instead of crashing.

### DIFF
--- a/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEvent.m
+++ b/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEvent.m
@@ -183,28 +183,7 @@
 #pragma mark - Public methods
 
 - (NSData *) jsonRepresentation {
-    NSMutableDictionary *dict = [[NSMutableDictionary alloc] init];
-    dict[kEventIdKey] = self.eventId;
-    dict[kStartTimeKey] = [NSNumber numberWithInteger:self.startTime];
-    dict[kEndTimeKey] = [NSNumber numberWithInteger:self.endTime];
-    dict[kNameKey] = self.name;
-    dict[kAttributesKey] = self.attributes;
-    dict[kSessionIdKey] = self.sessionId;
-    dict[kSequenceIdKey] = [NSNumber numberWithInteger:self.sequenceId];
-    dict[kSenderIdKey] = self.senderId;
-    dict[kSenderContextKey] = self.senderContext;
-    dict[kSchemaTypeKey] = [self stringValueOfSchemaType:self.schemaType];
-    dict[kEventTypeKey] = [self stringValueOfEventType:self.eventType];
-    dict[kErrorTypeKey] = [self stringValueOfErrorType:self.errorType];
-    if (self.deviceAppAttributes) {
-        dict[kDeviceAppAttributesKey] = [self.deviceAppAttributes jsonRepresentation];
-    }
-    dict[kConnectionTypeKey] = self.connectionType;
-    dict[kSenderParentIdKey] = self.senderParentId;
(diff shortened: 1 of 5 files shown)


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEvent.m
```diff
@@ -205,9 +205,14 @@ - (NSData *) jsonRepresentation {
...
-    NSError *error;
-    NSData *jsonData = [NSJSONSerialization dataWithJSONObject:dict options:0 error:&error];
-    return jsonData;
+    
+    if ([NSJSONSerialization isValidJSONObject:dict]) {
+        NSError *error;
+        NSData *jsonData = [NSJSONSerialization dataWithJSONObject:dict options:0 error:&error];
+        return jsonData;
+    } else {
+        return nil;
```
jwutke-sf: Should anything be logged in this case?

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Store/SFSDKEventStoreManager.m
```diff
@@ -75,13 +75,14 @@ - (void) storeEvent:(SFSDKInstrumentationEvent *) event {
...
     SFSDKInstrumentationEvent *eventCopy = [event copy];
-    if (!eventCopy || ![eventCopy jsonRepresentation]) {
+    NSData *eventJsonData = [eventCopy jsonRepresentation];
+    if (!eventCopy || !eventJsonData) {
         return;
     }
     if (![self shouldStoreEvent]) {
         return;
     }
-    NSData *encryptedData = self.dataEncryptorBlock([eventCopy jsonRepresentation]);
```
jwutke-sf: `[eventCopy jsonRepresentation]` was being called twice within this method, so moved it to a local variable `eventJsonData`.

bhariharan:
@jwutke-sf How are you producing invalid events? This should not be possible, because the library is designed to use [SFSDKInstrumentationEventBuilder](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEventBuilder.m#L42) to build events. We have validation built into [buildEvent](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/master/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEventBuilder.m#L58) that should prevent an invalid event from being formed. Since the validation is built in at build time, there should be no additional validation required while parsing/storing/reading the event - this is by design.

libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Store/SFSDKEventStoreManager.m
```diff
@@ -75,7 +75,7 @@ - (void) storeEvent:(SFSDKInstrumentationEvent *) event {
 
     // Copies event, to isolate data for I/O.
     SFSDKInstrumentationEvent *eventCopy = [event copy];
-    if (!eventCopy || ![eventCopy jsonRepresentation]) {
```
jwutke-sf: `jsonRepresentation` is nonnull, so we don't need to check it here.

PR #3115, opened 2020-01-02
Hitting back button on LoginViewController should automatically cancel the current auth request in flight

bhariharan:
Consumers shouldn't have to handle this since the login view controller is within the SDK.


bhariharan:
All these tests passed locally.

PR #3121, opened 2020-01-10
Add nada cache option

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -398,11 +399,10 @@ + (NSString *)launchActionsStringRepresentation:(SFSDKLaunchAction)launchActions
     return launchActionString;
 }
 - (void)setEncryptURLCache:(BOOL)encryptURLCache {
```
bhariharan: So, this method will go away in `9.0` based on the deprecation of the property?

PR #3129, opened 2020-01-24
Fix tests.

brandonpage:


libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientPublisherTests.swift
```diff
@@ -26,7 +26,6 @@ import XCTest
 import Combine
 @testable import SalesforceSDKCore
 
-@available(iOS 13.0, watchOS 6.0, *)
```
brandonpage: This annotation does not work for XCTest; the tests still run on iOS 12 and crash.

bbirman:
🎉 🎉 🎉

PR #3130, opened 2020-01-25
Don't follow redirects for authenticated requests

bbirman:


bhariharan:
It's possible, but the old behavior takes the user to a dead end with no recovery path. In that sense, that is the problem/bug. For endpoints that redirect to a different endpoint, `iOS` automatically recreates the request and sends it to the new endpoint. However, it appears to strip the authorization header while doing so. This results in the user getting back an unauthorized response with nowhere to go after.

With this change, the user will now get a `302` back with a response object that gives him/her the URL that they need to redirect to. This gives the user the ability to essentially replay the REST request with the new URL endpoint and the authorization header.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SalesforceRestAPITests.m
```diff
@@ -2465,6 +2465,22 @@ - (SFOAuthCredentials *)getTestCredentialsWithDomain:(nonnull NSString *)domain
...
+    SFNativeRestRequestListener *contactListener = [self sendSyncRequest:contactRequest];
+    NSString *contactId = ((NSDictionary *)contactListener.dataResponse)[LID];
+
+    // Authenticated request for contact image, shouldn't automatically redirect
+    NSString *path = [NSString stringWithFormat:@"/services/images/photo/%@", contactId];
+    SFRestRequest *request = [SFRestRequest requestWithMethod:SFRestMethodGET path:path queryParams:nil];
+    request.endpoint = @"";
+    SFNativeRestRequestListener *listener = [self sendSyncRequest:request];
+    NSInteger statusCode = [(NSHTTPURLResponse *)listener.rawResponse statusCode];
+    XCTAssertEqual(statusCode, 302, @"Request did not return 302");
```
bhariharan: Let's add a line here to read the URL that it's going to and assert it's not `nil`.

PR #3137, opened 2020-01-31
Fix some analyzer issues

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -135,7 +135,7 @@ - (instancetype)initWithIdentifier:(NSString *)theIdentifier clientId:(NSString*
 }
 
 - (instancetype)initWithIdentifier:(NSString *)theIdentifier clientId:(NSString *)theClientId encrypted:(BOOL)encrypted storageType:(SFOAuthCredentialsStorageType)type {
-    Class targetClass = self.class;
```
bbirman: Initialized value will always be overwritten by switch below

libs/SmartStore/SmartStore/Classes/SFStoreCursor.h
```diff
@@ -82,13 +82,13 @@ NS_SWIFT_NAME(StoreCursor)
  * Run query and resturn JSON serialized representation of the cursor.
  * @return JSON serialized representation of this object.
  */
-- (NSString*)getDataSerialized:(SFSmartStore*)store error:(NSError**)error;
+- (nullable NSString*)getDataSerialized:(SFSmartStore*)store error:(NSError**)error;
```
bbirman: Both of these methods have cases that return nil

brandonpage: @bbirman Isn't this technically a breaking change?  It's a good change because it forces the dev to handle this case which could avoid a crash but still.

bbirman: I thought we've changed nullability before in .x releases but I'm good either way

bhariharan: Leave it in. It's the correct value now anyway.

brandonpage: Agreed, just a gentle reminder for the future.

PR #3155, opened 2020-02-26
Adds Reusable, Generic pipeline

codefriar:
Includes a re-usable struct representing the JSON structure of a Salesforce Rest Response that includes an array of Salesforce Records.

The output of this pipeline is an array of swift objects hydrated from the Salesforce records.
The Salesforce records are decoded against the inferred type from the pipelines subscriber.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
@@ -179,4 +179,47 @@ extension RestClient {
             }
         }
     }
+  
+    /// Struct represents the JSON Structure of a Salesforce Response.
+    /// This struct requires a Model Object that conforms to Decodable
+    /// This model object's properties need to match the Salesforce Schema
+    ///   at least in part.
+    struct SFRestResponseWrapper<Record: Decodable>: Decodable {
+        var totalSize: Int
+        var done: Bool
+        var records: [Record]
+    }
+  
+    /// This method provides a reusuable, generic Combine pipeline for retrieving records
+    ///   from Salesforce. It relys on Swift Generics, and type inference to determine what
+    ///  models to create.
+    ///
+    /// Given a model object - Contact, you can use this pipeline like this:
+    /// contactsForCancellable = RestClient.shared.records(forRequest: request)
+    ///   .receive(on: RunLoop.main)
+    ///   .assign(to: \.contacts, on: self)
+    ///
+    /// This pipeline infers it's return type from the variable in the assign subscriber.
+    func records<Record: Decodable>(forRequest request:RestRequest ) -> AnyPublisher<[Record], Never> {
+      return RestClient.shared.publisher(for: request)
+        .tryMap({ (response) -> Data in
+          response.asData()
+        })
+        .decode(type: SFRestResponseWrapper<Record>.self, decoder: JSONDecoder())
+        .map({ (record) -> [Record] in
+          record.records
+        })
+        .catch({ _ in
+          Just([Record]())
+        })
+        .eraseToAnyPublisher()
+    }
+  
+    /// Reusable, generic Combine Pipeline returning an array of records of a local
+    /// model object that conforms to Decodable. This method accepts a query string and defers
+    /// to the method above.
+    func records<Record: Decodable>(forQuery query:String) -> AnyPublisher<[Record], Never> {
+      let request = RestClient.shared.request(forQuery: query, apiVersion: RestClient.apiVersion)
+      return self.records(forRequest: request)
+    }
 }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
@@ -178,7 +178,7 @@ extension RestClient {
     /// This method relies on the passed parameter ofModelType to infer the generic Record's
     /// concrete type.
     func fetchRecords<Record: Decodable>(ofModelType: Record.Type, forRequest request: RestRequest,
-                                       _ completionBlock: @escaping (Result<[Record], RestClientError>) -> Void) {
+                                       _ completionBlock: @escaping (Result<QueryResponse<Record>, RestClientError>) -> Void) {
       guard request.isQueryRequest else { return }
       RestClient.shared.send(request: request) { result in
           switch result {
@@ -186,9 +186,9 @@ extension RestClient {
                 do {
                   let decoder = JSONDecoder()
                   let wrapper = try decoder.decode(QueryResponse<Record>.self, from: response.asData())
-                  completionBlock(.success(wrapper.records))
+                  completionBlock(.success(wrapper))
                 } catch {
-                  completionBlock(.success([Record]()))
+                  completionBlock(.success(QueryResponse<Record>(totalSize: 0, done: true, records: [])))
               }
               case .failure(let err):
                   completionBlock(.failure(err))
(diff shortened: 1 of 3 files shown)


codefriar:
@trooper2013  -Just so we're on the same page, you're requesting three things:

1. A rename to the Wrapper Struct to QueryResponseWrapper
2. A more-restrictive type check on the incoming RestRequest to QueryRequest
3. A backwards compatible (iOS < 13.0) version that does not rely on Combine? 

I'll get started on those. please let me know if I missed anything.

codefriar:
@trooper2013 - I've updated this PR with non combine versions of the Records convenience api. There are tests failing with at this time, but they are seemingly unrelated to any of the changes here in. Not sure how to re-run the ci pipeline to see if it's an intermittent failure.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -87,7 +87,30 @@ public struct RestResponse {
...
 }
 
+extension RestRequest {
+  
+  /// Calculated property to determine if this request is a data retrieval request with a SOQL query.
+  /// All such queries will return a JSON decodable QueryResponseWrapper.
+  /// Implied contract is that all requests matching both properties here will be decodable via QueryResponseWrapper<Record>
+  public var isQueryRequest: Bool {
+    get {
+      return self.method == .GET && self.path.lowercased().hasSuffix("query")
```
trooper2013: @codefriar Isn't search supported?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -136,6 +159,70 @@ extension RestClient {
...
+    /// concrete type.
+    func fetchRecords<Record: Decodable>(ofModelType: Record.Type, forRequest request: RestRequest,
+                                       _ completionBlock: @escaping (Result<[Record], RestClientError>) -> Void) {
+      guard request.isQueryRequest else { return }
+      RestClient.shared.send(request: request) { result in
+          switch result {
+              case .success(let response):
+                do {
+                  let decoder = JSONDecoder()
+                  let wrapper = try decoder.decode(QueryResponseWrapper<Record>.self, from: response.asData())
```
trooper2013: @codefriar IMO we should return up the Query Response envelope instead. For e.g. in the case of pagination it would be important to know the total count.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -87,7 +87,30 @@ public struct RestResponse {
...
+  
+}
+
 extension RestClient {
+  
+    /// Struct represents the JSON Structure of a Salesforce Response.
+    /// This struct requires a Model Object that conforms to Decodable
+    /// This model object's properties need to match the Salesforce Schema
+    ///   at least in part.
+    struct QueryResponseWrapper<Record: Decodable>: Decodable {
```
trooper2013: @codefriar IMO Should we just rename this to QueryResponse?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -136,6 +159,70 @@ extension RestClient {
...
+      guard request.isQueryRequest else { return }
+      RestClient.shared.send(request: request) { result in
+          switch result {
+              case .success(let response):
+                do {
+                  let decoder = JSONDecoder()
+                  let wrapper = try decoder.decode(QueryResponseWrapper<Record>.self, from: response.asData())
+                  completionBlock(.success(wrapper.records))
+                } catch {
+                  completionBlock(.success([Record]()))
```
trooper2013: @codefriar I think we should either rethrow the error or not catch the decoder issues.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -136,6 +159,70 @@ extension RestClient {
...
+    ///       case .success(let records):
+    ///           do something with your array of model objects
+    ///       case .failure(let error):
+    ///           print(error)
+    ///       }
+    ///     }
+    ///
+    /// This method relies on the passed parameter ofModelType to infer the generic Record's
+    /// concrete type.
+    func fetchRecords<Record: Decodable>(ofModelType: Record.Type,
```
trooper2013: @codefriar Same as above. We should return up QueryResponse.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -179,4 +266,43 @@ extension RestClient {
...
+      }
+      return RestClient.shared.publisher(for: request)
+        .tryMap({ (response) -> Data in
+          response.asData()
+        })
+        .decode(type: QueryResponseWrapper<Record>.self, decoder: JSONDecoder())
+        .map({ (record) -> [Record] in
+          record.records
+        })
+        .catch({ _ in
```
trooper2013: @codefriar Similar comment about swallowing the error.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -179,4 +266,43 @@ extension RestClient {
...
+          Just([Record]())
+        })
+        .eraseToAnyPublisher()
+    }
+  
+    /// Reusable, generic Combine Pipeline returning an array of records of a local
+    /// model object that conforms to Decodable. This method accepts a query string and defers
+    /// to records<Record:Decodable>(forRequest request: RestRequest) -> AnyPublisher<[Record], Never>
+    func records<Record: Decodable>(forQuery query:String, withApiVersion version: String = SFRestDefaultAPIVersion ) -> AnyPublisher<[Record], Never> {
```
trooper2013: @codefriar Return QueryResponse instead?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: Search should work as well. Hand't thought about it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: renamed.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: @trooper2013  - I'm not sure I follow. You want to skip hydrating local models, and just return the response wrapper?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: Since the underlying Send Method doesn't throw, I can't throw here. I'm now returning an empty wrapper object.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: updated.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: This is now returning an empty query response as well

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: Yep, taken care of.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
codefriar: thats not true anymore. I just refactored so the pipeline is defined AnyPublisher<QueryResponse<Record>, RestClientError> so now'll throw the error up the stack.

PR #3160, opened 2020-03-09
Ensure that auth failures are processed on main thread

trooper2013:
Removes an old deprecated method for error handling. Auth flows typically have a some level of user interactions during failures.  Ensures that failure handling is always  dispatched on main (prior to this in some cases UI handling was being guarded and the other cases (adv. auth) was not). Addl.  removes  unnecessary dispatch_async for alert view presentation, the underlying mechanism dispatches on main thread anyway. cc: @qingqingliu

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
@@ -129,17 +129,6 @@ typedef void (^SFOAuthBrowserFlowCallbackBlock)(BOOL);
 - (void)oauthCoordinatorDidAuthenticate:(SFOAuthCoordinator *)coordinator authInfo:(SFOAuthInfo *)info;
 
 /**
- Sent if authentication fails due to an error. Note: This method is deprecated.  You should use the
- `oauthCoordinator:didFailWithError:authInfo` method instead.
- 
- @param coordinator The SFOAuthCoordinator instance processing this message
- @param error       The error message
- 
- @see SFOAuthCoordinator
- */
-- (void)oauthCoordinator:(SFOAuthCoordinator *)coordinator didFailWithError:(NSError *)error __attribute__((deprecated));
-
-/**
  Sent if authentication fails due to an error.
  
  @param coordinator The SFOAuthCoordinator instance processing this message.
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -291,12 +291,9 @@ - (void)notifyDelegateOfFailure:(NSError*)error authInfo:(SFOAuthInfo *)info
     self.authenticating = NO;
     self.advancedAuthState = SFOAuthAdvancedAuthStateNotStarted;
     if ([self.delegate respondsToSelector:@selector(oauthCoordinator:didFailWithError:authInfo:)]) {
-        [self.delegate oauthCoordinator:self didFailWithError:error authInfo:info];
-    } else if ([self.delegate respondsToSelector:@selector(oauthCoordinator:didFailWithError:)]) {
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated-declarations"
-        [self.delegate oauthCoordinator:self didFailWithError:error];
-#pragma clang diagnostic pop
+        dispatch_async(dispatch_get_main_queue(), ^{
```
qingqingliu: so the key fix is here. Thanks @trooper2013

qingqingliu: @mdarnall

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
```diff
@@ -128,17 +128,6 @@ typedef void (^SFOAuthBrowserFlowCallbackBlock)(BOOL);
...
-/**
- Sent if authentication fails due to an error. Note: This method is deprecated.  You should use the
- `oauthCoordinator:didFailWithError:authInfo` method instead.
- 
- @param coordinator The SFOAuthCoordinator instance processing this message
- @param error       The error message
- 
- @see SFOAuthCoordinator
- */
-- (void)oauthCoordinator:(SFOAuthCoordinator *)coordinator didFailWithError:(NSError *)error __attribute__((deprecated));
```
bhariharan: Shouldn't we wait for `Mobile SDK 9.0` to remove this, to avoid breaking semver?

trooper2013: @bhariharan This has been deprecated several versions ago. It should have been removed in 8.0.

bhariharan: Sounds good.

PR #3183, opened 2020-04-11
apply ios13 default pagesheet for login host list view

TLiu2121999:
Reverting change https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3052

We are ready to apply ios13 default modal view presentation style to pagesheet

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -339,11 +339,7 @@
 
 - (void)showHostListView {
     SFSDKNavigationController *navController = [[SFSDKNavigationController alloc] initWithRootViewController:self.loginHostListViewController];
-    if (@available(iOS 13.0, *)) {
-        navController.modalPresentationStyle = UIModalPresentationFullScreen;
-    } else {
-       navController.modalPresentationStyle = UIModalPresentationPageSheet;
-    }
+    navController.modalPresentationStyle = UIModalPresentationPageSheet;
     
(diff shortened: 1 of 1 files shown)


TLiu2121999:
@wmathurin

wmathurin:
@trooper2013 @bhariharan Do we want to change the default behavior?

TLiu2121999:
@wmathurin Just want to be clear, this change is the default behavior. The previous pr https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3052 was actually overwriting the default behavior. @bhariharan

wmathurin:
iOS 13 changed the way it renders UIModalPresentationPageSheet. They appear as a card (not full screen) that can be dismissed with a swipe. The previous pr https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3052 was meant to keep the UX the same. 
Are we ready to change the user experience? Is it okay if the user can swipe the screen away?

bhariharan:
@wmathurin Thanks for explaining. In that case, I don't think we should put this in. The screen should not be swipeable.

TLiu2121999:
ok, but just FYI, we are changing all other modal pages on Salesforce App to the default pageSheet, I've confirmed with our PM, she thinks it's ok to make this page swipe-able; But if msdk feel strongly about this, we are ok with either way. @bbirman @wmathurin

bhariharan:
I think non-swipeable would be my preferred option, but I'll leave it up to @wmathurin and @trooper2013 to decide.

TLiu2121999:
@wmathurin @trooper2013 what's the final decision here?

wmathurin:
(Let me 180 - I was too quick on my earlier review), I think it's fine for the **host list screen** to be swipe-able.

PR #3189, opened 2020-04-25
Make dateTime fields decodable as Date

woxtu:
Hi,

As the title suggests this PR makes dateTime fields decodable as `Date`.

```swift
struct Response: Decodable {
    struct Record: Decodable {
        struct Attributes: Decodable {
            let type: String
            let url: String
        }

        let attributes: Attributes
        let Id: String
        let CreatedDate: Date // Decode as `Date` instead of `String`
    }

    let totalSize: Int
    let done: Bool
    let records: [Record]
}

try response.asDecodable(type: Response.self)
```

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
@@ -77,7 +77,11 @@ public struct RestResponse {
     /// Decode the response as  a codable.
     /// - Parameter type: The type to use for decoding.
     public func asDecodable<T: Decodable>(type: T.Type) throws -> Decodable? {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ss.SSSZZZZZ"
+        
         let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .formatted(dateFormatter)
         do {
             let object = try decoder.decode(type, from: data)
             return object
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -77,7 +77,11 @@ public struct RestResponse {
     /// Decode the response as  a codable.
     /// - Parameter type: The type to use for decoding.
     public func asDecodable<T: Decodable>(type: T.Type) throws -> Decodable? {
+        let dateFormatter = DateFormatter()
```
trooper2013, line 80: @woxtu We should account for the fact that date format could occur in a  different format through query or maybe org locale prefs?  
e.g.```SELECT Id, LastModifiedDate, FORMAT(LastModifiedDate) formattedDate FROM Account ``` .   In fact my preference would be to pass in the JSONDecoder() itself. We should perhaps add something like  ```asDecodable(type: T.Type,  decoder: JSONDecoder)``` and this should let developers control decoding requirements not just for date but for other types as well.

trooper2013: In fact I think doing this is probably sufficient? 
```public func asDecodable<T: Decodable>(type: T.Type, decoder: JSONDecoder? = JSONDecoder()) throws -> T {```

woxtu: @trooper2013 Thank you for pointing that out. It seems better to pass a decoder, so I'll create a new PR later.

PR #3199, opened 2020-05-15
Fix layoutcomponents is an array of dictionaries

elvarliiv:
Thank you for fixing


bhariharan:
As I had mentioned in the previous pull request, we are not looking to add `Picklist` support currently (not in the way it's currently structured in this PR, anyway). Perhaps you meant to send this PR to your fork of the SDK?

Issue #3211, opened 2020-06-03
UserId vs. FullUserId (Android and iOS Parity)


luonglegs:
### Please fill out the following details:
This issue was brought up in the Salesforce Android SDK repo but I was told I should create a ticket for yall to address this. I believe the TL;DR is that iOS might need to standardize the 18 char on `userId`.

All of the discussions and details can be seen here:
https://github.com/forcedotcom/SalesforceMobileSDK-Android/issues/2071

1. Version of Mobile SDK Used: iOS
2. Issue found in Native App or Hybrid App: Native
3. OS Version: Latest
4. Device: iPhone X
5. Steps to reproduce:
6. Actual behavior:
7. Expected Behavior:
8. Error Log: N/A

bhariharan:
Thanks @luonglegs! We will take a look.

az-oolloow:
How did you get to `currentUser.userId` which returns 15 chars for you?

We get current user's id like this:
`UserAccountManager.shared.currentUserAccount?.idData?.userId` and this ALWAYS returns 18 chars for us.

bbirman:
This should be fixed by #3214 and will be in the 8.2 release

PR #3214, opened 2020-06-08
Standardize on 18 character user ID

bbirman:
For #3211

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager+Internal.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager+Internal.h
@@ -1,0 +1,42 @@
+//
+//  SFDirectoryManager+Internal.h
+//  SalesforceSDKCore
+//
+//  Created by Brianna Birman on 6/5/20.
+//  Copyright (c) 2020-present, salesforce.com, inc. All rights reserved.
+// 
+//  Redistribution and use of this software in source and binary forms, with or without modification,
+//  are permitted provided that the following conditions are met:
+//  * Redistributions of source code must retain the above copyright notice, this list of conditions
+//  and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright notice, this list of
+//  conditions and the following disclaimer in the documentation and/or other materials provided
+//  with the distribution.
+//  * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+//  endorse or promote products derived from this software without specific prior written
+//  permission of salesforce.com, inc.
+// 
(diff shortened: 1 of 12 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager+Internal.h
```diff
@@ -0,0 +1,42 @@
...
+static NSString * const kOrgPrefix = @"00D";
+
+// Prefix of a user ID
+static NSString * const kUserPrefix = @"005";
+
+@interface SFDirectoryManager ()
+
+// Starting in SDK 8.2, 18 character IDs are used instead of 15 character IDs.
+// This renames 15 character user ID directories to 18 characters.
++ (void)upgradeUserDirectories;
```
bhariharan, line 40: Add a `TODO:` here to remove this in `Mobile SDK 10.0`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.m
```diff
@@ -235,4 +237,45 @@ - (void)migrateFiles {
     [sharedDefaults synchronize];
 }
 
++ (void)upgradeUserDirectories {
+    [SFDirectoryManager upgradeUserDirectory:NSLibraryDirectory];
+    [SFDirectoryManager upgradeUserDirectory:NSDocumentDirectory];
+}
+
++ (void)upgradeUserDirectory:(NSSearchPathDirectory)type {
```
wmathurin, line 245: Do we have files on iOS with orgid_userid  in the name (like we do on Android)?

wmathurin, line 245: Could we add a test that exercise this method?

bhariharan: Oh, I should have mentioned this as well - we need upgrade steps for the Analytics data as well. The Analytics library takes a path from this library, so you'll need migration steps at both layers. I can talk to you offline about it.

bbirman: I only saw directories with the structure orgId/userId, though realized that I missed profile picture's userId (will add something for that). Is there anywhere else I should check for orgid_userid? Verified that the log files aren't dependent on userId and analytics's event_stores are covered by updating the userId directory in Documents. 
Screen Shot 2020-06-08 at 1 54 02 PM

bhariharan: Awesome! I can't think of anything else.

PR #3219, opened 2020-06-15
Enable objective-c compatibility header for swift

bbirman:
Skipped common and analytics because they don't have any swift so they don't have a swift section in build settings


shared/common/build/tools/generate_headers.sh
```diff
@@ -127,6 +127,9 @@ for file in $(find $wrapper_directory -type f); do
     if [[ $filename == $wrapper_filename ]]; then
         continue;
     fi
+    if [[ $filename == "$OPT_NAME-Swift.h" ]]; then
```
bbirman, line 130: Swift header shouldn’t be included in the umbrella header, the swift header imports the umbrella header

bhariharan: Circular dependency avoidance! Could you add a comment here, explaining that?

PR #3222, opened 2020-06-20
Refactoring of REST API Library and Simplification of Callbacks

bhariharan:
We have too many delegate callbacks currently in error cases that are non-standard, and complicate our code paths. They're there for legacy reasons and have been strung along as we switched network libraries for backwards compatibility. I took at stab at standardizing and simplifying the callbacks and deprecating the old ways. This allows us to clean up this area in `Mobile SDK 9.0`. In addition to this, a major drawback with the current error callbacks is that none of them return the response in the callback. I have added this in the new callbacks.

This PR:
- adds 2 simple callbacks for `onSuccess` and `onFailure`.
- adds simplified `send` methods to use the new callbacks instead of the old ones.
- deprecates the old callbacks.
- deprecates the old `send` methods.
- refactored the tests to use the new callbacks wherever possible (this will need one more round of cleanup as part of the removal effort in `Mobile SDK 9.0`).
- fixed a couple of minor issues with `RestExplorer`.
- updated all internal usage of the old APIs to use the new ones (in `Swift` extensions and `MobileSync` as well).

Testing:
- `MobileSyncExplorer` is working fine with the changes.
- All tests in `SalesforceSDKCore` are passing locally with the changes.

### DIFF
--- a/libs/MobileSync/MobileSync/Classes/Target/SFSyncUpTarget.m
+++ b/libs/MobileSync/MobileSync/Classes/Target/SFSyncUpTarget.m
@@ -276,10 +276,10 @@ typedef void (^SFSyncUpRecordModDateBlock)(SFRecordModDate *remoteModDate);
        completionBlock:(SFSyncUpTargetCompleteBlock)completionBlock
              failBlock:(SFSyncUpTargetErrorBlock)failBlock
 {
-    [SFMobileSyncNetworkUtils sendRequestWithMobileSyncUserAgent:request failBlock:^(NSError *e, NSURLResponse *rawResponse) {
+    [SFMobileSyncNetworkUtils sendRequestWithMobileSyncUserAgent:request failureBlock:^(id response, NSError *e, NSURLResponse *rawResponse) {
         self.lastError = e.description;
         failBlock(e);
-    } completeBlock:^(NSDictionary* d, NSURLResponse *rawResponse) {
+    } successBlock:^(NSDictionary* d, NSURLResponse *rawResponse) {
         completionBlock(d);
     }];
 }
@@ -296,12 +296,11 @@ typedef void (^SFSyncUpRecordModDateBlock)(SFRecordModDate *remoteModDate);
                                    fieldList:self.modificationDateFieldName
                                   apiVersion:nil];
 
-    [SFMobileSyncNetworkUtils
-            sendRequestWithMobileSyncUserAgent:request
-                                    failBlock:^(NSError *e, NSURLResponse *rawResponse) {
+    [SFMobileSyncNetworkUtils sendRequestWithMobileSyncUserAgent:request
+                                    failureBlock:^(id response, NSError *e, NSURLResponse *rawResponse) {
                                         completeBlock([[SFRecordModDate alloc] initWithTimestamp:nil isDeleted:e.code == 404]);
                                     }
-                                completeBlock:^(id response, NSURLResponse *rawResponse) {
+                                successBlock:^(id response, NSURLResponse *rawResponse) {
                                     completeBlock([[SFRecordModDate alloc] initWithTimestamp:response[self.modificationDateFieldName] isDeleted:FALSE]);
                                 }
     ];
--- a/libs/MobileSync/MobileSync/Classes/Util/SFMobileSyncNetworkUtils.h
+++ b/libs/MobileSync/MobileSync/Classes/Util/SFMobileSyncNetworkUtils.h
@@ -41,7 +41,16 @@ NS_SWIFT_NAME(NetworkUtils)
  @param failBlock The block to call if the request fails.
  @param completeBlock The block to call if the request succeeds.
  */
-+ (void)sendRequestWithMobileSyncUserAgent:(SFRestRequest *)request failBlock:(SFRestFailBlock)failBlock completeBlock:(SFRestResponseBlock)completeBlock;
++ (void)sendRequestWithMobileSyncUserAgent:(SFRestRequest *)request failBlock:(SFRestFailBlock)failBlock completeBlock:(SFRestResponseBlock)completeBlock SFSDK_DEPRECATED("8.2", "9.0", "Will be removed in Mobile SDK 9.0, use sendRequestWithMobileSyncUserAgent:request:failureBlock:successBlock instead.");
+
+/**
+ * Sends a REST request, after applying the MobileSync user agent string.
+ *
+ * @param request The request to send.
+ * @param failureBlock The block to call if the request fails.
+ * @param successBlock The block to call if the request succeeds.
+ */
++ (void)sendRequestWithMobileSyncUserAgent:(SFRestRequest *)request failureBlock:(SFRestRequestFailBlock)failureBlock successBlock:(SFRestResponseBlock)successBlock;
 
 @end
 
--- a/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Util/SFJsonUtils.m
+++ b/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Util/SFJsonUtils.m
@@ -70,14 +70,14 @@ static NSError *sLastError = nil;
     return [SFJsonUtils JSONRepresentation:obj options:options];
 }
 
-+ (NSString*)JSONRepresentation:(id)obj options:(NSJSONWritingOptions)options {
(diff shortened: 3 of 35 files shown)


libs/MobileSync/MobileSync/Classes/Util/SFMobileSyncNetworkUtils.h
```diff
@@ -41,7 +41,16 @@ NS_SWIFT_NAME(NetworkUtils)
...
++ (void)sendRequestWithMobileSyncUserAgent:(SFRestRequest *)request failBlock:(SFRestFailBlock)failBlock completeBlock:(SFRestResponseBlock)completeBlock SFSDK_DEPRECATED("8.2", "9.0", "Will be removed in Mobile SDK 9.0, use sendRequestWithMobileSyncUserAgent:request:failureBlock:successBlock instead.");
+
+/**
+ * Sends a REST request, after applying the MobileSync user agent string.
+ *
+ * @param request The request to send.
+ * @param failureBlock The block to call if the request fails.
+ * @param successBlock The block to call if the request succeeds.
+ */
++ (void)sendRequestWithMobileSyncUserAgent:(SFRestRequest *)request failureBlock:(SFRestRequestFailBlock)failureBlock successBlock:(SFRestResponseBlock)successBlock;
```
bhariharan, line 53: New method with new blocks.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Util/SFJsonUtils.m
```diff
@@ -70,14 +70,14 @@ + (NSString*)JSONRepresentation:(id)obj {
...
 }
 
-+ (NSString*)JSONRepresentation:(id)obj options:(NSJSONWritingOptions)options {
++ (NSString *)JSONRepresentation:(id)obj options:(NSJSONWritingOptions)options {
     NSString *result = nil;
-    
-    NSData *jsonData = [self JSONDataRepresentation:obj options:options];
-    if (nil != jsonData) {
-          result = [[NSString alloc] initWithData:jsonData encoding:NSUTF8StringEncoding];
+    if (obj) {
```
bhariharan, line 75: We need this check for some cases where `obj` could be `nil`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Instrumentation/SFRestAPI+Instrumentation.m
```diff
@@ -136,20 +168,33 @@ + (void)enableInstrumentation {
...
     os_log_t logger = self.class.oslog;
     os_signpost_id_t sid = sf_os_signpost_id_generate(logger);
     sf_os_signpost_interval_begin(logger, sid, "Send", "Method:%ld path:%{public}@", (long)request.method, request.path);
     id<SFRestDelegate> delegateWrapper = [SFRestDelegateWrapperWithInstrumentation wrapperWith:delegate signpost:sid logger:logger];
     request.instrDelegateInternal = delegateWrapper;
     return [self instr_send:request delegate:delegateWrapper];
- 
 }
 
+- (void)instrumentation_send:(SFRestRequest *)request requestDelegate:(id<SFRestRequestDelegate>)requestDelegate {
```
bhariharan, line 189: Signposting support for the new callbacks.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.h
```diff
@@ -53,88 +53,10 @@ extern NSString* const kSFRestDefaultAPIVersion NS_SWIFT_NAME(SFRestDefaultAPIVe
...
- Sending a request is done using `send:delegate:`.
- The class sending the request has to conform to the protocol `SFRestDelegate`.
- 
- A request can be obtained in two different ways:
-
- - by calling the appropriate `requestFor[...]` method
-
- - by building the `SFRestRequest` manually
- 
- Note: If you opt to build an SFRestRequest manually, you should be aware that
```
bhariharan, line 78: All this stuff was outdated, and it's in our dev guide anyway.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.h
```diff
@@ -80,6 +79,38 @@ extern NSString * const kSFDefaultRestEndpoint;
 // Forward declaration.
 @class SFRestRequest;
 
+/**
+ * Lifecycle events for REST requests.
+ */
+NS_SWIFT_NAME(RestRequestDelegate)
```
bhariharan, line 85: This is the new delegate - has a `didSucceed` and `didFail`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -136,10 +136,10 @@ extension RestClient {
     /// - See   [Composite](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_composite.htm).
     public func send(compositeRequest: CompositeRequest, _ completionBlock: @escaping (Result<CompositeResponse, RestClientError>) -> Void) {
         compositeRequest.parseResponse = false
-        __sendCompositeRESTRequest(compositeRequest, fail: { (error, urlResponse) in
+        __send(compositeRequest, failureBlock: { (response, error, urlResponse) in
```
wmathurin, line 139: Should the RestClientError also capture the response?

libs/MobileSync/MobileSync/Classes/Target/SFSyncUpTarget.m
```diff
@@ -276,10 +276,10 @@ - (void)sendRequest:(SFRestRequest*)request
        completionBlock:(SFSyncUpTargetCompleteBlock)completionBlock
              failBlock:(SFSyncUpTargetErrorBlock)failBlock
 {
-    [SFMobileSyncNetworkUtils sendRequestWithMobileSyncUserAgent:request failBlock:^(NSError *e, NSURLResponse *rawResponse) {
+    [SFMobileSyncNetworkUtils sendRequestWithMobileSyncUserAgent:request failureBlock:^(id response, NSError *e, NSURLResponse *rawResponse) {
         self.lastError = e.description;
```
wmathurin, line 280: lastError should get the body of the response (if available) instead of e.description.
That would address issues like https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/issues/162

bhariharan: Would that be considered a breaking change for this point release, even though it doesn't break public API? Should we do that in `9.0`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
bhariharan: Wouldn't the caller have access to both?

wmathurin: If you use RestClient.swift, the completion block in case of error only gets Result.failure(apiError).

libs/MobileSync/MobileSync/Classes/Target/SFSyncUpTarget.m
wmathurin: May be.
Although, the shape for lastError has never been well defined (the body of the response is in there but mixed up in the error message) and can changed depending of the NSError we build upstream.
Devs complained about the fact that is cannot be easily parsed. To extract any useful information, one has to do "string contains" type of logic - that logic might still work if we provide the clean body.

wmathurin: That change could/should be done in a separate dedicated PR.

bhariharan: True - @bbirman is going to look into this.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
bhariharan: Yes, I just realized. I focused on the `Objective C` side of things ;-)

PR #3224, opened 2020-06-24
Add key value store

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/KeyValueEncryptedFileStore.swift
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    /// - Returns: A Boolean value that indicates whether the store is empty.
+    @objc public func isEmpty() -> Bool {
+        return count() == 0
+    }
+
+    // MARK: Private
+    private static func storesDirectory(forUser user: UserAccount) -> String? {
+        return SFDirectoryManager.shared().directory(forUser: user, type: .libraryDirectory, components: [keyValueStoresDirectory])
```
bbirman, line 339: Did library to match sapp, not sure if we want to make it customizable at the shared instance level

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/KeyValueEncryptedFileStore.swift
```diff
@@ -0,0 +1,395 @@
+//
```
wmathurin, line 1: Could you put the file in a subdirectory?
E.g. `libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage` instead of `libs/SalesforceSDKCore/SalesforceSDKCore/Classes`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/KeyValueEncryptedFileStore.swift
```diff
@@ -0,0 +1,395 @@
...
+            try FileManager.default.removeItem(at: fileURL)
+            return true
+        } catch {
+            SFSDKCoreLogger.e(KeyValueEncryptedFileStore.self, message: "\(#function): Error removing file at path '\(fileURL.path)': \(error)")
+            return false
+        }
+    }
+
+    /// Removes all contents of the store.
+    @objc public func removeAll() {
```
bhariharan, line 318: We should call this when the user logs out, to ensure that all data is wiped when logout occurs.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/KeyValueEncryptedFileStore.swift
bhariharan: Let's line this up with the directory we use for `SmartStore`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/KeyValueEncryptedFileStore.swift
```diff
@@ -0,0 +1,395 @@
...
+        return count() == 0
+    }
+
+    // MARK: Private
+    private static func storesDirectory(forUser user: UserAccount) -> String? {
+        return SFDirectoryManager.shared().directory(forUser: user, type: .libraryDirectory, components: [keyValueStoresDirectory])
+    }
+
+    private static func globalStoresDirectory() -> String {
+        return SFDirectoryManager.shared().globalDirectory(ofType: .libraryDirectory, components: [keyValueStoresDirectory])
```
bhariharan, line 343: Same comment as above.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/KeyValueEncryptedFileStore.swift
```diff
@@ -0,0 +1,395 @@
+//
+//  KeyValueEncryptedFileStore.swift
```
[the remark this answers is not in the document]
wmathurin: KeyValueEncryptedFileStore is the name we have on Android currently.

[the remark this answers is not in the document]
bhariharan: Can we rename it there as well, since the release hasn't gone out yet? ;-)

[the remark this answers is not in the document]
bhariharan: Well, maybe it's not a big deal. We can leave it as-is.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -39,6 +39,7 @@
 #import "SFSDKNullURLCache.h"
 #import "UIColor+SFColors.h"
 #import "SFDirectoryManager+Internal.h"
+#import <SalesforceSDKCore/SalesforceSDKCore-Swift.h>
```
bhariharan, line 42: Could we use `#import "SalesforceSDKCore-Swift.h"` since this is within the same library?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -0,0 +1,398 @@
...
+    /// - Parameter name: Name of the store.
+    /// - Parameter user: The user account associated with the store.
+    @objc(sharedStoreWithName:user:)
+    public static func shared(withName name: String, forUserAccount user: UserAccount) -> KeyValueEncryptedFileStore? {
+        let userKey = self.userKey(forUser: user)
+        if userStores[userKey] == nil {
+            userStores[userKey] = SafeMutableDictionary<NSString, KeyValueEncryptedFileStore>()
+        }
+
+        if let store = userStores[userKey]?[name as NSString] {
```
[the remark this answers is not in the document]
bbirman: Since the store is being returned right after, I think guard would just move the return statement further down?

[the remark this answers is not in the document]
trooper2013: @bbirman intention was to clean up the number of return statements. But I see its not that much different. Its ok. We can leave it the way it is laid out currently

trooper2013:
@bbirman Do we plan on adding any combine extensions?

bbirman:
@trooper2013 no plans that I know of, but we can make them :P

wmathurin:
On Android, kv store lives in the smartstore lib and the factory method are on SmartStoreSDKManager.
Should we keep both platforms consistent? Should we change iOS or Android?

wmathurin:
Also wondering if we should add keys related methods (hasKey, allKeys). hasKey would be easy but allKeys will take more work the file name is a one way hash of the key. It might require us to store the key in a file?

bhariharan:
Maybe we should move this code to `SmartStore`. `SmartStore` is our offline + storage library - that seems to be the logical home for this. It could be done in the next PR along with `hasKey` and `allKeys`.

wmathurin:
I think the reason for having it lower was to leverage it for the secure cache.

bhariharan:
Gotcha. Maybe we leave the respective platforms the way they are then, since `Android` has no URL cache that we can control.

Issue #3200, opened 2020-05-15
Layout components is an array of dictionaries


elvarliiv:
The following code crashes at the moment on:     
if let c = ii.layoutComponents {. 

I also submited a fix as a pull request

guard let s = layout?.sections else {return}
                    for ss in s {
                        if let r = ss.layoutRows {
                            for rr in r {
                                if let i = rr.layoutItems {
                                    for ii in i {
                                        if let c = ii.layoutComponents {. 
                                            for cc in c {
                                                print(cc)
                                            }
                                        }
                                    }
                                }
                            }
                        }
                    }

elvarliiv:
@siilats

bbirman:
@elvarliiv thanks for reporting, fixed here: #3230

PR #3230, opened 2020-06-29
Fix layout components type

bbirman:
For #3200


bbirman:
@wmathurin looks like Android has the same thing, working on it now

PR #3236, opened 2020-07-14
Better handling of bad or missing sobject type during sync up

wmathurin:
Fixes the infinite loop reported here: https://github.com/forcedotcom/SalesforceMobileSDK-ReactNative/issues/168

### DIFF
--- a/libs/MobileSync/MobileSyncTests/SyncUpTargetTests.m
+++ b/libs/MobileSync/MobileSyncTests/SyncUpTargetTests.m
@@ -109,6 +109,22 @@
 }
 
 /**
+ * Sync up records missing sobject type
+ */
+- (void) testSyncUpWithNoType {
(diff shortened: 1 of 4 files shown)


libs/MobileSync/MobileSyncTests/SyncUpTargetTests.m
```diff
@@ -108,6 +108,22 @@ -(void) testSyncUpWithErrors
     [idToFields addEntriesFromDictionary:idToFieldsGoodNames];
 }
 
+/**
+ * Sync up records missing sobject type
+ */
+- (void) testSyncUpWithNoType {
```
wmathurin, line 114: New tests:
- no type
- bad typ
Those tests run for both SyncUpTarget and BatchSyncUpTarget.
Before the code change, the test would repro the infinite looping.

wmathurin:
Re-ran the tests - they passed ;-)

PR #3237, opened 2020-07-15
Match response keys with Android

bbirman:
First step to match changes here https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2095

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSURLResponse+SFAdditions.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSURLResponse+SFAdditions.m
@@ -37,7 +37,8 @@
     responseDictionary[@"textEncodingName"] = self.textEncodingName;
     responseDictionary[@"URL"] = self.URL.absoluteString;
     if ([self isKindOfClass:[NSHTTPURLResponse class]]) {
-        responseDictionary[@"allHeaderFields"] = ((NSHTTPURLResponse *)self).allHeaderFields;
+        responseDictionary[@"headers"] = ((NSHTTPURLResponse *)self).allHeaderFields;
+        responseDictionary[@"statusCode"] = [NSNumber numberWithInteger:((NSHTTPURLResponse *)self).statusCode];
     }
     return responseDictionary;
 }


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/NSURLResponse+SFAdditions.m
```diff
@@ -37,7 +37,8 @@ - (NSDictionary *)asDictionary {
     responseDictionary[@"textEncodingName"] = self.textEncodingName;
     responseDictionary[@"URL"] = self.URL.absoluteString;
     if ([self isKindOfClass:[NSHTTPURLResponse class]]) {
-        responseDictionary[@"allHeaderFields"] = ((NSHTTPURLResponse *)self).allHeaderFields;
+        responseDictionary[@"headers"] = ((NSHTTPURLResponse *)self).allHeaderFields;
```
bbirman, line 40: I noticed that the header name capitalization is different, Android is "content-type" vs iOS "Content-Type", should we make the iOS header keys lowercase?

bhariharan: Nah - should be fine.

wmathurin: It's also Content-Type on [Android](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/rest/RestResponse.java#L52), as it [should be](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type), isn't it?

bbirman: Interesting -- when going through .getAllHeaders they're lowercase
image

wmathurin: They are supposed to be case insensitive according to https://stackoverflow.com/a/5259004/11461823
Looking at the implementation on okHttp, it's indeed ignoring the case when getting headers by name: https://github.com/square/okhttp/blob/master/okhttp/src/main/kotlin/okhttp3/Headers.kt#L102

PR #3252, opened 2020-08-10
Fix warning and add import

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -26,6 +26,7 @@
 //  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import Foundation
+import SalesforceSDKCommon
```
bbirman, line 29: I didn’t catch this before because it compiles, but still adding it to be safe

PR #3253, opened 2020-08-11
SFSmartStore deprecations

bbirman:
- Deprecate old upgrade steps for removal
- Deprecate NS_SWIFT_UNAVAILABLE constants that are for internal

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStoreUpgrade.h
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStoreUpgrade.h
@@ -24,12 +24,14 @@
 
 #import <Foundation/Foundation.h>
 #import <SalesforceSDKCore/SFUserAccount.h>
+#import <SalesforceSDKCore/SalesforceSDKConstants.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
 /**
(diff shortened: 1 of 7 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStoreUpgrade.h
```diff
@@ -24,12 +24,14 @@
...
 #import <Foundation/Foundation.h>
 #import <SalesforceSDKCore/SFUserAccount.h>
+#import <SalesforceSDKCore/SalesforceSDKConstants.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
 /**
  Used internally for upgrading SmartStore.
  */
+SFSDK_DEPRECATED(8.3, 9.0, "Will be removed.")
```
bbirman, line 34: The upgrade steps are from 2014 except for `updateEncryptionSalt` which is from 7.1, if there's any reason to keep that method around longer I'm thinking the class could still be deprecated and be internal

bbirman:
@wmathurin Yup! I'd also been thinking about making a new internal set of of strings to duplicate the deprecated ones to get around the warnings but was on the fence about a name change being worth it for one release

PR #3257, opened 2020-08-18
Latest SQLCipher Binary

bhariharan:
I did some project cleanup and modernization on the `SQLCipher` project in our fork, in anticipation of supporting `Catalyst`. Original `SQLCipher` PR on the fork is [here](https://github.com/forcedotcom/sqlcipher/pull/17).


sgoldberg-sfdc:
Anything we should be aware of with new SQLCipher?

bhariharan:
@sgoldberg-sfdc This is not an upgrade of `SQLCipher` - nothing has changed. It's basically the same binary.

PR #3263, opened 2020-08-27
Allowing app name to be overridable

bhariharan:
`Android` already has this functionality.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
@@ -137,13 +137,19 @@ NS_SWIFT_NAME(SalesforceManager)
  *
  * @return App name.
  */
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -137,13 +137,19 @@ NS_SWIFT_NAME(SalesforceManager)
...
+ * Gets & sets the app name being used by the SDK for user agent and other parts within the SDK.
+ *
+ * @return App name.
+ */
+@property (class, nonatomic, strong) NSString *appName NS_SWIFT_NAME(appName);
 
 /**
  @return The singleton instance of the SDK Manager.
  */
-@property(class,nonatomic,readonly)SalesforceSDKManager *sharedManager NS_SWIFT_NAME(shared);
```
bbirman, line 146: Why is readonly going away?

bhariharan: Oh, good catch. This was accidentally changed. I'll revert it.

PR #3281, opened 2020-10-01
Move SmartStore constants to internal header

bbirman:


libs/SmartStore/SmartStore/Classes/SFSmartStore.h
```diff
@@ -300,17 +258,6 @@ NS_SWIFT_NAME(SmartStore)
...
 
-/**
- Creates a new soup or confirms the existence of an existing soup.
- @warning Deprecated. Use registerSoup:withIndexSpecs:error: instead.
-
- @param soupName The name of the soup to register.
- @param indexSpecs Array of one or more SFSoupIndex objects.
- @return YES if the soup is registered or already exists.
- */
-- (BOOL)registerSoup:(NSString*)soupName withIndexSpecs:(NSArray<SFSoupIndex*>*)indexSpecs
```
bbirman, line 311: I didn't see a release for when this would be removed but it looks like it was deprecated a while ago so I thought it was safe to take out, let me know if that's not the case

bhariharan (approved):

wmathurin (approved):

PR #3284, opened 2020-10-11
Remove SalesforceSDKManager deprecations

bbirman:
Ran tests locally

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SalesforceSDKCoreDefines.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SalesforceSDKCoreDefines.h
@@ -38,40 +38,6 @@ enum {
     kSalesforceSDKManagerErrorInvalidLaunchParameters
 };
 
-// Launch actions taken
-typedef NS_OPTIONS(NSInteger, SFSDKLaunchAction) {
-    SFSDKLaunchActionNone                 = 0,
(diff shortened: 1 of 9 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SalesforceSDKCoreDefines.h
```diff
@@ -38,40 +38,6 @@ enum {
     kSalesforceSDKManagerErrorInvalidLaunchParameters
 };
 
-// Launch actions taken
-typedef NS_OPTIONS(NSInteger, SFSDKLaunchAction) {
```
bbirman, line 42: I missed these when adding deprecation markers, but everything that uses them was deprecated

native/SampleApps/MobileSyncExplorer/MobileSyncExplorer/Classes/AppDelegate.m
```diff
@@ -184,10 +184,4 @@ - (void)resetViewState:(void (^)(void))postResetBlock
     }
 }
 
-- (void)sdkManagerWillResignActive {
```
bbirman, line 187: I tried locking/unlocking with the keyboard up and snapshot enabled and didn't notice any issues without this, let me know if there's a more specific scenario I should test

bhariharan (approved):
Awesome! 👍

PR #3297, opened 2020-11-05
check nullability before adding to dictionary

liguopku:
@bhariharan @sgoldberg-sfdc
we run into an crash when account is null
Screen Shot 2020-11-05 at 2 55 01 PM

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
@@ -104,10 +104,12 @@
     if (self.completionBlock) {
         dispatch_async(dispatch_get_main_queue(), ^{
             SFUserAccount *account = [[SFUserAccountManager sharedInstance] accountForCredentials:self.credentials];
-            NSDictionary *userInfo = @{ kSFNotificationUserInfoAccountKey : account };
-            [[NSNotificationCenter defaultCenter] postNotificationName:kSFNotificationUserDidRefreshToken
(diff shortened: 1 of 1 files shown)


bhariharan:
@liguopku I'm not sure this is the right fix. If control comes to `completeWithSuccess`, the value of `account` should never be `nil`. How do you reproduce this? It seems to indicate an underlying issue. What do you think, @trooper2013?

trooper2013:
> @liguopku I'm not sure this is the right fix. If control comes to `completeWithSuccess`, the value of `account` should never be `nil`. How do you reproduce this? It seems to indicate an underlying issue. What do you think, @trooper2013?

I have seen this issue prop up in the past too. We should find out why the userAccount is nil. Adding a check may prevent a crash but it's not the right solution. Why would a userAccount not be found during the refresh? There could be an issue with the lookup of the user account, in which case we should find out why (maybe community handling related)?  Is it possible that a logout is in progress?

liguopku:
@bhariharan @trooper2013 Thanks for the feedback! I have put the cause of the logout and refresh flow into an internal quip document and we can discuss the details there. For long term goals, we will turn off refresh on 403 from the app by default. If certain cases where refresh is required, app side need to handle them by itself.

bhariharan:
@liguopku There are 3 issues here:

- Refresh token expires immediately.
- API access is disabled for the user, which causes a `403` response.
- Auto-refresh on `403`.

Since we don't refresh on `403` anymore, this situation should not arise going forward. We will independently augment our code to ensure that refresh flow doesn't occur when logout is in progress. I'm going to close this for now, based on an internal discussion @trooper2013, @bbirman, and I had.

PR #3307, opened 2020-11-20
Makes writes to user account file atomic.

trooper2013:
Write to a temp and rename. Apply file protections. In the worst case scenario NSFileProtection may not match the set defaults.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFDefaultUserAccountPersister.m
```diff
@@ -233,8 +218,10 @@ - (BOOL)saveUserAccount:(SFUserAccount *)userAccount toFile:(NSString *)filePath
                                      userInfo:@{ NSLocalizedDescriptionKey : reason } ];
         return NO;
     }
-
-    return YES;
+    //lets add the file protection now
+    NSString *fileProtections = [SFFileProtectionHelper fileProtectionForPath:filePath];
+    [[SFFileProtectionHelper sharedInstance] addProtection:fileProtections forPath:filePath];
+    return saveFileSuccess;
```
bhariharan, line 224: This will always be `YES` at this point.

bhariharan (approved):

PR #3308, opened 2020-11-24
Improved nullability notations

AsceticMonk:
Added missing nullability notations

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
@@ -94,7 +94,7 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (NSString*)directoryOfCurrentUserForType:(NSSearchPathDirectory)type components:(NSArray*)components;
+- (NSString*)directoryOfCurrentUserForType:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
 
 /** Returns the path to the global directory of the specified type. For example, NSCachesDirectory will
  return "Library/Caches/<bundleIdentifier>/"
@@ -102,7 +102,7 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (NSString*)globalDirectoryOfType:(NSSearchPathDirectory)type components:(NSArray*)components;
+- (NSString*)globalDirectoryOfType:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
 
 @end
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
@@ -94,7 +94,7 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (NSString*)directoryOfCurrentUserForType:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
+- (nullable NSString*)directoryOfCurrentUserForType:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
 
 /** Returns the path to the global directory of the specified type. For example, NSCachesDirectory will
  return "Library/Caches/<bundleIdentifier>/"
@@ -102,7 +102,7 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (NSString*)globalDirectoryOfType:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
+- (nullable NSString*)globalDirectoryOfType:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
 
 @end
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
@@ -79,7 +79,7 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (nullable NSString*)directoryForUser:(SFUserAccount *)user scope:(SFUserAccountScope)scope type:(NSSearchPathDirectory)type components:(nullable NSArray *)components;
+- (nullable NSString*)directoryForUser:(nullable SFUserAccount *)user scope:(SFUserAccountScope)scope type:(NSSearchPathDirectory)type components:(nullable NSArray *)components;
 
 /** Returns the path to the directory type for the specified user.
  @param account The user account to use. If nil, the path returned corresponds to the global path type
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
@@ -79,7 +79,7 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (nullable NSString*)directoryForUser:(nullable SFUserAccount *)user scope:(SFUserAccountScope)scope type:(NSSearchPathDirectory)type components:(nullable NSArray *)components;
+- (nullable NSString*)directoryForUser:(SFUserAccount *)user scope:(SFUserAccountScope)scope type:(NSSearchPathDirectory)type components:(nullable NSArray *)components;
 
 /** Returns the path to the directory type for the specified user.
  @param account The user account to use. If nil, the path returned corresponds to the global path type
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
```diff
@@ -89,8 +89,8 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) NSString *accessToken;
 @property (nonatomic, readonly) NSString *refreshToken;
 @property (nonatomic, readonly) NSDate *issuedAt;
-@property (nonatomic, readonly) NSURL *instanceUrl;
-@property (nonatomic, readonly) NSURL *identityUrl;
+@property (nonatomic, readonly, nullable) NSURL *instanceUrl;
```
bbirman, line 92: If I'm remembering right, these were intended to be nonnull but in that case we could fix the implementations so they don't have a nil return, @trooper2013 would know better :)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
```diff
@@ -94,15 +94,15 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (NSString*)directoryOfCurrentUserForType:(NSSearchPathDirectory)type components:(NSArray*)components;
+- (NSString*)directoryOfCurrentUserForType:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
```
bbirman, line 97: I think this one and the one below have nullable return strings too

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
AsceticMonk: Thanks @bbirman! @trooper2013 could you provide tips here? Should I keep this `nonnull` and fix the implementation?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
AsceticMonk: You are right! Fixed.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
```diff
@@ -79,7 +79,7 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (nullable NSString*)directoryForUser:(SFUserAccount *)user scope:(SFUserAccountScope)scope type:(NSSearchPathDirectory)type components:(nullable NSArray *)components;
+- (nullable NSString*)directoryForUser:(nullable SFUserAccount *)user scope:(SFUserAccountScope)scope type:(NSSearchPathDirectory)type components:(nullable NSArray *)components;
```
bhariharan, line 82: `user` should be `nonnull`, so should the directory returned. Are we making `user nullable` because the implementation falls back on `currentUser` if `nil` is passed in?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
bhariharan: Yes, this should be `nonnull`.

bhariharan (changes requested):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
AsceticMonk: @bhariharan If `instanceUrl` is designed to be `nonnull`, then we should fix the implementation:

```objc
- (NSURL *)instanceUrl {
    if (self.values[kSFOAuthInstanceUrl]) {
        return [NSURL URLWithString:self.values[kSFOAuthInstanceUrl]];
    }
    return nil;
}
```

Any tip on what should it return if `kSFOAuthInstanceUrl` is not found in `values` dictionary?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
AsceticMonk: Looking at the current implementation of this method, we can probably keep `user` parameter to be `nonnull`. However, if we want to annotate the directory returned to be `nonnull`, we would have to update the implementation, currently it does return `nil` in several logic branches:

```objc
if (nil == user.credentials.organizationId && scope != SFUserAccountScopeGlobal) {
        // do nothing
        return nil;
}

switch (scope) {
        case SFUserAccountScopeGlobal:
            return [self directoryForOrg:nil user:nil community:nil type:type components:components];
            
        case SFUserAccountScopeOrg:
            if (!user.credentials.organizationId) {
                [SFSDKCoreLogger w:[self class] format:@"Credentials missing for user"];
                return nil;
            }
            return [self directoryForOrg:user.credentials.organizationId user:nil community:nil type:type components:components];
            
        case SFUserAccountScopeUser:
            if (!user.credentials.organizationId || !user.credentials.userId) {
                [SFSDKCoreLogger w:[self class] format:@"Credentials missing for user"];
                return nil;
            }
            return [self directoryForOrg:user.credentials.organizationId user:user.credentials.userId community:nil type:type components:components];
            
        case SFUserAccountScopeCommunity:
            if (!user.credentials.organizationId || !user.credentials.userId) {
                [SFSDKCoreLogger w:[self class] format:@"Credentials missing for user"];
                return nil;
            }
            // Note: if the user communityId is nil, we use the default (internal) name for it.
            return [self directoryForOrg:user.credentials.organizationId user:user.credentials.userId community:user.credentials.communityId?:kDefaultCommunityName type:type components:components];
    }
```

bhariharan: - `user` should be `nonnull`.
- Ideally, we should remove all the `userId` and `organizationId` checks and that would make `directory` `nonnull` as well, at that point. There's no reason a `user` object would be empty without a proper `organizationId`  or `userId`.

Could you make those changes, @AsceticMonk? cc: @trooper2013 @bbirman

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
bhariharan: Remove the check, and directly return `[NSURL URLWithString:self.values[kSFOAuthInstanceUrl]]`. Instance URL should always be present.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
AsceticMonk: What about this check `scope != SFUserAccountScopeGlobal` at the beginning?

AsceticMonk: Also, all cases converge to the method `[self directoryForOrg:user:community:type:components:]`, and this method has a `nullable` return type, so it makes sense that the return value of this method is `nullable` too, no?

bhariharan: Alright, let's leave the return value as `nullable`, but `user` should be `nonnull` and those other checks can go.

AsceticMonk:
@bhariharan Take another look please, I've made changes based on your feedback.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.h
```diff
@@ -87,22 +87,22 @@ FOUNDATION_EXTERN NSString * const kDefaultCommunityName NS_SWIFT_NAME(SFDirecto
  @param components The additional path components to be added at the end of the directory (eg ['mybundle', 'common'])
  @return The path to the directory
  */
-- (nullable NSString*)directoryForUser:(nullable SFUserAccount*)account type:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
+- (nullable NSString*)directoryForUser:(SFUserAccount*)account type:(NSSearchPathDirectory)type components:(nullable NSArray*)components;
```
AsceticMonk, line 90: @bhariharan Based on the discussion, I assume this should be `nonnull` also?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFDirectoryManager.m
```diff
@@ -124,29 +124,18 @@ - (NSString*)directoryForUser:(SFUserAccount *)user scope:(SFUserAccountScope)sc
...
-                [SFSDKCoreLogger w:[self class] format:@"Credentials missing for user"];
-                return nil;
-            }
             return [self directoryForOrg:user.credentials.organizationId user:user.credentials.userId community:nil type:type components:components];
             
         case SFUserAccountScopeCommunity:
-            if (!user.credentials.organizationId || !user.credentials.userId) {
-                [SFSDKCoreLogger w:[self class] format:@"Credentials missing for user"];
-                return nil;
-            }
```
AsceticMonk, line 149: I removed these checks.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
```diff
@@ -90,7 +90,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) NSString *refreshToken;
 @property (nonatomic, readonly) NSDate *issuedAt;
 @property (nonatomic, readonly) NSURL *instanceUrl;
-@property (nonatomic, readonly) NSURL *identityUrl;
+@property (nonatomic, readonly, nullable) NSURL *identityUrl;
```
bhariharan, line 93: @AsceticMonk Last change - `identityUrl` is also `nonnull`. I don't think the service allows for a user to exist without a corresponding `identityUrl`.

bhariharan (changes requested):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
AsceticMonk: Done! Also updated the implementation.

bhariharan (approved):
Perfect! LGTM!

PR #3313, opened 2020-12-08
Fixing build issues with xcode 12

wmathurin:
According to https://github.com/facebook/react-native/issues/29633#issuecomment-694187116

One should always explicitly depend on the lib that provides the APIs you rely on, rather than relying on transitive dependencies—but it seems like in previous Xcode versions the build system worked differently and having the transitive dependencies around just happened to work.


bhariharan:
@wmathurin That's not true. Libraries are meant to use transitive dependencies, as long as the `semver` in their Podfiles are set up correctly. Those are the PRs I made a couple of months ago to make sure our Podfiles have their `semver` set up correctly, i.e. `8.3.x` will transitively only pull in `8.3.x`. Only apps need to specify dependencies directly, to avoid the `.x` changes. What doesn't work with `Xcode 12`? We're using `Xcode 12` in `SecuritySDK` and `Communities` with the transitive dependencies and it works fine.

wmathurin:
Right now the ReactNativeTemplate does not build in xcode 12 if I point to forcedotcom/SalesforceMobileSDK-iOS#dev (original pod specs) but compiles against wmathurin/SalesforceMobileSDK-iOS#dev (these changes).
Screen Shot 2020-12-08 at 11 24 55 AM

bhariharan:
@wmathurin Hmm I think it's because we're not using version specifiers [here](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/dev/MobileSyncExplorerReactNative/ios/Podfile#L45):
```
pod 'SalesforceSDKCommon', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
pod 'SalesforceAnalytics', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
pod 'SalesforceSDKCore', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
pod 'SmartStore', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
pod 'MobileSync', :path => '../mobile_sdk/SalesforceMobileSDK-iOS'
pod 'SalesforceReact', :path => '../node_modules/react-native-force'
```
Should we change the pods in the template to pull them from the Github repo directly, like the other templates?

bhariharan (approved):
This solution is fine too, @wmathurin. 👍🏻

PR #3317, opened 2020-12-09
Remove or internalize deprecated classes/methods for passcode

bbirman:
- Removes passcode based encryption key
- Moves some headers from public to internal
- Removes passcode provider options and uses PBKDF2
- Consolidates passcode management into SFSecurityLockout

Tested fresh install, user switching and upgrading from 8.3-- let me know if there's anything else I should check


bhariharan (approved):
Excellent! 👍🏻

PR #3325, opened 2021-01-14
Expand window manager to support multiple scenes

bbirman:
This is the first step to supporting multiple scenes, more PRs would come after this to use these APIs

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.h
@@ -87,7 +87,7 @@
  */
 @property (readonly,nonatomic,strong) SFSDKWindowContainer * _Nonnull snapshotWindow;
 
-/** SDK uses this window to present the passocde View.
+/** SDK uses this window to present the passcode View.
  */
 @property (readonly,nonatomic,strong) SFSDKWindowContainer * _Nonnull passcodeWindow;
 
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -48,85 +48,26 @@
  Stash(nullify) the rootviewcontroller when the window is presented and unstash(restore) when dismissed. Extended UIWindow (SFSDKUIWindow) to handle the stash and unstash.
  Windows are created lazily and the references are removed when the windows are dismissed.
  */
-@interface SFSDKUIWindow ()
```
bbirman, line 51: This is just getting moved within the file so that it can access internal SFSDKWindowManager methods

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.h
```diff
@@ -87,7 +87,7 @@
  */
 @property (readonly,nonatomic,strong) SFSDKWindowContainer * _Nonnull snapshotWindow;
 
-/** SDK uses this window to present the passocde View.
+/** SDK uses this window to present the passcode View.
  */
 @property (readonly,nonatomic,strong) SFSDKWindowContainer * _Nonnull passcodeWindow;
```
bbirman, line 92: @bhariharan we originally talked about getting rid of these window properties in favor of the methods that take the nullable scene. Since we're not supporting passcode for the scene case in 9.0, should I keep the passcode one and get rid of the others?

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKWindowManagerTests.m
```diff
@@ -25,6 +25,7 @@
 
 #import <XCTest/XCTest.h>
 #import "SFSDKWindowManager.h"
+#import "SFApplicationHelper.h"
 
 @interface SFSDKWindowManagerTests: XCTestCase{
```
bbirman, line 30: I'm still looking at ways to unit test with multiple scenes but for now added tests for the new APIs for single scenes and manually tested multiple scenes

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.h
bhariharan: I think we would still go with the original plan for now, i.e. getting rid of the passcode window property as well. We can always deprecate the new method we create if needed.

bhariharan (approved):
LGTM! Only comment is to go with our original plan of getting rid of the passcode property and replacing it with a method instead.

PR #3333, opened 2021-01-28
 Relocate deprecated methods

bbirman:


libs/SmartStore/SmartStore/Classes/NSData+SFAdditions.h
```diff
@@ -0,0 +1,37 @@
...
+//  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+/** Provides MD5 conversion support.
+ */
+@interface NSData (SFSHA)
+
+/**Derives  a 16-byte  hex encoded string .
+ @return 16-byte  hex encoded string.
+ */
+-(NSString *)digest;
```
wmathurin, line 35: Should this live in the SalesforceSDKCommon library instead?

bhariharan (approved):

libs/SmartStore/SmartStore/Classes/NSData+SFAdditions.h
bhariharan: This should be completely removed in `9.1`. NCC keeps flagging it because of `MD5` usage. The only reason we're making it internal for now is to give us time to ensure that we have all the required upgrade steps in place.

bbirman: It was originally in core but I moved it here because SmartStore couldn't access it after I took the public API out, but if there's a better way to handle that I can switch

wmathurin: I see. It makes sense to have it here because it needs to be internal.

wmathurin (approved):

bhariharan (approved):

PR #3344, opened 2021-03-04
Dev picklist

siilats:
Support picklists (similar to layouts)


bhariharan:
Thank you for your contribution. However, we will probably take a different approach or simply derive inspiration from this, if we decide to support this. Our status hasn't changed from the last time you sent a PR [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3191). I'm closing this for now.

PR #3347, opened 2021-03-08
Dev picklist

siilats:
Adding picklist support to IOS


bhariharan:
@siilats Please stop making these PRs against the upstream repo. As we've mentioned in your previous PRs, we're not looking to add picklist support at this time. Please add this to your fork of this repo if you need to. Thanks!

PR #3349, opened 2021-03-13
SmartSql now supports selecting or filtering by non-indexed field.

wmathurin:
Thanks to json1, we can query by or return non-indexed field (as long as the data is stored in the database).

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStore.m
@@ -1125,27 +1125,28 @@ NSUInteger CACHES_COUNT_LIMIT = 1024;
 }
 
 - (NSString*)columnNameForPath:(NSString*)path inSoup:(NSString*)soupName withDb:(FMDatabase*) db {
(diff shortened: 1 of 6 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStore.m
```diff
@@ -1125,27 +1125,28 @@ - (void)updateTable:(NSString*)tableName values:(NSDictionary*)map entryId:(NSNu
 }
 
 - (NSString*)columnNameForPath:(NSString*)path inSoup:(NSString*)soupName withDb:(FMDatabase*) db {
-    //TODO cache these with soupName:path ? if slow...
     NSString *result = nil;
-    if (nil == path) {
-        return result;
+    NSArray* indexSpecs = [self indicesForSoup:soupName withDb:db];
```
wmathurin, line 1129: Took care of the old TODO on the left by calling indicesForSoup (which does use a cache).

bhariharan (approved):

PR #3354, opened 2021-03-24
Minor improvements to KeyValueEncryptedFileStore

AsceticMonk:
**Changes**
- Use `isEmpty` instead of `count`
- Use `Data` initializer to convert `String` to `Data` to avoid dealing with optional
- Call super `setUp` and `tearDown` in test

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
@@ -34,7 +34,7 @@ public class KeyValueEncryptedFileStore: NSObject {
     @objc(storeName) public let name: String
     @objc public static let maxStoreNameLength = 96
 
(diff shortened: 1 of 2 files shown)


wmathurin (approved):

bbirman (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -61,7 +61,7 @@ public class KeyValueEncryptedFileStore: NSObject {
         self.encryptionKey = encryptionKey
     }
 
-    // MARK: Store management
+    // MARK: - Store management
```
bbirman, line 64: What does the dash change?

AsceticMonk: It adds line separator.
line-separator

PR #3355, opened 2021-03-26
Add new public method allKeys()

AsceticMonk:
**Summary**
Add a new public method `allKeys()` to retrieve all keys in the file-based key-value storage. The design of the technical solution is identical to the one in Android (refer to [this PR](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2168)).

**Change Details**
- Creates an additional file when saving a value, which contains the key. The name of this new file is `hash(key)_key`.
- The name of the file that contains the value is changed to `hash(key)_value`.
- A v2 store contains another new file to store version number. This file is simply named `version`, name is not hashed.
- Version file is never deleted, unless the store is removed.
- When removing a value, both key and value files are deleted.
- New public `allKeys()` added.
- Updated existing tests and added tests for new methods.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
@@ -270,7 +270,7 @@ public class KeyValueEncryptedFileStore: NSObject {
         }
     }
 
-    // MARK: - Storage operations
+    // MARK: - Store operations
 
     /// Returns whether the given store name is valid.
     /// - Parameter name: name of the store.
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -59,6 +89,13 @@ public class KeyValueEncryptedFileStore: NSObject {
         self.name = name
         self.directory = URL(fileURLWithPath: fullPath)
         self.encryptionKey = encryptionKey
+        super.init()
+        
+        if isNewlyCreated {
+            _ = writeFile(key: KeyValueEncryptedFileStore.storeVersionFileName,
```
bbirman, line 95: Maybe writeFile (and readFile) should have a discardable result?

bbirman (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
AsceticMonk: `writeFile` makes sense. Thanks!

bbirman: ugh, I meant to say "(and removeFile)", that would make more sense :P

AsceticMonk: Makes sense too! Updated.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -366,28 +413,92 @@ public class KeyValueEncryptedFileStore: NSObject {
...
             return try FileManager.default.contentsOfDirectory(atPath: directory)
         } catch {
-            SFSDKCoreLogger.e(KeyValueEncryptedFileStore.self, message: "\(function): Error getting contents of directory: \(error)")
+            SFSDKCoreLogger.e(KeyValueEncryptedFileStore.self, message: "\(callingFunction): Error getting contents of directory: \(error)")
             return []
         }
     }
+    
+    @discardableResult
+    private func writeFile(key: String, value: String, fileType: FileType, callingFunction: String = #function) -> Bool {
```
wmathurin, line 422: `writeFile`, `removeFile` take a `key`+`fileType` and compute the fileURL but `readFile` takes a `fileURL`.
For symmetry, could all three take a `fileURL` (and leave it to their caller to compute fileURL from key+fileType)?

AsceticMonk: Updated those methods, check out the changes.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -49,16 +68,38 @@ public class KeyValueEncryptedFileStore: NSObject {
...
+            if !versionFileCreated {
+                return nil
+            } else {
+                self.version = 2
+            }
+        } else {
+            if let version = readVersion(versionFileURL) {
+                self.version = version
+            } else {
+                return nil
```
wmathurin, line 100: :+1:
I need to update Android to also fail if version could not be read.

wmathurin (approved):

PR #3358, opened 2021-03-29
Minor revisions of `KeyValueEncryptedFileStore`

AsceticMonk:
**Changes**
- Simplify public access of store version
- Correct a typo in `allKeys()` method documentation

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
@@ -368,7 +368,7 @@ public class KeyValueEncryptedFileStore: NSObject {
     }
     
     /// All keys in the store
-    /// - Returns: all keys of stored values in a v2 store, nil if it's a v2 store
+    /// - Returns: all keys of stored values in a v2 store, nil if it's a v1 store
     @objc public func allKeys() -> [String]? {
         guard storeVersion >= 2 else {
             SFSDKCoreLogger.e(KeyValueEncryptedFileStore.self, message: "This store does not have this capability!")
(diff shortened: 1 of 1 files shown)


bhariharan (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -33,13 +33,13 @@ import SalesforceSDKCommon
 public class KeyValueEncryptedFileStore: NSObject {
     @objc(storeDirectory) public let directory: URL
     @objc(storeName) public let name: String
+    @objc public private(set) var storeVersion = 1
```
wmathurin, line 36: Shouldn't that default to 2?

AsceticMonk: It doesn't matter really, because the version needs to be determined during initialization. This default value is only assigned to work around the initialization requirement in Swift.

wmathurin: True, but still it could show the current version the code runs in.

AsceticMonk: Also here are the reasons why default value is 1:
- Internally the initialization of `KeyValueEncryptedFileStore` does not always translate to creation of new stores (new folders). In fact, most likely for apps that already uses this type of storage, will likely be initializing an instance of it with existing folders, and existing stores are v1 stores.
- v2 stores are newly created ones with version file in it. At this point, version file is not yet checked, its existence is unknown, so the default value is 1.

Let me know if these reasons make sense to you. If not, I can change it to 2. Likely I said, it doesn't really matter. The version determination in initialization does not take this default value into account.

wmathurin (approved):

AsceticMonk:
@wmathurin I thought more of your feedback, and made the changes to avoid default value. Please take another look and let me know if it's better.

wmathurin (approved):

PR #3363, opened 2021-04-09
Catalyst support

bbirman:
Enables Catalyst for all the frameworks and sample apps.

Call out: I tested with the templates and pods and ran into this issue with SalesforceSDKResources https://github.com/CocoaPods/CocoaPods/issues/8891. I was able to build locally by setting the team in the template app.

Next round:
- Enable all the unit test apps -> The tests are passing locally but need to figure out signing
- Need to investigate the advanced auth flow more, it works but could be smoother


SalesforceAnalytics.podspec
```diff
@@ -14,6 +14,8 @@ Pod::Spec.new do |s|
                      :tag => "v#{s.version}",
                      :submodules => true }
 
+  s.frameworks   = 'CoreTelephony'
```
bbirman, line 17: CoreTelephony is supported in Catalyst but needs to be explicitly linked to build successfully (for example, on core we didn’t and that worked for everything else but not Catalyst).   

So far I just added CoreTelephony to the frameworks here but if we go with that route I’m thinking we should include the other frameworks (UIKit, Fundation) to be consistent. The other option is to skip the CoreTelephony related code for Catalyst, usages [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/9f4daa6f28fb1dbeaeee0e611bd550ece66e408e/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Model/SFSDKInstrumentationEventBuilder.m#L109) and [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/9f4daa6f28fb1dbeaeee0e611bd550ece66e408e/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice%2BSFHardware.m#L640). Any preferences?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -254,6 +258,11 @@ - (UIInterfaceOrientationMask)supportedInterfaceOrientations {
 - (void)layoutWebView {
```
bbirman, line 258: Changed because this is what it looked like before
nav bar webview

iPhone/iPad still look normal with the change
iphone-ipad

wmathurin (approved):

SalesforceAnalytics.podspec
bhariharan: Let's add it in for now - we need those methods.

bhariharan (approved):

PR #3369, opened 2021-04-20
Fixing bug when advanced auth is cancelled

bhariharan:
Currently, if advanced auth is cancelled, it automatically takes the user to the server picker screen. The only workaround is for the app to specify where to go instead, through a completion block. However, we accidentally make this internal. This PR makes it public again.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.h
```diff
@@ -233,6 +233,11 @@ NS_SWIFT_NAME(UserAccountManager.NotificationUserInfo)
 NS_SWIFT_NAME(UserAccountManager)
 @interface SFUserAccountManager : NSObject
 
+/**
+ * Completion block for when auth is cancelled.
+ */
+@property (nonatomic, readwrite, copy, nonnull) void (^authCancelledByUserHandlerBlock)(void);
```
bhariharan, line 239: This is the only change.

bbirman (approved):

PR #3378, opened 2021-05-11
Fixing advanced authentication in iOS 14

bhariharan:
The current code works on `iOS 13`, but crashes on `iOS 14`. This fix allows it to work on `iOS 14` as well.


bbirman (approved):

wmathurin (commented):
👍

PR #3394, opened 2021-06-11
Add encrypt/decrypt using CryptoKit

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift
```diff
@@ -0,0 +1,180 @@
...
+    
+    static func decrypt(data: Data, privateKey: Data) throws -> Data {
+        let key = try derivedKey(from: privateKey)
+        return try decrypt(data: data, using: key)
+    }
+    
+    static func derivedKey(from privateKey: Data) throws -> SymmetricKey {
+        var sharedSecret: SharedSecret
+        if SecureEnclave.isAvailable && !UIDevice.current.isSimulator() {
+            let key = try SecureEnclave.P256.KeyAgreement.PrivateKey(dataRepresentation: privateKey)
```
bbirman, line 81: The only way I saw to encrypt with the EC key is to derive a symmetric key from it and then use that, let me know if there are any other API routes I missed

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift
```diff
@@ -0,0 +1,180 @@
...
+            sharedSecret = try key.sharedSecretFromKeyAgreement(with: key.publicKey)
+        } else {
+            let key = try P256.KeyAgreement.PrivateKey(rawRepresentation: privateKey)
+            sharedSecret = try key.sharedSecretFromKeyAgreement(with: key.publicKey)
+        }
+        return sharedSecret.x963DerivedSymmetricKey(using: SHA256.self, sharedInfo: Data(), outputByteCount: 32)
```
bbirman, line 87: Because the key is being derived from a single key pair, I didn't think it mattered what the sharedInfo is and used `Data()`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift
wmathurin: The existing [encryptUsingECforData](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSDKCryptoUtils.m#L331) is used by [SFSecureEncryptionKey]( https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/9f4daa6f28fb1dbeaeee0e611bd550ece66e408e/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFSecureEncryptionKey.m#L141) with the private key stored in secure enclave. So the key is never actually in memory, the data to be encrypted/decrypted is sent to that dedicated chip and the resulting data is returned. That makes it very secure. Even on a jailbroken device, you won't to get the key.
If we use a derive key, then that extra protection is lost.

bbirman: thanks! Sounds like we'd only be able to move the symmetric keys to CryptoKit then, I'll experiment with a combo approach

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift
```diff
@@ -0,0 +1,228 @@
...
+            let key = try symmetricKey(for: label)
+            keyCache[label] = key
+            return key
+        }
+    }
+    
+    static func symmetricKey(for label: String, keySize: SymmetricKeySize = .bits256) throws -> SymmetricKey {
+        let storedLabel = "\(KeyGenerator.keyStoreService).\(label)"
+        let result = KeychainHelper.read(service: storedLabel, account: nil)
+        if let encryptedKeyData = result.data {
```
trooper2013, line 130: ```if let encryptedKeyData = KeychainHelper.read(service: storedLabel, account: nil).data```?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift
```diff
@@ -0,0 +1,228 @@
...
+public class Encryptor: NSObject {
+    
+    enum EncryptorError: Error {
+        case combinedBoxFailed
+        case encryptionFailed(underlyingError: Error?)
+        case decryptionFailed(underlyingError: Error?)
+    }
+
+    // MARK: Symmetric Encrypt/Decrypt
+    @objc @available(swift, obsoleted: 1.0) // Objective-c only wrapper
```
trooper2013, line 43: Add documentation here for all the public functions

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift
```diff
@@ -0,0 +1,228 @@
...
+    public static func encrypt(data: Data, using key: SymmetricKey) throws -> Data {
+        let sealedBox = try AES.GCM.seal(data, using: key, nonce: AES.GCM.Nonce())
+        if let combined = sealedBox.combined {
+            return combined
+        } else {
+            throw EncryptorError.combinedBoxFailed
+        }
+    }
+    
+    @objc @available(swift, obsoleted: 1.0) // Objective-c only wrapper
```
trooper2013, line 58: Doc

trooper2013 (commented):
@bbirman This looks good. Added a  few minor comments.We should add docs for all public apis.

trooper2013 (approved):

wmathurin (approved):

bhariharan (approved):

PR #3408, opened 2021-08-12
Fix for Xcode 12.4 compiler and dynamicKeys variable (#866)

mbwimmer:
* Fix for dynamicKeys variable and Xcode 12.4 compiler.


trooper2013 (approved):
Thanks @mbwimmer

PR #3418, opened 2021-09-09
Fixing compilation on Xcode 12

wmathurin:


brandonpage (approved):

wmathurin:
Hmm, I merged that too quickly. Without changing the deployment target, the ifdef is never entered (even if running from xcode 13 onto a iOS 15 device), so now the code is never run...

wmathurin:
This one should do the job: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3419

PR #3421, opened 2021-09-17
Using regexp for smartsql parsing instead of NSScanner

wmathurin:
Implementation is now closer to Android's
Added new test that was failing with NSScanner implementation

### DIFF
--- a/libs/SmartStore/SmartStoreTests/SFSmartSqlTests.m
+++ b/libs/SmartStore/SmartStoreTests/SFSmartSqlTests.m
@@ -245,6 +245,11 @@
                           [self.store convertSmartSql:@"select {employees:education} from {employees} where {employees:education} like ' { { { } } '"]);
 }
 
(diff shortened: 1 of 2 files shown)


libs/SmartStore/SmartStoreTests/SFSmartSqlTests.m
```diff
@@ -245,6 +245,11 @@ - (void) testConvertSmartSqlWithQuotedUnbalancedCurlyBrace {
                           [self.store convertSmartSql:@"select {employees:education} from {employees} where {employees:education} like ' { { { } } '"]);
 }
 
+- (void) testConvertOtherComplexSmartSql {
```
wmathurin, line 248: I added the same test on Android (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2234/commits/66d95aecacf263e21977ccb46fdd600e47c34c58).
It passes there without code changes.

bbirman (approved):

PR #3423, opened 2021-09-23
SmartStore encryption update

bbirman:
- Create new encrypt/decrypt stream classes
- SmartStore to use new encryption key and new salt to get off MD5 + migration steps
- Remove old upgrade steps for going from single user -> multiuser


bhariharan (approved):

wmathurin (approved):
LGTM

PR #3424, opened 2021-09-24
Replace Passcode with Screen Lock

brandonpage:
- All Passcode related public API has been deprecated for removal in 10.0.
- Passcode is replace with the OS's Screen Lock. This approach is far simpler for the end user and will be much easier to maintain over time.
- Removed pin length requirement. If customers want to enforce more strict policy than Android's minimum they can do this via MDM.
- Removed timeout requirement. The app locks on background consistently like users expect.
- Removed biometric override from connected app.
- Removed auth attempts, this is handled by the OS now.
- No more passcode lock-in until you uninstall or logout all users. When the last user that requires screen lock is removed, the lock is too. Logout button on screen lock removes all (and only) users that require the lock.
- ScreenLockUiView is simple lock screen that displays the app icon, a setup button if applicable, a retry auth button if applicable, logout button if applicable and an OS provided error message if applicable.


zaitsman:
@brandonpage Apologies if I am wrong, this looks like an unintended change:

https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3424/files#diff-4beeb8e410a97932ae9081dd91ecb260e5f3ec12a0a233bf5e349eb4b2b1b504R175

a format specifier ` %@` got dropped from a log error message in SFSDKSalesforceSDKUpgradeManager.m on line 165/175(after change).

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -2754,6 +2774,7 @@
 			baseConfigurationReference = CEBA94EF1C1B5D4C00C0B2E0 /* SalesforceSDKCore-Dynamic-iOS-Debug.xcconfig */;
 			buildSettings = {
 				ALWAYS_SEARCH_USER_PATHS = NO;
+				APPLICATION_EXTENSION_API_ONLY = NO;
```
bbirman, line 2777: What's not safe for extensions?

brandonpage: `UIApplication.shared.open(settingsUrl)` in ScreenLockUIView.  The button that navigates the user to settings.  I could remove the button if this is a big problem.

bhariharan: This might be a problem for some of  our apps that use extensions, like SApp, for instance.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSalesforceSDKUpgradeManager.m
```diff
@@ -173,4 +183,12 @@ + (void)updateEncryptionForUserAccountPath:(NSString *)userAccountPath {
     [encryptedArchiveData writeToFile:userAccountPath atomically:YES];
 }
 
++ (void)upgradePasscode {
```
bhariharan, line 186: Add `TODO:` to remove this in `10.0`.

shared/resources/SalesforceSDKResources.bundle/en.lproj/Localizable.strings
```diff
@@ -48,11 +48,17 @@
...
 "accessibilityPasscodeLengthHint" = "Passcode is %d numbers long.";
 "accessibilityUnlockAnnouncement" = "Passcode confirmed.";
+
+// Screen Lock
+"biometricReason" = "Please authenticate to proceed.";
+"logoutButtonTitle" = "Log out";
+"retryButtonTitle" = "Retry Unlock";
+"setUpPasscodeButton" = "Set Up Passcode";
+"setUpPasscodeMessage" = "%@ reqires you to set up an iOS passcode to use the app.";
+"fallbackErrorMessage" = "An unexpected error occured.";
```
bhariharan, line 61:
```suggestion
"fallbackErrorMessage" = "An unexpected error occurred.";
```

bhariharan (changes requested):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSalesforceSDKUpgradeManager.m
brandonpage: Shouldn't we remove this in `11.0`?  If we remove in `10.0` customers will not be safe to upgrade from `9.1.1` to `10.0`.

bhariharan: Sure, `11.0` is fine too. Let's add a `TODO:` here so we don't forget.

shared/resources/SalesforceSDKResources.bundle/en.lproj/Localizable.strings
```diff
@@ -48,11 +48,17 @@
...
 "accessibilityPasscodeFieldLabel" = "Passcode field";
 "accessibilityPasscodeLengthHint" = "Passcode is %d numbers long.";
 "accessibilityUnlockAnnouncement" = "Passcode confirmed.";
+
+// Screen Lock
+"biometricReason" = "Please authenticate to proceed.";
+"logoutButtonTitle" = "Log out";
+"retryButtonTitle" = "Retry Unlock";
+"setUpPasscodeButton" = "Set Up Passcode";
+"setUpPasscodeMessage" = "%@ reqires you to set up an iOS passcode to use the app.";
```
bhariharan, line 60:
```suggestion
"setUpPasscodeMessage" = "%@ requires you to set up an iOS passcode to use the app.";
```

bhariharan (changes requested):

bhariharan (approved):

PR #3443, opened 2021-11-03
Make key cache thread safe

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift
```diff
@@ -291,6 +287,42 @@ public class KeyGenerator: NSObject {
     }
 }
 
+class Cache<Key: Hashable, Value> {
```
bbirman, line 290: SFSDKSafeMutableDictionary only works with objects rather than structs so I made this using the same queue properties (thinking we could expand it later if we need it for other cases)

bhariharan (approved):

wmathurin (commented):
👍

Issue #3437, opened 2021-10-20
iOS 15 wakes the app up in the background, and SFDefaultUserAccountPersister deletes the UserAccount.plist


az-oolloow:
Please fill out the following details:

1. Version of Mobile SDK Used: 9.2.0 *using recommended `forceios` npm package method*, 8.3.0 with legacy migration removed (also confirmed the code is the same in latest `dev` branch)
2. Issue found in Native App or Hybrid App: Native App
3. OS Version: 15.0 + (so also 15.0.1 and 15.0.2)
4. Device: iPad (all sorts of iPad devices, new iPad mini 6, iPad Pro 9.7", iPad Pro 2018 etc.)
5. Steps to reproduce: 

Connect the device to the mac
Open Console
In the console, filter by `Any` `AppName`
Build an app using SF SDK
Login
Make sure device has passcode/faceid/touch id
Kill the app
Lock the screen
Wait until iOS wakes up your app in the background (typically under 30 minutes, I found no way to speed this up, sometimes launching and killing the app rapidly a few times causes iOS to decide to re-launch it just 1 minute after you killed it and sometimes it does take over 30 minutes).
The symbol of this is the following log:
```
2247	default	14:14:22.847892+1100	logoutRepro	Initializing connection
```
(you can see full log below, basically there will be a lot of messages, > 20 with the specified filter settings)
Open the app again

6. Actual behavior:
User is logged out and is presented with a 'please log in'

7. Expected Behavior:
User is not logged out

8. Analysis:

This issue stems from the fact that `SFDefaultUserAccountPersister` saves `UserAccount.plist` using `NSFileProtectionType.NSFileProtectionComplete` (which is really weird because on top of OS encryption SDK already uses it's own AES encryption so this like... `extra protection`?)

It seems that Apple made some changes in iOS 15 that cause the app to be launched more frequently in the background (?) and this happens.

You can see it clearly in the log here: `2247	default	14:14:22.961899+1100	logoutRepro	CLASS: SFDefaultUserAccountPersister User account data could not be decrypted. Can't load account.`

Here's where the `protection` is specified for the file:
https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFDefaultUserAccountPersister.m#L219

Here's where the encrypted file is actually nil when you do `MobileSyncSDKManager.initializeSDK()` in the background:
[...]

PR #3447, opened 2021-11-29
Change default keychain accessibility and file protection

bbirman:
For #3437 
Tested with fresh install and upgrade from 9.2.0 on MobileSyncExplorer

### DIFF
--- a/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainManager.swift
+++ b/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainManager.swift
@@ -65,7 +65,7 @@ internal class KeychainItemManager: NSObject {
     /// Initializer for kSecClassGenericPassword with accessgroup. Will create a keychain item manager
     /// for kSecClassGenericPassword operations
     convenience init(service: String, account: String?) {
-        self.init(service: service, account: account, accessibilityAttribute: kSecAttrAccessibleWhenUnlockedThisDeviceOnly)
+        self.init(service: service, account: account, accessibilityAttribute: kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly)
     }
 
     /// Initializer for kSecClassGenericPassword with accessgroup. Will create a keychain item manager
@@ -247,12 +247,15 @@ public class KeychainResult: NSObject {
 
 @objc(SFSDKKeychainHelper)
 public class KeychainHelper: NSObject {
+    @objc public private(set) static var accessibilityAttribute: CFString?
+
+    private static var keychainAccessibleAttribute: CFString {
+        return accessibilityAttribute ?? kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
+    }
 
     private static let upgrade: Void = {
         _ = KeychainUpgradeManager.init().upgradeManagedKeys()
     }()
-
-    private static var keychainAccessibleAttribute: CFString = kSecAttrAccessibleWhenUnlockedThisDeviceOnly
 
     //pre 9.1 keys are updated to have the creator tag.
     private static func upgradeIfRequired() {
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSalesforceSDKUpgradeManager.m
```diff
@@ -42,18 +43,26 @@ @implementation SFSDKSalesforceSDKUpgradeManager
...
     NSString *lastVersion = [SFSDKSalesforceSDKUpgradeManager lastVersion];
     NSString *currentVersion = [SFSDKSalesforceSDKUpgradeManager currentVersion];
-    
+
     if ([currentVersion isEqualToString:lastVersion]) {
         return;
     }
     
-    if (!lastVersion || [lastVersion doubleValue] < 9.2) {
+    if (!lastVersion || [lastVersion compare:@"9.2.0" options:NSNumericSearch] == -1) {
```
wmathurin, line 51: Or is that test equivalent? "version older from 9.2" changed to "version is not 9.2". Do we want to enter the if with 9.2.1, what about future releases?

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainManager.swift
```diff
@@ -380,19 +383,15 @@ public class KeychainHelper: NSObject {
...
+    @objc @discardableResult
+    public class func setAccessibleAttribute(_ secAttrAccessible: KeychainItemAccessibility) -> KeychainResult {
        
+        accessibilityAttribute = secAttrAccessible.asCFString
         if accessibleAttributeMatches(secAttrAccessible) {
             return KeychainResult.init(data: Data(), status: errSecSuccess)
         }
         
         CachedWrapper.clearAllCaches()
-        let accessibleAttribute = secAttrAccessible.asCFString
```
wmathurin, line 390: Why drop this?

bbirman: I moved it up before the `accessibleAttributeMatches(secAttrAccessible)` check so that even if it matches what's already there, we'll know that the app set the accessible attribute

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKSalesforceSDKUpgradeManager.m
bbirman: It should be -1 (NSOrderedAscending) for older, 0 (NSOrderedSame), and 1 (NSOrderedDescending) for newer. Looking at it again I'm thinking I should use the enum name instead of the raw value. Let me know if you think there's a better way to do the comparison, too

wmathurin: My bad, I forgot how compare works. I was thinking (indexOf() == -1).

bhariharan (approved):
@bbirman @wmathurin If we're doing a patch release, should this go against `master` and be a part of that patch release?

wmathurin:
@bhariharan That makes sense. Let's have it go to master. I'm updating the release script for patches (it won't do the dev to master merge initial step).

wmathurin (approved):

az-oolloow:
@bbirman sorry to barge in like this - how does this PR handle the scenario where user's app is upgraded but not launched?

e.g. for us the app is pushed by the MDM systems to thousands of devices. Users do not launch it sometimes for days after install - this is how we came across the original issue, come Monday lots of users were logged out.

The `updateUserAccountFileProtection` method would fail to find the files that had been created previously with protectionComplete flag set unless `UIApplication.shared.isProtectedDataAvailable` returns true.

Apologies if I am missing where this is checked for.

bbirman:
@az-oolloow thanks for chiming in! I didn't try that, will check it out

PR #3449, opened 2021-11-30
Address issues with UINavigationBarAppearance scoped to certain view flows

bcriscuolo:
In iOS 15 (building with Xcode 13.x), SFS has seen issues of slight navigation bar positioning which are due to the limited usage of `UINavigationBarAppearance` in MSDK. This PR expands the limited support already in place to include `configureWithOpaqueBackground` and `compactAppearance` & `scrollEdgeAppearance`.

Additionally, as the `styleNavigationBar:config:` method is only used by four VCs, appearance setting is scoped to _those_ classes only - meaning that an application which uses its own navigation bar appearance will not have its settings overridden by MSDK (depending on call lifecycle).

Screenshot shows primary issue.
Simulator Screen Shot - iPhone 13 Pro - 2021-11-30 at 09 06 38

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKViewUtils.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKViewUtils.m
@@ -28,6 +28,11 @@
 #import "SFSDKViewUtils.h"
 #import "SFSDKViewControllerConfig.h"
 
+#import "SFLoginViewController.h"
+#import "SFSDKAppLockViewController.h"
+#import "SFSDKLoginFlowSelectionViewController.h"
+#import "SFSDKLoginHostListViewController.h"
+
 @implementation SFSDKViewUtils
 
 + (void)styleNavigationBar:(UINavigationBar *)navigationBar config:(SFSDKViewControllerConfig *) config {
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKViewUtils.m
```diff
@@ -28,6 +28,11 @@
 #import "SFSDKViewUtils.h"
 #import "SFSDKViewControllerConfig.h"
 
+#import "SFLoginViewController.h"
```
bhariharan, line 31: Do we want to introduce dependency on internal view controllers for a utility class here? Is there a better way we could do this? What do you think, @bbirman and @wmathurin?

bcriscuolo: An idea - not sure how it will work though, but can make an attempt.

This helper function can be passed a new param, a `Class` representing the class intended to be "controlled".

That would break the dependency (now, more generic) and it is possible `[UINavigationBar appearanceWhenContainedInInstancesOfClasses...` will _add_ the class rather than replacing others.

bbirman: I like that idea

bcriscuolo: I may have some cycles this afternoon to look more closely at that, to see if it works.

bcriscuolo:
Closing in favor of https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3457.

PR #3455, opened 2021-12-14
Fix URL cache

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -797,16 +796,15 @@ - (void)setURLCacheType:(SFURLCacheType)URLCacheType {
         _URLCacheType = URLCacheType;
         [NSURLCache.sharedURLCache removeAllCachedResponses];
         NSURLCache *cache;
-        NSURL *defaultURL = [NSURL URLWithString:kDefaultCachePath];
         switch (URLCacheType) {
             case kSFURLCacheTypeEncrypted:
-                cache = [[SFSDKEncryptedURLCache alloc] initWithMemoryCapacity:kDefaultCacheMemoryCapacity diskCapacity:kDefaultCacheDiskCapacity directoryURL:defaultURL];
+                cache = [[SFSDKEncryptedURLCache alloc] initWithMemoryCapacity:kDefaultCacheMemoryCapacity diskCapacity:kDefaultCacheDiskCapacity directoryURL:nil];
```
bbirman, line 801: The old API took the subdirectory name as a relative path and I missed the difference when changing over

bhariharan (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
wmathurin: Should we put that in master as part of 9.2.1?

PR #3460, opened 2022-01-04
Bumping up deployment target to 14.0 and some clean up

bhariharan:
Deployment target is now `14.0`.


libs/SalesforceAnalytics/SalesforceAnalytics.xcodeproj/project.pbxproj
```diff
@@ -813,8 +815,8 @@
 				GCC_PREFIX_HEADER = "$(PROJECT_NAME)/Supporting Files/$(PROJECT_NAME)-Prefix.pch";
 				INFOPLIST_FILE = "SalesforceAnalytics/Supporting Files/Info.plist";
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
-				IPHONEOS_DEPLOYMENT_TARGET = 13.0;
-				"IPHONEOS_DEPLOYMENT_TARGET[sdk=macosx*]" = 14.5;
```
bhariharan, line 817: The `Mac OS` ones were messed up for some reason.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccount+Internal.h
```diff
@@ -30,7 +30,7 @@ extern NSString * __nonnull const kUserAccountPhotoEncryptionKeyLabel;
 
 @property (nonatomic, readwrite, getter = isUserDeleted) BOOL userDeleted;
 @property (nonatomic, readwrite, assign) SFUserAccountLoginState loginState;
-@property (nonatomic, strong) NSMutableDictionary *customData;
+@property (nonatomic, strong, nullable) NSMutableDictionary *customData;
```
bhariharan, line 33: Got rid of a warning.

libs/SalesforceAnalytics/SalesforceAnalytics.xcodeproj/project.pbxproj
bbirman: macosx was at 14.5 because that was the latest when we introduced it and we didn't test any earlier versions

wmathurin: NB: It's not possible to downgrade to anything below 14.8.1 right now (it's the only 14.x still signed).

bhariharan: @wmathurin Not sure I understand. Even if this says `14.0`, it should still work, as long as you have the latest `Xcode`, right? Isn't signing related to the `Xcode` version?

wmathurin: Sure it will work. Although what is the point of supporting a version Apple no longer supports?

The only way for someone to still be on 14.x (with x<8.1) is to have avoided updates. It is not possible to downgrade to such versions because they are no longer signed by Apple.

Apple aggressively stops signing older version to block people for downgrading.
https://9to5mac.com/2021/10/04/apple-stops-signing-ios-14-8-for-some-devices-blocking-downgrades-from-ios-15/

bhariharan: Ah, I see what you mean. Should I move to `14.8.1` across the board then?

bhariharan: I'm worried about blowback from internal apps (and maybe even some external ones), if we do that.

bhariharan: Also, I was considering devices in APAC and EMEA who may have declined OS updates.

wmathurin: If it's going to cause trouble, then we can just go with 14.0.

wmathurin (approved):

.circleci/config.yml
```diff
@@ -1,11 +1,11 @@
 anchors:
-  - &latest-xcode "12.5.1"
-  - &latest-ios   "14.5"
-  - &min-ios      "13.7"
-  - &chruby       "2.7.3"
-  - &beta-xcode   "13.0.0"
+  - &latest-xcode "13.2.1"
+  - &latest-ios   "15.0"
```
brandonpage, line 3: If you are bumping Xcode to `13.2.1` you need to bump latest iOS to `15.2`, min iOS to `14.5` and chruby any of the available versions: `2.6.9, 2.7.5, 3.0.3` as per the [what is available in the xcode 13.2.1 container](https://circle-macos-docs.s3.amazonaws.com/image-manifest/v6690/index.html).  (link to this [in the circleci anncounment post](https://discuss.circleci.com/t/xcode-13-2-1-released/42334)).

.circleci/config.yml
```diff
@@ -140,7 +140,7 @@ workflows:
               nightly-test: [true]
               ios: [*latest-ios]
           requires:
-            - test << matrix.lib >> iOS 13.7
+            - test << matrix.lib >> iOS 14.0
       - run-tests:
```
brandonpage, line 144: There is no beta Xcode available at the moment so we can ignore/comment out/delete this.

bbirman (approved):

PR #3463, opened 2022-01-05
Moving to SQLCipher 4.5.0

wmathurin:
Like on Android (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2265), had to tweak tests doing explain query plan.

NB, they are two failures I need to look into:
* testUpgradeNoSaltLegacyKey
* testUpgradeLegacySaltLegacyKey

### DIFF
--- a/SmartStore.podspec
+++ b/SmartStore.podspec
@@ -22,7 +22,7 @@ Pod::Spec.new do |s|
 
       smartstore.dependency 'SalesforceSDKCore', "~>#{s.version}"
       smartstore.dependency 'FMDB/SQLCipher', '~> 2.7.5'
-      smartstore.dependency 'SQLCipher/fts', '~> 4.4.3'
+      smartstore.dependency 'SQLCipher/fts', '~> 4.5.0'
       smartstore.source_files = 'libs/SmartStore/SmartStore/Classes/**/*.{h,m,swift}', 'libs/SmartStore/SmartStore/SmartStore.h'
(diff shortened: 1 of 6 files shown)


bhariharan (approved):

wmathurin:
The two tests failures happen with SQLCipher 4.5.0 but not with 4.4.3. Looking into them.

wmathurin:
Here is what I found out for the failing tests.

When we unlock the database we do the following to automatically migrate settings if dealing with a SQLCipher 2.x or 3.x database:
```objective-c
[[db executeQuery:@"PRAGMA cipher_migrate"] close];
```
We never checked the result set to see whether the migration was successful. 
It turns out, it returns an error (non zero integer) in those tests that change encryption. It can be seen by using the following code instead:
```objective-c
FMResultSet* res = [db executeQuery:@"PRAGMA cipher_migrate"];
if ([res next]) {
  int migrationResult = [res intForColumnIndex:0];
  if (migrationResult != 0) {
  // migration failed
  [SFSDKSmartStoreLogger e:[self class] format:@"Error running cipher_migrate for store '%@'", [db databasePath]];
  }
}
[res close];
```
It is happening with both SQLCipher 4.5.0 and 4.4.3.
In SQLCipher 4.5.0, however, `PRAGMA cipher_migrate` causes the database to permanently enter an error state if a migration fails (see [release notes](https://www.zetetic.net/blog/2021/10/28/sqlcipher-4.5.0-release/). That's why the tests are now failing.

Since it's a 4.x database, I'm not sure why cipher_migrate is complaining.
Considering that we started using 4.x in 7.1, this call would only be useful for customer attempting to jump from Mobile SDK 7.0 to 10.0. I'm not even sure that's possible given the deprecated code we have removed.

I recommend we drop that call to cipher_migrate. Thoughts? @bhariharan @bbirman

bbirman:
@wmathurin sounds good to me

bbirman (approved):

bhariharan:
Sounds good to me as well.

PR #3465, opened 2022-01-08
Update interface orientation method and fix API deprecation

AsceticMonk:
Fix `statusBarOrientation` deprecation

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
@@ -599,20 +599,20 @@
 }
 
 - (UIInterfaceOrientation)interfaceOrientation {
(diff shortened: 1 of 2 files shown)


bhariharan (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
```diff
@@ -599,20 +599,20 @@ - (NSString *) macaddress
 }
 
 - (UIInterfaceOrientation)interfaceOrientation {
```
bbirman, line 601: it looks like the comment in the header for this method mentions falling back on the status bar, could you update that too?

bbirman (approved):
🎉

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
AsceticMonk: Yes, updated.

PR #3478, opened 2022-01-27
Fix for KV store initialization

bbirman:
If the legacy key is present new stores that didn’t use the legacy key could have an encryption update triggered when it shouldn’t

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
@@ -111,13 +111,17 @@ public class KeyValueEncryptedFileStore: NSObject {
         let versionFileURL = directory.appendingPathComponent(KeyValueEncryptedFileStore.storeVersionFileName)
         if isNewlyCreated {
             let versionFileCreated = writeFile(versionFileURL, content: KeyValueEncryptedFileStore.storeVersionString)
-            if !versionFileCreated {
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Storage/KeyValueEncryptedFileStore.swift
```diff
@@ -111,13 +111,17 @@ public class KeyValueEncryptedFileStore: NSObject {
         let versionFileURL = directory.appendingPathComponent(KeyValueEncryptedFileStore.storeVersionFileName)
         if isNewlyCreated {
             let versionFileCreated = writeFile(versionFileURL, content: KeyValueEncryptedFileStore.storeVersionString)
-            if !versionFileCreated {
+            let encryptionFileURL = directory.appendingPathComponent(KeyValueEncryptedFileStore.storeGCMEncryptionFileName)
```
wmathurin, line 114: How is that file use?

bbirman: It's an empty file called "gcmEncryption", we're writing it upgrade on but I didn't do it for new stores. Having the legacy key will trigger the upgrade method but then the method will check for that file's existence to determine if the upgrade to gcm needs to happen or not

wmathurin: 👍

wmathurin (approved):

bhariharan (approved):

PR #3479, opened 2022-01-29
Remove or internalize deprecated encryption classes

bbirman:
- Kept any classes that we need for upgrade paths, making a bunch internal and leaving some of the entry points public so that they can be used for external upgrade steps (for removal in 11.0)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/SFDecryptStream.h
```diff
@@ -33,7 +33,7 @@ NS_ASSUME_NONNULL_BEGIN
  * `SFDecryptStream` implements an input stream that decrypts data immediately after it's read.
  * It uses `SFCryptChunks` to perform the decryption.
  */
-SFSDK_DEPRECATED(9.2, 10.0, "Will be used internally for upgrade steps only. Use SFSDKDecryptStream instead")
+SFSDK_DEPRECATED(9.2, 11.0, "Will be removed, use SFSDKDecryptStream instead. This should only be used for upgrade steps")
```
bbirman, line 36: @wmathurin I ended up leaving this as public because SFS mentioned it would be helpful for their upgrade steps besides the SmartStore case we talked about, let me know if you want to change it

wmathurin: 👍

wmathurin (approved):

bhariharan (approved):

PR #3487, opened 2022-02-21
Set header fields, body and method on redirect request

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.m
```diff
@@ -154,7 +154,11 @@ - (void)URLSession:(NSURLSession *)session task:(NSURLSessionTask *)task willPer
         // Don't auto follow redirects in authenticated case
         completionHandler(nil);
     } else {
```
bhariharan, line 156: A few questions:

- Shouldn't the request in this case fall in the `if (...)` condition block on `line 153`, since it's an authenticated request?
- I'm also not sure why we prevent redirection from happening in the authenticated case. Any idea why?
- So, it looks like redirection strips out all the headers and everything else, based on your code fix. Am I correct in assuming that? If so, do we still need the `if (...)` condition above? Could we get rid of it?

bhariharan: Spoke with Brianna offline, this is good to go.

bhariharan (approved):

PR #3488, opened 2022-02-22
Fix for MDM host update when app groups enabled

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFManagedPreferences.m
```diff
@@ -93,7 +93,7 @@ - (void)dealloc {
 }
 
 - (void)configurePreferences {
-    self.rawPreferences = [[NSUserDefaults msdkUserDefaults] dictionaryForKey:kManagedConfigurationKey];
+    self.rawPreferences = [[NSUserDefaults standardUserDefaults] dictionaryForKey:kManagedConfigurationKey];
```
bbirman, line 96: It looks like the MDM app configuration is only pushed to standard user defaults, so when app groups are enabled the login hosts from standard user defaults are migrated to the app groups user defaults suite but subsequent host changes are stored in standard user defaults while managed preferences is reading from the other one through msdkUserDefaults

bhariharan (approved):

PR #3503, opened 2022-05-06
Fix for decrypt stream optimization issue

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/DecryptStream.swift
```diff
@@ -95,6 +95,11 @@ public class DecryptStream: InputStream {
             do {
                 let data = Data(bytes: encryptedBuffer, count: bytesRead)
                 let decryptedData = try Encryptor.decrypt(data: data, using: key)
+                guard decryptedData.count <= CryptStream.chunkSize else {
```
bbirman, line 98: SFS upgraded to Xcode 13.3.x and started seeing file decryption issues only in their release builds. 

It seems like something weird is happening here where we know that the decrypted data that’s being added to the buffer is bound by our encryption block size but the compiler doesn’t and is handling it differently somehow unless we have an explicit check for it

`precondition(decryptedData.count <= CryptStream.chunkSize)` also works instead but I didn’t want to be that aggressive even if it should never trigger

I have a bigger change that does more refactoring, but opened this on `master` so they could pick up the least intrusive fix now

bcriscuolo (approved):
This has been validated within production code and addresses a blocker to our release. 👍

bhariharan (approved):

Issue #3508, opened 2022-05-19
Tel: URL in Login View shows Unsupported URL dialog


reubenbijl:
Please fill out the following details:

1. Version of Mobile SDK Used: 
  - 8.1.0
2. Issue found in Native App or Hybrid App:
 - Native app
3. OS Version:
 - All
4. Device:
 - All
5. Steps to reproduce:
 - Include a telephone number in the login webview
 - Tap the telephone number in the login
 - Error, cannot load page
6. Actual behavior:
 - Error message, cannot load page
7. Expected Behavior:
  - Call number
8. Error Log:
[SalesforceSDKCore] CLASS: SFOAuthCoordinator webView:didStartProvisionalNavigation: host=(null) : path=(null)
[ProcessSwapping] 0x1100f14a0 - [pageProxyID=31, webPageID=40, PID=14295, navigationID=5] ProvisionalPageProxy::didFailProvisionalLoadForFrame: frameID=3
[Process] 0x128815e18 - [pageProxyID=31, webPageID=32, PID=14292] WebPageProxy::didFailProvisionalLoadForFrame: frameID=3, domain=NSURLErrorDomain, code=-1002, isMainFrame=1
[SalesforceSDKCore] CLASS: SFOAuthCoordinator SFOAuthCoordinator:didFailLoadWithError: error code: -1002, description: unsupported URL, URL: ***
[SalesforceSDKCore] CLASS: SFSDKAuthHelper Authentication failed: unsupported URL.

wmathurin:
Our navigation delegate does not handle tel: URL. See https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m#L715

Marked as "enhancement".

valentin-salamatov:
Hi there! I've got the same issue when navigating to the Privacy page (https://www.salesforce.com/us/company/privacy) and clicking telephone hyperlinks in different places.

I’m wondering if there is a way to get around this problem and open the Privacy page in some browser application instead of WKWebView?

<img width="384" alt="Screenshot 2023-01-11 at 12 41 55" src="https://user-images.githubusercontent.com/122388932/211788602-8f7f6711-ad8a-490a-99c2-3f9d4be3c736.png">

wmathurin:
[...]

PR #3509, opened 2022-05-19
Added URL support for external URLs

reubenbijl:
Added support for tel and mailto URLs
Resolves issue https://github.com/forcedotcom/SalesforceMobileSDK-iOS/issues/3508

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -715,7 +715,12 @@
 - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler {
     NSURL *url = navigationAction.request.URL;
     NSString *requestUrl = [url absoluteString];
-    if ([self isRedirectURL:requestUrl]) {
+    
+    if ([url.scheme isEqual:@"mailto"] || [url.scheme isEqual:@"tel"]){
+        [[UIApplication sharedApplication] openURL:url options:@{} completionHandler:nil];
+        decisionHandler(WKNavigationActionPolicyCancel);
+    }
+    else if ([self isRedirectURL:requestUrl]) {
         [self handleUserAgentResponse:url];
         decisionHandler(WKNavigationActionPolicyCancel);
     } else if ([self isSPAppRedirectURL:requestUrl]){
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -715,7 +715,13 @@ - (NSURLSession*)session {
 - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler {
     NSURL *url = navigationAction.request.URL;
     NSString *requestUrl = [url absoluteString];
-    if ([self isRedirectURL:requestUrl]) {
+    
+    
+    if ([url.scheme isEqual:@"mailto"] || [url.scheme isEqual:@"tel"]){
```
bhariharan, line 720: This block is only meant to handle auth-related redirects, to detect the completion of the login flow. In the code displayed below:

- `Code block in line 724`: Checks if the callback URL has been hit, signaling the end of the user agent flow, hence intercept and cancel.
- `Code block in line 727`: Checks if the callback URL has been hit in the IDP flow, triggering further processing to complete the IDP login flow, after which we cancel webview redirects.
- `Code block in line 730`: Default processing of redirects within the webview, hence `WKNavigationActionPolicyAllow`.

We wouldn't want to explicitly add logic to support `tel:` or `mailto:` here, within the login page. If those redirects are processed, the login flow will be incomplete and will need to be restarted within this webview. Besides, if we want to support this in the future, we would have to provide an override so apps can choose to support it in their subclass, we wouldn't want to support it out of the box (by default, we leave it to the standard webview processing). Hope this helps!

cc: @wmathurin @bbirman

bhariharan (changes requested):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
reubenbijl: Thanks for the feedback. Where are you suggesting that we do handle this use case?

reubenbijl: I've spent some more time looking at this today and am struggling to work out what you mean by supporting it in a subclass? Currently this function looks like it gets called through the SFSDKAuthHelper loginIfRequired, which calls, SFUserAccountManger, SFSDKAuthSession and then the SFOAuthCoordinator.


Are you able to share an example of what you mean by subclassing this?

wmathurin:
This https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3674 will allow apps to handle such URLs.

PR #3513, opened 2022-05-26
Trust work on SalesforceSDKCommon tests

AsceticMonk:
**Changes**
- Fix minor formatting issues in both Swift and Objective-C test codes
- Use `@testable import` in Swift tests
- Deleted empty setup or teardown methods
- Got rid of unnecessary `throws`

### DIFF
--- a/libs/SalesforceSDKCommon/SalesforceSDKCommonTests/KeychainItemManagerTests.swift
+++ b/libs/SalesforceSDKCommon/SalesforceSDKCommonTests/KeychainItemManagerTests.swift
@@ -26,29 +26,17 @@
 //  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import XCTest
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCommon/SalesforceSDKCommonTests/KeychainItemManagerTests.swift
```diff
@@ -26,29 +26,17 @@
...
 
 import XCTest
-import SalesforceSDKCommon
-enum KeychainTestError: Error {
-    case failed(osStatus: OSStatus)
-    case failedToRead
-}
-
-class KeychainItemManagerTests: XCTestCase {
+@testable import SalesforceSDKCommon
```
bbirman, line 29: Did something need internal access?

AsceticMonk: You are right, not needed since we are only testing public methods.

wmathurin (approved):

PR #3517, opened 2022-05-27
Upgrade sqlcipher to 4.5.1 and add arm64 for simulators

bbirman:


external/ThirdPartyDependencies/sqlcipher/SQLCipher.xcframework/Info.plist
```diff
@@ -6,23 +6,22 @@
 	<array>
 		<dict>
 			<key>LibraryIdentifier</key>
-			<string>ios-x86_64-simulator</string>
+			<string>ios-arm64_arm64e</string>
```
bbirman, line 9: The diff looks messier because the order of the entries changed, but the catalyst and ios entries are the same. the simulator change is `ios-x86_64-simulator` -> `ios-arm64_x86_64-simulator`

AsceticMonk (approved):

bhariharan (approved):

PR #3518, opened 2022-05-27
New sync up target using sobject collection

wmathurin:
- New sync up target CollectionSyncUpTarget (a subclass of BatchSyncUpTarget, itself a subclass of SyncUpTarget) uses sobject collection create/update/upsert/delete instead of composite batch. 
- It's the new default for sync up (when implementation class is not specified) replacing BatchSyncUpTarget (which was the default since 7.1).
- Most of the work went into refactoring CompositeRequestHelper (and converting it from Objective-C to swift) to allow it to work with requests/responses to/from composite/batch or composite/sobjects, as a result CollectionSyncUpTarget is very short (constructors plus a one line overriding method sendRecordRequests).
- New test suite: CollectionSyncUpTargetTests (a subclass of SyncUpTargetTest) so running the same tests as existing sync up tests to make sure parity of behavior.
- Added a test that syncs up "many" (500) records at once, on my machine the sync up ran in:
  - 146s with SyncUpTarget
  - 40s with BatchSyncUpTarget
  - 8s with CollectionSyncUpTarget (5x faster than batch sync up and 18x faster than single record sync up)


wmathurin:
In draft since I have a couple of test failures to sort out.

libs/MobileSync/MobileSync/Classes/Util/CompositeRequestHelper.swift
```diff
@@ -0,0 +1,330 @@
...
+        
+        static func getRefIds(recordRequests:Array<RecordRequest>, requestType:RequestType) -> Array<String> {
+            return recordRequests
+                .filter { $0.requestType == requestType }
+                .map { $0.referenceId! }
+        }
+
+        static func getIds(recordRequests:Array<RecordRequest>, requestType:RequestType) -> Array<String> {
+            return recordRequests
+                .filter { $0.requestType == requestType && $0.objectId != nil }
```
bbirman, line 196: I think you could also do this if you wanted to get around the nil check/force unwrap
```
 .filter { $0.requestType == requestType }
 .compactMap { $0.objectId }
 ```

**Review by bbirman — APPROVED** [review]

**Comment by mjmasn** [comment]
@wmathurin can this be highlighted as a breaking change in the release notes for iOS/Android/ReactNative please as all the error formats have changed with this new API.

`/services/data/v56.0/sobjects/Account`
```json
{
    "Name" : "Mobile SDK 10.2 is an interim release that features non-breaking API changes and modernized iOS support"
}
```
```json
[ {
  "message" : "60 Character Limit",
  "errorCode" : "FIELD_CUSTOM_VALIDATION_EXCEPTION",
  "fields" : [ "Name" ]
} ]
```

`/services/data/v56.0/composite/sobjects`
```json
{
   "allOrNone" : false,
   "records" : [{
      "attributes" : {"type" : "Account"},
      "Name" : "Mobile SDK 10.2 is an interim release that features non-breaking API changes and modernized iOS support"
   }]
}
```
```json
[ {
  "success" : false,
  "errors" : [ {
    "statusCode" : "FIELD_CUSTOM_VALIDATION_EXCEPTION",
    "message" : "60 Character Limit",
    "fields" : [ "Name" ]
  } ]
} ]
```

wmathurin:
You can change the API version your client is using, you don't have to use the default that the Mobile SDK ships with.
Of course if you do, some of the more recent REST wrapper might not work as they expect the API version to be above a certain value.

We don't duplicate in the Mobile SDK documentation what is already published in the Salesforce REST API docs. To see what is possible and not possible with a given API version, I recommend you check the [Salesforce REST API docs](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/intro_rest.htm) directly.

PR #3521, opened 2022-06-09
iOS 15 window crash

bbirman:
There's a crash reported on iOS 15 where `[UIWindow dealloc]` triggers `[SFSDKUIWindow disableWindow]` and then looks like it hits [this check](https://github.com/apple-oss-distributions/objc4/blob/8701d5672d3fd3cd817aeb84db1077aafe1a1604/runtime/objc-weak.mm#L420) when setting `super.rootViewController = nil` which crashes because UIWindow is deallocating. We don't have any repro steps with an app but I could hit the same crash point using the unit test in this PR on iOS 15 while it runs successfully on iOS 14 without any SDK changes. Adding a guard for now until we can get more information

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
@@ -528,12 +528,15 @@ static NSString *const kSFScreenLockWindowKey = @"screenlock";
 }
 @end
 
-@implementation SFSDKUIWindow
+@implementation SFSDKUIWindow {
+    BOOL deallocating;
+}
 
 - (instancetype)initWithFrame:(CGRect)frame {
     self = [super initWithFrame:frame];
     if (self) {
         _windowName = @"NONAME";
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
```diff
@@ -597,6 +606,10 @@ - (void)disableWindow {
     }
 }
 
+- (void)dealloc {
+    deallocating = YES;
+}
```
wmathurin, line 611: I thought you were missing [super dealloc] but you are not (https://stackoverflow.com/questions/14390652/overriding-the-dealloc-method-in-objective-c).

wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.m
bbirman: thanks for double checking :)

PR #3523, opened 2022-06-14
Bulk checking last modified date during collection sync up

wmathurin:
During collection sync up, we no longer check records last modified date one at a time.
Instead we leverage sobject collection retrieve to get last modified date by batches of up to 2000.

SFAdvancedSyncUpTask and its super classes are project objective-c classes. So I made the changes needed there in objective-c. However CollectionSyncUpTarget is written in Swift so the new code to get last modified date using sobject collection retrieve is written in Swift.

Other changes in this PR:
- using short hand notations for array and dictionary (e.g. [String:String] instead of Dictionary<String, String>),
- RecordRequest and RecordResponse are no longer inner classes of CompositeRequestHelper,
- made changes in CompositeRequestHelper (from last PR) to fix compilation errors in templates.

### DIFF
--- a/libs/MobileSync/MobileSync/Classes/CollectionSyncUpTarget.swift
+++ b/libs/MobileSync/MobileSync/Classes/CollectionSyncUpTarget.swift
@@ -158,16 +158,17 @@ public class CollectionSyncUpTarget: BatchSyncUpTarget {
                                 recordIdToLastModifiedDate[storeId] = SFRecordModDate(timestamp: nil, isDeleted: true)
                             }
                         }
+                        print("--> leaving group")
                         group.leave()
                     } else {
 //                        errorBlock(nil)
                     }
                 }
             }
+        }
                 
-            group.notify(queue: DispatchQueue.global()) {
-                completeBlock(recordIdToLastModifiedDate)
-            }
+        group.notify(queue: DispatchQueue.global()) {
+            completeBlock(recordIdToLastModifiedDate)
         }
     }
 }
--- a/libs/MobileSync/MobileSync/Classes/CollectionSyncUpTarget.swift
+++ b/libs/MobileSync/MobileSync/Classes/CollectionSyncUpTarget.swift
@@ -27,8 +27,9 @@
 
 import Foundation
 import SalesforceSDKCore
-import Combine
 
+public typealias SyncUpRecordsNewerThanServerBlock = (Dictionary<AnyHashable,Any>) -> ()
+typealias FetchLastModDatesBlock = (Dictionary<NSNumber, SFRecordModDate>) -> ()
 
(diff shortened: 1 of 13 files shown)


wmathurin:
NB: there is a test flapping testSyncUpManyLocallyCreatedRecords which I'm looking into.

libs/MobileSync/MobileSync/Classes/CollectionSyncUpTarget.swift
```diff
@@ -38,35 +39,136 @@ public class CollectionSyncUpTarget: BatchSyncUpTarget {
...
+
+            resultBlock(storeIdToNewerThanServer)
+        }
+    }
+    
+    func getRecordType(_ record:[AnyHashable: Any]) -> String? {
+        return SFJsonUtils.project(intoJson: record, path: MobileSync.kObjectTypeField) as? String
+    }
+    
+    func fetchLastModifiedDates(_ syncManager:SyncManager,
```
wmathurin, line 107: This is the new method that gets last modified dates in bulk.

libs/MobileSync/MobileSync/Classes/Manager/SFAdvancedSyncUpTask.m
```diff
@@ -32,72 +33,87 @@ -(instancetype) init:(SFMobileSyncSyncManager*)syncManager sync:(SFSyncState*)sy
...
+    SFSyncUpTarget *target = (SFSyncUpTarget *)sync.target;
+    NSString* soupName = sync.soupName;
+    sync.totalSize = recordIds.count;
+    
+    NSArray<NSDictionary*>* records = [target getFromLocalStore:self.syncManager
+                                                       soupName:soupName
+                                                       storeIds:recordIds];
+    
+    // Figuring out what records need to be synced up based on merge mode and last mod date on server
+    [self shouldSyncUpRecords:self.syncManager target:target records:records options:sync.options resultBlock:^(NSDictionary * _Nonnull recordIdToShouldSyncUp) {
```
wmathurin, line 45: Same refactoring we did on Android.

wmathurin: To allow sync up target to decide if records should be synched in bulk (but only CollectionSyncUpTarget actually does the checking in bulk). Existing sync up target check the old way. Two reasons for leaving them alone:
- more motivation for customer to adopt the new CollectionSyncUpTarget
- not break customer that might have store multiple types of objects in the sam e soup: it's not forbidden and works with BatchSyncUpTarget but is not allowed with the new CollectionSyncUpTarget.

bbirman (approved):

PR #3529, opened 2022-08-04
Screen lock timeout

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/ScreenLock/ScreenLockManager.swift
```diff
@@ -123,7 +155,7 @@ public class ScreenLockManager: NSObject {
         
         userAccounts?.forEach({ account in
             let hasMobilePolicy = account.idData.mobileAppPinLength > 0 && account.idData.mobileAppScreenLockTimeout != -1
-            self.storeMobilePolicy(userAccount: account, hasMobilePolicy: hasMobilePolicy)
+            self.storeMobilePolicy(userAccount: account, hasMobilePolicy: hasMobilePolicy, lockTimeout: account.idData.mobileAppScreenLockTimeout)
```
brandonpage, line 158: Should we extend this out to Mobile SDK 12.0?   If a user is on `10.1.0` and goes straight to `11.0` they will miss this upgrade.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1544,14 +1544,15 @@ - (void)retrievedIdentityData:(SFSDKAuthSession *)authSession {
...
     
     __weak typeof(self) weakSelf = self;
     [self dismissAuthViewControllerIfPresentForScene:authSession.oauthRequest.scene completion:^{
           __strong typeof(weakSelf) strongSelf = weakSelf;
         [strongSelf finalizeAuthCompletion:authSession];
         if (authSession.authInfo.authType != SFOAuthTypeRefresh) {
-            [[SFScreenLockManager shared] storeMobilePolicyWithUserAccount:self.currentUser hasMobilePolicy:hasMobilePolicy];
-            [[SFScreenLockManager shared] handleAppForeground];
+            [[SFScreenLockManager shared] storeMobilePolicyWithUserAccount:self.currentUser hasMobilePolicy:hasMobilePolicy lockTimeout:lockTimeout];
+            [[SFScreenLockManager shared] handleAppForegroundWithLockTimeout:NO];
```
brandonpage, line 1555: I don't quite understand why the `lockTimeout` boolean is needed.  Just to force lock on initial login?

bbirman: yup, though since you mentioned that it should only be with a passcode user, this should be updated. Or could go with the other suggestion to  reset the timestamp to 0 in `storeMobilePolicy`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/ScreenLock/ScreenLockManager.swift
bcriscuolo-sf: Is this something that would be revisited at the time of MobileSDK 12? “If a user is on X and goes straight to Y…” may hold over time, until a baseline version is established. We know that SDK upgrades lag significantly.

brandonpage: @bcriscuolo-sf  No.  This is the exact reason why apps are not allowed to skip major versions.  You cannot upgrade from 9.X to 11.X without upgrading users to some 10.X version in between.  You have an entire year between major releases, there is absolutely no excuse.

bcriscuolo-sf: @brandonpage EXCELLENT. Fully justified there. Is that documented somewhere (not that it will be a problem for SFS ever, but rather so "you" don't have to explain it again)?

brandonpage (approved):

PR #3534, opened 2022-08-25
@W-11296703 - Crash when writing event in analytics event store manager.

mbwimmer:
Crash that is being reported through AppCenter with 179 users and 208 reports:

```
0   CoreFoundation          __exceptionPreprocess + 220
1   libobjc.A.dylib         objc_exception_throw + 56
2   Foundation              -[NSData(NSData) writeToFile:options:error:] + 308
3   SalesforceAnalytics     -[SFSDKEventStoreManager storeEvent:] (SFSDKEventStoreManager.m:100)
4   SalesforceSDKCore       +[SFSDKEventBuilderHelper createAndStoreEvent:userAccount:className:attributes:] (SFSDKEventBuilderHelper.m:64)
5   SalesforceSDKCore       -[SFUserAccountManager handleAnalyticsAddUserEvent:account:] (SFUserAccountManager.m:1707)
6   SalesforceSDKCore       -[SFUserAccountManager finalizeAuthCompletion:] (SFUserAccountManager.m:1637)
7   SalesforceSDKCore       __46-[SFUserAccountManager retrievedIdentityData:]_block_invoke (SFUserAccountManager.m:1553)
8   SalesforceSDKCore       -[SFSDKWindowManager makeTransparentWithCompletion:completion:] (SFSDKWindowManager.m:400)
9   SalesforceSDKCore       -[SFSDKWindowManager dismissWindow:animated:withCompletion:] (SFSDKWindowManager.m:359)
10  SalesforceSDKCore       -[SFSDKWindowContainer dismissWindowAnimated:withCompletion:] (SFSDKWindowContainer.m:101)
11  SalesforceSDKCore       -[SFUserAccountManager dismissAuthViewControllerIfPresentForScene:completion:] (SFUserAccountManager.m:621)
12  SalesforceSDKCore       -[SFUserAccountManager retrievedIdentityData:] (SFUserAccountManager.m:1551)
13  SalesforceSDKCore       -[SFUserAccountManager identityCoordinatorRetrievedData:] (SFUserAccountManager.m:773)
14  SalesforceSDKCore       __48-[SFIdentityCoordinator notifyDelegateOfSuccess]_block_invoke (SFIdentityCoordinator.m:200)
```

@bbirman @wmathurin @bhariharan

### DIFF
--- a/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Store/SFSDKEventStoreManager.m
+++ b/libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Store/SFSDKEventStoreManager.m
@@ -96,14 +96,26 @@
     if (encryptedData) {
         NSString *filename = [self filenameForEvent:eventCopy.eventId];
         NSString *parentDir = [filename stringByDeletingLastPathComponent];
-        [[NSFileManager defaultManager] createDirectoryAtPath:parentDir withIntermediateDirectories:YES attributes: @{ NSFileProtectionKey: NSFileProtectionCompleteUntilFirstUserAuthentication } error:&error];
-        [encryptedData writeToFile:filename options:NSDataWritingFileProtectionCompleteUntilFirstUserAuthentication error:&error];
+        NSFileManager *fileManager = [NSFileManager defaultManager];        
+        [fileManager createDirectoryAtPath:parentDir withIntermediateDirectories:YES attributes: @{ NSFileProtectionKey: NSFileProtectionCompleteUntilFirstUserAuthentication } error:&error];
+        
         if (error) {
-            [SFSDKAnalyticsLogger w:[self class] format:@"Error occurred while writing to file: %@", error.localizedDescription];
-        } else {
-            @synchronized (self.eventCountMutex) {
-                self.numStoredEvents++;
+            [SFSDKAnalyticsLogger w:[self class] format:@"Error occurred while trying to create directory: %@", error.localizedDescription];
+            return;
+        }
+        
+        @try {
+            [encryptedData writeToFile:filename options:NSDataWritingFileProtectionCompleteUntilFirstUserAuthentication error:&error];
+            if (error) {
+                [SFSDKAnalyticsLogger w:[self class] format:@"Error occurred while writing to file: %@", error.localizedDescription];
+            } else {
+                @synchronized (self.eventCountMutex) {
+                    self.numStoredEvents++;
+                }
             }
+        }
+        @catch (NSException *e) {
+            [SFSDKAnalyticsLogger w:[self class] format:@"Exception occurred while writing to file: %@", e.reason];
         }
     }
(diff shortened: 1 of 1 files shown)


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Store/SFSDKEventStoreManager.m
```diff
@@ -96,15 +96,27 @@ - (void) storeEvent:(SFSDKInstrumentationEvent *) event {
...
         if (error) {
-            [SFSDKAnalyticsLogger w:[self class] format:@"Error occurred while writing to file: %@", error.localizedDescription];
-        } else {
-            @synchronized (self.eventCountMutex) {
-                self.numStoredEvents++;
+            [SFSDKAnalyticsLogger w:[self class] format:@"Error occurred while trying to create directory: %@", error.localizedDescription];
+            return;
+        }
+        
+        @try {
```
bhariharan, line 107: Other than restructuring the code, it looks like the only change here is adding a `try-catch` block. Do we know what exception is thrown, and under what circumstances this happens? Are we able to reproduce this issue? I'm wondering if we could address the root cause without adding a `try-catch`.

mbwimmer: @bhariharan I did some more digging, and I can reproduce the exact same stack trace with a crash if the `fileName` property is nil. Digging a little more, it looks like the file name is constructed from a constructor property that comes from the `SFSDKSalesforceAnalyticsManager` init code:
```
NSString *rootStoreDir;
if (_userAccount != nil) {
    rootStoreDir = [[SFDirectoryManager sharedManager] directoryForUser:userAccount type:NSDocumentDirectory components:@[ kEventStoresDirectory ]];
} else {
    rootStoreDir = [[SFDirectoryManager sharedManager] globalDirectoryOfType:NSDocumentDirectory components:@[ kEventStoresDirectory ]];
}
```

It looks like the directory manager has several paths that could return nil in either of these two cases. I have also confirmed that the code changes in this PR prevent the crash. I'm assuming the side effects are that analytics events are not recorded. I would probably suggest someone with more domain knowledge of these singleton classes take a look as to why the directory manager returns a nil store directory.

bhariharan: @mbwimmer Thank you for the analysis. I understand that this could occur if `filename` is `nil`, but looking at that code, `filename` should never be `nil`. Are you able to repro a situation where `filename` is `nil`? In other words, `filename` could only be `nil` is the `userAccount` object is not `nil` but it's corrupted inside and has `nil` properties, and this isn't valid. So, I'm wondering how this ever happens.

wmathurin:
Closing for lack of activity / response to our last questions.

PR #3539, opened 2022-09-09
Minor refactoring of EncryptionTests

AsceticMonk:
**Changes**
- Removed unnecessary import
- Removed some unnecessary assertions
- Removed some overlapping tests

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/EncryptionTests.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/EncryptionTests.swift
@@ -27,7 +27,6 @@
 
 import XCTest
 @testable import SalesforceSDKCore
-import CryptoKit
 
 class EncryptionTests: XCTestCase {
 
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCoreTests/EncryptionTests.swift
```diff
@@ -48,35 +46,21 @@ class EncryptionTests: XCTestCase {
         
         let decryptedData = try Encryptor.decrypt(data: encryptedData, using: keyAgain)
         let decryptedString = String(data: decryptedData, encoding: .utf8)
-        
         XCTAssertEqual(decryptedString, sensitiveInfo)
     }
     
     func testEncryptDecryptWrongKey() throws {
         let key = try KeyGenerator.encryptionKey(for: "test1")
-        XCTAssertNotNil(key)
```
AsceticMonk, line 57: I don't think this is needed because `encryptionKey(for:)` method does not return an optional.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/EncryptionTests.swift
```diff
@@ -48,35 +46,21 @@ class EncryptionTests: XCTestCase {
...
-    func testKeyRetrieval() throws {
-        let key = try KeyGenerator.encryptionKey(for: "test1")
-        XCTAssertNotNil(key)
-        let keyAgain = try KeyGenerator.encryptionKey(for: "test1")
-        XCTAssertEqual(key, keyAgain)
-        
-        let differentKey = try KeyGenerator.encryptionKey(for: "test2")
-        XCTAssertNotNil(differentKey)
-        XCTAssertNotEqual(key, differentKey)
-    }
```
AsceticMonk, line 78: These are already covered in previous 2 tests.

bbirman (approved):

PR #3542, opened 2022-09-21
Update sqlcipher to 4.5.2 and depend on sqlcipher instead of sqlcipher/fts in podspec

wmathurin:


bbirman (approved):

libs/SmartStore/SmartStoreTests/SFSmartStoreTests.m
```diff
@@ -88,7 +88,6 @@ - (void) testCompileOptions
     XCTAssertTrue([options containsObject:@"ENABLE_FTS4"]);
     XCTAssertTrue([options containsObject:@"ENABLE_FTS3_PARENTHESIS"]);
     XCTAssertTrue([options containsObject:@"ENABLE_FTS5"]);
-    XCTAssertTrue([options containsObject:@"ENABLE_JSON1"]);
```
wmathurin, line 91: Flag no longer returned on both iOS and Android / json1 is now always "baked in".

PR #3548, opened 2022-09-30
Set kdf_iter for current connection, not global default

threeve:
Using `PRAGMA cipher_default_kdf_iter` changes the global default value, which affects all databases opened from that point forward. This is dangerous when Mobile SDK is used in an application where any other code uses SQLCipher directly, as connections to those databases will unexpectedly have a non-default value for `kdf_iter`.

This change sets `kdf_iter` only for the specific database connection being opened. Per the SQLCipher docs, this needs to happen after the key is set.

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
@@ -199,12 +199,12 @@ static NSString * const kSFSmartStoreVerifyReadDbErrorDesc = @"Could not read fr
 
 + (FMDatabase*) unlockDatabase:(FMDatabase*)db key:(NSString*)key salt:(NSString *)salt {
     if ([db open]) {
-        // Using sqlcipher 2.x kdf iter because 3.x default (64000) and 4.x default (256000) are too slow
-        [[db executeQuery:@"PRAGMA cipher_default_kdf_iter = 4000"] close];
-       
         if (key)
            [db setKey:key];
         
+        // Using sqlcipher 2.x kdf iter because 3.x default (64000) and 4.x default (256000) are too slow
+        [[db executeQuery:@"PRAGMA kdf_iter = 4000"] close];
+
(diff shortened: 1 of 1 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
```diff
@@ -199,12 +199,12 @@ + (FMDatabase*) setKeyForDb:(FMDatabase*) db key:(NSString *)key salt:(NSString
 
 + (FMDatabase*) unlockDatabase:(FMDatabase*)db key:(NSString*)key salt:(NSString *)salt {
     if ([db open]) {
-        // Using sqlcipher 2.x kdf iter because 3.x default (64000) and 4.x default (256000) are too slow
-        [[db executeQuery:@"PRAGMA cipher_default_kdf_iter = 4000"] close];
-       
         if (key)
            [db setKey:key];
         
+        // Using sqlcipher 2.x kdf iter because 3.x default (64000) and 4.x default (256000) are too slow
```
wmathurin, line 205: Should it happen before setKey?

wmathurin (commented):
We should do the same on Android (https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SmartStore/src/com/salesforce/androidsdk/smartstore/store/DBOpenHelper.java#L382).

libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
threeve: @wmathurin The [docs explicitly say](https://www.zetetic.net/sqlcipher/sqlcipher-api/#kdf_iter) it should happen after setting the key: 

> PRAGMA kdf_iter must be called after PRAGMA key and before the first actual database operation or it will have no effect.

wmathurin: 👍 
That makes sense. cipher_default_kdf_iter being "global" could happen earlier.

wmathurin (approved):

wmathurin:
I can take care of the Android change.

PR #3558, opened 2022-12-06
Handle shareBrowserSessionIOS flag

sdavalaSF:
**Background:**

- For MyDomain an additional Auth Config: "shareBrowserSessionIOS" was added in: [W-11606434](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000014FkazYAC/view)

**Changes:**
To handle "shareBrowserSessionIOS",  `prompt=login` should be added to the approvalUrl when this setting is false, when setting is true nothing should be added to url. If `shareBrowserSessionIOS` is not present treat as true.


bhariharan:
@sdavalaSF What is this feature, and why are we adding this here? I'd like more information on this before we add this in the public SDK. Also, why's this specific to iOS?

sdavalaSF:
- Closing PR after discussions of being Platform specific and needing to verify additional flows.

PR #3562, opened 2022-12-20
Pma 2270 token fix

HEBPartnerNetDeveloper:
Apply patch from version 10.2 to fix access token error


HEBPartnerNetDeveloper:
Opened by accident

Issue #3572, opened 2023-02-20
SFPushNotificationManager deviceToken always nil at startup


HannesVDB:
Please fill out the following details:

1. Version of Mobile SDK Used: 10.1
2. Issue found in Native App or Hybrid App: Native iOS App
3. OS Version: 16.2
4. Device: iPhone 13 and iPhone 14 Pro Max Simulator
5. Steps to reproduce:
- Authenticate the user via `AuthHelper.loginIfRequired`

```
        AuthHelper.loginIfRequired {
        //// Some internal code
         }
```
- Enable push notifications:
```    func requestPushNotificationPermission() {
        PushNotificationManager.sharedInstance().registerForRemoteNotifications()
        let center = UNUserNotificationCenter.current()
        center.requestAuthorization(options: [.alert, .sound, .badge]) { [weak self] granted, error in
           //// Some internal code
            self?.retreiveNotificationSettings()
        }
    }
    
    func retreiveNotificationSettings() {
      UNUserNotificationCenter.current().getNotificationSettings { settings in
          guard settings.authorizationStatus == .authorized else {
              DispatchQueue.main.async {
                  self.openMain()
              }
              return
          }
          DispatchQueue.main.async {
              UIApplication.shared.registerForRemoteNotifications()
              self.openMain()
          }
      }
    }
```
- Register and save push notification token
```
    func application(_ application: UIApplication, didRegisterForRemoteNotificationsWithDeviceToken deviceToken: Data) {
        let tokenParts = deviceToken.map { data in String(format: "%02.2hhx", data) }
        let token = tokenParts.joined()
        FirebaseLogger().log(event: "register_push_notifications", parameters: ["token": token])
            PushNotificationManager.sharedInstance().didRegisterForRemoteNotifications(withDeviceToken: deviceToken)
                PushNotificationManager.sharedInstance().registerForSalesforceNotifications { result in
                    switch result {
                    case .success(let success):
                        Logger.notifications.info("📲 Registerd successfully \(success)")
                        Analytics.logEvent("registered_for_push", parameters: ["Success": success])
                    case .failure(let error):
                         Logger.notifications.error("📲 Failed register: \(error.localizedDescription)")
                    }
                }
    }
```
- Later in app, if the app has restarted: request the notifiation token:
`var savedToken = PushNotificationManager.sharedInstance().deviceToken`
This token is always nil!

bbirman:
[...]

PR #3573, opened 2023-02-21
Persist values for push notifications

bbirman:
For #3572

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
@@ -109,6 +109,7 @@ static NSString * const kSFAppFeaturePushNotifications = @"PN";
     [SFSDKCoreLogger i:[self class] format:@"Registration with Apple for remote push notifications succeeded"];
     _deviceToken = [NSString stringWithHexData:deviceTokenData];
     [[SFPreferences currentUserLevelPreferences] setObject:_deviceToken forKey:kSFDeviceToken];
+    [[SFPreferences currentUserLevelPreferences] synchronize];
 }
 
 #pragma mark - Salesforce registration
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m
```diff
@@ -109,6 +109,7 @@ - (void)didRegisterForRemoteNotificationsWithDeviceToken:(NSData*)deviceTokenDat
     [SFSDKCoreLogger i:[self class] format:@"Registration with Apple for remote push notifications succeeded"];
     _deviceToken = [NSString stringWithHexData:deviceTokenData];
     [[SFPreferences currentUserLevelPreferences] setObject:_deviceToken forKey:kSFDeviceToken];
+    [[SFPreferences currentUserLevelPreferences] synchronize];
```
bbirman, line 112: It looks like both of these values are intended to be persisted based on the init below but I couldn't find any other calls to `synchronize` https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/a29fdb259318e45180e69547aefeeb054f6905eb/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m#L71-L75

wmathurin: So they were never saved to disk? What scenario does it break? How come we didn't notice for so long?

bbirman: It looks like our sample apps always register for push notifications at start up so it would never rely on the restored value. For a local test I changed it so that I logged in -> registered -> killed the app -> launched again without registering and hit the issue, but we can also ask what the customer's flow was

wmathurin: The fix makes sense. The problem probably got unnoticed because it is hard to hit.

wmathurin (approved):

HannesVDB:
Thanks for fixing this!

In our application we do not ask notification permission immediately at startup. It's requested later on in some kind of onboarding.
On the other hand I had a bug in my `AppDelegate` not correctly registering the push notification token to salesforce due to an Authorisation Error.. So I wanted to solve the bug by registering again with the stored value in the preferences.. But the preferences were always empty. So that's why I stumbled onto this issue.

PR #3582, opened 2023-04-14
IDP - SP auth using keychain

bbirman:
- Expanded keychain APIs to configure cache and access group, deprecated similar access group that isn't used
- Added new command for IDP to SP with keychain, renamed existing commands to line up with Android
- If the keychain group is set on the SP, the SP will store a value in the keychain to track if it's logged in or not so that the IDP knows when to send a user hint command and when to send an auth code


libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainHelper.swift
```diff
@@ -103,9 +168,13 @@ public class KeychainHelper: NSObject {
...
     /// - Returns: KeychainResult
     @objc public class func reset(service: String, account: String?) -> KeychainResult {
+        return reset(service: service, account: account, accessGroup: accessGroup, cacheMode: .unspecified)
+    }
+    
+    @objc public class func reset(service: String, account: String?, accessGroup: String? = nil, cacheMode: CacheMode = .unspecified) -> KeychainResult {
         self.upgradeIfRequired()
         
-        return CachedWrapper.wrapRemoves(service, account) { service, account in
+        let keychainReset: KeychainOperation = { service, account in
```
wmathurin, line 177: No access group passed to KeychainItemManager?

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainHelper.swift
```diff
@@ -115,17 +184,28 @@ public class KeychainHelper: NSObject {
...
     /// - Returns: KeychainResult
     @objc public class func remove(service: String, account: String?) -> KeychainResult {
+        return remove(service: service, account: account, accessGroup: accessGroup, cacheMode: .unspecified)
+    }
+    
+    @objc public class func remove(service: String, account: String?, accessGroup: String? = nil, cacheMode: CacheMode = .unspecified) -> KeychainResult {
         self.upgradeIfRequired()
         
-        return CachedWrapper.wrapRemoves(service, account) { service, account in
+        let keychainRemove: KeychainOperation = {service, account in
```
wmathurin, line 208: No access group passed to KeychainItemManager?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/IDP/SFSDKAuthResponseCommand.m
```diff
@@ -60,4 +60,20 @@ - (NSString *)command {
...
 
+- (NSString *)keychainReference {
+    return [self paramForKey:kSFKeychainReferenceParam];
+}
+
+- (void)setKeychainReference:(NSString *)keychain {
+    [self setParamForKey:keychain key:kSFKeychainReferenceParam];
+}
+
+- (NSString *)userHint {
```
wmathurin, line 71: Isn't the auth response command going from SP to IDP (in response to a IDP initiated login request)?
If that's true, why is the user hint there?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -319,15 +325,81 @@ - (void) setLoginViewControllerConfig:(SFSDKLoginViewControllerConfig *)config {
 
 #pragma  mark - login & logout
 
-- (BOOL)handleIDPAuthenticationResponse:(NSURL *)appUrlResponse options:(nonnull NSDictionary *)options {
-    [SFSDKCoreLogger d:[self class] format:@"handleIDPAuthenticationResponse %@",[appUrlResponse description]];
+- (BOOL)handleIDPAuthenticationResponse:(NSURL *)appUrlResponse options:(NSDictionary *)options {
```
wmathurin, line 328: I think Android and iOS did not name the various requests / responses which is okay, may be some comments might help.

For reference, on Android we have:
- IDPLoginRequest: sent by IDP to SP for IDP initiated login flow (contains a user hint)
- IDPLoginResponse: sent by SP to IDP at the end of the flow (either right after the IDPLoginRequest if the user hinted was available, or after getting the code and logging in) => probably no equivalent on iOS 
- SPLoginRequest: sent by SP to IDP for SP initiated login flow or during IDP initiated login flow if SP does not have the hinted user (contains code challenge)
- SPLoginResponse: sent by IDP to IDP in response to SPLoginRequest (contains code)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.h
```diff
@@ -513,6 +513,10 @@ Use this method to stop/clear any authentication which is has already been start
  */
 - (BOOL)handleIDPAuthenticationResponse:(NSURL *)url options:(nonnull NSDictionary *)options NS_SWIFT_NAME(handleIdentityProviderResponse(from:with:));
 
+- (BOOL)handleIDPAuthenticationResponse:(NSURL *)url options:(nonnull NSDictionary *)options completion:(nullable void (^)(void))completion NS_SWIFT_NAME(handleIdentityProviderResponse(from:with:completion:));
+
+- (void)initiateSPAuthentication:(NSString *)spConsumerKey redirectURI:(NSString *)spRedirectURI scopes:(NSSet<NSString *> *)spScopes;
```
wmathurin, line 518: Where does the IDP get the sp consumer key and redirect uri?

bbirman: I thought the IDP app would know that and pass it in?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/IDP/SFSDKAuthResponseCommand.m
bbirman: I'm not sure about the naming history, but it's used to open the SP app  here https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/664d49497bcd6d02282a6c6d4fc853640c8981ac/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/IDP/SFSDKIDPAuthHelper.m#L91

I was also thinking about breaking them into more specific commands instead of one, I could try aligning the naming more to Android

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
bbirman: thanks!

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainHelper.swift
```diff
@@ -66,18 +97,34 @@ public class KeychainHelper: NSObject {
...
+    /// Create an item in the keychain if not present.
+    /// - Parameters:
+    ///   - service: Identifier to use for keychain service key.
+    ///   - account: Identifier to use for keychain account key.
+    /// - Returns: KeychainResult
+    @objc public class func createIfNotPresent(service: String, account: String?, accessGroup: String? = nil, cacheMode: CacheMode = .unspecified) -> KeychainResult {
         self.upgradeIfRequired()
         
-        return CachedWrapper.wrap(service, account) { service, account in
+        let keychainRead: KeychainOperation = { service, account in
```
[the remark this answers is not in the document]
bbirman: I updated the wrapper func param name to be more generic too since it's shared between reads and creates

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.h
wmathurin: You are right :-)
On Android, SalesforceSDKManager has a setAllowedSPApps(List<SPConfig> allowedSPApps) method.
To kick off a IDP initiated login flow, you just pass the SP app name (part of the SPConfig).
During SP initiated login flow, IDP only services a request from a SP app if it is one of the allowed ones.
But I guess SP initiated login is out of scope.

bbirman: ahh, I thought SP initiated would use the existing two hop flow since iOS doesn't have the broadcast option, did I misunderstand? also want me to make the params into an SPConfig object instead to match more?

wmathurin: I think you are right. (I will end up having to re-introduce the two hops for Android also since the code cannot be obtained without user interactions in all cases).
Would the SP's access group name be part of the SPConfig then?

bbirman: Having the group on SPConfig sounds good -- so the IDP would use the group from the config and the SP would still set the group on SalesforceSDKManager so it can write the "loggedIn" key?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
bbirman: I updated existing command names to line up more with Android, and then added `SFSDKIDPAuthCodeLoginRequestCommand` => sent by IDP to SP for IDP initiated login flow using the keychain to store the code verifier

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/IDP/Commands/SFSDKSPLoginRequestCommand.h
```diff
@@ -43,12 +44,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic,copy) NSString *spUserHint;
 
-@property (nonatomic,copy) NSString *callingAppUrl;
```
bbirman, line 46: These are internal and I couldn't find anywhere they're used

brandonpage (approved):
At a relatively hight level, LGTM.

PR #3586, opened 2023-04-27
Biometric Authentication

brandonpage:
- New connected app controlled "Biometric Authentication" lock screen that forces user re-authentication or biometric to unlock.  Auth token is not refreshed when the app is locked.
- Added functions in SalesforceSDKManager that return public ScreenLockManager and BiometricAuthenticationManager protocols backed by internal classes.
- Biometric enrollment dialog is shown upon initial login.  Can be displayed again by the app.
- Native biometric unlock button shown by default but can be disabled by the app.

TODO:
- [x] Handle web scheme to prompt biometric.
- [x] Revoke old sessions if user re-authenticates before session is expired.
- [x] Cleanup on logout.
- [x] Handle multiuser scenarios.
- [x] State Notifications
- [x] Write Tests.

<p float="left">


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -257,18 +270,65 @@ - (UIInterfaceOrientationMask)supportedInterfaceOrientations {
...
+            UIButton *button = [UIButton buttonWithType:UIButtonTypeCustom];
+            [button setTitle:[SFSDKResourceUtils localizedString:@"biometricLoginButton"] forState:UIControlStateNormal];
+            x = 0;
+            y = (self.view.bounds.size.height - y);
+            w = self.view.bounds.size.width;
+            h = ((self.view.bounds.size.height - y) * 0.1);
+            button.frame = CGRectMake(x, y, w, h);
+            [button addTarget:self action:@selector(presentBioAuthAction:) forControlEvents:UIControlEventTouchUpInside];
+            [self.view addSubview:button];
+        }
```
brandonpage, line 295: There is probably a better way to do this.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1119,15 +1121,6 @@ - (SFUserAccount *)userAccountForUserIdentity:(SFUserAccountIdentity *)userIdent
     [_accountsLock lock];
     result = (self.userAccountMap)[userIdentity];
 
-    // TODO: Remove this block in Mobile SDK 10.0, needed for upgrade step after changing from 15 character to 18 character user IDs
```
bbirman, line 1122: 🙌

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -311,11 +313,12 @@ - (void)enqueueRequest:(SFRestRequest *)request requestDelegate:(id<SFRestReques
                 [strongSelf notifyDelegateOfSuccess:requestDelegate request:request data:dataForDelegate rawResponse:response];
             } else {
                 if (shouldRetry && statusCode == 401) {
-
-                    // 401 indicates refresh is required.
-                    [strongSelf replayRequest:request response:response requestDelegate:requestDelegate];
+                    // Do not refresh token if biometric authentiction lock is enabled
```
bbirman, line 316: does `notifyDelegateOfFailure:` need to be called if the lock is enabled? or is it not the same as a failure?

brandonpage: Good call, I will combine the if surrounding the replay request with the one above.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
wmathurin: @bbirman do we have a "Login with IDP" button shown on the Login screen on iOS? If there is, how are we handling the case where "Login with IDP" and "Biometric Login" buttons both need to be shown?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -711,6 +723,33 @@ - (NSURLSession*)session {
     return _session;
 }
 
+- (void)presentBioAuth {
```
wmathurin, line 726: Isn't that a copy of the method in SFLoginViewController ?
Should we have a single implementation in a helper class?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/ScreenLock/ScreenLockManagerInternal.swift
```diff
@@ -0,0 +1,284 @@
...
+                return try JSONDecoder().decode(MobilePolicy.self, from: resultData).hasPolicy
+            } catch {
+                SFSDKCoreLogger.e(ScreenLockManagerInternal.self, message: "Failed to read Mobile policy for user.")
+            }
+        }
+        
+        return false
+    }
+
+    private struct MobilePolicy: Encodable, Decodable {
```
wmathurin, line 279: Would it make sense to have a single method that reads from the keychain and returns the MobilePolicy object.
In all the other places, we would call that method and then access the field of interest.
That way we wouldn't have the following repeated multiple times:
```swift
let result = KeychainHelper.read(service: kScreenLockIdentifier, account: userId)
if let resultData = result.data, result.success {
  do {
    try JSONDecoder().decode(MobilePolicy.self, from: resultData)
  } catch {
      SFSDKCoreLogger.e(ScreenLockManagerInternal.self, message: "Failed to read Mobile policy for user.")
  }
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
brandonpage: I almost did but looked a bit comical.  The utility would just be a method that makes a LAContext and calls evaluate on it.  There is significantly more code in the blocks the caller would need to supply for success or failure.  I will give it another shot though.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
bbirman: This is how it looks with IDP, so it might work out that if you select "Login Using the App", then you'd get the "Biometric Login" button on the regular login screen?
Screenshot 2023-05-09 at 9 16 49 PM

brandonpage: Yeah, I figured that screen had not changed.  It would be nice to have more consistent UX between Android and iOS.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
```diff
@@ -0,0 +1,222 @@
...
+        }
+        return nil
+    }
+    
+    func lock() {
+        locked = true
+        let accountManager = UserAccountManager.shared
+        let currentAccount = accountManager.currentUserAccount
+        
+        accountManager.switchToNewUserAccount { result in
```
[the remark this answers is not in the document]
brandonpage: Originally I had a reason for calling `switchToNewUserAccount` but the code has changed so I updated this to be more direct.

wmathurin (commented):
LGTM what's the "State Notification" TODO you have in the PR's description?

brandonpage:
> LGTM. What's the "State Notification" TODO you have in the PR's description?

Done.

wmathurin (approved):

PR #3590, opened 2023-05-11
Add option for web server auth in web view

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
```diff
@@ -264,7 +264,7 @@ typedef void (^SFOAuthBrowserFlowCallbackBlock)(BOOL);
 /**
  The current state of any in-progress advanced authentication flow.
  */
-@property (nonatomic, readonly) SFOAuthAdvancedAuthState advancedAuthState;
+@property (nonatomic, readonly) SFOAuthAdvancedAuthState advancedAuthState SFSDK_DEPRECATED(11.0, 12.0, "Will be removed");
```
bbirman, line 267: I don't see any internal app usage of this, thinking it could be retired?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
```diff
@@ -54,7 +54,7 @@ typedef NS_ENUM(NSUInteger, SFOAuthAdvancedAuthState) {
      initiated a token exchange request.
      */
     SFOAuthAdvancedAuthStateTokenRequestInitiated
-};
+} SFSDK_DEPRECATED(11.0, 12.0, "Will be removed");
```
wmathurin, line 57: That's going away ?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -398,10 +388,10 @@ - (void)continueNativeBrowserFlow {
     [self.delegate oauthCoordinator:self didBeginAuthenticationWithSession:_asWebAuthenticationSession];
 }
 
-- (void)beginUserAgentFlow {
+- (void)beginWebViewFlow {
```
bbirman, line 391: I was trying to figure out a name to encompass  user-agent and web server, went with this to pair up with `beginNativeBrowserFlow`, open to anything else

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
```diff
@@ -71,7 +71,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (NSString *)generateApprovalUrlString;
 
-- (void)beginUserAgentFlow;
+- (void)beginWebViewFlow;
```
wmathurin, line 74: It should probably be called beginWebServerFlow ?

wmathurin: Never mind. I misunderstood.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -508,7 +500,7 @@ - (void)updateCredentials:(NSDictionary *) params {
     [self.credentials updateCredentials:params];
 }
 
-- (void)beginTokenEndpointFlow:(SFOAuthTokenEndpointFlow)flowType {
+- (void)beginTokenEndpointFlow {
```
bbirman, line 503: doesn't look like the parameter is used here

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -653,42 +645,55 @@ - (void)handleUserAgentResponse:(NSURL *)requestUrl {
...
+    return [self approvalURLForEndpoint:[self brandedAuthorizeURL]
+                            credentials:self.credentials
+                          webServerFlow:[[[SalesforceSDKManager sharedManager] appConfig] useWebServerAuthentication]
+                               protocol:nil
+                                 domain:nil];
 }
 
-- (NSString *)generateApprovalUrlString:(SFOAuthCredentials *)credentials {
-    NSAssert(nil != credentials.domain, @"credentials.domain is required");
+- (NSString *)approvalURLForEndpoint:(NSString *)authorizeEndpoint
```
bbirman, line 655: I tried to consolidate the approval url logic here, currently the intent is to keep the same url each path was generating before but there might be more consolidation possible?
- I wasn't sure why the IDP protocol was hardcoded to "https" vs credentials.protocol
- There was one case that didn't use the branded authorize endpoint but I wasn't sure if that was intentional

wmathurin (commented):
LGTM
I wonder if web server flow should be default on or off.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.h
wmathurin: 👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
wmathurin: 👍

wmathurin:
Also I wonder if we should offer methods to set that flag (in addition to having it come from the config). That would allow customer to slowly roll it out through feature flags.

wmathurin:
I deviated slightly from your PR on Android.
See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2402:
- By default it will use "web server flow".
- By calling SalesforceSDKManager.getInstance().setUseWebServerAuthentication(false), an app can opt to use the (deprecated) user agent flow instead.
- The selected flow is shown in the dev info screen.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
```diff
@@ -66,6 +68,20 @@ - (SFAuthErrorHandlerList *)populateDefaultAuthErrorHandlerList
     __weak typeof(self) weakSelf = self;
     SFAuthErrorHandlerList *authHandlerList = [[SFAuthErrorHandlerList alloc] init];
     
+    
+    // User agent disabled handler
+    self.userAgentErrorHandler =
```
bbirman, line 73: Before, this error would fall into the invalid grant category which only logs on error, so I made a separate handler to catch it but then reusing the alert from the generic error handler block
Screenshot 2023-05-11 at 7 08 41 PM

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
```diff
@@ -66,6 +68,20 @@ - (SFAuthErrorHandlerList *)populateDefaultAuthErrorHandlerList
...
+        [[SFAuthErrorHandler alloc] initWithName:kSFUserAgentErrorHandler authSessionBlock:^BOOL(NSError *error, SFSDKAuthSession *authSession, NSDictionary *options) {
+            if ([error.localizedDescription containsString:@"user-agent flow has been disabled"]) {
```
bbirman, line 75: even though it's in "localizedDescription", it looks like it's always english (even if the device, org language and user language are not)

bbirman:
@wmathurin I updated to match you-- moved the toggle to SalesforceSDKManager, changed the default and added it in dev info

wmathurin (approved):

PR #3592, opened 2023-05-16
Using prompt=login during advanced authentication so that user must re-enter credentials

wmathurin:
Previously if a user did a logout followed by a login, they would go straight to the allow/deny screen. 

Changes were made initially in the internal fork of the Mobile SDK (see [here](https://git.soma.salesforce.com/iOS/SalesforceMobileSDK-iOS/pull/892)).

Configurable through auth config.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.m
@@ -26,6 +26,7 @@
 
 static NSString * const kAuthConfigMobileSDKKey        = @"MobileSDK";
 static NSString * const kAuthConfigUseNativeBrowserKey = @"UseiOSNativeBrowserForAuthentication";
+static NSString * const kAuthConfigShareBrowserSession = @"shareBrowserSessionIOS";
 static NSString * const kAuthConfigSamlProvidersKey    = @"SamlProviders";
 static NSString * const kAuthConfigAuthProvidersKey    = @"AuthProviders";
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthOrgAuthConfiguration.m
```diff
@@ -54,6 +55,15 @@ - (BOOL)useNativeBrowserForAuth {
     return [self.authConfigDict[kAuthConfigMobileSDKKey][kAuthConfigUseNativeBrowserKey] boolValue];
 }
 
+- (BOOL)shareBrowserSession {
+    if ([self.authConfigDict[kAuthConfigMobileSDKKey] objectForKey:kAuthConfigShareBrowserSession] != nil) {
+        return [self.authConfigDict[kAuthConfigMobileSDKKey][kAuthConfigShareBrowserSession] boolValue];
+    } else {
+        // return false which means prompt=login being appended to login url
+        return false;
```
bbirman, line 63: why's the fallback different than the internal fork?

wmathurin: The fork had a different fallback for iOS and Android.
I wanted the fallback to be prompt=login.
Also I checked with Maggie and she said "it doesn't matter today as servers are all upgraded to 242+, and server side will return false by default".

bbirman: 👍 thanks!

PR #3611, opened 2023-06-13
Bump deployment target to iOS 15

bbirman:


wmathurin:
We should also change 14->15 [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/mobilesdk_pods.rb#L70).

wmathurin:
In the package repo, I moved to xcode 14.3 (see https://github.com/forcedotcom/SalesforceMobileSDK-Package/commit/33bc3f1f531a31a72073f053e01a809a5e0c4a46).

.circleci/config.yml
```diff
@@ -2,12 +2,12 @@
 #  Each post contains a full image manifest, including iOS runtimes, devices, CocoaPods version, etc.
 #  NOTE: When updaing Xcode check the manifest for compatible chruby versions.
 anchors:
-  - &latest-xcode    "13.4.1"
-  - &latest-ios      "15.5"
-  - &min-ios         "14.5"
-  - &requres-string  "test << matrix.lib >> iOS 14.5"  # Keep this version up to date
+  - &latest-xcode    "14.1.0"
```
bbirman, line 5: It looks like the CircleCI 14.2.0+ images have some breaking changes so I updated to `14.1.0` to unblock the iOS 15 bump and then can play around with the config more for updating to `14.3.1`

wmathurin: 👍

wmathurin (approved):

PR #3613, opened 2023-06-14
Move keychain group to URL and ensure IDP doesn't trigger advanced auth

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -211,8 +211,6 @@ NS_SWIFT_NAME(SalesforceManager)
  */
 @property (nonatomic, copy, nullable) NSString *idpAppURIScheme NS_SWIFT_NAME(identityProviderURLScheme);
 
-@property (nonatomic, copy, nullable) NSString *idpKeychainGroup NS_SWIFT_NAME(identityProviderKeychainGroup);
```
bbirman, line 214: Originally added this so that the SP app could use it for user tracking, but we don’t need that anymore so we could pass the keychain group from SPConfig through in the URL

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/URLHandlers/SFSDKAdvancedAuthURLHandler.m
```diff
@@ -40,13 +42,13 @@ - (BOOL)canHandleRequest:(NSURL *)url options:(NSDictionary *)options {
    
     // Not IDP enabled but Auth Code is present represents an Advanced Auth Flow situations. Ensure that we are not looking at any errors in either case.
     return (!preferences.idpEnabled &&
```
bbirman, line 44: `idpEnabled` is based on the IDP url scheme being set but that’s not used for the keychain based IDP -> SP flow. I did this because we know the format of the request for that case but we could also keep the `idpKeychainGroup` property or something similar

wmathurin (approved):

PR #3614, opened 2023-06-16
Deprecate classes related to inactivity tracking

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFApplication.h
```diff
@@ -23,13 +23,15 @@
...
+#import <SalesforceSDKCore/SalesforceSDKConstants.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
 /**
  * Mobile SDK extension of the UIApplication object.  Currently just gathers
  * date/time information about user events.
  */
+SFSDK_DEPRECATED(11.0, 12.0, "Will be removed")
 @interface SFApplication : UIApplication
```
bbirman, line 35: Now the screen lock is based on time since backgrounding instead of activity, I think both of these classes can be removed. Bonus is that SDK apps could use `@main` instead of initializing `UIApplicationMain` with `SFApplication`

brandonpage (approved):
This is fine, but seems unnecessary.  No one is using this right?  Should we just remove it now instead of deprecating?

bbirman:
Good question, I doubt `SFInactivityTimerCenter` has much usage. `SFApplication` isn't necessary anymore but I think all of the consumers would have a reference to it in their main files like [this one](https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/77bdce367c5486e2347874290c93a2baa85afb6e/iOSNativeSwiftTemplate/iOSNativeSwiftTemplate/main.swift#L32) (or are subclassing it)

PR #3615, opened 2023-06-19
Remove old upgrade steps

bbirman:


libs/SmartStore/SmartStore/Classes/SFSmartStore+Internal.h
```diff
@@ -184,21 +184,11 @@ static NSString *const EXPLAIN_ROWS = @"rows";
  */
 - (NSNumber*)currentTimeInMilliseconds;
 
-/**
- @return The legacy key used to encrypt the store.
- */
-+ (NSString *)legacyEncKey;
```
bbirman, line 190: I removed the implementation in https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3591 but missed this internal header and the upgrade step that calls it

wmathurin (approved):

PR #3636, opened 2023-08-16
Update custom URL detection

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
```diff
@@ -39,7 +39,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, readonly, nullable) NSURLSession *session;
 @property (nonatomic, strong , nullable) NSMutableData *responseData;
 @property (nonatomic, assign) BOOL initialRequestLoaded;
-@property (nonatomic, assign) BOOL domainUpdated;
```
bbirman, line 42: This one was to make sure it didn't loop but with the more specific URL and another tweak the case is already avoided

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -763,11 +763,11 @@ - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigati
         if (bioAuthManager.locked && bioAuthManager.hasBiometricOptedIn) {
             [bioAuthManager presentBiometricWithScene:self.view.window.windowScene];
         }
-    } else if (!self.domainUpdated && [self shouldUpdateDomain:url]) {
+    } else if ([SalesforceSDKManager sharedManager].shouldInferCustomDomain && [self shouldUpdateDomain:url]) {
         // To support case where my domain is entered through "Use Custom Domain"
-        self.domainUpdated = YES;
         decisionHandler(WKNavigationActionPolicyCancel);
         [self stopAuthentication];
+        [[SFUserAccountManager sharedInstance] setLoginHost:url.host];
```
bbirman, line 770: The host was already getting updated when the current user is set, but I added it here too so that a user can see the change in the host picker before logging in

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -776,9 +776,11 @@ - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigati
 }
 
 - (BOOL)shouldUpdateDomain:(NSURL *)webviewURL {
-    NSURL *currentDomain = [NSURL URLWithString:self.credentials.domain];
     NSString *myDomainSuffix = @".my.salesforce.com";
-    return (![currentDomain.host hasSuffix:myDomainSuffix] && [webviewURL.host hasSuffix:myDomainSuffix]);
+    return (![self.credentials.domain hasSuffix:myDomainSuffix] &&
+            [webviewURL.host hasSuffix:myDomainSuffix] &&
+            [webviewURL.path isEqualToString:@"/"] &&
```
wmathurin, line 782: @brandonpage we should check for an empty path on Android as well, as it is done here (line 782).

wmathurin (commented):
LGTM
Maybe we could have an example regexp in a comment in the code?

wmathurin (approved):

PR #3642, opened 2023-09-17
iOS 17 test fixes

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKURLCacheTests.m
```diff
@@ -77,6 +77,12 @@ - (void)testNilURL {
     NSString *contentString = @"This is my content";
     NSData *contentData = [contentString dataUsingEncoding:NSUTF8StringEncoding];
     NSURL *url = [[NSURL alloc] initWithString:@"bad string -- will create nil URL"];
+    #if __IPHONE_OS_VERSION_MAX_ALLOWED >= 170000
+    if (@available(iOS 17.0, *)) {
+        url = [[NSURL alloc] initWithString:@"bad string -- will create nil URL" encodingInvalidCharacters:NO];
```
wmathurin, line 82: Why do we need the iOS version check now?

bbirman: It looks like they changed the behavior of the API for invalid URLs, from [release notes](https://developer.apple.com/documentation/Xcode-Release-Notes/xcode-15-release-notes):

> Fixed: For apps linked on or after iOS 17 and aligned OS versions, URL parsing has updated from the obsolete RFC 1738/1808 parsing to the same RFC 3986 parsing as URLComponents. This unifies the parsing behaviors of the URL and URLComponents APIs. Now, URL automatically percent- or IDNA-encodes invalid characters to help create a valid URL.

> To check if a urlString is strictly valid according to the RFC, use the new URL(string: urlString, encodingInvalidCharacters: false) initializer. This init leaves all characters as they are and will return nil if urlString is explicitly invalid. (93368104)

wmathurin: 👍 
May be a little comment in the code so we don't forget why we added the check.

brandonpage (approved):

PR #3676, opened 2024-01-12
Native Login

brandonpage:
### This PR includes:
- async `login(username: String, password: String)` function
- `fallbackToWebAuthentication` function
- `shouldShowBackButton` and `cancelAuthentication` functions
- Functions to enable Biometric Authentication via Native Login.
- New `nativeLogin` flag on UserAccount's idData.

### Usage: 
App calls `useNativeLogin` in AppDelegate :
```
SalesforceManager.shared.useNativeLogin(withConsumerKey: clientId, callbackUrl: redirectUri, communityUrl: loginUrl, nativeLoginViewController: vsc, scene: nil)
```
_Or_ the app can call the same function from `SceneDelegate` and provide a scene to enable multi-window support.

While this function takes a `UIViewController`, creating one from SwiftUI with `UIHostingController` works as expected.

In their own UI App calls:
SwiftUI
```
 let result = await SalesforceManager.shared.nativeLoginManager().login(username: username, password: password)
```
Objective c View Controller
```
[[[SalesforceSDKManager sharedManager] nativeLoginManager] loginWithUsername:username password:password completionHandler:^(enum NativeLoginResult result) {
        // handle results
}];
```

Login $~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~$ Webview Fallback $~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~$              Adv Auth Fallback
<p float="left">
  
   
  
</p>

### Working Related Features
- Screen Lock (a.k.a Passcode)
- Biometric Authentication - APIs added to give app full control.
- Deferred Login
- Multi-Window Support
- IDP works for standard user logins but is not possible for naive login users due to the nature of the profile configuration.

### Reviewer Considerations
- A couple APIs were made public due to `NativeLoginManagerInternal` being written in Swift.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -0,0 +1,204 @@
...
+        let codeVerifier = generateCodeVerifier()
+        guard let challenge = generateChallenge(codeVerifier: codeVerifier) else { return .unknownError }
+        let authRequestBody = "\(kSFOAuthResponseType)=\(kSFOAuthCodeCredentialsParamName)&\(kSFOAuthClientId)=\(self.clientId)&\(kSFOAuthRedirectUri)=\(redirectUri)&\(kSFOAuthCodeChallengeParamName)=\(challenge)"
+        authRequest.customHeaders = customHeaders
+        authRequest.setCustomRequestBodyString(authRequestBody, contentType: kHttpPostContentType)
+        authRequest.requiresAuthentication = false
+        authRequest.endpoint = ""
+   
+        // First REST Call - Authorization
+        let authResult = await withCheckedContinuation { continuation in
```
brandonpage, line 89: `await withCheckedContinuation` is straight up magic 🪄.  

I think we should consider wrapping all of our RestClient APIs so customers get that sweet Swift 5.5 syntax.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -289,19 +289,40 @@ NS_SWIFT_NAME(SalesforceManager)
...
+ * @param remoteAccessConsumerKey The Conneted App consumer key.
+ * @param oauthRedirectURI the Conneted App redirect URI.
+ * @param loginUrl the login url for native login
+ * @param nativeLoginViewController the UI instance that will be presented to the user and
+ *  is responsible for using the returned Native Login Manager to initiate the login process.
+ *
+ * @return The Native Login Manager.
+ */
+- (id <SFNativeLoginManager>)useNativeLogin:(nonnull NSString *)remoteAccessConsumerKey :(nonnull NSString *)oauthRedirectURI
+                                           :(nonnull NSString *)loginUrl :(nonnull UIViewController *)nativeLoginViewController;
```
brandonpage, line 317: Is this the approach we want going forward? Or do we want to make a new `bootconfig` equivalent?  The main thing that is new here is that each connected app config is tied to and only works with one login url/org.  So we would still need a function like this to pass in the view controller and a name for which connected app to use.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager+Internal.h
```diff
@@ -25,6 +25,7 @@
 @property (nonatomic, assign) SFAppType appType;
 @property (nonatomic, weak, nullable) id<SalesforceSDKManagerFlow> sdkManagerFlow;
 @property (nonatomic, strong, nonnull) SFSDKSafeMutableDictionary<NSString *, UIViewController *> *snapshotViewControllers;
+@property (nonatomic, strong, nullable) UIViewController* nativeLoginViewController;
```
bbirman, line 28: Is the multiple window scenario covered? If not, could potentially make this into a dictionary like `snapshotViewControllers`, there's also the AuthHelper methods that take a scene parameter and can be called from a scene delegate

brandonpage: I have not tested multi-window.  Because I am using the standard way of kicking off the login flow I would expect it to act the same way as our LoginViewController does today (for better or worse).  I'll add it to the considerations list so we don't loose track of it!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
brandonpage: Update:  This worked flawlessly in Xcode 5.1 but now fails (with no code change) in Xcode 5.2 😡.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
bbirman: No preference on my end! If they fallback to the web view, does the web view use this too or the bootconfig?

small question, is there a reason the parameters aren't named? It looks like the auto complete on the objective-c side doesn't have any other context besides the type

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -313,6 +314,12 @@ - (IBAction)backToPreviousHost:(id)sender {
 - (void)handleBackButtonAction {
     UIScene *scene = self.view.window.windowScene;
     [[SFUserAccountManager sharedInstance] stopCurrentAuthentication:nil];
+    
+    if ([SFUserAccountManager sharedInstance].shouldFallbackToWebAuthentication) {
+        [SFUserAccountManager sharedInstance].shouldFallbackToWebAuthentication = NO;
+        [[SFUserAccountManager sharedInstance] loginWithCompletion:^(SFOAuthInfo* authInfo, SFUserAccount* user) { } failure:^(SFOAuthInfo* authInfo, NSError* error) { }];
```
bbirman, line 320: I think loginWithCompletion has the blocks as nullable, can this be nil instead of empty?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1136,6 +1182,35 @@ - (SFUserAccount*)createUserAccount:(SFOAuthCredentials *)credentials {
...
+    
+    UIScene *scene = SalesforceSDKManager.sharedManager.nativeLoginViewController.view.window.windowScene;
+    SFSDKAuthSession *authSession = self.authSessions[scene.session.persistentIdentifier];
+    
+    // Dummy request is necessary for flow
+    SFSDKOAuthTokenEndpointRequest *request = [[SFSDKOAuthTokenEndpointRequest alloc] init];
+    request.additionalOAuthParameterKeys = [NSArray new];
+    
+    // Let SFSDKOAuth2 do error handling
+    [(SFSDKOAuth2 *)self.authClient() handleTokenEndpointResponse:^(SFSDKOAuthTokenEndpointResponse *response) {
```
bbirman, line 1201: could `handleTokenEndpointResponse` be added to `SFSDKOAuthProtocol` instead of casting?

brandonpage: Perhaps, but that would make it public.  I was trying to make as little private API public as possible.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -289,19 +289,40 @@ NS_SWIFT_NAME(SalesforceManager)
...
  *
- * @return the Biometric Authentication Manager
+ * @return The Biometric Authentication Manager.
  */
 - (id <SFBiometricAuthenticationManager>)biometricAuthenticationManager;
 
+/**
+ * Creates the NativeLoginManager instance.
+ *
+ * @param remoteAccessConsumerKey The Conneted App consumer key.
```
JohnsonEricAtSalesforce, line 308: Totally random, but I noticed `Connected` is misspelled while I was just reading for context in the upcoming OTP effort.  There are a couple of instances.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
brandonpage: Falling back to webview uses the the values from the existing bootconfig for a few reasons but most notably the callback for the new connected app needs to be **a real url** for the org to do the token exchange (and we have seen apps have issues with that).  

RE named parameters:  I'll add it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -828,6 +867,13 @@ - (void)oauthCoordinatorDidCancelBrowserAuthentication:(SFOAuthCoordinator *)coo
         return;
     }
     
+    if (self.nativeLoginEnabled && self.shouldFallbackToWebAuthentication) {
+        self.shouldFallbackToWebAuthentication = NO;
+        [self stopCurrentAuthentication:^(BOOL result) { }];
```
bbirman, line 872: can the blocks on 872-873 be nil instead of empty?

libs/SalesforceSDKCore/SalesforceSDKCoreTests/ScreenLockManagerTests.m
```diff
@@ -87,15 +87,9 @@ - (void)testLockScreenTriggers {
     XCTAssertTrue([[[SFSDKWindowManager sharedManager] screenLockWindow] isEnabled]);
     
     // Backgrounding/foregrounding on lock -- lock screen should remain & callback shouldn't be called
-    XCTestExpectation *expectation = [[XCTestExpectation alloc] initWithDescription:@"Callback"];
```
bbirman, line 90: I think this expectation was added for a bug fix, did the behavior change?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -846,6 +851,31 @@ - (void)biometricAuthenticationFlowDidComplete:(NSNotification *)notification {
...
+#pragma mark - Native Login
+
+- (id <SFNativeLoginManager>)useNativeLoginWithConsumerKey:(nonnull NSString *)consumerKey
+                                               callbackUrl:(nonnull NSString *)callbackUrl
+                                              communityUrl:(nonnull NSString *)communityUrl
+                                 nativeLoginViewController:(nonnull UIViewController *)nativeLoginViewController
+                                                     scene:(nullable UIScene *)scene {
+    
+    [SFSDKAppFeatureMarkers registerAppFeature:kSFAppFeatureNativeLogin];
+    NSString *key = scene ? scene.session.persistentIdentifier : kSFDefaultNativeLoginViewControllerKey;
```
bbirman, line 863: if there's no scene you could also default to `[[SFSDKWindowManager sharedManager] defaultScene]` and use its identifier so you don't have to track a separate key

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -846,6 +851,31 @@ - (void)biometricAuthenticationFlowDidComplete:(NSNotification *)notification {
...
+    [_nativeLoginViewControllers setObject:nativeLoginViewController forKey:key];
+    nativeLogin = [[SFNativeLoginManagerInternal alloc] initWithClientId:consumerKey redirectUri:callbackUrl loginUrl:communityUrl scene:scene];
+    [SFUserAccountManager sharedInstance].nativeLoginEnabled = true;
+    
+    return nativeLogin;
+}
+
+- (id <SFNativeLoginManager>)nativeLoginManager {
+    if (nativeLogin == nil) {
+        [[SFLogger defaultLogger] e:[self class] message:@"You must call 'useNativeLogin' to create the Native Login Manager instance before retrieving it."];
```
bbirman, line 873: I think you'd want `SFSDKCoreLogger` for this one

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -846,6 +851,31 @@ - (void)biometricAuthenticationFlowDidComplete:(NSNotification *)notification {
...
+- (id <SFNativeLoginManager>)useNativeLoginWithConsumerKey:(nonnull NSString *)consumerKey
+                                               callbackUrl:(nonnull NSString *)callbackUrl
+                                              communityUrl:(nonnull NSString *)communityUrl
+                                 nativeLoginViewController:(nonnull UIViewController *)nativeLoginViewController
+                                                     scene:(nullable UIScene *)scene {
+    
+    [SFSDKAppFeatureMarkers registerAppFeature:kSFAppFeatureNativeLogin];
+    NSString *key = scene ? scene.session.persistentIdentifier : kSFDefaultNativeLoginViewControllerKey;
+    [_nativeLoginViewControllers setObject:nativeLoginViewController forKey:key];
+    nativeLogin = [[SFNativeLoginManagerInternal alloc] initWithClientId:consumerKey redirectUri:callbackUrl loginUrl:communityUrl scene:scene];
```
bbirman, line 865: it looks like if a couple of scenes were opened in a row that manager would get overwritten and only have the latest one? I wasn't thinking about that when we were talking about it before, wondering if login could also have a scene parameter so you don't have to store scene directly on the manager and this could only be initialized once? or could have a property so the SDK can create the view controllers as needed like `SnapshotViewControllerCreationAction` or `SFLoginViewControllerCreationBlock`

brandonpage: I will think that over for unexpected side effects, but I did design it like this intentionally.  I'm betting that 90%+ of apps that use this will have one native login manager so we should hold one reference they can retrieve for convenience.  However, since these managers are **per community/org** there will be a more rare use case for multiple (prod + sandbox or perhaps publisher's public playground app).  Apps that need that can maintain references to managers themselves.  Although I haven't given much thought to that being a memory leak.  

Regardless I _think_ I need to store the scene with the manager because it links the result of the rest calls being made from that login VC to the auth session in the SDK that is necessary for correctly creating the user account.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/ScreenLockManagerTests.m
brandonpage: Can you please DM me a link to the WI?  I removed this because it causes the test runner to crash.  If you run the tests in xcode it's fine but on the command line (like we do in CI) it has a bunch of side effects that cause other tests to fail.  It took me a long time to figure this out but we weren't seeing it before because of the run order of tests.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -0,0 +1,214 @@
...
+                case .failure(let error):
+                    SFSDKCoreLogger().e(self.classForCoder, message: "error: \(error)")
+                    return .unknownError
+                }
+            } catch {
+                SFSDKCoreLogger().e(self.classForCoder, message: "error: \(error)")
+                return .unknownError
+            }
+        case .failure(let error):
+            // You will catch the error here in the event of auth failure or if the use cannot login this way.
```
JohnsonEricAtSalesforce, line 128: I randomly noticed the `use` was probably supposed to be `user` in this comment.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
brandonpage: Oh, nice.  Thank you.

brandonpage: As it turns out this isn't viable since `defaultScene` is nil at this point so I can't use it as a key. 😞

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.m
```diff
@@ -317,6 +318,11 @@ - (NSDate *)lastModifiedDate
         return nil;
 }
 
+- (BOOL)nativeLogin 
+{
+    return  [self.dictRepresentation objectForKey:kSFNativeLoginDateKey];
```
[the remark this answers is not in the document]
brandonpage: It seems to return `0x0000000000000001` regardless 🤔

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
bbirman: ooo, where in the app delegate / scene delegate is it called from?

brandonpage: I have been calling it from AppDelegate init (for non-multi-window apps).  If it's called from SceneDelegate it isn't a problem.  

However, if I called it in `didFinishLaunchingWithOptions` it would work so I guess if we document it that is fine.  There is no way that `defaultScene` will change throughout a run of the app though, correct?

bbirman: not if it's a single window app, but no preference about changing it! I was just curious about where the nil case was coming from

libs/SalesforceSDKCore/SalesforceSDKCore/NativeLoginManagerTests.swift
```diff
@@ -0,0 +1,111 @@
...
+    }
+    
+    func testShouldShowBackButton() {
+        let accountManager = UserAccountManager.shared
+        XCTAssertNil(accountManager.currentUserAccount)
+        XCTAssertFalse(nativeLoginManager.shouldShowBackButton(), "Should not show back button by default.")
+        _ = createUser()
+        XCTAssertTrue(nativeLoginManager.shouldShowBackButton(), "Should show back button when there is a logged in user.")
+        
+        // Clear accuount
```
wmathurin, line 79: typo 'accuount'

libs/SalesforceSDKCore/SalesforceSDKCore/NativeLoginManagerTests.swift
```diff
@@ -0,0 +1,111 @@
...
+    func testShouldShowBackButtonWithBioAuth() {
+        let user = createUser()
+        let bioAuthManager = BiometricAuthenticationManagerInternal.shared
+        bioAuthManager.storePolicy(userAccount: user, hasMobilePolicy: true, sessionTimeout: 1)
+        XCTAssertTrue(nativeLoginManager.shouldShowBackButton(), "Should show back button when there is a logged in user (but not locked).")
+
+        bioAuthManager.locked = true
+        XCTAssertFalse(nativeLoginManager.shouldShowBackButton(), "Should not show back button when bio auth is locked.")
+        
+        // Clear accuount
```
wmathurin, line 94: typo 'accuount'

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h
```diff
@@ -289,19 +289,44 @@ NS_SWIFT_NAME(SalesforceManager)
...
+ * @param consumerKey The Connected App consumer key.
+ * @param callbackUrl The Connected App redirect URI.
+ * @param communityUrl The login url for native login
+ * @param nativeLoginViewController The UI instance that will be presented to the user and
+ *  is responsible for using the returned Native Login Manager to initiate the login process.
+ * @param scene Optional UIScene to enable multi-window support.
+ *
+ * @return The Native Login Manager.
+ */
+- (id <SFNativeLoginManager>)useNativeLoginWithConsumerKey:(nonnull NSString *)consumerKey
```
wmathurin, line 317: So an app can setup a native login that uses a different consumer key from the one in bootconfig?
Are we storing the consumer key alongside the refresh token or are we refreshing using the one from bootconfig?

brandonpage: [I think it has always been stored with the user](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/d95c81823341a9076ae549fc3ea5b1d8e094b370/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m#L273).

wmathurin: I guess I forgot - it's also the case on [Android](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/dev/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/AuthenticatorService.java#L236).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthSession.h
```diff
@@ -45,6 +45,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable)NSError *authError;
 @property (nonatomic, strong) SFOAuthInfo *authInfo;
 @property (nonatomic, copy, nullable) void (^authCoordinatorBrowserBlock)(BOOL);
+@property BOOL nativeLogin;
```
wmathurin, line 48: I think that without specifier it means (atomic), is that the intention?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1417,7 +1493,11 @@ - (void)setCurrentUserInternal:(SFUserAccount*)user {
                 [self willChangeValueForKey:@"currentUser"];
                 _currentUser = user;
                 [self setCurrentUserIdentity:user.accountIdentity];
-                if (user.credentials.domain)
+                
+                BOOL isNativeLogin = self.nativeLoginEnabled && !self.shouldFallbackToWebAuthentication;
+                // Native Login uses a secondary Connected App tied to a specifc community url.  If the
```
wmathurin, line 1498: Back to my earlier comment:
- web login does not require the app to programmatically pass the client id (it's read from bootconfig)
- native login expects the app to programmatically pass the client id
I wonder for consistency if we should have the native login client id also be defined in bootconfig?

wmathurin: Makes sense. We can always add a bootconfig entry + method that does not take the client id later, if we think it has value.

wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.m
```diff
@@ -65,6 +65,7 @@
 NSString * const kSFIdentityCustomAttributesKey           = @"custom_attributes";
 NSString * const kSFIdentityCustomPermissionsKey          = @"custom_permissions";
 NSString * const kSFIdentityLastModifiedDateKey           = @"last_modified_date";
+NSString * const kSFNativeLoginDateKey                    = @"native_login";
```
bbirman, line 68: could this be renamed if it's not used to store a date?

bbirman (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityData.m
brandonpage: Yes!

PR #3679, opened 2024-02-13
Fix for dismissing snapshot with screen mirroring

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -278,7 +278,7 @@ - (instancetype)init {
         [[NSNotificationCenter defaultCenter] addObserver:self.sdkManagerFlow selector:@selector(handleAppForeground:) name:UIApplicationWillEnterForegroundNotification object:nil];
         [[NSNotificationCenter defaultCenter] addObserver:self.sdkManagerFlow selector:@selector(handleAppBackground:) name:UIApplicationDidEnterBackgroundNotification object:nil];
         [[NSNotificationCenter defaultCenter] addObserver:self.sdkManagerFlow selector:@selector(handleAppTerminate:) name:UIApplicationWillTerminateNotification object:nil];
-        [[NSNotificationCenter defaultCenter] addObserver:self.sdkManagerFlow selector:@selector(handleSceneDidActivate:) name:UISceneDidActivateNotification object:nil];
+        [[NSNotificationCenter defaultCenter] addObserver:self.sdkManagerFlow selector:@selector(handleSceneWillEnterForeground:) name:UISceneWillEnterForegroundNotification object:nil];
```
bbirman, line 281: It looks like `UISceneDidActivateNotification` isn’t called for the AirPlay scene when the app is foregrounded, but `UISceneWillEnterForegroundNotification` is called for it. 

From the below description of `UISceneDidActivateNotification`, I’m thinking it’s not called because the AirPlay scene is non interactive 

> "A notification that indicates that the scene is now onscreen and responding to user events."

brandonpage (approved):

PR #3688, opened 2024-02-28
@W-15045641: [iOS] OTP implementation

JohnsonEricAtSalesforce:
🚧 _Under Construction: Not Ready For Review_ 🧱

This one's still work-in-progress, particularly the practice app code which will be removed and formalized at a later time.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
```diff
@@ -71,4 +71,36 @@ public protocol NativeLoginManager {
     ///
     /// Note: this call will dismiss your login view controller.
     @objc func biometricAuthenticationSuccess()
+    
+    // MARK: Headerless, Password-Less Login Via One-Time-Passcode
+    
+    /// Submits a request for a one-time-passcode to the Salesforce headerless password-less login flow.
```
brandonpage, line 77: "headerless" should be "headless" I assume.

bbirman:
What's the background on adding RecaptchaEnterprise as a dependency? Its .xcframework doesn't look like it supports all of the architectures that Mobile SDK does

external/fmdb
bbirman, line 1: What version of FMDB is this? It looks like it doesn't include `privacy/PrivacyInfo.xcprivacy` anymore?

brandonpage:
> What's the background on adding RecaptchaEnterprise as a dependency? Its .xcframework doesn't look like it supports all of the architectures that Mobile SDK does

I don't think it should be committed into our project at all.  This is something the customer will need to add to their project to utilize passwordless login (similar to how we don't have Google Services in our Android project despite it being necessary for Push).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
```diff
@@ -71,4 +71,36 @@ public protocol NativeLoginManager {
...
+    ///   - isReCaptchaEnterprise: Specifies if reCAPTCHA uses the enterprise license
+    ///   - otpVerificationMethod: The delivery method for the OTP
+    /// - Returns: A login result indicating the outcome of the OTP request
+    ///
+    @objc func submitOtpRequest(
+        username: String,
+        reCaptchaToken: String,
+        reCaptchaSiteKeyId: String?,
+        googleCloudProjectId: String?,
+        isReCaptchaEnterprise: Bool,
```
brandonpage, line 98: If these are also used for registration and forgot password maybe they should be params in the setup of NativeLoginManager.  You could have a second constructor (vs making all of them optional on the original constructor) if these are all or nothing.

external/fmdb
wmathurin: This PR should not change the fmdb submodule. It looks recent changes are lost. @JohnsonEricAtSalesforce, please get latest from dev and re-run install.sh (to update submodules).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -210,5 +213,166 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+        case .sms: "sms"
+        }
+        // Generate the OTP request body.
+        let requestBodyString = try {
+            do { return String(
+                data: try JSONEncoder().encode(
+                    OtpRequestBody(
+                        recaptcha: reCaptchaParameter,
+                        recaptchaevent: reCaptchaEventParameter,
+                        username: username,
```
brandonpage, line 277: Does the server trim whitespace and new lines?  I guess I never checked.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -373,6 +373,8 @@ - (NSError*) prepareErrorForDelegate:(NSData *)data response:(NSURLResponse *)re
 
     // Parse error from data if any.
     if (data) {
+        NSString* newStr = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+        NSLog(@"%@", newStr);
```
brandonpage, line 377: Is this debug code?

native/SampleApps/MobileSyncExplorer/MobileSyncExplorer/bootconfig.plist
```diff
@@ -8,8 +8,8 @@
 		<string>api</string>
 	</array>
 	<key>oauthRedirectURI</key>
-	<string>testsfdc:///mobilesdk/detect/oauth/done</string>
+	<string>https://msdk-enhanced-dev-ed.my.site.com/services/oauth2/echo</string>
 	<key>remoteAccessConsumerKey</key>
-	<string>3MVG98dostKihXN53TYStBIiS8FC2a3tE3XhGId0hQ37iQjF0xe4fxMSb2mFaWZn9e3GiLs1q67TNlyRji.Xw</string>
+	<string>3MVG9CEn_O3jvv0wTqRT0Le6tmzX.EQ9ZvtHL1TG3gHFV.4IvKZyXw5SgdiVPi61mXrpu40mCOhKevEfYNMOm</string>
```
brandonpage, line 13: This is not necessary for Native Login (and in fact will prevent fallback to web based auth).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
JohnsonEricAtSalesforce: I have proven unable to type headless this past week. Thanks!

JohnsonEricAtSalesforce:
Just to note, this is still a draft and not ready for review 🙃 There's a lot of things here that will be removed and before this is ready for review.  Hold off on going through the details until then.

external/fmdb
JohnsonEricAtSalesforce: I hadn't reviewed the diff at the time, since a week ago was "More draft PR than draft PR" (Sing that like Rob Zombie in the 90's - oh, yeah!)

I had to revert it manually since the `install.sh` didn't want to make any changes.  The new diff looks better.

Note: Still draft status so don't get into the details yet ⏳ This is only open so I can watch the pull request tests.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
JohnsonEricAtSalesforce: That's a dandy idea.  Thanks! 👉🏻 https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3688/commits/3d25dc70b0a41f660cb0106a57e82c14c2088af1

JohnsonEricAtSalesforce: This is still in draft status, so hold on further review just a little longer.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: I confirmed the service is _not_ trimming white space for these, alas.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
JohnsonEricAtSalesforce: Yessir - I'll be removing all this before marking this ready for review.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -210,5 +222,342 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+        let reCaptchaParameter: String? = if (isReCaptchaEnterprise) {
+            nil
+        } else {
+            reCaptchaToken
+        }
+        // Determine the reCAPTCHA "event" parameter for enterprise reCAPTCHA
+        let reCaptchaEventParameter: OtpRequestBodyReCaptchaEvent? = if (isReCaptchaEnterprise) {
+            OtpRequestBodyReCaptchaEvent(
+                token: reCaptchaToken,
+                siteKey: try reCaptchaSiteKeyId ?? { throw NSError(
```
JohnsonEricAtSalesforce: From a code-style perspective, this default-to-throwing-closure style is something I've used before since it's something of a call-site/in-expression guard and reads a bit like the Elvis `?:` operator in Kotlin.  I'd be happy to use whatever style rule we have in place, though.

I thought about adding specific cases for each of these errors as cases in `NativeLoginResult`.  I really wanted to provide detailed developer feedback about why their parameters were wrong, however.  So I went for a throw with a detailed and developer-friendly message.  If `NativeLoginResult` were a Swift enum and could have the message as an associated value, that would be the ideal.  We have Objective-C compatibility blocking that - alas!

Thoughts? After running the code a bit, I thought it might even be nice to retrofit the `.unknownError`, `.invalidPassword` and `.invalidUsername` cases to a similar thrown error.  The existing code outside of native login seems to mix and match error control styles, so I wasn't sure we had a hard style in place.

JohnsonEricAtSalesforce (commented):
Here's the code-walk-through from my self-review to get others started on what's changed 📚

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
```diff
@@ -71,4 +71,64 @@ public protocol NativeLoginManager {
...
+    ///
+    @objc func submitPasswordlessAuthorizationRequest(
+        otp: String,
+        otpIdentifier: String,
+        otpVerificationMethod: OtpVerificationMethod
+    ) async throws -> NativeLoginResult
+}
+
+/// An Objective-C compatible OTP request result
+@objc public class OtpRequestResult: NSObject {
```
bbirman, line 113: Thinking the objective-c version of the class name should have a prefix, maybe `@objc(SFOtpRequestResult)` to match SFNativeLogin?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
```diff
@@ -71,4 +71,64 @@ public protocol NativeLoginManager {
...
+    
+    /// The overall result of the OTP request.
+    public let nativeLoginResult: NativeLoginResult
```
bbirman, line 116: should `nativeLoginResult` and `otpIdentifier` be exposed to objc too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -39,25 +39,70 @@ let maximumPasswordLengthInBytes = 16000
 /// See https://developer.apple.com/documentation/swift/importing-swift-into-objective-c#Import-Code-Within-a-Framework-Target
 @objc(SFNativeLoginManagerInternal)
 public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
+    
+    static let errorDomain = "com.salesforce.security.nativeLoginException"
```
bbirman, line 43: What's this used for?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -210,5 +222,333 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+    }
+    
+    // MARK: Private Implementation
+    
+    /// Generates a Base64 encoded value by concatinating the provided values with a colon, which is
+    /// commonly used in the Headless Identity API request headers.
+    /// - Parameters:
+    ///   - value1: The left-side value
+    ///   - value2: The right-side value
+    private func generateColonConcatinatedBase64String(
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: Thanks!  I wonder how many times I've missed that one? 🧐

libs/SalesforceSDKCore/SalesforceSDKCore/NativeLoginManagerTests.swift
```diff
@@ -108,4 +136,36 @@ final class NativeLoginManagerTests: XCTestCase {
         
         return user
     }
+    
+    /// A convenience for Native Login Manager's login method that discards errors.
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: When I set up the login method to throw, I didn't want to introduce too much boilerplate error handling code to the tests which will fail on a nil result just as easily.  So, I was hoping this would improve the readability of the tests.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
JohnsonEricAtSalesforce: Good thinking!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
JohnsonEricAtSalesforce: Is there a way to cascade the declaration from the class?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: ...Absolutely _nothin'_ since I made the error handling smoother in a newer commit during self-review.  I'll remove this.  Good catch!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
bbirman: I haven't tried it before but there's this https://www.hackingwithswift.com/example-code/language/what-is-the-objcmembers-attribute

JohnsonEricAtSalesforce: That's the one I was thinking of.  I have used that quite a bit and this type likely won't have members that are Swift-only.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
brandonpage: So should we be using the trimmed version here?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
```diff
@@ -44,7 +44,7 @@ public protocol NativeLoginManager {
     ///   - username: User provided Salesforce username.
     ///   - password: User provided Salesforce password.
     /// - Returns: NativeLoginResult
-    @objc func login(username: String, password: String) async -> NativeLoginResult
+    @objc func login(username: String, password: String) async throws -> NativeLoginResult
```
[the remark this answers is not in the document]
brandonpage: Personally, I don't think login should throw.  Maybe there is a case to be made for a purely dev setup scenario like `reCaptchaSiteKeyId` is not provided and they try to imitate OTP.  But having the ability for the app to crash on the user because the dev doesn't catch it feels wrong.  IMO a string message about something that went wrong is just as useful in a log as it is in a throwable at runtime.

libs/SalesforceSDKCore/SalesforceSDKCore/NativeLoginManagerTests.swift
```diff
@@ -43,29 +43,57 @@ final class NativeLoginManagerTests: XCTestCase {
     }
     
     func testUsernameValidation() async {
-        var result = await nativeLoginManager.login(username: "", password: "")
+        var result = await login(username: "", password: "")
```
brandonpage, line 46: Dumb question:  How/why does this compile?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
```diff
@@ -71,4 +71,66 @@ public protocol NativeLoginManager {
...
+        nativeLoginResult: NativeLoginResult,
+        otpIdentifier: String? = nil
+    ) {
+        self.nativeLoginResult = nativeLoginResult
+        self.otpIdentifier = otpIdentifier
+    }
+}
+
+/// The possible OTP verification methods.
+@objc public enum OtpVerificationMethod: Int {
```
brandonpage, line 133: Should this and `OtpRequestResult` be up at the top of the file with `NativeLoginResult`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
```diff
@@ -71,4 +71,66 @@ public protocol NativeLoginManager {
...
+        otp: String,
+        otpIdentifier: String,
+        otpVerificationMethod: OtpVerificationMethod
+    ) async throws -> NativeLoginResult
+}
+
+/// An Objective-C compatible OTP request result
+@objc(SFOtpRequestResult)
+@objcMembers
+public class OtpRequestResult: NSObject {
```
brandonpage, line 115: The reason this class exists is so we can pass the `otpIdentifier` back to the view from `submitOtpRequest` so that they can pass it back to us in `submitPasswordlessAuthorizationRequest`?  Couldn't we just hold onto it or am I missing something?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I could roll with just logging a warning and returning one of the int enum types as well.

libs/SalesforceSDKCore/SalesforceSDKCore/NativeLoginManagerTests.swift
JohnsonEricAtSalesforce: This `login` method is a utility method of `NativeLoginManagerTests` that returned `nil` on any thrown error.  That failed the test and kept any new error handling code to a minimum.

This goes away though it we revert to using integer enums and logging on failure per an another thread on this pull request.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
JohnsonEricAtSalesforce: It could be.  I always look for the type the source is named for at the top of a given file, so that was my style precedent.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
JohnsonEricAtSalesforce: That's actually a neat idea but might also have some consequences in scalability.  I usually think of library classes as being more stateless and leaving data and state to the app or test between methods.  In that case, app or test wouldn't have to retain or prime a native login manager instance between calls or tests - they'd always have enough state to let the object go and start over from their own state at any point.

This also lets us skip any null/mutability/synchronization checks on a native login manager property on the second method since the app provides a required value (rather than it being an mutable optional on native login manager).

I could be flexible 'cuz it'd work either way, but those're some reasons I always lean towards returning data to the app/test rather than retaining lots of member variables.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
brandonpage: Fair enough.  I don't really care if they are all at the top or bottom of the file as long as it is consistent.

libs/SalesforceSDKCore/SalesforceSDKCore/NativeLoginManagerTests.swift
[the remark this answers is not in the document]
bbirman: I think in that scenario you could add `throws` to the test's signature and then any error would become a test failure automatically

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
brandonpage: I agree it is great to avoid maintaining state in a library whenever possible, but this OTP flow has state inherently just due to waiting for a response from the user.  I suppose it is okay to offload the state management to the app but that also makes consuming the result of the first call more annoying since they get back a separate class that has the result (all the other calls return directly) and on success this value they need to give back to us.  

As you said it will work either way and just seems like a differing of opinion/preference/priority.  I am absolutely _not_ dying on this hill 😄.

libs/SalesforceSDKCore/SalesforceSDKCore/NativeLoginManagerTests.swift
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: That really does make sense.  I'll have to give it a try, possibly for fun.  I'm also trying a version that doesn't throw 🙂

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I'm trying out the suggestion by @brandonpage to see what it would look like if the native login manager methods catch all error and unexpected conditions, simply log them and return one of the NativeLoginResult int codes.  If that looks good, that'll be our way forward on this one.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -39,25 +39,68 @@ let maximumPasswordLengthInBytes = 16000
...
         let code: String
     }
     
-    @objc public init(clientId: String, redirectUri: String, loginUrl: String, scene: UIScene?) {
+    @objc public init(
+        clientId: String,
+        redirectUri: String,
+        loginUrl: String,
+        scene: UIScene?
+    ) {
```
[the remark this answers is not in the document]
JohnsonEricAtSalesforce: I keep my line wrap at 80 since I primarily use my laptop's display and a few Xcode tool windows open here and there, but they invented scrollbars for a reason. 💡 I'll revert this.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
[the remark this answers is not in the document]
bbirman: I don't think we have a specific style set, but looking through the other swift code, we have throwing methods in [Encryptor.swift](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift) and Swift only result methods in extensions like [SmartStore.swift](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SmartStore/SmartStore/Classes/Extensions/SmartStore.swift). I could see any of those (with an objective-c/swift combo if needed) or the original signature as long as all the errors are in one place (like if throwing, move over the errors from `NativeLoginResult`)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManager.swift
JohnsonEricAtSalesforce: Also, happy to change it if we want 🏁

wmathurin (approved):

brandonpage (approved):

bbirman (approved):

PR #3703, opened 2024-04-11
Respect useHybridAuthentication option in Native Login.

brandonpage:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -505,9 +505,10 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
                 // Decode the authorization response.
                 let authorizationResponseBody = try successfulResponse.asDecodable(
                     type: AuthorizationResponseBody.self)
+                let grantType = (SalesforceManager.shared.useHybridAuthentication) ? kSFOAuthGrantTypeHybridAuthorizationCode: kSFOAuthGrantTypeAuthorizationCode
```
bbirman, line 508: nit: I don't think you need the parentheses around the first section

bbirman (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
brandonpage: It isn't needed.  For whatever reason my brain is a little faster to recognize `SalesforceManager.shared.useHybridAuthentication` is a boolean expression in parentheses.  I'll fix it since I need to fix the space around the `:` also.

bbirman (approved):

PR #3705, opened 2024-04-12
Fix Bio Auth Policy being overridden.  Update biometric login button location.

brandonpage:
All of the `SFLoginViewController` updates were shamelessly stolen from @bbirman, thank you!

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -97,13 +97,17 @@
     }
 }
 
+- (CGFloat) belowFrame:(CGRect) frame {
+    return frame.origin.y + frame.size.height;
+}
+
 - (void)viewWillLayoutSubviews {
     CGFloat heightOffsetMultiplier = _biometricButton ? 0.9 : 1.0;
     CGFloat bottomOffset = _biometricButton ? self.view.safeAreaInsets.bottom : 0;
     
     // Web view
     CGFloat x = 0;
-    CGFloat y = self.view.safeAreaInsets.top;
+    CGFloat y = [self belowFrame:self.navBar.frame];
     CGFloat w = self.view.bounds.size.width;
     CGFloat h = ((self.view.bounds.size.height - y) * heightOffsetMultiplier) - bottomOffset;
     self.oauthView.frame = CGRectMake(x, y, w, h);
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
```diff
@@ -81,12 +81,16 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
...
     @objc public func storePolicy(userAccount: UserAccount, hasMobilePolicy: Bool, sessionTimeout: Int32) {
-        let policyData = try! JSONEncoder().encode(
-            BioAuthPolicy(hasPolicy: hasMobilePolicy, timeout: sessionTimeout)
-        )
-        let result = KeychainHelper.write(service: kBioAuthPolicyIdentifier, data: policyData, account: userAccount.idData.userId)
-        if result.success {
-            SFSDKCoreLogger.i(BiometricAuthenticationManagerInternal.self, message: "Biometric authentication policy stored for user.")
+        let existingPolicy = readBioAuthPolicy(userId: userAccount.idData.userId)
+        if let policyData = try? JSONEncoder().encode(
+            BioAuthPolicy(hasPolicy: hasMobilePolicy, timeout: sessionTimeout, optIn: existingPolicy?.optIn)
```
brandonpage, line 86: This is the bugfix.  Retain existing optIn status if it exists.  Also I don't think we should retain native button preference with user since it should be set for the app, not the user.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -72,25 +73,45 @@ - (void)viewDidLoad {
...
-
-- (void)viewDidLayoutSubviews {
-    [super viewDidLayoutSubviews];
+- (void)viewWillLayoutSubviews {
+    CGFloat heightOffsetMultiplier = _biometricButton ? 0.9 : 1.0;
+    CGFloat bottomOffset = _biometricButton ? self.view.safeAreaInsets.bottom : 0;
+    
+    // Web view
+    CGFloat x = 0;
+    CGFloat y = self.view.safeAreaInsets.top;
```
bbirman, line 106: This is the part where I'm not sure how it would play with [this line](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/030dec5e2ab53574ff011365952d93a10f4e5e86/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m#L195) where the nav bar is added to the view directly, so we could try tracking that down or use the old nav bar calculation with the new `viewWillLayoutSubviews` and bottom inset pieces?

brandonpage: My brain has sort of shut down for the weekend lol.  Are you saying we should hard code `y` to 0 instead?

bbirman: sounds good!

bbirman (approved):

PR #3724, opened 2024-06-06
Add logout reason to revoke call

bbirman:


wmathurin (commented):
Is the logout reason parameter already supported in production?

bbirman:
> Is the logout reason parameter already supported in production?

It's in 250 so it looks like the last roll out is next week (6/15)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.h
```diff
@@ -515,12 +515,20 @@ Use this method to stop/clear any authentication which is has already been start
...
  This throws out the OAuth refresh token.
  */
 - (void)logout;
 
+/**
+ Forces a logout from the current account, redirecting the user to the login process.
+ This throws out the OAuth refresh token.
+ @param reason The reason that log out was initiated.
+ */
+- (void)logout:(SFLogoutReason)reason;
```
brandonpage, line 530: I am actually a little surprised iOS didn't have a logout without user param until now.  Maybe that's because Android has (too) many overloads :/

brandonpage (approved):
LGTM!

No new tests though?  And are there existing tests using any functions that are now deprecated?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1517,7 +1525,7 @@ - (void)setCurrentUserInternal:(SFUserAccount*)user {
             for (SFUserAccountIdentity *identity in keys) {
                 // Logout any other user with Biometric Authentication
                 if ([bioAuthManager checkForPolicyWithUserId:identity.userId] && ![identity isEqual:[self currentUserIdentity]]) {
-                    [self logoutUser:[self userAccountForUserIdentity:identity]];
+                    [self logoutUser:[self userAccountForUserIdentity:identity] reason:SFLogoutReasonUnknown];
```
brandonpage, line 1528: I think this is a candidate for a new logout reason in the future.  It's a stretch but does `SFLogoutReasonUserInitiated` make any sense?  The user is choosing to login again (instead of biometric) and we are revoking the old one because we have a new one.

brandonpage: Actually the `REFRESH_TOKEN_ROTATED` seems to fit.

bbirman: I can't remember, do we tell them that's what's happening? I was initially thinking user initiated would be them clicking a log out button but I could see it here if they're choosing to proceed knowing a logout would happen

brandonpage: Yeah, user initiated isn't the right fit.  However I think `REFRESH_TOKEN_ROTATED` is perfect.

bbirman: It looks like that one is used in server side logs, I'm wondering if we can use it too, can double check tomorrow

bbirman: Updated!

wmathurin (approved):

PR #3728, opened 2024-06-18
@W-14662360/W-15061210: [iOS] Add additional native login feature(s): forgot password/(user registration)

JohnsonEricAtSalesforce:
🎸 _*Ready For Review!_ 🥁

Here's a two-in-one: forgot password and user registration via the Salesforce Headless Identity API.  This will look really familiar to previous client implementations we've done for iOS for this API.  It will also look really familiar compared to the Android implementation since most of the logic is identical along with very similar naming conventions and code layout.

Keep an eye on the template implementation incubating in https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/414

Thanks for reading!

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
@@ -431,7 +431,7 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         // React to the start password reset response.
         switch startPasswordResetResponse {
             
-        case .success(let startPasswordResetResponse):
+        case .success:
             return .success
             
         case .failure(let error):
@@ -496,7 +496,7 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         // React to the complete password reset response.
         switch completePasswordResetResponse {
             
-        case .success(let completePasswordResetResponse):
+        case .success:
             return .success
             
         case .failure(let error):
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
@@ -4,7 +4,7 @@
 //
 //  Created by Brandon Page on 12/13/23.
 //  Copyright (c) 2023-present, salesforce.com, inc. All rights reserved.
-// 
+//
 //  Redistribution and use of this software in source and binary forms, with or without modification,
 //  are permitted provided that the following conditions are met:
 //  * Redistributions of source code must retain the above copyright notice, this list of conditions
@@ -15,7 +15,7 @@
 //  * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
 //  endorse or promote products derived from this software without specific prior written
 //  permission of salesforce.com, inc.
-// 
+//
 //  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
 //  IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
 //  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
@@ -39,7 +39,7 @@ let maximumPasswordLengthInBytes = 16000
 /// See https://developer.apple.com/documentation/swift/importing-swift-into-objective-c#Import-Code-Within-a-Framework-Target
 @objc(SFNativeLoginManagerInternal)
 public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
-        
+    
     @objc public let clientId: String
     @objc public let redirectUri: String
     @objc public let loginUrl: String
@@ -92,13 +92,9 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         let trimmedUsername = username.trimmingCharacters(in: .whitespacesAndNewlines)
         let trimmedPassword = password.trimmingCharacters(in: .whitespacesAndNewlines)
         
-        if !isValidUsername(username: trimmedUsername) {
-            return .invalidUsername
-        }
+        if let it = mapInvalidUsernameToResult(trimmedUsername) { return it }
+        if let it = mapInvalidPasswordToResult(trimmedPassword) { return it }
         
-        if !isValidPassword(password: trimmedPassword) {
-            return .invalidPassword
-        }
         guard let credentials = generateColonConcatenatedBase64String(
             value1: trimmedUsername,
             value2: trimmedPassword) else {
@@ -128,7 +124,7 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
                 continuation.resume(returning: result)
             }
         }
-
+        
         // Second REST Call - Access token request with code verifier
         return await submitAccessTokenRequest(
             authorizationResponse: authorizationResponse,
@@ -196,7 +192,7 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         
         return containsNumber && containsLetter && password.count >= minimumPasswordLength && password.utf8.count <= maximumPasswordLengthInBytes
     }
- 
+    
     private func urlSafeBase64Encode(data: Data) -> String {
         return data.base64EncodedString()
             .replacingOccurrences(of: "/", with: "_")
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -215,7 +211,303 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+                        classForCoder,
+                        message: "Cannot JSON decode start registration response body due to a decoding error with description '\(error.localizedDescription)'.")
+                    return nil
+                }}() else { return StartRegistrationResult(nativeLoginResult: .unknownError) }
+            return StartRegistrationResult(
+                nativeLoginResult: .success,
+                email: startRegistrationResponseBody.email,
+                requestIdentifier: startRegistrationResponseBody.identifier)
+            
+        case .failure(let error):
```
bbirman, line 303: In here and some other places, it looks like we're swallowing a potentially more detailed error and returning it as "unknown", thoughts on throwing the original errors instead of wrapping them?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -215,7 +211,303 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+                        recaptchaevent: reCaptchaParameterGenerationResult.enterpriseReCaptchaEvent,
+                        userData: UserData(
+                            email: email,
+                            username: username,
+                            password: newPassword,
+                            firstName: firstName,
+                            lastName: lastName),
+                        otpVerificationMethod: otpVerificationMethodString)
+                ),
+                encoding: .utf8)!
```
bbirman, line 254: Why is this force unwrapped? Is there any risk of hitting it?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -215,7 +211,303 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+                request: completePasswordResetRequest
+            ) { result in
+                continuation.resume(returning: result)
+            }
+        }
+        
+        // React to the complete password reset response.
+        switch completePasswordResetResponse {
+            
+        case .success(let completePasswordResetResponse):
```
bbirman, line 499: same warning as start

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -215,7 +211,303 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+    }
+    
+    /// A data class for the Salesforce Identity API start registration request body.
+    private struct StartRegistrationRequestBody : Codable {
+        
+        /// The reCAPTCHA token provided by the reCAPTCHA iOS SDK.  This is not used with reCAPTCHA Enterprise
+        var recaptcha: String?
+        
+        /// The reCAPTCHA parameters for use with reCAPTCHA Enterprise
+        var recaptchaevent: ReCaptchaEventRequestParameter?
```
bbirman, line 331: maybe?
```suggestion
        var recaptchaEvent: ReCaptchaEventRequestParameter?
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -215,7 +211,303 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+                request: startPasswordResetRequest
+            ) { result in
+                continuation.resume(returning: result)
+            }
+        }
+        
+        // React to the start password reset response.
+        switch startPasswordResetResponse {
+            
+        case .success(let startPasswordResetResponse):
```
bbirman, line 434: Another warning here, could potentially use `case .success:`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
```diff
@@ -558,4 +1000,56 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+        
+        return nil
+    }
+    
+    /// Validates a string is a valid password.
+    /// - Parameters:
+    ///    - string: The string to validate
+    /// - Returns: nil for valid passwords - The invalid password login result otherwise
+    func mapInvalidPasswordToResult(_ string: String) -> NativeLoginResult? {
+        switch (!isValidPassword(password: string.trimmingCharacters(in: .whitespacesAndNewlines))) {
```
bbirman, line 1039: potentially more of a style thing, why switch on a boolean? are other cases coming in?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: That's a real great question and maybe a conversation starter.  At the onset of our Salesforce Headless Identity API implementations in https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3676 `NativeLoginResult.unknown` was introduced in way I interpreted as a general catch-all so none of the public methods of `NativeLoginManager` would have to throw.  There was some discussion later on https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3688 when we tried setting a few of the methods to throw that we later decided to discard and have return one of the `NativeLoginResult` error cases (they're all error except for success).

My long-term recommendation actually would be to have these methods throw and ensure the catch can access the original failure response from the API for their decision.  That lets us avoid having to add detailed `NativeLoginResult` cases, like I just added for invalid email, each time we want to let a caller see or respond to something different.  I don't think native login manager has any client or server side error case it handles with its own internal logic - it always passes the decision on to the caller. Using throw we also don't have to maintain a pass-through type like `NativeLoginResult` for every error case.  We might even be able to eliminate that enum type entirely and have a simple success or fail result for each action.  Using throw we might also be able to pass-through the actual API results to the caller since there are a lot of them and they could change at any time without an MSDK release in which we create a surrogate `NativeLoginResult` to represent them.

I was also thinking about our style for client-side validation since many mobile projects I've worked don't attempt things like email validation on the client - the API takes responsibility for that in a single place so clients don't have to duplicate the logic.  I'm just thinking aloud on this one, but the Headless Identity API docs do validate parameters such as username, password and email.

The only trouble is that's a different pattern from 12.x, so we'd either have an API change or some methods that throw while others use an error enum case.

I'm really flexible to anything we want to do for the moment.  Shall we keep it as-is so it's consistent with the released API, set the net new code up to throw or something else either now or later?

By the by, I confirmed the API is already handling email validation and returning that in a way we could probably throw meaningfully.

Screenshot 2024-06-18 at 15 02 13

JohnsonEricAtSalesforce: This might be interesting when compared to Kotlin's pattern since Swift exceptions are checked at compile time.  Though, there're not when called from Objective-C if I recall.  One conversation point in #3688 was that we never wanted to throw since an exception not checked at compile time might not be caught by the caller.  Maybe we solve for that with documentation?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: That is totally left-over from earlier code.  Great catch!

brandonpage: Can you decode JSON without the force unwrap?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: I'm so used to everything being a `when` in Kotlin instead of `if` that this just sort of happened on muscle memory.  It probably _could_ be an `if` instead.  Shall we "switch" it to "if"? (get it? 🤣)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: I just verified the unwrap isn't needed 👍🏻

JohnsonEricAtSalesforce:
@bbirman - With the "exception" about throwing (Gah! I'm on a pun streak!), this should have all your requests.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
brandonpage: Great conversation here, throwing in my 2¢:

I have never liked the idea of intentionally throwing if it has a possibility of happening in production.  I do however, absolutely see the value in passing that server side error message to the client (one way or another).  IMO it would be great if we could attach a dynamic string to the unknown enum value for the app to log of whatever.  I am absolutely open to ideas on how to pass the error.

I do however think there is value in the result enum and do not want to see that go away.  A lot of the value we are adding here with this API is that we have thought through and tested the error cases.  I see returning a defined results enum so that the app can simply use a switch on to handle all cases is a real convenience when all they (probably) care about is the UI/UX.

brandonpage: Oh and I think client side validation is just nice to have.  It's pretty trivial to implement and allows the app to give the user instant feedback instead of waiting for the (somewhat slow) API call to come back.  It also allows us to give the app a specific error back without having to parse the response and map it to a result.

bbirman: I like your long term recommendation and what you mentioned about the enum, I think `NativeLoginResult` made sense for the earlier cases but it seems like it's getting more complicated with nested result types and more cases.

I'd lean towards making the net new code throw and then could have separate work to adjust the old if wanted? 

For the exception part-- for a method that throws in Swift, the generated Objective-c version has an error argument instead so I don't think there's any safety issue there (apps that want the error information would provide one to potentially be populated). Here's an example of the original vs generated version and [usage](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/a8d90a6be90897319513dbfc1a3d0abcef859640/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFSDKEncryptedURLCache.m#L91-L92)

```
public static func encrypt(data: Data, key: Data?) throws -> Data

+ (NSData * _Nullable)encryptData:(NSData * _Nonnull)data key:(NSData * _Nullable)key error:(NSError * _Nullable * _Nullable)error
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
bbirman: sounds good and of course props for the pun :P

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: It sounds like we might have a plan as I paraphrase through the notes above and try to find the best common ground.  I'll try to summarize a direction, but am still open to letting others choose since everything on the table "works" for the customer (if I were the customer, that's what I'd say) and we're just choosing style.

1. We can't attach any values to this enum, unfortunately, since it's an Objective-C enum rather than Swift.  It only gets to be an int in the end.  
2. Thanks for the clarification regarding throwing Swift functions and how they handle in Objective-C.
3. Do we have a majority consensus to introduce throws on the new methods here when the API call fails and ensure that the API provided error is available to the caller? If we do, I can push a commit with that.  If we don't I believe this pull request merges as-is.

Option "C" is we can revisit this topic later as well, which might be good so we can brown bag on our error styles in our own time without incrementing our cycle time on this issue.

Thanks for the ideas and help in moving this one to the goal 🚀

JohnsonEricAtSalesforce: I kind of like option "C" so we could pull in the remainder of the team to discuss our error styles on our own time.

bbirman: I'm good with throwing or option C, @brandonpage what do you think?

brandonpage: @JohnsonEricAtSalesforce @bbirman I was going to suggest something along the lines of "option C".  I think we leave it as is for now.  If we want to remove the return value/type and throw (which I am still not a huge fan of) we should do it across the board -- which would be breaking.  So we might as well lump that in with the name changes Eric wanted to make to password-less login APIs in a major release.  We can discuss as a team what we want the API to be for iOS and Android.   

The timeline I propose:
- Leave this for now.
- Deprecate in 12.1 or 12.2.  
- Implement new API in 12.2 or 13.0.
- Remove existing API in 13.0

bbirman (approved):

PR #3729, opened 2024-06-19
Add CodeCov.

brandonpage:


brandonpage:
I will contact CircleCI about the the error that causes CI to fail:

`Resource class macos for macos.m1.medium.gen1, image xcode:15.4.0 is not available for your project, or is not a valid resource class.`

`macos.x86.medium.gen2` (Intel) has been deprecated for a long time and we were supposed to get access to the above M1 resource class in May....

wmathurin (approved):

PR #3730, opened 2024-06-19
Move tokenRefresh event

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
```diff
@@ -283,7 +283,10 @@ - (void)accessTokenForRefresh:(SFSDKOAuthTokenEndpointRequest *)endpointReq comp
                 }
             });
             return;
+        } else {
```
bbirman, line 286: Moved it to this method because this is the one the rest call into. I was thinking it should only be logged if there's not a network error-- assuming we want to track how many times it was actually refreshed vs attempted (let me know if anyone thinks otherwise!)

wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
smcnulty-sfdc: It should log whenever we refresh token; that's what is happening on Android side.

PR #3734, opened 2024-06-27
Move to SQLCipher 4.6.0

wmathurin:


wmathurin:
All tests passing except for a couple of BriefcaseSyncDown tests - those failures are unrelated to the SQLCipher upgrade and will be investigated separately.

PR #3741, opened 2024-07-16
Don't fetch auth config for login pools

bbirman:


brandonpage (approved):
Nice.  This came up recently when looking at logs, people were concerned and questioning why we would get a 404 from this call.  This should cut down on some of the noise.

wmathurin (commented):
Do we need the same on Android?

brandonpage:
> Do we need the same on Android?

Android has done this for years.

wmathurin (approved):

bbirman:
Investigating the test failures separately, one has started flapping before this and the other seems based on iOS version

PR #3747, opened 2024-08-02
Add property to be able to show login screen earlier

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -187,6 +187,7 @@ - (id)init {
         _authPreferences = [SFSDKAuthPreferences  new];
         _errorManager = [[SFSDKAuthErrorManager alloc] init];
         _shouldFallbackToWebAuthentication = NO;
+        _showAuthWindowWhileLoading = NO;
```
bbirman, line 190: I can't remember if we made a decision about what the default behavior should be, I can flip this and have consumers opt out if they want as well

brandonpage: This is fine IMO.  If we thing it is safe we can always flip the default in the next major.

brandonpage (approved):

PR #3752, opened 2024-08-13
Fix testNotificationTransformNonRSASecret failure

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKEncryptedPushNotificationTests.m
```diff
@@ -190,7 +190,11 @@ - (void)testNotificationTransformNonRSASecret {
     NSError *nonRSASecretError = nil;
     BOOL result = [SFSDKPushNotificationDecryption decryptNotificationContent:notifContent error:&nonRSASecretError];
     XCTAssertFalse(result);
-    XCTAssertEqual(nonRSASecretError.code, SFSDKPushNotificationErrorSecretDecryptionFailed);
+    if (@available(iOS 17.4, *)) {
```
bbirman, line 193: It looks like there’s a behavior change between 17.2 and 17.4 with [this line](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/f38dfb6dcf517c33ad199ff86147176c0926d563/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationDecryption.m#L195), starting in 17.4 it returns data instead of nil when decrypting a bad secret. On 17.4 it still ultimately fails when trying to use the decrypted secret as an encryption key so the end result is the same but the reported reason changes

brandonpage (approved):
TY!  I was just starting to look at this.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKEncryptedPushNotificationTests.m
wmathurin: Should MSDK hide the difference in behavior and always return the same error??

bbirman: I'm not sure if there are still cases which could still trigger `SecretDecryptionFailed` before `ContentDecryptionFailed` happens, could make a generic one like `DecryptionFailed` to replace both of them? Here's [the full enum for reference](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/f38dfb6dcf517c33ad199ff86147176c0926d563/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationError.h)

wmathurin: I re-read your comment and looked at SFSDKPushNotificationDecryption.m - I think it's fine. Maybe just a comment in the test code with the explication above?

wmathurin (approved):

PR #3758, opened 2024-09-12
Using Single Access API to get front door URL

wmathurin:
- Added `SFRestAPI requestForSingleAccess` (with new test in `SalesforceRestAPITests`) which calls `/services/oauth2/singleaccess` to generate a frontdoor URL to bridge into UI sessions.
- Added `shouldRefresh` method in `SFRestAPI: it now refreshes when getting either a 401 or a 403 with reason `Bad_OAuth_Token` while going to `/services/oauth2` (as long as the biometric is not enabled and locked).
- Using `requestForSingleAccess` in `SFOAuthCoordinator beginIDPFlow` instead of building 
a frontdoor URL directly


wmathurin:
Note: I'm in the middle of testing the IDP-SP flows.

bbirman (approved):

PR #3760, opened 2024-09-20
Add support for visionOS

bbirman:
Need to test for SPM and audit any other APIs we may not want to support

### DIFF
--- a/SalesforceAnalytics.podspec
+++ b/SalesforceAnalytics.podspec
@@ -8,12 +8,14 @@ Pod::Spec.new do |s|
   s.license      = { :type => "Salesforce.com Mobile SDK License", :file => "LICENSE.md" }
   s.author       = { "Bharath Hariharan" => "bhariharan@salesforce.com" }
 
-  s.platform     = :ios, "16.0"
+  s.platforms    =  { :ios => "16.0", :visionos => "2.0" }
 
   s.source       = { :git => "https://github.com/forcedotcom/SalesforceMobileSDK-iOS.git",
                      :tag => "v#{s.version}" }
 
-  s.frameworks   = 'CoreTelephony'
+  s.xcconfig = { 'OTHER_LDFLAGS[sdk=iphoneos*]' => '$(inherited) -framework "CoreTelephony"',
+                 'OTHER_LDFLAGS[sdk=iphonesimulator*]' => '$(inherited) -framework "CoreTelephony"',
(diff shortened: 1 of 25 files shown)


SalesforceSDKCore.podspec
```diff
@@ -8,14 +8,12 @@ Pod::Spec.new do |s|
...
   s.author       = { "Kevin Hawkins" => "khawkins@salesforce.com" }
 
-  s.platform     = :ios, "16.0"
+  s.platforms    =  { :ios => "16.0", :visionos => "2.0" }
   s.swift_versions = ['5.0']
 
   s.source       = { :git => "https://github.com/forcedotcom/SalesforceMobileSDK-iOS.git",
                      :tag => "v#{s.version}" }
 
-  s.frameworks   = 'CoreTelephony'
```
bbirman, line 17: Realized we had a CoreTelephony import in SalesforceSDKCore but it wasn't being used in the importing class

SalesforceAnalytics.podspec
```diff
@@ -8,12 +8,14 @@ Pod::Spec.new do |s|
   s.license      = { :type => "Salesforce.com Mobile SDK License", :file => "LICENSE.md" }
   s.author       = { "Bharath Hariharan" => "bhariharan@salesforce.com" }
 
-  s.platform     = :ios, "16.0"
+  s.platforms    =  { :ios => "16.0", :visionos => "2.0" }
 
   s.source       = { :git => "https://github.com/forcedotcom/SalesforceMobileSDK-iOS.git",
                      :tag => "v#{s.version}" }
 
-  s.frameworks   = 'CoreTelephony'
```
bbirman, line 16: Background on this: 
- I originally added `s.frameworks = 'CoreTelephony'` because it needed to be explicitly linked for catalyst (old comment [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3363#discussion_r610842486)), though testing catalyst again now without this line works
- VisionOS doesn't support CoreTelephony and including in the podspec causes an error 
- I made the inclusion conditional through the xcconfig, could also potentially take out the line completely because because it was only an issue for catalyst and that doesn't seem to be the case anymore

wmathurin (commented):
So there is no need for a post-install step to change SQLCipher's platform in mobilesdk_pods.rb ?

wmathurin (approved):

bbirman:
> So there is no need for a post-install step to change SQLCipher's platform in mobilesdk_pods.rb ?

Just added it

PR #3764, opened 2024-10-15
Fix testLockScreenTriggers

bbirman:
~~Working locally but need to see how it does on CI~~
update: passing on CI!


libs/SalesforceSDKCore/SalesforceSDKCore/ScreenLockManagerTests.swift
```diff
@@ -0,0 +1,95 @@
...
+//  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import Testing
+
+import XCTest
+@testable import SalesforceSDKCore
+
+class ScreenLockManagerTests: XCTestCase {
+    override func setUpWithError() throws {
+        ScreenLockManagerInternal.shared.backgroundTimestamp = 0
```
bbirman, line 35: This is the fix, I moved the test to Swift so that we could reset this as an internal variable instead of trying to expose it to Objective-C / change something in the public API footprint

wmathurin (approved):

brandonpage (commented):
Awesome!

PR #3769, opened 2024-11-02
GitHub Actions

brandonpage:
Github Actions implementation with:
- PR workflow
  - Static Analysis - issues found in files modified by your PR will be reported back in a comment.
  - Real dynamic test matrix - only modified libs tests will be run (and no more hacky starting all of them and canceling what we don't need like w/ CircleCI).  Modifying CI files causes all tests to run.
  - Test Results summary and in-line test failure comments on your PR!
- Nightly workflow - run twice a week on a cron.
- Fastlane has been completely removed in favor of actions/bash.  I tried to use Fastlane Swift but SPM was a nightmare.
- Re-written Danger (Ruby).  I tried to switch to Danger Swift but again, SPM for CI and not in the root of the project is a real pain.
  - Danger now uses the more secure generated `GITHUB_TOKEN` instead of a personal access token.
- Tests are run with [mxcl/xcodebuild](https://github.com/marketplace/actions/mxcl-xcodebuild) which has the awesome ability to take in semver inputs.  This means if we keep running on the `macos-latest` image and run our tests with xcode set to `^16` and ios set to `^18` we should automatically run against the latest available with no code changes until Xcode 17/iOS 19 are released.
- CodeCov
- Dispatch workflow so one off test runs can be manually trigged from the browser.

Example dynamic test matrix:
Screenshot 2024-11-03 at 7 18 41 PM

What's not in this PR:
- ~Adding a dispatch workflow that can be trigged manually from the browser w/ inputs.~
- ~CodeCov~
- SwiftLint?
- Removal of CircleCi 😉


wmathurin (approved):
LGTM!!

.circleci/fastlane/Fastfile
```diff
@@ -1,127 +0,0 @@
-$git_pr_api = "https://api.github.com/repos/%s/SalesforceMobileSDK-iOS/pulls/%s/files"
```
brandonpage, line 1: I actually didn't mean to delete this, will revert it.

.github/DangerFiles/StaticAnalysis.rb
```diff
@@ -0,0 +1,42 @@
...
+
+# Parse Clang Plist files and report issues associated with files modified or added in this PR.
+for file in files;
+    report = Plist.parse_xml(file)
+    report_file_name = File.basename(file, File.extname(file))
+    print "File with clang report: #{report_file_name}\n"
+
+    if modified_file_names.include?(report_file_name) || added_file_names.include?(report_file_name)
+        issues = report['diagnostics']
+        print "File modified in PR: #{file}, has #{issue.count} issues.\n"
```
brandonpage, line 29: I am going to leave the various prints in this file so we can check it is working.  It's not super obvious if it breaks because it only reports issues in files modified in the PR.

brandonpage:
`mxcl/xcodebuild` just added support for `xcbeautify` (instead of the default `xcpretty`), which comes with in line warnings and test failures: 

Screenshot 2024-11-14 at 12 14 58 PM

PR #3771, opened 2024-11-08
Saving cookie client src, cookie sid client and sid cookie name with user credentials

wmathurin:
Also cleaned up some old code

### DIFF
--- a/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainHelper.swift
+++ b/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainHelper.swift
@@ -38,6 +38,8 @@ public class KeychainHelper: NSObject {
     @objc public static var cacheEnabled: Bool = true
     @objc public private(set) static var accessibilityAttribute: CFString?
     
+    static let baseAppIdentifierKey = "com.salesforce.security.baseappid"
+    
     @objc public enum CacheMode: Int {
         case unspecified
         case enabled
@@ -46,15 +48,6 @@ public class KeychainHelper: NSObject {
 
     private static var keychainAccessibleAttribute: CFString {
         return accessibilityAttribute ?? kSecAttrAccessibleAfterFirstUnlockThisDeviceOnly
-    }
-
-    private static let upgrade: Void = {
-        _ = KeychainUpgradeManager.init().upgradeManagedKeys()
-    }()
-
-    //pre 9.1 keys are updated to have the creator tag.
-    private static func upgradeIfRequired() {
-        self.upgrade
     }
 
     /// Read a value from the keychain.
@@ -73,8 +66,6 @@ public class KeychainHelper: NSObject {
     ///   - accessGroup: kSecAttrAccessGroup attribute for keychain item
     /// - Returns: KeychainResult
     @objc public class func read(service: String, account: String?, accessGroup: String? = nil, cacheMode: CacheMode) -> KeychainResult {
-        self.upgradeIfRequired()
-        
         let keychainRead: KeychainOperation = { service, account in
             let keychainManager = KeychainItemManager(service: service,
                                                       account: account,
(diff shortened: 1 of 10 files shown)


libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainHelper.swift
```diff
@@ -417,118 +400,6 @@ public class KeychainHelper: NSObject {
             }
         }
     }
-
```
wmathurin, line 420: Removing that code means that apps won't be upgrade from 9.0 to 13.0 - but I don't think it would work anyway.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthKeychainCredentials.m
```diff
@@ -128,12 +128,6 @@ - (void)setRefreshToken:(NSString *)token withEncryptionKey:(NSData *)encryption
         if (self.isEncrypted) {
             tokenData = [SFSDKEncryptor encryptData:tokenData key:encryptionKey error:nil];
         }
-    } else {
```
wmathurin, line 131: All the callers of setRefreshToken that sets the refresh token to nil also take care of clearing the other fields. Having such side effect in a setter is not good practice.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -97,17 +97,18 @@ - (id)initWithCoder:(NSCoder *)coder {
             _encrypted = (encryptedBool
                           ? [encryptedBool boolValue]
                           : [coder decodeBoolForKey:@"SFOAuthEncrypted"]);
-            if ([self isMemberOfClass:[SFOAuthCredentials class]]) {
```
wmathurin, line 100: That check is [ancient](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/3e48d5b5e1f336d91fbda99da36dddc4c152a8d2/libs/SalesforceOAuth/SalesforceOAuth/Classes/SFOAuthCredentials.m) and when introduced it was only to guard the setting of refresh and access token - maybe we had a different credentials class early on??

brandonpage (approved):

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Keychain/KeychainHelper.swift
bbirman: this will be nice not to have in the logs anymore

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthKeychainCredentials.m
```diff
@@ -80,65 +113,33 @@ - (NSData *)tokenForService:(NSString *)service
...
-        self.communityUrl = nil;
-        self.issuedAt    = nil;
-        self.identityUrl = nil;
     }
     
-    BOOL updateSucceeded = [self updateKeychainWithTokenData:tokenData forService:kSFOAuthServiceRefresh];
+    BOOL updateSucceeded = [self updateKeychainWithTokenData:tokenData forService:service];
     if (!updateSucceeded) {
-        [SFSDKCoreLogger w:[self class] format:@"%@:%@ - Failed to update refresh token.", [self class], NSStringFromSelector(_cmd)];
+        [SFSDKCoreLogger w:[self class] format:@"%@:%@ - Failed to update %@.", [self class], service];
```
bbirman, line 142: the format string looks like it has more string specifiers than string arguments, maybe this instead?
`[SFSDKCoreLogger w:[self class] format:@"%@:%@ - Failed to update %@.", [self class], NSStringFromSelector(_cmd), service];`

bbirman (approved):

PR #3772, opened 2024-11-14
@W-16580504: [iOS] Hook to propagate MSDK logs to other systems

JohnsonEricAtSalesforce:
🎸 _*Ready For Review!*_ 🥁

  This is a counterpart to MSDK Android's [logging update](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2635).  Everything about MSDK's current logging API and support for file logging in downstream repositories remains the same.  What's new is the app can register "log receivers" via a new factory interface.  These will receive every log issued by MSDK in parallel with the existing logging infrastructure, if the app so desires.

  The big discussion topic here is if this meets both the requirements we have in the work item _plus_ those we may also know of in other threads.  @bbirman gave me quite a bit of the context, so I believe this is an abstract enough solution that most any app need could be met with this API.  I'd be really curious to see if that still looks like that case once we get eyes on this code update.

 Thanks, as always, for reading and I'll incorporate any updates we'll need based on our reviews.

### DIFF
--- a/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
+++ b/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
@@ -28,6 +28,8 @@
 #import <Foundation/Foundation.h>
 #import <os/log.h>
 
+@protocol SFLogReceiverFactory;
+
 typedef NS_ENUM(NSUInteger, SFLogLevel){
     SFLogLevelDefault =  OS_LOG_TYPE_DEFAULT,
     SFLogLevelInfo    =  OS_LOG_TYPE_INFO,
@@ -364,6 +366,12 @@ NS_SWIFT_NAME(SalesforceLogger)
  * @param logger Class.
  */
 + (void)setInstanceClass:(Class<SFLogging>)logger;
+
+/**
+ * Sets the log receiver factory which creates log receivers.
+ * @param logReceiverFactory The log receiver factory
+ */
++ (void)setLogReceiverFactory:(id<SFLogReceiverFactory>)logReceiverFactory;
 
 /**
  * Get the default Logger.
--- a/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
+++ b/libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
@@ -27,24 +27,61 @@
  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#import <SalesforceSDKCommon/SalesforceSDKCommon-Swift.h>
+
 #import "SFLogger.h"
 #import "SFDefaultLogger.h"
 #import "SFSDKSafeMutableDictionary.h"
+
+@protocol SFLogReceiver;
+@protocol SFLogReceiverFactory;
+
 static NSString * const kDefaultComponentName = @"SFSDK";
 
 static Class InstanceClass;
+
+/**
+ * The factory for Salesforce log receivers.
+ * - Parameters
+ *   - logReceiverFactory: The factory for Salesforce log receivers
+ */
+static id<SFLogReceiverFactory> _Nullable _logReceiverFactory;
+
 static SFSDKSafeMutableDictionary *loggerList = nil;
 
 @interface SFLogger()
 - (instancetype)init:(NSString *)componentName;
+- (instancetype)init:(NSString *)componentName logReceiver:(id<SFLogReceiver>)logReceiver;
+
 + (void)clearAllComponents;
+
 @property id<SFLogging> logger;
+
+/** The Salesforce log receiver to receive all logs issued by this Salesforce logger */
+@property id<SFLogReceiver> logReceiver;
 
 @end
 
 @implementation SFLogger
 
+/**
+ * Creates a new Salesforce logger.
+ *
+ * @param componentName The named component the logger will be associated with
+ */
 - (instancetype)init:(NSString *)componentName {
+    return [self init:componentName logReceiver:nil];
+}
+
+/**
+ * Creates a new Salesforce logger.
+ *
+ * @param componentName The named component the logger will be associated with
+ * @param logReceiver The Salesforce log receiver to receive all logs issued by the Salesforce logger
+ */
+- (instancetype)init:(NSString *)componentName
+         logReceiver:(id<SFLogReceiver>)logReceiver
+{
     self = [super init];
     if (self) {
         //for backward compatibility with SFSDKLogger
@@ -54,6 +91,7 @@ static SFSDKSafeMutableDictionary *loggerList = nil;
         } else {
             self.logger = [[InstanceClass alloc] initWithComponent:componentName];
         }
+        self.logReceiver = logReceiver;
     }
     return self;
 }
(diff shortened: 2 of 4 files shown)


libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.h
```diff
@@ -365,6 +367,12 @@ NS_SWIFT_NAME(SalesforceLogger)
  */
 + (void)setInstanceClass:(Class<SFLogging>)logger;
 
+/**
+ * Sets the log receiver factory which creates log receivers.
+ * @param logReceiverFactory The log receiver factory
+ */
++ (void)setLogReceiverFactory:(id<SFLogReceiverFactory>)logReceiverFactory;
```
JohnsonEricAtSalesforce, line 374: On Android we also created a static method for the registration of the new factory instance.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
```diff
@@ -54,6 +91,7 @@ - (instancetype)init:(NSString *)componentName {
         } else {
             self.logger = [[InstanceClass alloc] initWithComponent:componentName];
         }
+        self.logReceiver = logReceiver;
```
JohnsonEricAtSalesforce, line 94: Here's the assignment of the new log receiver for this logger.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
```diff
@@ -67,82 +105,132 @@ -(void)setLogLevel:(SFLogLevel) logLevel {
 }
 - (void)log:(nonnull Class)cls message:(nonnull NSString *)message {
     [self.logger log:cls level:SFLogLevelDefault message:message];
+    
+    [self.logReceiver receiveWithLevel:SFLogLevelDefault cls:cls component:self.logger.componentName message:message];
```
JohnsonEricAtSalesforce, line 109: Here's the parallel logging for the new log receiver.  There's quite a few points for this.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
```diff
@@ -67,82 +105,132 @@ -(void)setLogLevel:(SFLogLevel) logLevel {
...
+    
+    NSString *formattedMessage = [[NSString alloc] initWithFormat:format arguments:args];
+    [self.logReceiver receiveWithLevel:level cls:cls component:self.logger.componentName message:formattedMessage];
+    
     va_end(args);
 }
 
 - (void)log:(Class)cls level:(SFLogLevel)level format:(NSString *)format args:(va_list)args {
+    va_list args_copy;
+    va_copy(args_copy, args);
```
JohnsonEricAtSalesforce, line 142: Note: We have to copy the va_list since the logger will probably end it before we can parallel log it to the log receiver.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
```diff
@@ -272,7 +364,13 @@ + (nonnull instancetype)loggerForComponent:(nonnull NSString *)componentName {
         }
         id logger = [loggerList objectForKey:componentName];
         if (!logger) {
-            logger =  [[self alloc] init:componentName];
+            id<SFLogReceiver> logReceiver = nil;
+            if (_logReceiverFactory) {
+                logReceiver = [_logReceiverFactory createWithComponentName:componentName];
```
JohnsonEricAtSalesforce, line 369: This is where the new factory injects the log receiver.  Much like for Android, I'm hoping an app could implement their dependency injection in their factory if desired.

JohnsonEricAtSalesforce (commented):
Here's some self-review notes to kickstart the discussion.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
JohnsonEricAtSalesforce: If the logger ends the va_list, then the second call would result in a crash.  So, this local copy guards for that.

JohnsonEricAtSalesforce:
Here's a screenshot of how this looked in my app.
Screenshot 2024-11-13 at 16 52 56

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
bbirman: Could the parallel calls be consolidated to a few helper methods (like one each for message, format, etc) that the top levels could call into?

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
```diff
@@ -65,84 +103,106 @@ -(SFLogLevel)logLevel {
 -(void)setLogLevel:(SFLogLevel) logLevel {
     [self.logger setLogLevel:logLevel];
 }
+
+- (void)submitLogEntryWithClass:(nonnull Class)cls level:(SFLogLevel)level message:(nonnull NSString *)message {
```
JohnsonEricAtSalesforce, line 107: @bbirman - Here are two new private submit log entry methods that help increase our code reuse.  How do they look?  We do need at least two since they have to fork to the two different signatures in the logger member.  There _are_ ways we could have a single method, but I though the parameter or runtime handling looked less intuitive when I tried it for fun.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
```diff
@@ -65,84 +103,106 @@ -(SFLogLevel)logLevel {
...
     va_list args;
     va_start(args, format);
-    [self.logger log:cls level:level format:format args:args];
+    [self submitLogEntryWithClass:cls level:level format:format args:args];
     va_end(args);
 }
 
 - (void)log:(Class)cls level:(SFLogLevel)level format:(NSString *)format args:(va_list)args {
+    va_list args_copy;
+    va_copy(args_copy, args); // Note: This log convenience cannot use the common submit log entry methods since it requires this local memory copy due to the caller already having run `va_start` on `args`.
```
JohnsonEricAtSalesforce, line 144: @bbirman - I added a new comment here so future maintainers will know why this one method still has its own implementation.

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
JohnsonEricAtSalesforce: I just pushed a commit to that effect 🌈

bbirman (approved):
👍 could you also add your usage example to one of the sample apps in the workspace? (RestAPIExplorer or MobileSyncExplorer)

brandonpage (approved):

wmathurin (approved):

JohnsonEricAtSalesforce:
I verified the static analyzer runs successfully locally with no new issues for the modified code.  We'll have to monitor and review what's changed in Circle CI separately, I reckon.

PR #3773, opened 2024-11-19
Support for JWT-based access token

wmathurin:
## Overview
Connected app / external client app can be configured to issue JWT-based access token (for more information see [here](https://help.salesforce.com/s/articleView?id=sf.connected_app_enable_jwt.htm&type=5)).

The mobile application does not need to know if it is using a JWT-based access token or an opaque token **in most cases**.
Login and refresh will just work as before.

However, front door URL cannot be built directly using the JWT-based access token. Instead the [single access API](https://help.salesforce.com/s/articleView?id=sf.frontdoor_singleaccess.htm&type=5) should be called to generate front door URL.

We already made changes in the Mobile SDK to be JWT-based access token ready:
- We introduced a REST wrapper for single access API and leveraged it in the IDP-SP flow (see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3758)
- We moved away from front-door and hydrated the web view session in hybrid remote apps using the cookies returned by the oauth hybrid flow (https://github.com/forcedotcom/SalesforceMobileSDK-iOS-Hybrid/pull/256)

## Changes in this PR
- We are saving the token format and parent sid returned by the token end point during login and refresh in the user account.
- We added a class `JwtAccessToken` to allow apps to inspect the JWT encoded in the JWT-based access token.

## Testing
- We added tests to check that token format and parent sid are properly captured and saved.
- We added tests for `JwtAccessToken`.
- We manually verified that login/refresh works in native applications.
- We manually verified that the IDP-SP flows work when JWT-based access token are being used.

## TODO 
Will happen in a separate PR on the [SalesforceMobileSDK-iOS-Hybrid repo](https://github.com/forcedotcom/SalesforceMobileSDK-ios-hybrid)
- Modify `SalesforceWebViewCookieManager` i.e. the class responsible for hydrating the web view sessions in hybrid remote apps when JWT-based access token are in use.
- Manually verify that login/refresh works in hybrid remote applications.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
@@ -36,6 +36,7 @@ NSString * const kSFOAuthServiceLightningSid    = @"com.salesforce.mobilesdk.oau
 NSString * const kSFOAuthServiceVfSid           = @"com.salesforce.mobilesdk.oauth.vfSid";
 NSString * const kSFOAuthServiceContentSid      = @"com.salesforce.mobilesdk.oauth.contentSid";
 NSString * const kSFOAuthServiceCsrf            = @"com.salesforce.mobilesdk.oauth.csrf";
+NSString * const kSFOAuthServiceParentSid       = @"com.salesforce.mobilesdk.oauth.parentSid";
 
 NSString * const kSFOAuthServiceLegacyAccess    = @"com.salesforce.oauth.access";
 NSString * const kSFOAuthServiceLegacyRefresh   = @"com.salesforce.oauth.refresh";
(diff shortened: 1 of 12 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -140,15 +143,12 @@ - (void)encodeWithCoder:(NSCoder *)coder {
     [coder encodeObject:self.issuedAt           forKey:@"SFOAuthIssuedAt"];
     [coder encodeObject:self.protocol           forKey:@"SFOAuthProtocol"];
     [coder encodeObject:self.lightningDomain    forKey:@"SFOAuthLightningDomain"];
-    [coder encodeObject:self.lightningSid       forKey:@"SFOAuthLightningSID"];
```
bbirman, line 143: just curious, why do these encode calls go away?

bbirman (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
wmathurin: The sensitive stuff goes to the keychain and is handled by SFOAuthKeychainCredentials.
I stopped decoding them from the archive in the [last PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3771/files) but forgot to take out the encoding part.

PR #3778, opened 2024-11-25
Fix for snapshot dismissal

bbirman:
In https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3679 I changed the notification we use to dismiss snapshot from `UISceneDidActivateNotification` to `UISceneWillEnterForegroundNotification` because `UISceneDidActivateNotification` isn't called for screen mirroring and I thought `UISceneWillEnterForegroundNotification` could cover both cases.  There's a new issue with using `UISceneWillEnterForegroundNotification` so this PR separates it into using `UISceneDidActivateNotification` for the main flow like it was originally and `UISceneWillEnterForegroundNotification` for the non-interactive/mirrored case


wmathurin (approved):

wmathurin (approved):

brandonpage (approved):

bbirman:
Merging because the failing test (`testCredentialsCopying`) looks like it started flapping before this PR. Looking into it but I haven't been able to repro locally yet

PR #3793, opened 2025-01-08
Add labels and resizing to fields for accessibility on add connection…

bbirman:
… screen

Screen with new design:
after

and example with large text:
large


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -0,0 +1,110 @@
...
+        .background(Color(uiColor: .secondarySystemBackground))
+        .scrollDisabled(true)
+        .listStyle(.plain)
+        .navigationTitle(SFSDKResourceUtils.localizedString("LOGIN_ADD_SERVER"))
+        .toolbar {
+            ToolbarItemGroup(placement: .topBarTrailing) {
+                Button {
+                    save()
+                } label: {
+                    Text("Done").bold()
```
bbirman, line 100: I realized there’s a snag here— the old code uses `[[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemDone…` and the system takes care of localization but I haven’t been able to find an equivalent in Swift UI so far (I found someone else’s tweet with the same question and no answers :grimacing:). I could also add this to our localizable strings which feels a little annoying but we already needed new entries to split the field labels and placeholders so maybe not as bad to do all at once?

wmathurin: There does not seem to be an equivalent in Swift UI, so you probably have to add a localized string for done.

bbirman: 👍 thanks for looking!

wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
bbirman: I have one more change for the tint color of the button, I'll open another PR because I touched a bunch of files

PR #3796, opened 2025-01-10
Bump deployment target to iOS 17

bbirman:


install.sh
```diff
@@ -24,8 +24,8 @@ then
 fi
 
 # Check for Xcode minimum version
-XCODE_MIN_VERSION=130
-XCODE_MIN_VERSION_STR="Xcode 13.0"
+XCODE_MIN_VERSION=160
```
bbirman, line 27: I think this and the readme got out of sync with the min version spreadsheet, assuming we want Xcode 16 for 13.0?

wmathurin (approved):

bbirman:
merging because the test failures are flappers, will make work items to investigate them more

PR #3799, opened 2025-01-11
Add nil check for crash

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
```diff
@@ -104,7 +104,12 @@ - (void)completeWithSuccess {
     if (self.completionBlock) {
         dispatch_async(dispatch_get_main_queue(), ^{
             SFUserAccount *account = [[SFUserAccountManager sharedInstance] accountForCredentials:self.credentials];
-            NSDictionary *userInfo = @{ kSFNotificationUserInfoAccountKey : account };
```
bbirman, line 107: There’s a crash reported at this line with reason `-[__NSPlaceholderDictionary initWithObjects:forKeys:count:]: attempt to insert nil object from objects[0]` but we don’t have app logs or repro steps. It looks like the crash would be because the account is nil so I added a check for that for now but I’d like to dig deeper into why the account is nil if we get any more info

wmathurin (approved):

PR #3805, opened 2025-01-15
Add property for WKUIDelegate method

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -953,6 +953,13 @@ - (void)webView:(WKWebView *)webView runJavaScriptConfirmPanelWithMessage:(NSStr
     }
 }
 
+- (nullable WKWebView *)webView:(WKWebView *)webView createWebViewWithConfiguration:(WKWebViewConfiguration *)configuration forNavigationAction:(WKNavigationAction *)navigationAction windowFeatures:(WKWindowFeatures *)windowFeatures {
```
wmathurin, line 956: So WebKit calls that method? And when nil is returned, it creates the web view itself?

bbirman: Yup that WebKit calls it. When I tested it didn't look like it auto created when nil. One of the publisher cases uses it to intercept the URL and open in Safari and I saw another generic example [here](https://betterprogramming.pub/handling-new-tabs-and-window-popup-with-wkwebview-using-swiftui-22795c5cee92) where they redirect to load in the existing webview while returning nil

wmathurin (approved):

PR #3806, opened 2025-01-16
@W-15363709 Changes to RestAPI for modern Async/Await

Crebs:


wmathurin (commented):
Shouldn't we keep the tests for the old APIs since they are deprecated and not (yet) removed?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -115,6 +116,7 @@ extension RestClient {
     /// Execute a prebuilt request.
     /// - Parameter request: `RestRequest` object.
     /// - Parameter completionBlock: `Result` block that handles the server's response.
+    @available(*, deprecated, message: "Use the async/await version of `send(request:)` instead.")
```
bbirman, line 119: I think we should add something to the deprecation messages about it being removed in 14.0. We have [this](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/40bcc28b8b96b5764870e306db9c4143b9a7c007/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKConstants.h#L103) for deprecations in Objective-C, I'm not sure if there's a way to do something similar to have a standard message in Swift

Crebs: Got it makes sense.  I'll add that.

wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
bbirman: one small thing, can you use 13.0 and 14.0 (vs 13 and 14) in the deprecation messages?

bbirman:
👍 for getting rid of the warnings in NativeLoginManager. I still see two in RestAPIExplorer (ignore the traitCollection one)
rest api explorer warnings

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
Crebs: Got it.  I'll update that.

native/SampleApps/RestAPIExplorer/RestAPIExplorer/ViewControllers/RootViewController.swift
```diff
@@ -568,13 +568,10 @@ class RootViewController: UIViewController {
...
         let request = RestRequest(method: method, path: path, queryParams: queryParams)
-        RestClient.shared.send(request: request) { result in
-            switch result {
-                case .success(let response):
-                    self.handleSuccess(request: request, response: response)
-                case .failure(let error):
-                    self.handleError(request: request, error: error)
-            }
+        
+        self.view.endEditing(true)
```
bbirman, line 572: what's this change for? is the same call at 562 still needed?

Crebs: @bbirman OK, Fixed my goof.  Thanks for catching that .

bbirman (approved):
lgtm! It looks like one of the templates uses the old way too so please update that as well, let me know if you have any questions about that repo!
https://github.com/forcedotcom/SalesforceMobileSDK-Templates/blob/358f7cace478f6458c32f57283e8364e493b8cd5/iOSNativeSwiftEncryptedNotificationTemplate/EncryptedNotificationTemplate/RootViewController.swift#L39

Crebs:
@bbirman I made update to the template repo.  PR is here https://github.com/forcedotcom/SalesforceMobileSDK-Templates/pull/441

PR #3810, opened 2025-01-23
Guards for analytics crash

bbirman:
Open to any ideas / changes to approach on this one

Backstory:
There's a crash with no repro steps that was originally reported on Sapp when `[encryptedData writeToFile:filename options:NSDataWritingFileProtectionCompleteUntilFirstUserAuthentication error:&error];` is called. @mbwimmer investigated and found that the fileName being nil could cause an exception but it was unclear how that scenario could happen ([more discussion here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3534/files)). That fix went into Sapp's fork but now MP is also seeing the same issue (still no repro steps). 

It looks like we do have a nullability conflict where SFSDKEventStoreManager's init method specifies a non-null storeDirectory but the method used to generate the directory that's eventually passed into that init method can return nil which would make the original `fileName` argument nil. Note: I still don't know what user scenario would cause that.

Two parts to the guards here:
- Fail initialization of the analytics manager earlier if a nil directory is provided since events will never successfully write without it, avoids invalid init with SFSDKEventStoreManager 
- Add a guard around the method to catch the exception like @mbwimmer originally had in case there's another cause


wmathurin (approved):

Crebs (approved):
LGTM.   I'm curious our the deprecated Method warning.  Do we need to create another working to fix these warning I missed from last time?

bbirman:
> LGTM. I'm curious our the deprecated Method warning. Do we need to create another working to fix these warning I missed from last time?

I don't remember see it then either :) I ran into it when I was doing something else recently so I have a local change I can send out soon

PR #3815, opened 2025-01-27
@W-14212091 Remove userLogout AILTN event

Crebs:
Based on discussion [here](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001akoOhYAI/view) and in stand up, the event doesn't work and it wouldn't be possible to use AILTN to log the scenario where the user's token is revoked because we need the token to publish to AILTN

What to change

Remove this line: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/821bc4f0eab189fb62d979e041f8715c3f3b13dc/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m#L257

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
@@ -254,7 +254,6 @@ static dispatch_once_t pred;
             NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
             attributes[@"errorCode"] = [NSNumber numberWithInteger:error.code];
             attributes[@"errorDescription"] = error.localizedDescription;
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -494,7 +493,6 @@ - (void)createAndStoreLogoutEvent:(NSError *)error user:(SFUserAccount*)user {
     NSMutableDictionary *attributes = [[NSMutableDictionary alloc] init];
     attributes[@"errorCode"] = [NSNumber numberWithInteger:error.code];
     attributes[@"errorDescription"] = error.localizedDescription;
-    [SFSDKEventBuilderHelper createAndStoreEvent:@"userLogout" userAccount:user className:NSStringFromClass([self class]) attributes:attributes];
```
bbirman, line 497: I didn't notice the second line before 👍 . Looks like the containing method isn't needed either?

PR #3818, opened 2025-01-28
Fix nav bar color for user detail

bbirman:
before & after
<p>
Screenshot 2025-01-27 at 3 40 27 PM
Screenshot 2025-01-27 at 3 40 27 PM
</p>


wmathurin (approved):

Crebs:
I noticed these test-ios runs are failing on my PR too.  Is that related to any recent changes that might have gottne merged while my User permission stuff was failing?

bbirman:
> I noticed these test-ios runs are failing on my PR too. Is that related to any recent changes that might have gottne merged while my User permission stuff was failing?

If it's these four, I don't think so, they've been flapping for a bit. That's another thing we should look at before releasing if you want to work on that
```
testPublicApiCalls
testCredentialsCopying
testUserAccountEncoding
testLockScreenTriggers
```

Crebs (approved):

PR #3819, opened 2025-01-28
@W-12514054 Update device support and identification logic.

Crebs:
This PR updates the device identification logic to include support for newly released devices and refines existing mappings. Key changes include:
	•	Added identification support for the latest iPhone and iPad models
	•	General code cleanup.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
@@ -218,78 +218,126 @@
     
     // Placeholder for unknown platforms or iFPGA (rare and mostly unused)
     if ([platform isEqualToString:@"iFPGA"]) return UIDeviceIFPGA;
-
-    // iPhones (iOS 17 and newer supported)
-    if ([platform isEqualToString:@"iPhone12,8"]) return UIDeviceSE2iPhone;           // iPhone SE (2nd generation)
-    if ([platform isEqualToString:@"iPhone14,6"]) return UIDeviceSE3iPhone;           // iPhone SE (3rd generation)
-    if ([platform isEqualToString:@"iPhone11,8"]) return UIDeviceXRiPhone;            // iPhone XR
-    if ([platform isEqualToString:@"iPhone11,2"]) return UIDeviceXsiPhone;            // iPhone XS
-    if ([platform isEqualToString:@"iPhone11,4"] || [platform isEqualToString:@"iPhone11,6"]) return UIDeviceXsMaxiPhone; // iPhone XS Max
-    if ([platform isEqualToString:@"iPhone12,1"]) return UIDevice11iPhone;           // iPhone 11
-    if ([platform isEqualToString:@"iPhone12,3"]) return UIDevice11ProiPhone;        // iPhone 11 Pro
-    if ([platform isEqualToString:@"iPhone12,5"]) return UIDevice11ProMaxiPhone;     // iPhone 11 Pro Max
-    if ([platform isEqualToString:@"iPhone13,1"]) return UIDevice12MiniiPhone;       // iPhone 12 Mini
-    if ([platform isEqualToString:@"iPhone13,2"]) return UIDevice12iPhone;           // iPhone 12
-    if ([platform isEqualToString:@"iPhone13,3"]) return UIDevice12ProiPhone;        // iPhone 12 Pro
-    if ([platform isEqualToString:@"iPhone13,4"]) return UIDevice12ProMaxiPhone;     // iPhone 12 Pro Max
-    if ([platform isEqualToString:@"iPhone14,4"]) return UIDevice13MiniiPhone;       // iPhone 13 Mini
-    if ([platform isEqualToString:@"iPhone14,5"]) return UIDevice13iPhone;           // iPhone 13
-    if ([platform isEqualToString:@"iPhone14,2"]) return UIDevice13ProiPhone;        // iPhone 13 Pro
-    if ([platform isEqualToString:@"iPhone14,3"]) return UIDevice13ProMaxiPhone;     // iPhone 13 Pro Max
-    if ([platform isEqualToString:@"iPhone14,7"]) return UIDevice14iPhone;           // iPhone 14
-    if ([platform isEqualToString:@"iPhone14,8"]) return UIDevice14PlusiPhone;       // iPhone 14 Plus
-    if ([platform isEqualToString:@"iPhone15,2"]) return UIDevice14ProiPhone;        // iPhone 14 Pro
-    if ([platform isEqualToString:@"iPhone15,3"]) return UIDevice14ProMaxiPhone;     // iPhone 14 Pro Max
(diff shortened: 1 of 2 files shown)


wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
```diff
@@ -17,72 +17,117 @@
...
- iPhone2,1 ->    iPhone 3GS, N89/N88?
- 
- iPhone3,1 ->    iPhone 4/AT&T, N90
- iPhone3,2 ->    iPhone 4/Other Carrier?, ??
- iPhone3,3 ->    iPhone 4/Verizon, N92
- 
- iPhone4,1 ->    iPhone 4S/GSM, N94
- iPhone4,2 ->    iPhone 4S/???, ???
- iPhone4,3 ->    iPhone 4S/???, ???
+ // iPhone Models
```
bbirman, line 23: Can we take out the ones we don't support in the commented list too?

Crebs: Yes. I'll update that

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
```diff
@@ -187,119 +232,81 @@ - (NSNumber *)sfsdk_freeDiskSpace {
...
+    if ([platform isEqualToString:@"AppleTV14,1"]) return UIDeviceAppleTV4k3G; // Apple TV 4K (3rd generation)
+
+    // Simulators
     #if !TARGET_OS_VISION
-        if ([platform hasSuffix:@"86"] || [platform isEqual:@"x86_64"]) {
-            BOOL smallerScreen = [[UIScreen mainScreen] bounds].size.width < 768;
-            return smallerScreen ? UIDeviceSimulatoriPhone : UIDeviceSimulatoriPad;
-        }
+    UIDevice *device = [UIDevice currentDevice];
+    if (device.userInterfaceIdiom == UIUserInterfaceIdiomPhone) {
```
bbirman, line 284: This reads to me like it could also be an unknown iPhone instead of a simulator?

bbirman: could potentially use this to replace the architecture check https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/40de3762b541473a55685ff28013928a1cd30b9d/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice%2BSFHardware.m#L457

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
```diff
@@ -187,119 +216,81 @@ - (NSNumber *)sfsdk_freeDiskSpace {
...
+    // Apple TVs
+    if ([platform isEqualToString:@"AppleTV5,3"]) return UIDeviceAppleTV4; // Apple TV HD
+    if ([platform isEqualToString:@"AppleTV6,2"]) return UIDeviceAppleTV4k; // Apple TV 4K (1st generation)
+    if ([platform isEqualToString:@"AppleTV11,1"]) return UIDeviceAppleTV4k2G; // Apple TV 4K (2nd generation)
+    if ([platform isEqualToString:@"AppleTV14,1"]) return UIDeviceAppleTV4k3G; // Apple TV 4K (3rd generation)
+
+    // Simulators
     #if !TARGET_OS_VISION
-        if ([platform hasSuffix:@"86"] || [platform isEqual:@"x86_64"]) {
-            BOOL smallerScreen = [[UIScreen mainScreen] bounds].size.width < 768;
```
bbirman, line 297: iirc`#if !TARGET_OS_VISION` might be able to go away too if we're not using UIScreen anymore

Crebs: I'm still seeing a few place where UIScreen is being used in SalesforceSDKCore.  So maybe I'll leave this for now. Unless you feel strongly?

bbirman: This shouldn't impact the other places. The `#if !TARGET_OS_VISION` check is there because UIScreen isn't available on visionOS, so any use of UIScreen has a check around it at some level (either like this or the class being marked as unavailable on visionOS) to make sure it will compile on visionOS

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
```diff
@@ -187,119 +216,129 @@ - (NSNumber *)sfsdk_freeDiskSpace {
...
+    if ([self sfsdk_isSimulator]) {
+        UIDevice *device = [UIDevice currentDevice];
+        
+        switch (device.userInterfaceIdiom) {
+            case UIUserInterfaceIdiomPhone:
+                return UIDeviceSimulatoriPhone;
+            case UIUserInterfaceIdiomPad:
+                return UIDeviceSimulatoriPad;
+            default:
+                return UIDeviceSimulator; // Default case for unknown idioms (e.g., visionOS or future devices)
```
bbirman, line 233: I still think `#if !TARGET_OS_VISION` should be removed but if not the comment is confusing because visionOS will never execute here while that check in place

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/UIDevice+SFHardware.m
```diff
@@ -187,119 +216,127 @@ - (NSNumber *)sfsdk_freeDiskSpace {
...
-    // Simulator thanks Jordan Breeding
-    #if !TARGET_OS_VISION
-        if ([platform hasSuffix:@"86"] || [platform isEqual:@"x86_64"]) {
-            BOOL smallerScreen = [[UIScreen mainScreen] bounds].size.width < 768;
-            return smallerScreen ? UIDeviceSimulatoriPhone : UIDeviceSimulatoriPad;
-        }
-    #endif
-
-    return UIDeviceUnknown;
+    return UIDeviceUnknown;  // Unknown platform fallback
```
bbirman, line 339: I was just testing the latest and I noticed that the old code used to have more unknown subcategories (snippet below), thinking we should check those too before returning `UIDeviceUnknown` 

```
if ([platform hasPrefix:@"iPhone"]) return UIDeviceUnknowniPhone;
if ([platform hasPrefix:@"iPad"]) return UIDeviceUnknowniPad;
```

Crebs: ah, so add these before we return straight unknow?  OK, I'll add that back to make sure I don't regress expected behaviour here.

bbirman (approved):
thanks for getting all this updated!

PR #3821, opened 2025-01-29
@W-17694365 - Disabling flapping tests while we fix them

Crebs:
Disabling flapping tests and will add them back as they are fixed

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
@@ -263,48 +263,48 @@ class RestClientTests: XCTestCase {
         }
     }
     
-    func testAsyncBatchRequestStopOnFailure() async throws {
-        do {
-            // Create account
-            let accountName = self.generateRecordName()
-            let contactName = self.generateRecordName()
-            let apiVersion = RestClient.shared.apiVersion
-            
-            let requestBuilder = BatchRequestBuilder()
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
```diff
@@ -263,48 +263,48 @@ class RestClientTests: XCTestCase {
...
-            
-            XCTAssertTrue(resp1["statusCode"] as? Int == 201, "Wrong status for first request")
-            XCTAssertTrue(resp2["statusCode"] as? Int == 201, "Wrong status for first request")
-            XCTAssertTrue(resp3["statusCode"] as? Int == 400, "Wrong status for first request")
-            XCTAssertTrue(resp4["statusCode"] as? Int == 412, "Request processing should have stopped on error")
-        } catch {
-            XCTFail("Send Batch Request should not throw an error")
-        }
-    }
+//    func testAsyncBatchRequestStopOnFailure() async throws {
```
wmathurin, line 266: Couldn't you just rename it to _testAsyncBatchRequestStopOnFailure to have it skipped?
That way the body of the test would still be compiled (or get refactored if we rename some methods etc).

brandonpage: @Crebs I think you missed the _ instead of commenting on this test.

Crebs: ah yes, I missed this one. I'll fix that. 👍

brandonpage (approved):

PR #3822, opened 2025-01-29
Fix Github Actions issues.

brandonpage:
- Fix an issue where `pull_request_target` makes the default checkout ref the base branch instead of PR code.
- Add `workflow_dispatch` trigger so nightly tests can be run from the browser.
- Use Ruby gem cache. 
- Add a couple automated PR messages.

### DIFF
--- a/.github/DangerFiles/TestOrchestrator.rb
+++ b/.github/DangerFiles/TestOrchestrator.rb
@@ -1,3 +1,11 @@
+#!/usr/bin/env ruby
+
+# Warn when there is a big PR
+warn("Big PR, try to keep changes smaller if you can.", sticky: true) if git.lines_of_code > 1000
+
+# Redirect contributors to PR to dev.
+fail("Please re-submit this PR to the dev branch, we may have already fixed your issue.", sticky: true) if github.branch_for_base != "dev"
+
 # List of supported xcode schemes for testing
 SCHEMES = ['SalesforceSDKCommon', 'SalesforceAnalytics', 'SalesforceSDKCore', 'SmartStore', 'MobileSync']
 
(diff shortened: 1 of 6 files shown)


.github/DangerFiles/TestOrchestrator.rb
```diff
@@ -1,3 +1,11 @@
+#!/usr/bin/env ruby
+
+# Warn when there is a big PR
+warn("Big PR, try to keep changes smaller if you can.", sticky: true) if git.lines_of_code > 1000
+
+# Redirect contributors to PR to dev.
+fail("Please re-submit this PR to the dev branch, we may have already fixed your issue.", sticky: true) if github.branch_for_base != "dev"
```
bbirman, line 7: Does the whole check fail? Wondering about when we do PRs to master for cherry picks

brandonpage: I believe it only fails the Danger status check on the PR.  It should not stop the execution of tests or do much of anything else negative.

However, while writing this comment I realized it is actually completely pointless with the current configuration since CI is only triggered on `dev`.  So I think we have two choices:

1. Add `master` as a branch to trigger PRs.  We probably want to keep the line then and ignore the fail.  (Although I thought we typically just merge dev to master without a PR?).
2. Remove the warning since CI doesn't run against `master`.  

I think I like option #1 more?

bbirman: sounds good! it's been a while but this was the scenario I was thinking of https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3686

brandonpage: I double checked this on the Android testing setup.  It just displays this error, which is fine in the rare event we are cherry picking to master.

Screenshot 2025-01-29 at 5 27 23 PM

bbirman (approved):

PR #3825, opened 2025-01-30
Fix Nightly Test Runs

brandonpage:
Use Xcode 15 and 16 on macos-15 instead of macos-14 and and macos-15 to run tests on iOS 17 and 18 for Nightly test runs.


Crebs:
I'm slow at review this.  Missed the party.  Looks good 👍 .  Thanks Brandon.

PR #3827, opened 2025-02-04
@W-17749645 Resolve failing unit tests caused by handling of strongSelf

Crebs:
Resolve failing unit tests caused by handling of strongSelf in the accessTokenForRefresh:completion: method. Ensure the completion handler executes safely when self is deallocated, improving test stability and robustness of asynchronous operations.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
@@ -281,6 +281,7 @@ const NSTimeInterval kSFOAuthDefaultTimeout  = 120.0; // seconds
     NSURLSession *session = [self createURLSessionWithIdentifier:instanceIdentifier];
 
     __weak typeof(self) weakSelf = self;
+    NSString *className = NSStringFromClass([self class]);
     [[session dataTaskWithRequest:request completionHandler:^(NSData *data, NSURLResponse *urlResponse, NSError *error) {
         __strong typeof(weakSelf) strongSelf = weakSelf;
         SFSDKOAuthTokenEndpointResponse *endpointResponse = nil;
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
```diff
@@ -293,21 +294,29 @@ - (void)accessTokenForRefresh:(SFSDKOAuthTokenEndpointRequest *)endpointReq comp
...
                 }
             });
             return;
-        } else {
-            [SFSDKEventBuilderHelper createAndStoreEvent:@"tokenRefresh" userAccount:[SFUserAccountManager sharedInstance].currentUser className:NSStringFromClass([strongSelf class]) attributes:nil];
         }
         
-        [strongSelf handleTokenEndpointResponse:completionBlock request:endpointReq data:data urlResponse:urlResponse];
+        [SFSDKEventBuilderHelper createAndStoreEvent:@"tokenRefresh" userAccount:[SFUserAccountManager sharedInstance].currentUser className:className attributes:nil];
+        if (strongSelf) {
```
bbirman, line 310: What was the test failure this was causing?

Crebs: Everytime I'd run SalesforceSDKCore UnitTests, I was getting an Error that "a method should be called on main" .  But after doing some debugging and tracking it down, it was because this we were setting a nil value into a dictionary during a test.  Inside of SFSDKEventBuilderHelper line 57.   The reason it was crashing every time was because the test strongSelf was nil every time. The way the life cycle was set up for the test would cause self to be released and nil to be inserted in the dictionary.  It was happening because of SalesforceSDKIdentityTests and TestSetupUtils synchronousAuthRefresh lifecycle.   Let me know if you want sync up and chat about this too.

wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
bbirman: Is there a urlResponse when strongSelf is nil? Wondering if something is being deallocated too early

PR #3829, opened 2025-02-05
[DO NOT MERGE] @W-17736609 adding back test

Crebs:
Adding back flappy test to see if another issue we fix resolves this issue.


wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCoreTests/ScreenLockManagerTests.swift
```diff
@@ -50,37 +49,37 @@ class ScreenLockManagerTests: XCTestCase {
         ScreenLockManagerInternal.shared.handleAppForeground()
         XCTAssertTrue(SFSDKWindowManager.shared().screenLockWindow().isEnabled())
         wait(for: [expectation], timeout: 1)
-        
+//        
         // Unlock
         SFSDKWindowManager.shared().screenLockWindow().dismissWindow()
```
Crebs, line 54: So the tests run fine locally.  However, when run on CI, this line is causing an exception to be thrown.  I had to basically do a bit of a committing bisect where I’d comment out some code and comment the other half in to narrow down where the failure seems to be happening. When you log at the CI logs it doesn't give a lot of informaiton as what is happening.

PR #3838, opened 2025-02-25
No longer returning a db object if encryption can't be used

wmathurin:
- Using FMDB's goodConnection method to check the state of the database
- Removed fixFor530Bug which would automatically encrypt a non-encrypted database - with the use of goodConnection, app won't be able to get a non-encrypted database

### DIFF
--- a/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
+++ b/libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
@@ -146,28 +146,7 @@ static NSString * const kSFSmartStoreVerifyReadDbErrorDesc = @"Could not read fr
     return [[self class] openDatabaseWithPath:fullDbFilePath key:key salt:salt error:error];
 }
 
-// If you created your database with an app based on Mobile SDK 5.3.0 using cocoapod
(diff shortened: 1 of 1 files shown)


libs/SmartStore/SmartStore/Classes/SFSmartStoreDatabaseManager.m
```diff
@@ -440,21 +419,16 @@ + (FMDatabase *)encryptOrUnencryptDb:(FMDatabase *)db
 
 + (BOOL)verifyDatabaseAccess:(FMDatabase *)db error:(NSError **)error
 {
-    NSString *sqlCommand = @"SELECT name FROM sqlite_master LIMIT 1";
-    FMResultSet *rs = [db executeQuery:sqlCommand];
-    if (rs == nil) {
-        // May not be results, but rs should never be nil coming back.
+    if (![db goodConnection]) {
```
wmathurin, line 422: FMDB's goodConnection: https://github.com/ccgus/fmdb/blob/2.7.12/src/fmdb/FMDatabase.m#L512

brandonpage (approved):

bbirman (approved):

PR #3847, opened 2025-03-25
Bug Fix - popup loop after connected app oauth access is blocked

Crebs:
Refactored and modernized SFSDKWebViewStateManager into a Swift-based WebViewStateManager, added async/await support, improved Objective-C compatibility, and wrote comprehensive Swift unit tests to validate cookie and session management behavior.

I made a new Method to forceully reset cookies. So we basically do the same thing that Android can do.


wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
```diff
@@ -0,0 +1,102 @@
...
+
+    @objc
+    @MainActor
+    public static func removeSessionForcefully() async {
+        sharedProcessPool = nil
+        await removeWKWebViewCookies()
+    }
+
+    @objc
+    public static func resetSessionCookie() async {
```
bbirman, line 55: I think you also need to have a non-async `resetSessionCookie` so that it doesn't break the original API footprint, you can check out the generated Objective-C header to compare it to the old class. 

The generated one is `+ (void)resetSessionCookieWithCompletionHandler:(void (^ _Nonnull)(void))completionHandler;` while the original was `+ (void)removeSession;`

Screenshot 2025-03-25 at 3 34 59 PM

bbirman: updated for my typo :)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
```diff
@@ -0,0 +1,102 @@
...
+        set {
+            if newValue !== processPool {
+                SFSDKCoreLogger.i(WebViewStateManager.self, message: "[\(Self.self) setSharedProcessPool]: Changing from \(String(describing: processPool)) to \(String(describing: newValue))")
+                processPool = newValue
+            }
+        }
+    }
+
+    @objc
+    public static func setSessionCookieManagementDisabled(_ disabled: Bool) {
```
bbirman, line 94: It looks like this and the method below are a breaking API change too since it was a property before. Could it be a `static var` or was there an issue with it?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
Crebs: ah, Good ctach.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
Crebs: Ah, yes.  I'll fix that too.  Thank you.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
```diff
@@ -0,0 +1,101 @@
...
+//  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+//  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+//  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import Foundation
+import WebKit
+
+@objc(SFSDKWebViewStateManager)
+public class WebViewStateManager: NSObject {
```
bbirman, line 27: I noticed that in the old code we didn't have a Swift name for it so even in Swift the class was `SFSDKWebViewStateManager`. I agree `WebViewStateManager` is the better name for Swift but it would also mean a class name change in a minor release, do we want to do that?

bbirman: one option could be to have a wrapping Swift only class with the old name and a deprecation warning to move to the new name

wmathurin: > one option could be to have a wrapping Swift only class with the old name and a deprecation warning to move to the new name

I like that idea.

bbirman (approved):
excited for the conversion!

PR #3862, opened 2025-05-07
@W-18250531 Refactor SFRestAPI Block Support and Migrate to Swift-Based Async Interface

Crebs:
✅ What Changed
	•	✅ Removed SFRestAPI+Blocks.m (legacy Objective-C block support)
	•	✅ Replaced with Swift implementations:
	•	RestClient+Async.swift – introduces async/await methods for sending REST, composite, and batch requests
	•	RestAPI+Blocks.swift – adds Swift-compatible block-based APIs
	•	✅ Added RestClient+Async.swift to Xcode project
	•	✅ Removed unused headers (SFSDKBatchResponse+Internal.h, etc.)
	•	✅ Updated SFRestAPI to:
	•	Provide native Objective-C block-based request sending (send:failureBlock:successBlock:)
	•	Route delegate and block requests through a unified enqueue pipeline
	•	✅ Rewrote the token refresh logic in replayRequest:response: to:
	•	Use a serialized dispatch_queue_t with a dispatch_semaphore_t to ensure one refresh at a time
	•	Block all request retries until refresh completes

⸻

⚠️ Why
	•	Legacy SFRestAPI+Blocks.m was using objc_setAssociatedObject, which was fragile and hard to debug
	•	Bringing request control into Swift makes block and async usage much safer, clearer, and more modern
	•	Fixes a long-standing issue where only one request was retried after a failed token refresh, while others were dropped
	•	Greatly improves testability and Swift interop

⸻

🧪 How to Test
	•	Run PushNotificationManagerTests.swift, especially testInvalidAccessAndRefreshToken_MultipleRequests
	•	Ensure multiple simultaneous requests get retried correctly after an auth refresh
	•	Validate that Swift and ObjC callers can use block and async methods without regressions

⸻

🧹 Cleanup
	•	Removed unused or redundant headers and references from .pbxproj
	•	Consolidated initWith: methods in BatchResponse and CompositeResponse into public headers

⸻


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFSDKBatchResponse.h
```diff
@@ -29,6 +29,7 @@ NS_SWIFT_NAME(BatchResponse)
 @interface SFSDKBatchResponse : NSObject
 @property (nonatomic,readonly) BOOL hasErrors;
 @property (nonatomic,readonly) NSArray<id> *results;
+-(instancetype)initWith:(NSDictionary *)dict;
```
Crebs, line 32: Compositie Response had it's int in the normal header file.  I need this Batch one to be in the heard file as well so I could call it from swift.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -231,10 +228,39 @@ - (void)send:(SFRestRequest *)request requestDelegate:(nullable id<SFRestRequest
...
+                                 data:response
+                          rawResponse:rawResponse
+                                error:e];
+    } successBlock:^(id  _Nullable response, NSURLResponse * _Nullable rawResponse) {
+        [self notifyDelegateOfSuccess:requestDelegate
+                              request:request
+                                 data:response
+                          rawResponse:rawResponse];
+    } shouldRetry: shouldRetry];
+}
```
Crebs, line 248: 💬 This change bridges the legacy delegate-based request path with the new block-based pipeline.

Previously, block-based requests were injected into the delegate system at runtime using objc_setAssociatedObject, which was fragile and hard to reason about. Now, everything below this point uses native block handling without relying on Objective-C runtime hacks.

This method preserves the public delegate API but internally forwards to the new unified block-based implementation, allowing us to consolidate logic, improve maintainability, and ensure consistent behavior across both patterns.

See the related diff for the full context on the transition from SFRestAPI+Blocks.m to Swift-native block and async support.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Blocks.m
```diff
@@ -1,124 +0,0 @@
...
-#import "SFSDKBatchResponse+Internal.h"
-#import "SFSDKCompositeResponse+Internal.h"
-#import <objc/runtime.h>
-
-// Pattern demonstrated in the Apple documentation. We use a static key
-// whose address will be used by the objc_setAssociatedObject (no need to have a value).
-static char FailureBlockKey;
-static char SuccessBlockKey;
-
-@implementation SFRestAPI (Blocks)
```
bbirman, line 41: It looks like `SFRestAPI+Blocks.m` is deleted here, what about its header file?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -407,48 +447,54 @@ - (NSError*) prepareErrorForDelegate:(NSData *)data response:(NSURLResponse *)re
...
+                
+                // Make sure we call logout on the main thread.
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    [[SFUserAccountManager sharedInstance] logoutUser:strongSelf.user reason:SFLogoutReasonTokenExpired];
+                });
+            }
+            dispatch_semaphore_signal(refreshGate); // Also mark done on failure
+        }];
+        dispatch_semaphore_wait(refreshGate, DISPATCH_TIME_FOREVER); // Wait until signal
+    });
```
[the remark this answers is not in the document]
bbirman: In the new code, what triggers sending the remaining requests in the pending queue after a successful refresh like `resendActiveRequestsRequiringAuthentication` did?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -292,29 +321,40 @@ - (void)enqueueRequest:(SFRestRequest *)request requestDelegate:(id<SFRestReques
...
                 return;
             }
             NSInteger statusCode = [(NSHTTPURLResponse *)response statusCode];
 
             // 2xx indicates success.
             if ([SFRestAPI isStatusCodeSuccess:statusCode]) {
                 id dataForDelegate = [strongSelf prepareDataForDelegate:data request:request response:response];
-                [strongSelf notifyDelegateOfSuccess:requestDelegate request:request data:dataForDelegate rawResponse:response];
+                if (successBlock) {
+                    successBlock(dataForDelegate, response);
```
bbirman, line 345: Is `[self removeActiveRequestObject:request];` needed for the success case too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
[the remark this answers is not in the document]
Crebs: Hey  @bbirman great question 👋 — in the new block-based implementation, we no longer need to manually resend all pending requests like resendActiveRequestsRequiringAuthentication did.

Each request now carries its own retry logic via its block closure. So when a request fails due to an expired token and triggers a refresh, it waits for the refresh to complete (via a dispatch queue + semaphore) and then replays only itself on line 475 in the code block above.

This means:
	•	No shared delegate callbacks to juggle
	•	No global replay of all pending requests
	•	Each request is self-contained, simpler, and more predictable

Let me know if you’d like a visual breakdown or want to walk through one of the flows 🙌

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
Crebs: you might be right.  Let me see if I can move [self removeActiveRequestObject:request]; up a level in the callback so we know it's not getting missed anywhere.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Blocks.m
Crebs: We removed SFRestAPI+Blocks.m because its Objective-C implementation was redundant after moving to the new Swift-based block and async API. However, we kept the header file SFRestAPI+Blocks.h for legacy support and compatibility reasons.

There are still a number of references to the header across tests and in our umbrella headers, and keeping it around ensures we don’t break existing consumers that may still include it. Since the underlying functionality is now handled by Swift, any calls made through the interface will correctly route through the Swift implementation.

That said, if we ever decide to fully clean house, we can do a follow-up cleanup PR to remove the .h and fix the dependent references. For now, it’s a pragmatic choice to keep the interface stable.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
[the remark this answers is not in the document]
bbirman: If multiple requests are sent at the same time and they all get a 401, we should only trigger the token refresh once. Based on what you said and testing locally it looks this is triggering a token refresh for every 401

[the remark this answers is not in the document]
Crebs: Ah OK. So maybe I need to add the  back in resendActiveRequestsRequiringAuthentication.  I'll through through this again.  thanks for keeping me honest here.

wmathurin (approved):
LGTM

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
bbirman: just saw the new way and it looks way cleaner!

wmathurin:
Some AI told me that if the request is deallocated early, successBlock/failureBlock might not be retained and as a result the system might fail to call them causing the continuation to not resume.

Here is the modified implementation of  https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift#L140 that holds on to the request to prevent that situation from arising. Other methods might need to be changed also.

```swift
public func send(request: RestRequest) async throws -> RestResponse {
    request.parseResponse = false
    
    return try await withCheckedThrowingContinuation { continuation in
        var strongRequest: RestRequest? = request // NEW
        
        __send(strongRequest!,
               failureBlock: { rawResponse, error, urlResponse in
            defer { strongRequest = nil } // NEW
            let apiError = RestClientError.apiFailed(
                response: rawResponse,
                underlyingError: error ?? RestClientError.apiResponseIsEmpty,
                urlResponse: urlResponse
            )
            continuation.resume(throwing: apiError)
        },
               successBlock: { rawResponse, urlResponse in
            defer { strongRequest = nil } // NEW
            if let data = rawResponse as? Data,
               let urlResponse = urlResponse {
                let result = RestResponse(data: data, urlResponse: urlResponse)
                continuation.resume(returning: result)
            } else {
                continuation.resume(throwing: RestClientError.apiResponseIsEmpty)
            }
        })
    }
}
```

wmathurin:
I tried dev (never see the issue reported by @bbirman), Riley's branch (saw the issue / but not every time) and Riley's branch with the strongRequest change (not seeing the issue).
We probably need more testing.

Crebs:
So I tried it with the fix that Wolf gave above.  Seems to fix the issue.  @bbirman not sure if you wanted to double check my work there.

bbirman:
I still see the issue on the latest. I also see a warning: `Variable 'strongRequest' was written to, but never read`. How did you test it?

Crebs:
@bbirman OK, so I was able to reproduce the issue with the strong reference still as well.  Looks like the block was getting lost when we’d go to the retry after the token was revoked and it'd call back the delegate method, not any of the blocks would get called back in this case.  So I changed it so the delegate calls are only called if the request was sent from a delegate in the first place; otherwise, we use the blocks as the default behaviour to call back for the request. 


I ran this senario on dev and the logs would out put:

```
CLASS: SFJsonUtils WARNING error parsing json: Error Domain=NSCocoaErrorDomain Code=3840 "No string key for value in object around line 1, column 1." UserInfo={NSDebugDescription=No string key for value in object around line 1, column 1., NSJSONSerializationErrorIndex=1}
CLASS: SFRestAPI response request path: /v63.0
CLASS: SFRestAPI response code: 401
CLASS: SFRestAPI replayRequest:response:requestDelegate:: REST request failed due to expired credentials. Attempting to refresh credentials.
CLASS: SFSDKOAuth2 accessTokenForRefresh:completion:: Initiating refresh token flow.
CLASS: SFOAuthSessionRefresher completeWithSuccess Session was successfully refreshed.
CLASS: SFRestAPI replayRequest:response:requestDelegate:: Credentials refresh successful. Replaying original REST request.
CLASS: SFJsonUtils invalid object passed to JSONDataRepresentation???
```

No with the changes I've made the longs now out put:

```
CLASS: SFJsonUtils WARNING error parsing json: Error Domain=NSCocoaErrorDomain Code=3840 "No string key for value in object around line 1, column 1." UserInfo={NSDebugDescription=No string key for value in object around line 1, column 1., NSJSONSerializationErrorIndex=1}
CLASS: SFRestAPI response request path: /v63.0
CLASS: SFRestAPI response code: 401
CLASS: SFRestAPI replayRequest:response:failureBlock:successBlock:: REST request failed due to expired credentials. Attempting to refresh credentials.
CLASS: SFSDKOAuth2 accessTokenForRefresh:completion:: Initiating refresh token flow.
CLASS: SFOAuthSessionRefresher completeWithSuccess Session was successfully refreshed.
CLASS: SFRestAPI replayRequest:response:failureBlock:successBlock:: Credentials refresh successful. Replaying original REST request.
CLASS: SFJsonUtils invalid object passed to JSONDataRepresentation???
```

The log stacks are basically the same, but calling blocks back instead of the delegates.  I've ran this seveerl times now and I no longer see the `SWIFT TASK CONTINUATION MISUSE: send(request:) leaked its continuation!`.

wmathurin:
So much for AI debugging !

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI+Blocks.m
bbirman: Hmm, I think there might be an issue there since at least one of the methods are redefined in the new Swift files.

I tried calling the error method in Swift and it gives me this error unless I delete the signature in SFRestAPI+Blocks.h, I haven't tried in Objective-c.

Screenshot 2025-05-15 at 12 13 06 PM

If we want to be able to rewrite public headers in place, we might need to do something like leave SFRestAPI+Blocks.h + SFRestAPI+Blocks.m but delete any method that's rewritten in Swift and have `#import <SalesforceSDKCore/SalesforceSDKCore-Swift.h>` in the .m file to link to the new ones for the interface stability you mentioned until we can delete them in 14.0. Haven't tried it and would need more testing to confirm

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Analytics/SFSDKAILTNPublisher.m
```diff
@@ -32,10 +32,10 @@
 #import "SalesforceSDKManager.h"
 #import "NSData+SFAdditions.h"
 #import "SFRestAPI+Blocks.h"
+#import "SalesforceSDKCore/SalesforceSDKCore-Swift.h"
 
 static NSString* const kCode = @"code";
 static NSString* const kAiltn = @"ailtn";
-static NSString* const kSchemaTypeKey = @"schemaType";
```
bbirman, line 38: why is this deleted?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -551,9 +551,9 @@ - (BOOL)authenticateWithCompletion:(SFUserAccountManagerSuccessCallbackBlock)com
                             codeVerifier:codeVerifier];
 }
 
-- (SFSDKAuthRequest *)defaultAuthRequestWithLoginHost:(nullable NSString *)loginHost {
+-(SFSDKAuthRequest *)defaultAuthRequestWithLoginHost:(nullable NSString *)loginHost {
     SFSDKAuthRequest *request = [[SFSDKAuthRequest alloc] init];
-    request.loginHost = loginHost != nil ? loginHost : self.loginHost;
+    request.loginHost = loginHost != nil ? loginHost : self.loginHost;;
```
bbirman, line 556:
```suggestion
    request.loginHost = loginHost != nil ? loginHost : self.loginHost;
```

bbirman (approved):

PR #3863, opened 2025-05-08
Keep prior add connection functionality with removing scheme.

mbwimmer:
The current behavior if a scheme is included in the add connection host url:
Input: `https://test.salesforce.com`
Result: `https://t`
Expected: `test.salesforce.com`

Fix: 
If a scheme is included in the host input field, the scheme and "://" are removed from the host url variable.

@bbirman

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
@@ -82,9 +82,9 @@ struct NewLoginHostView: View {
     
     func save() {
         var hostToSave = host.trimmingCharacters(in: .whitespaces)
-        if let httpsRange = hostToSave.range(of: "://"),
-           let components = URLComponents(string: hostToSave),
-           let scheme = components.scheme {
+        if hostToSave.contains("://"),
+           let components = URLComponents(string: hostToSave) {
+            let scheme = components.scheme ?? ""
             hostToSave = String(hostToSave.dropFirst("\(scheme)://".count))
         }
         saveAction(hostToSave, hostLabel.trimmingCharacters(in: .whitespaces))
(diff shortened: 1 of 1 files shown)


mbwimmer:
@bbirman no problem. I just updated the PR to account for a nil or missing scheme. Prior to MSDK 13, you can add some other ill formed URLs (ie `:test.salesforce.com, //test.salesforce.com`, etc.) I don't know if you'd want to account for those as well.

Crebs:
Do we know if this wil affect existing cases of URL scheme validation?

mbwimmer:
> Do we know if this wil affect existing cases of URL scheme validation?

@Crebs I'm not sure what other scheme validation we do. I'm pretty sure the reason why we pull off the scheme in this case is so the SDK can control the scheme and always set it to `https`.  As far as I know, once the modified URL is passed to the completion closure, any other validation code has been unchanged. This fix is specifically for removing the scheme in `NewLoginHost` which preserves the same functionality previously in [SFSDKNewLoginHostViewController](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/a8addb3d9570f377f6e8c4c76cbc7d16466d62fe/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKNewLoginHostViewController.m#L77)
```
- (void)addNewServer:(id)sender {
    NSString *hostName = [self.name.text stringByTrimmingCharactersInSet:[NSCharacterSet whitespaceCharacterSet]];
    NSString *host = [self.server.text stringByTrimmingCharactersInSet:
                          [NSCharacterSet whitespaceCharacterSet]];
    NSRange httpsRange = [host rangeOfString:@"://"];
    if (host && httpsRange.length > 0) {
        host = [host substringFromIndex:httpsRange.location + httpsRange.length];
    }
    [self.loginHostListViewController addLoginHost:[SFSDKLoginHost hostWithName:hostName host:host  deletable:YES]];
}
```

PR #3866, opened 2025-05-16
Add tests for login host handling

bbirman:
Add test for the login host based on this issue & fix: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3863


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -80,7 +80,7 @@ struct NewLoginHostView: View {
         }
     }
     
-    func save() {
+    func save(host: String, hostLabel: String) {
```
bbirman, line 83: This is to make it testable, could also move it into a view model but this is the only logic in the view so I was on the fence if it was worth it

wmathurin (approved):

PR #3870, opened 2025-05-20
@W-18542896 Provide a way for the application to specify notification types / actions it actually supports

Crebs:
# Add Support for Filtering Notification Types and Actions

## Problem
Currently, all notification types and actions received from the server are automatically registered with the system, even if the app doesn't support or handle them. This can lead to a poor user experience where users see notification actions that don't work or aren't relevant to the app's functionality.

## Solution
Added a new property `filterSupportedNotificationTypes` to `UserAccountManager` that allows apps to specify which notification types they support. This filter is applied when creating notification categories, ensuring that only supported types and their actions are registered with the system.

### Key Changes
- Added `filterSupportedNotificationTypes` property to `UserAccountManager`
- Modified `setNotificationCategories` in `PushNotificationManager` to apply the filter
- Added unit tests to verify filtering behavior

### Usage
Apps can now filter notification types by setting the filter property:

```swift
// Only allow specific notification types
UserAccountManager.shared.filterSupportedNotificationTypes = { types in
    return types.filter { $0.apiName == "supported_type" }
}

// Or allow all types (default behavior)
UserAccountManager.shared.filterSupportedNotificationTypes = nil
```

### Testing
Added comprehensive test coverage:
- `testSetNotificationCategories_WithFilter`: Verifies that only supported notification types are registered
- `testSetNotificationCategories_NoFilter`: Verifies that all types are registered when no filter is set

## Impact
This change improves the user experience by ensuring that only relevant notification actions are displayed to users. It also provides apps with more control over which notification types they support and handle.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/PushNotificationManagerTests.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/PushNotificationManagerTests.swift
@@ -645,6 +645,90 @@ class PushNotificationManagerTests: XCTestCase {
         XCTAssertEqual(mockUserAccount.notificationTypes?.count, 1)
         XCTAssertEqual(mockUserAccount.notificationTypes?.first?.apiName, "cached_type")
     }
+
+    func testSetNotificationCategories_WithFilter() {
+        // Given
+        let mockTypes = [
+            NotificationType(type: "supported_type", apiName: "supported_type", label: "Supported Type", actionGroups: [
+                ActionGroup(name: "group1", actions: [
+                    Action(name: "action1", identifier: "group1__action1", label: "Action 1", type: .notificationApiAction)
+                ])
+            ]),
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCoreTests/PushNotificationManagerTests.swift
```diff
@@ -645,6 +645,90 @@ class PushNotificationManagerTests: XCTestCase {
         XCTAssertEqual(mockUserAccount.notificationTypes?.count, 1)
         XCTAssertEqual(mockUserAccount.notificationTypes?.first?.apiName, "cached_type")
     }
+
```
wmathurin, line 648: Could we have a test / example code showing a specific action being filtered out?

wmathurin: Maybe a helper method to create a NotificationType from an existing NotificationType but keeping only specific actions?

wmathurin (approved):
LGTM!

PR #3871, opened 2025-05-27
@W-18250537 Add support for direct URLRequest support to MSDK REST client

Crebs:
🚀 Summary

This PR adds support for sending native URLRequest instances directly through the Salesforce Mobile SDK REST client. This feature simplifies networking for developers who already use URLRequest, eliminating the need to wrap requests in SFRestRequest for basic use cases.

✅ What’s Included

# New Async Await
```
public func send(urlRequest: URLRequest) async throws -> (Data, URLResponse)
```
# New Objc block equivalent 
```
- (void)sendURLRequest:(NSURLRequest *)urlRequest
            completion:(void (^)(NSData *data, NSURLResponse *response, NSError *error))completion;
```

See [Test Plan: Direct URLRequest to MSDK](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/compare/link-if-applicable) for full details.

⸻

🛡️ Compatibility
	•	SFRestRequest support remains unchanged and continues to be the default abstraction.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
@@ -115,11 +115,15 @@ extension RestClient {
     
     /// Sends a URLRequest
     /// - Parameter urlRequest: The URLRequest to send
+    /// - Parameter networkServiceType: Specify a service type or use the default.
+    /// - Parameter requiresAuthentication: Specify requires auth, default to true.
     /// - Returns: A RestResponse containing the response data and metadata
     /// - Throws: A RestClientError if the request fails
-    @objc(sendURLRequest:completion:)
-    public func send(urlRequest: URLRequest) async throws -> (Data, URLResponse) {
-        guard let restRequest = urlRequest.toRestRequest() else {
+    @objc(sendURLRequest:networkServiceType:requiresAuthentication:completion:)
+    public func send(urlRequest: URLRequest,
+                     networkServiceType: RestRequest.NetWorkServiceType = RestRequest.NetWorkServiceType.SFNetworkServiceTypeDefault,
+                     requiresAuthentication: Bool = true) async throws -> (Data, URLResponse) {
+        guard let restRequest = urlRequest.toRestRequest(networkServiceType, requiresAuthentication) else {
             throw RestClientError.invalidRequest("Request is not a valid MSDK REST request.")
         }
         let response = try await send(request: restRequest)
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
@@ -113,6 +113,19 @@ extension RestClient {
         var records: [Record]?
     }
     
+    /// Sends a URLRequest
+    /// - Parameter urlRequest: The URLRequest to send
+    /// - Returns: A RestResponse containing the response data and metadata
+    /// - Throws: A RestClientError if the request fails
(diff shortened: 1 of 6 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -113,6 +113,19 @@ extension RestClient {
         var records: [Record]?
     }
     
+    /// Sends a URLRequest
+    /// - Parameter urlRequest: The URLRequest to send
+    /// - Returns: A RestResponse containing the response data and metadata
+    /// - Throws: A RestClientError if the request fails
+    @objc(sendURLRequest:completion:)
+    public func send(urlRequest: URLRequest) async throws -> (Data, URLResponse) {
```
bbirman, line 121: Should we expose parameters for [networkServiceType](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/0b0eda4f92258530d4e3925d07873fc74f5ba828/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.h#L129) and [requiresAuthentication](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/0b0eda4f92258530d4e3925d07873fc74f5ba828/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.h#L188)?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -113,6 +113,19 @@ extension RestClient {
...
+        guard let restRequest = urlRequest.toRestRequest() else {
+            throw RestClientError.invalidRequest("Request is not a query request.")
```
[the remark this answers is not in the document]
Crebs: OK, let me think of a better generanl error message.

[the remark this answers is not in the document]
Crebs: What doyou think of something simple like, "Not able to create Rest Request."?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
Crebs: I'm actualy torn between two options. These are specific to a RestRequest, not a URLRequest. Therefore, I’m inclined to opt for the simpler approach and avoid exposing more details to maintain the API. However, I’m open to changing my mind if there’s a clear use case that I’m overlooking.

wmathurin: Could we add them as optional parameters on sendRequest ?
```swift
@objc(sendURLRequest:completion:)
public func send(
    urlRequest: URLRequest,
    networkServiceType: RestRequest.NetWorkServiceType = .default,
    requiresAuthentication: Bool = true
) async throws -> (Data, URLResponse) {
...
}
```

Could we also fix the comment for https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestRequest.h#L41C1-L49C49 (it's copied over from SFRestMethod.

Also, it's unfortunate that the swift name for network service type is NetWorkServiceType but we can't fix that :-(

wmathurin: I see. Hopefully there is nobody out there that want to use the new API from objective-c code.

wmathurin (approved):

PR #3876, opened 2025-06-04
@W-18250526 Add Websocket Support for MSDK RestClient

Crebs:
This PR adds foundational WebSocket support to the Salesforce Mobile SDK for iOS. It allows consumers to open URLSessionWebSocketTask connections by passing a fully configured URLRequest, enabling use cases such as streaming audio to services like the Einstein Transcription API. The implementation is lightweight and non-opinionated, delegating responsibility for authentication and request construction to the application layer. Includes mock-based unit tests for verifying correct WebSocket setup and request handling.


brandonpage (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient+WebSocket.swift
```diff
@@ -0,0 +1,40 @@
...
+    /// Creates a new WebSocket from a URLRequest.
+    /// - Parameters:
+    ///   - request: The URLRequest.
+    ///   - delegate: The WebSocket delegate.
+    /// - Returns: A configured URLSessionWebSocketTask.
+    public func newWebSocket(
+        from request: URLRequest,
+        delegate: URLSessionWebSocketDelegate
+    ) -> URLSessionWebSocketTask {
+        let session = URLSession(configuration: .default, delegate: delegate, delegateQueue: .main)
```
bbirman: I think creating a URLSession for each request could be expensive. We have [SFNetwork](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFNetwork.h) for maintaining shared sessions for different configurations, is it possible to integrate with it?

bbirman:
> @bbirman
> 
> What do you think of this approcach to updated the token when making the websocket?
> 
> ```
>     public func newWebSocket(
>         from request: URLRequest,
>         delegate: URLSessionWebSocketDelegate
>     ) async throws -> URLSessionWebSocketTask {
>         let signedRequest = try await prepareWebSocketRequest(request)
>         let session = URLSession(configuration: .default, delegate: delegate, delegateQueue: .main)
>         return session.webSocketTask(with: signedRequest)
>     }
> ```
> 
> ```
>     private func prepareWebSocketRequest(_ request: URLRequest) async throws -> URLRequest {
>         var mutableRequest = request
>         if let account = UserAccountManager.shared.currentUserAccount {
>             try await refreshCredentials(for: account)
>             if let token = account.credentials.accessToken {
>                 mutableRequest.setValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
>             }
>         }
>         return mutableRequest
>     }
> ```
> 
> ```
>     private func refreshCredentials(for account: UserAccount) async throws {
>         try await withCheckedThrowingContinuation { continuation in
>             _ = UserAccountManager.shared.refresh(credentials: account.credentials) { result in
>                 switch result {
>                 case .success:
>                     continuation.resume()
>                 case .failure(let error):
>                     continuation.resume(throwing: error)
>                 }
>             }
>         }
>     }
> }
> ```

I don't think we'd want to refresh on every call, 

> @bbirman
> 
> What do you think of this approcach to updated the token when making the websocket?
> 
> ```
>     public func newWebSocket(
>         from request: URLRequest,
>         delegate: URLSessionWebSocketDelegate
>     ) async throws -> URLSessionWebSocketTask {
>         let signedRequest = try await prepareWebSocketRequest(request)
>         let session = URLSession(configuration: .default, delegate: delegate, delegateQueue: .main)
>         return session.webSocketTask(with: signedRequest)
>     }
> ```
> 
> ```
>     private func prepareWebSocketRequest(_ request: URLRequest) async throws -> URLRequest {
>         var mutableRequest = request
>         if let account = UserAccountManager.shared.currentUserAccount {
>             try await refreshCredentials(for: account)
>             if let token = account.credentials.accessToken {
>                 mutableRequest.setValue("Bearer \(token)", forHTTPHeaderField: "Authorization")
>             }
>         }
>         return mutableRequest
>     }
> ```
> 
> ```
>     private func refreshCredentials(for account: UserAccount) async throws {
>         try await withCheckedThrowingContinuation { continuation in
>             _ = UserAccountManager.shared.refresh(credentials: account.credentials) { result in
>                 switch result {
>                 case .success:
>                     continuation.resume()
>                 case .failure(let error):
>                     continuation.resume(throwing: error)
>                 }
>             }
>         }
>     }
> }
> ```

Instead of making a refresh call every time I would check out SFRestAPI. It sends the request and then checks the response code ([code](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/6d8147b6856482f0197d9468a27c1b0de5a1f85f/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m#L378)) to see if the token needs to be refreshed. Could you see how much of that we can reuse?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient+WebSocket.swift
Crebs: Ah OK.  I'll look into that and see how it works there.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient+WebSocket.swift
```diff
@@ -0,0 +1,40 @@
...
+//  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+//  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+//  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+//  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+extension RestClient {
+    /// Creates a new WebSocket from a URLRequest.
+    /// - Parameters:
+    ///   - request: The URLRequest.
```
wmathurin, line 32: Should we have a method that starts with a RestRequest too?

Crebs: yeah, I can just do it with a rest reqest only if we want

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient+WebSocket.swift
Crebs: Thanks Brianna — great point about URLSession reuse.

In our current implementation, we’re creating a new URLSession for each WebSocket connection to allow the caller to specify a custom URLSessionWebSocketDelegate. However, if the session is provided by SFNetwork, the delegate is already assigned at the shared session level — so we can’t override it per request.

That creates a bit of a constraint: if we were to use the shared session from SFNetwork, we’d need a way to route all WebSocket delegate callbacks through the existing shared delegate. That could be doable, but it would require us to design and manage our own routing layer inside the delegate — especially to support things like per-task state.

So in short:
	•	Yes, we can integrate with SFNetwork, but it would require a different pattern for delegate management.
	•	Right now, using a dedicated session per connection gives us a straightforward path to support custom delegates per WebSocket without affecting the rest of the SDK’s networking stack.

Would love thoughts if we want to explore routing via a shared delegate or keep it simple for now.

@wmathurin @bbirman

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient+WebSocket.swift
Crebs: We can I guess with this token refresh work we might have to just switch to a RestRequest method. Would we need both or can we just us one Method with RestRequest only?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient+WebSocket.swift
Crebs: Or what if SFNetwork has two sessions one for the curerent flow and one that we use for websockets?  Thoughts?

wmathurin: If we reuse the URLSession, then MSDK could setup the URLSessionWebSocketDelegate and let the app provide an optional delegate (some new MSDK type) that we would call into if the app cares about websocket lifecycle event.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/WebSocketClientTask.swift
```diff
@@ -0,0 +1,127 @@
...
+//  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import Foundation
+
+enum WebSocketAuthError: Error {
+    case tokenExpired
+    case unrecoverable(Error)
+}
+
+func isAuthError(_ error: Error) -> Bool {
```
wmathurin: Should we mirror the conditions under which we refresh that are [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m#L378) or better refactor code so there is only one implementation called from both places?

Crebs: Good question — since WebSocket doesn’t expose the HTTP handshake response, we can’t mirror the exact same refresh conditions as standard requests. We’re using similar token refresh logic, but with separate error detection tailored for WebSocket failures (e.g., checking for 403, Bad_OAuth_Token, etc. in the error message). Let me know if you’d prefer we extract the shared refresh logic into a common utility.

Crebs: | Feature                      | Standard URLSession Request             | URLSessionWebSocketTask                     |
|-----------------------------|-----------------------------------------|---------------------------------------------|
| Access to HTTP Status Code  | ✅ Yes (via `URLResponse`)              | ❌ No direct access after handshake         |
| Error Handling              | Uses `HTTPURLResponse.statusCode`       | Relies on `Error` No  URLResponse     |

Crebs: The main issue I'm seeing is a WebSocket doesn't have any access or knowledge of a URLResponse on an errror.

Crebs: So because of this difference - mirroring the conditions under which we refresh - while we are still looking for similar conditions, the way we do it in behavior and data structure is different.

Crebs: Actually I was able to update so the retry logic is shared between the two differen Session types using a Refresh Policy:

https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3876/files#diff-f9d38ab10347f5bec4f6b475dd308aa1fa61abb6674f484691640e7e633f648eR33

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient+WebSocket.swift
Crebs: I added both methods to support URLRequest and RestRequest.

wmathurin (approved):
LGTM
Was there a live web socket API you could try it against?

bbirman (approved):

PR #3878, opened 2025-06-12
@W-18766706 Fix bug of mixed global and local state, just staying with Global state

Crebs:
We need to move back to global state dependence on the original design.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
@@ -80,7 +80,6 @@ public class PushNotificationManager: NSObject {
     var isSimulator: Bool = false
     private let notificationRegister: RemoteNotificationRegistering
     private var preferences: SFPreferences?
-    
     private var loginObserver: NSObjectProtocol?
     private var enterForegroundObserver: NSObjectProtocol?
     
@@ -91,7 +90,7 @@ public class PushNotificationManager: NSObject {
     @objc
     public override convenience init() {
         self.init(notificationRegister: DefaultRemoteNotificationRegistrar(),
-                  preferences: SFPreferences.sharedPreferences(for: .user, user: UserAccountManager.shared.currentUserAccount))
+                  preferences: SFPreferences.currentUserLevel())
     }
     
     /// Internal initializer used for testing and dependency injection.
@@ -276,7 +275,7 @@ public class PushNotificationManager: NSObject {
                 
                 Task {
                     do {
-                        try await self.fetchAndStoreNotificationTypes(restClient: restClient)
+                        try await self.fetchAndStoreNotificationTypes(restClient: restClient, account: user)
                     } catch {
                         SFSDKCoreLogger.e(Self.self, message: "Get Notification Types Error: \(error.localizedDescription)")
                     }
@@ -360,10 +359,12 @@ public class PushNotificationManager: NSObject {
     ///
     /// - Parameters:
     ///   - restClient: The `RestClient` to use for the API call.
+    ///   - account: The user account to associate notification types with.
     /// - Throws: An error if the types cannot be retrieved from server or cache.
-    @objc(fetchAndStoreNotificationTypesWithRestClient:completionHandler:)
-    public func fetchAndStoreNotificationTypes(restClient: RestClient = RestClient.shared) async throws {
-        guard let account = UserAccountManager.shared.currentUserAccount else {
+    @objc(fetchAndStoreNotificationTypesWithRestClient:account:completionHandler:)
+    public func fetchAndStoreNotificationTypes(restClient: RestClient = RestClient.shared,
+                                               account: UserAccount? = UserAccountManager.shared.currentUserAccount) async throws {
+        guard let account = account else {
             throw PushNotificationManagerError.currentUserNotDetected
         }
         
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
@@ -79,45 +79,33 @@ public class PushNotificationManager: NSObject {
     
     var isSimulator: Bool = false
     private let notificationRegister: RemoteNotificationRegistering
-    private var apiVersion: String
     private var restClient: RestClient?
-    private var currentUser: UserAccount?
     private var preferences: SFPreferences?
     
     /// Convenience initializer that sets up the PushNotificationManager with default values:
     /// - notificationRegister: DefaultRemoteNotificationRegistrar() - Handles APNS registration
-    /// - apiVersion: RestClient.shared.apiVersion - Uses the current API version from RestClient
     /// - restClient: RestClient.shared - Uses the shared RestClient instance
-    /// - currentUser: UserAccountManager.shared.currentUserAccount - Uses the current logged-in user
     /// - preferences: SFPreferences.sharedPreferences(for: .user, user: currentUser) - Uses user-level preferences
     @objc
     public override convenience init() {
         self.init(notificationRegister: DefaultRemoteNotificationRegistrar(),
-                   apiVersion: RestClient.shared.apiVersion,
                    restClient: RestClient.shared,
-                   currentUser: UserAccountManager.shared.currentUserAccount,
                    preferences: SFPreferences.sharedPreferences(for: .user, user: UserAccountManager.shared.currentUserAccount))
     }
     
     /// Internal initializer used for testing and dependency injection.
     /// This initializer allows for customizing the dependencies of PushNotificationManager:
     /// - Parameter notificationRegister: The registrar that handles APNS registration. Defaults to DefaultRemoteNotificationRegistrar.
-    /// - Parameter apiVersion: The Salesforce API version to use. Defaults to RestClient.shared.apiVersion.
     /// - Parameter restClient: The REST client for making API calls. Defaults to RestClient.shared.
-    /// - Parameter currentUser: The user account to use. Defaults to UserAccountManager.shared.currentUserAccount.
     /// - Parameter preferences: The preferences store to use. Defaults to user-level SFPreferences.
     internal init(notificationRegister: RemoteNotificationRegistering = DefaultRemoteNotificationRegistrar(),
-                  apiVersion: String = RestClient.shared.apiVersion,
                   restClient : RestClient? = RestClient.shared,
-                  currentUser: UserAccount? = UserAccountManager.shared.currentUserAccount,
                   preferences: SFPreferences? = nil) {
         self.notificationRegister = notificationRegister
-        self.apiVersion = apiVersion
         self.restClient = restClient
-        self.currentUser = currentUser
         
         if preferences == nil {
-            self.preferences = SFPreferences.sharedPreferences(for: .user, user: currentUser)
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
```diff
@@ -473,4 +475,14 @@ private extension PushNotificationManager {
         SFSDKCoreLogger.i(Self.self, message: "App entering foreground, re-registering push")
         _ = registerSalesforceNotifications(completionBlock: nil, failBlock: nil)
     }
+    
+    func refreshDependencies() {
```
bbirman: Are these values being stored to make it more testable vs calling something like `RestClient.restClient(for: user)` directly in the registration method?

Crebs: yeah, it was because I was trying to do some dependency injection for unit tests.

bbirman: I'm not sure if we can always assume the current user if we expose a user parameter in `public func registerForSalesforceNotifications(user: UserAccount, completionBlock:@escaping (Result<Bool, PushNotificationManagerError>)->())`

Crebs: Ah, I think I miss read this last comment.  Yeah, I think I see what you are saying here.  Is it that we have a local state and we are still using the gloabl state instead of the global state in the method?

Crebs: I cleaned up the parameter interface and global state dependence a bit.  Since we kind of depend on the global state for determining our login flow for notification and things like that, I just moved back to using global state as it's the original pattern.  Let me know if you think this looks better? 

I updated the unit tests as well to mock the global state as well to be more accurate to our original design for Push Notification Manager.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -670,7 +670,7 @@ - (void)logoutUser:(SFUserAccount *)user reason:(SFLogoutReason)reason {
     
     // Before starting actual logout (which will tear down SFRestAPI), first unregister from push notifications if needed
     __weak typeof(self) weakSelf = self;
-    [[SFPushNotificationManager sharedInstance] unregisterSalesforceNotificationsWithCompletionBlock:user completionBlock:^void() {
```
Crebs, line 673: Objc and Swift overlook here.  Looks like the original method was this as well.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
```diff
@@ -317,64 +305,14 @@ public class PushNotificationManager: NSObject {
...
     
-    /// Unregisters the device from Salesforce push notifications for a specific user.
-    ///
-    /// - Parameters:
-    ///   - user: The user account to unregister.
-    ///   - completionBlock: A block executed when unregistration is complete.
-    /// - Returns: `true` if unregistration started successfully, otherwise `false`.
-    @discardableResult
-    @objc(unregisterSalesforceNotificationsWithCompletionBlock:completionBlock:)
-    public func unregisterSalesforceNotifications(for user: UserAccount,
```
bbirman: Making this private looks like it would be a breaking change from 13.0.x

https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/096b4e515c4f61487c03ed0279363eb57c09c092/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.h#L100

bbirman:
I'm running into the same error on login with the latest:  "Cannot register for notifications with Salesforce: no restClient"

Crebs:
@bbirman I ran it locally again and was able to see a notificaiton on after login.  Did a couple times as well.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
bbirman: I think there's a bug even in the old code where `registerSalesforceNotificationsWithCompletionBlock:(SFUserAccount *)user` would use the RestClient for the current user instead of the RestClient for the user in the parameter. It looks like the global state here would have the same issue if the current user and the user in the parameter were different. Is there a way to keep the testability improvements you made and fix that?

cc @wmathurin if you have a different take

If we kept the global state approach I think user switching would also need to be observed

wmathurin: @bbirman You are right, the code has the issue before this PR.
The issue existed also in the original Objective-C code.

See https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/v12.2.0/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFPushNotificationManager.m#L163

We were sending the request using the default rest client `[SFRestAPI sharedInstance]` (which is bound to the current user  `[SFRestAPI sharedInstanceWithUser:[SFUserAccountManager sharedInstance].currentUser]`).
Instead we should have used the rest client for the user being passed through.

Maybe we shouldn't have a `restClient` member in the class and instead get the right one in each method that needs one using the user passed in?

Crebs: OK, in this update I introduced a protocol-based approach for remote notification registration to keep testability — making it easier to inject mocks in unit tests.

What’s the Fix?

We now fetch the correct RestClient for the user being passed in (instead of relying on the global currentUserAccount). To do this, I updated the protocol to:

func clientForRemoteNotification(for user: UserAccount?) -> RestClient?

This fixes the bug from the original implementation and keeps the code testable and clean.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
```diff
@@ -224,6 +347,8 @@ public class PushNotificationManager: NSObject {
             return false
         }
         
+        let restClient = notificationRegister.clientForRemoteNotification(for: user)
```
Crebs: @wmathurin @bbirman I’ve updated the implementation to use a protocol for interfacing with the rest client. I removed the restClient property from the class and now inject a dependency that conforms to the protocol. This allows us to provide either a mock rest client in tests or fetch the appropriate client from the UserAccountManager.shared (as was done previously), but now in a cleaner and testable way.

bbirman: nice!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
bbirman: Same comment on the latest, I only see a private method for unregistering for a user?

Crebs: Ah, you’re right — I think I was looking at the Objective-C header and got tripped up by the naming. It does look like I goofed that… seeing completionBlock twice in there threw me off. 🫨

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
```diff
@@ -198,16 +190,147 @@ public class PushNotificationManager: NSObject {
...
+              !UserAccountManager.shared.isLogoutSettingEnabled,
+              deviceToken != nil else {
+            return
+        }
+        
+        SFSDKCoreLogger.i(Self.self, message: "App entering foreground, re-registering push")
+        _ = registerSalesforceNotifications(completionBlock: nil, failBlock: nil)
+    }
+    
+    private func registerSalesforceNotifications(for user: UserAccount,
```
bbirman: Like unregistering, registering notifications for a given user was a public API in 13.0

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
bbirman: Can this go back to passing the user parameter?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
Crebs: Yeah, Kept thinking those were typo errors by me.  with `completionBlock` in teh Objc header naming.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
Crebs: yes

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
```diff
@@ -79,45 +79,31 @@ public class PushNotificationManager: NSObject {
...
-    /// - currentUser: UserAccountManager.shared.currentUserAccount - Uses the current logged-in user
     /// - preferences: SFPreferences.sharedPreferences(for: .user, user: currentUser) - Uses user-level preferences
     @objc
     public override convenience init() {
         self.init(notificationRegister: DefaultRemoteNotificationRegistrar(),
-                   apiVersion: RestClient.shared.apiVersion,
-                   restClient: RestClient.shared,
-                   currentUser: UserAccountManager.shared.currentUserAccount,
-                   preferences: SFPreferences.sharedPreferences(for: .user, user: UserAccountManager.shared.currentUserAccount))
+                  preferences: SFPreferences.currentUserLevel())
```
bbirman: I think the user preferences could get out of sync on user logout / login / switch if it's set at init, is it possible to not store it at the class level like you did with the user account and rest client?

Crebs: Yes.  I'll do a similar pattern for preferences

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
```diff
@@ -224,6 +202,8 @@ public class PushNotificationManager: NSObject {
             return false
         }
         
+        let restClient = notificationRegister.client(for: user)
```
bbirman: sorry last one, could you also add a guard or something for the rest client and log an error if it's nil for some reason? I ran into a similar case recently and just want to make sure it can't fail silently. same for the similar block in the unregister method

Crebs: Yes.  I'll update that this morning.

bbirman (approved):

PR #3880, opened 2025-06-24
Pass level to logger

bbirman:


libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
```diff
@@ -146,7 +146,7 @@ - (void)log:(Class)cls level:(SFLogLevel)level format:(NSString *)format args:(v
     [self.logger log:cls level:level format:format args:args];
     
     NSString *formattedMessage = [[NSString alloc] initWithFormat:format arguments:args_copy];
-    [self.logReceiver receiveWithLevel:SFLogLevelInfo cls:cls component:self.logger.componentName message:formattedMessage];
+    [self.logReceiver receiveWithLevel:level cls:cls component:self.logger.componentName message:formattedMessage];
```
bbirman, line 149: @JohnsonEricAtSalesforce ran into a scenario today where debug logs were tagged as info and just wanted to double check this change is okay

wmathurin (approved):

libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Logger/SFLogger.m
JohnsonEricAtSalesforce: This looks like a bug you've fixed, for sure.  Thanks!

PR #3881, opened 2025-06-25
@W-18892644 Change Action to String to match Android implementation

Crebs:
Matching Android Action types to just be strings to have it match server side and match Android handling as well.


bbirman (approved):

Crebs:
The test seem to be flappy again.  I'll have to make another story maybe to look into this again.  When I run these tests locally they're passing.  So for now I'll merge this PR.  I can revert if we feel different otherwise.

PR #3883, opened 2025-06-30
@W-18777800: [qr code scan][ios] we should prevent a successful login of client A by scanning QR code of client B

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This is the counterpart to MSDK Android's https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2731

  QR Code Login now throws when the consumer key ("client id") of start URL query parameter of the front door bridge URL doesn't match the app's consumer key in boot config.

  This is a bit different than the Android implementation.  On Android, the app starts QR Code Log In by starting MSDK's `LoginActivity` without the ability to get a result or an error.  On iOS, there are direct method calls for this task.  So, we can throw from the MSDK methods and let the app catch.  The app then has complete control over the error UX and the QR Code Log In is cancelled.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController+QrCodeLogin.swift
```diff
@@ -51,9 +51,25 @@ public extension SalesforceLoginViewController {
     class func loginWithFrontdoorBridgeUrl(
         _ frontdoorBridgeUrlString: String,
         pkceCodeVerifier: String?
-    ) {
+    ) throws {
         guard let frontdoorBridgeUrl = URL(string: frontdoorBridgeUrlString) else { return }
         
+        // Only allow use of front door bridge URLs with matching consumer keys.
+        guard let frontdoorBridgeUrlComponents = NSURLComponents(
```
JohnsonEricAtSalesforce: This is the crux of the update where the consumer keys are compared.  It is a bit unwieldy to fetch the value from the front door bridge URL since it is actually in another URL, start URL, that itself a query string parameter of the front door bridge URL.  That said, it is a one off task so I kept the code simple as possible.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController+QrCodeLogin.swift
```diff
@@ -51,9 +51,25 @@ public extension SalesforceLoginViewController {
...
+            resolvingAgainstBaseURL: true),
+              let frontdoorBridgeUrlQueryItems = frontdoorBridgeUrlComponents.queryItems,
+              let startUrlString = frontdoorBridgeUrlQueryItems.first(where: { $0.name == "startURL"})?.value,
+              let startUrl = URL(string: startUrlString),
+              let startUrlComponents = NSURLComponents(url: startUrl, resolvingAgainstBaseURL: true),
+              let startUrlQueryItems = startUrlComponents.queryItems,
+              let frontdoorBridgeUrlClientId = startUrlQueryItems.first(where: { $0.name == "client_id"})?.value,
+              frontdoorBridgeUrlClientId == SalesforceManager.shared.bootConfig?.remoteAccessConsumerKey
+        else {
+            let message = "Cannot use another app's login QR Code.  Please log in to this app."
```
JohnsonEricAtSalesforce: This is the text Android's UX uses in the toast displayed by LoginActivity.  Here, we're just using it as the error message.  The app can provide text at their whim.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -401,6 +401,7 @@
 		CED452ED1D808DEE009266EB /* SFNativeRestRequestListener.h in Headers */ = {isa = PBXBuildFile; fileRef = CED452E91D808DEE009266EB /* SFNativeRestRequestListener.h */; };
 		CED452EF1D808E0A009266EB /* SFNativeRestRequestListener.m in Sources */ = {isa = PBXBuildFile; fileRef = CED452EA1D808DEE009266EB /* SFNativeRestRequestListener.m */; };
 		CED452F01D808E0F009266EB /* SalesforceRestAPITests.m in Sources */ = {isa = PBXBuildFile; fileRef = CED452E81D808DEE009266EB /* SalesforceRestAPITests.m */; };
+		D325B5DB2E135242007F8221 /* LoginError.swift in Sources */ = {isa = PBXBuildFile; fileRef = D325B5DA2E135242007F8221 /* LoginError.swift */; };
```
JohnsonEricAtSalesforce: At last I remembered I can't _only_ update the pod-generated project references in my template app.  Practice makes perfect 🤩

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginError.swift
```diff
@@ -0,0 +1,37 @@
...
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import Foundation
+
+/// An error for login.
+public struct LoginError: Error {
```
JohnsonEricAtSalesforce: I didn't see a Swift error for general login use.  Did I miss it?

wmathurin (approved):
LGTM

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginError.swift
Crebs: Would it be SPLoginError?

Crebs: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.h#L183

@bbirman  Thoughts?

bbirman: I think that's more specific to the[ IDP-SP flow](https://developer.salesforce.com/docs/platform/mobile-sdk/guide/auth-identity-providers.html)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
```diff
@@ -54,6 +54,9 @@ NS_ASSUME_NONNULL_BEGIN
 /// For Salesforce Identity UI Bridge API support, the optional web server flow code verififer accompaning the front door bridge URL.  This can only be used with `overrideWithfrontDoorBridgeUrl`.
 @property (nonatomic, strong, nullable) NSString *overrideWithCodeVerifier;
 
+/// For Salesforce Identity UI Bridge API support, indicates if overriding front door bridge URL has a consumer key value that matches the app config, which is also known as the boot config.
+@property (nonatomic, assign) BOOL overridingFrontDoorBridgeUrlMatchesConsumerKey;
```
JohnsonEricAtSalesforce: Changing approach here, this new property lets MSDK track if a front door URL was attempted that mismatched the consumer key.   That'll let the login view display and alert when needed a little later.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -595,8 +595,15 @@ - (BOOL)authenticateWithRequest:(SFSDKAuthRequest *)request
     authSession.authFailureCallback = failureBlock;
     authSession.authSuccessCallback = completionBlock;
     authSession.oauthCoordinator.delegate = self;
-    authSession.oauthCoordinator.overrideWithCodeVerifier = codeVerifier;
-    authSession.oauthCoordinator.overrideWithFrontDoorBridgeUrl = frontDoorBridgeUrl;
+
+    // Only allow use of front door bridge URLs with matching consumer keys.
+    if (frontDoorBridgeUrl != nil) {
+        authSession.oauthCoordinator.overridingFrontDoorBridgeUrlMatchesConsumerKey = [self validateBootConfigConsumerKeyMatches:frontDoorBridgeUrl];
```
JohnsonEricAtSalesforce: Here's a new home for the same consumer key validation logic we saw in the earlier commit.

JohnsonEricAtSalesforce: Note the front door URL will not be set if it is invalid and the "matches" property is `NO` so the error alert can be displayed later by the login view.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -906,20 +913,34 @@ - (void)oauthCoordinatorDidFetchAuthCode:(SFOAuthCoordinator *)coordinator authI
...
     loginViewController.oauthView = view;
     SFSDKAuthViewHolder *viewHolder = [SFSDKAuthViewHolder new];
     viewHolder.loginController = loginViewController;
     viewHolder.scene = coordinator.authSession.oauthRequest.scene;
+    
+    void (^authViewDisplayBlock)(void) = ^{
+        
+        self.authViewHandler.authViewDisplayBlock(viewHolder);
+        if (!coordinator.overridingFrontDoorBridgeUrlMatchesConsumerKey) {
+            UIAlertController* alertController = [UIAlertController
```
JohnsonEricAtSalesforce, line 925: With the addition of the alert code, I moved that and the invocation of the display block into this single inline function.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -906,20 +913,34 @@ - (void)oauthCoordinatorDidFetchAuthCode:(SFOAuthCoordinator *)coordinator authI
...
+                                                  alertControllerWithTitle:@"Error"
+                                                  message:@"Cannot use another app's login QR Code.  Please log in to this app."
```
wmathurin: We shouldn't hardcode the message. Instead it should come from [Localizable.strings](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/shared/resources/SalesforceSDKResources.bundle/en.lproj/Localizable.strings)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
wmathurin: Maybe we could have a little class to hold the three properties:
- overridingFrontDoorBridgeUrlMatchesConsumerKey
- overrideWithFrontDoorBridgeUrl 
- overrideWithCodeVerifier
And some code around manipulating them as well.

So we wouldn't have the "if this one is NO, then this one will be nil" or at least that would be hidden inside that class.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
```diff
@@ -33,6 +33,21 @@ typedef NS_ENUM(NSUInteger, SFOAuthTokenEndpointFlow) {
...
+/// For Salesforce Identity UI Bridge API support, an overriding front door bridge URL to use in place of the default initial URL.
+@property (nonatomic, strong, nullable) NSURL *frontdoorBridgeUrl;
+
+/// For Salesforce Identity UI Bridge API support, the optional web server flow code verififer accompaning the front door bridge URL.  This can only be used with `overrideWithfrontDoorBridgeUrl`.
+@property (nonatomic, strong, nullable) NSString *codeVerifier;
+
+/// For Salesforce Identity UI Bridge API support, indicates if overriding front door bridge URL has a consumer key value that matches the app config, which is also known as the boot config.
+@property (nonatomic, assign) BOOL matchesConsumerKey;
+
+- initWithFrontdoorBridgeUrl:(NSURL *)frontdoorBridgeUrl codeVerifier:(NSString *)codeVerifier;
```
bbirman: I haven't seen the empty return type before, what's the difference between that and `- (instancetype)`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
```diff
@@ -33,6 +33,21 @@ typedef NS_ENUM(NSUInteger, SFOAuthTokenEndpointFlow) {
 };
 NS_ASSUME_NONNULL_BEGIN
 
+@interface SFOAuthCoordinatorFrontdoorBridgeLoginOverride: NSObject
```
wmathurin: Should it have its own .h and .m?
What's the best practice @bbirman ?

bbirman: 👍 I think it's usually each class having their own with Objective-C

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
JohnsonEricAtSalesforce: That was an interesting omission when the code generated.  It actually _was_ returning the object, though I'd prefer the explicit return type for consistency's sake.  I added it.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
JohnsonEricAtSalesforce: Done.h and Done.m 😆

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
```diff
@@ -0,0 +1,77 @@
...
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import Foundation
+
+@objc public class AuthCoordinatorFrontdoorBridgeLoginOverride: NSObject {
```
JohnsonEricAtSalesforce: Note: Yes, this entire Obj-C to Swift conversion, refactoring of the existing Obj-C plus the Xcode project update was all done by Cursor.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
```diff
@@ -25,6 +25,7 @@
 #import "SFOAuthCoordinator.h"
 #import "SFSDKAuthSession.h"
 @class SFOAuthInfo;
+@class AuthCoordinatorFrontdoorBridgeLoginOverride;
```
JohnsonEricAtSalesforce: Cursor seems to like forward declarations.  I'm going to replace this with the actual imports.

wmathurin (approved):
LGTM

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
bbirman: Can you add SFSDK to the Objective-C name?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
bbirman: We've done that in other places to to avoid a circular dependency with the Swift header, but maybe not an issue here since it's an internal header

bbirman (approved):

PR #3885, opened 2025-07-03
Updating keychain operation scope

bbirman:
Background: 
- Some of our keychain operations aren't scoped to only apply to MSDK keys which can cause issues with keys managed outside of MSDK

Changes:
- Update KeychainHelper.setAccessibleAttribute query scope and apply updates to keys individually instead of batch update (first commit)
- Have SecItem operations (add, copy, delete, update) go through `SecItemOperations` class so we isolate usage and add MSDK scope (second commit)
    - Note: the EC & RSA key operations in `SFSDKCryptoUtils` and `Encryptor` don't have the creator tag but the queries are by ApplicationTag and the internal queries for encryption and push notifications have `com.salesforce` prefixes. The symmetric keys in `Encryptor` use `KeychainHelper` which now should all go through `SecItemOperations`.

Manual testing on device ~~in progress~~:
- [x] Base scenario with biometric key outside of MSDK
- [x] Push notification registration with biometric key outside of MSDK
- [x] Upgrade from 13.0 
- [x] Login with multiple users


Crebs (approved):

bbirman:
There's one test failure but I made a work item for it because it looks like it's been flapping

PR #3889, opened 2025-07-17
Revoke refresh token with a HTTP POST instead of a HTTP GET call

wmathurin:
Manually tested by:

- stepping through debugger on the client
- watching the session management page in the browser at the same time to see the oauth2 session disappear


bbirman (approved):

wmathurin:
Only one test failure in WebViewStateManagerTests which passes locally. It does not seem related to this change.

PR #3890, opened 2025-07-18
Use OauthCoordinator as source of truth for auth info & provide default value

bbirman:
Background:
There's a crash log that looks it's coming from [this line](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/f1be8efd1305cc2f43eff7a81bbe0008e9774d06/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m#L1993) because `authInfo` is nil
```
     NSDictionary *userInfo = @{kSFNotificationUserInfoAccountKey: userAccount,
                                kSFNotificationUserInfoAuthTypeKey: authInfo};
```
The change so far is to make authInfo non-null and use the existing `SFOAuthTypeUnknown` as the default value.  I noticed that there's an `authInfo` property on SFSDKAuthSession [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/f1be8efd1305cc2f43eff7a81bbe0008e9774d06/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKAuthSession.h#L46) and another `authInfo` property on the SFOAuthCoordinator [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/f1be8efd1305cc2f43eff7a81bbe0008e9774d06/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator%2BInternal.h#L46) and we update the one on the session based on the value from the coordinator. Since the session also has a reference to the coordinator, I think we could use the value on the coordinator as the source of truth and remove authInfo on the session.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
@@ -810,7 +810,6 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
 
 #pragma mark - SFOAuthCoordinatorDelegate
 - (void)oauthCoordinatorWillBeginAuthentication:(SFOAuthCoordinator *)coordinator authInfo:(SFOAuthInfo *)info {
-    coordinator.authSession.authInfo  = info;
     NSDictionary *userInfo = @{ kSFNotificationUserInfoCredentialsKey: coordinator.credentials,
                                 kSFNotificationUserInfoAuthTypeKey: coordinator.authInfo };
     [[NSNotificationCenter defaultCenter] postNotificationName:kSFNotificationUserWillLogIn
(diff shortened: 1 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -810,7 +810,6 @@ - (void)dismissAuthViewControllerIfPresentForScene:(UIScene *)scene completion:(
 
 #pragma mark - SFOAuthCoordinatorDelegate
 - (void)oauthCoordinatorWillBeginAuthentication:(SFOAuthCoordinator *)coordinator authInfo:(SFOAuthInfo *)info {
-    coordinator.authSession.authInfo  = info;
```
Crebs, line 813: So this is to just remove unneeded reference?  Basically looks like we just call the SFOAuthCoordinator directly to the auth info? Seems like a good opportunity to convert to Swift 👍.

bbirman: yup! convert what part?

Crebs: Oh, just in general. When we have time.  Was just continuing my Swfit Rant.  Nothing supper urgent at this point.

Crebs (approved):
LGTM.

PR #3892, opened 2025-07-22
@W-17964653 Discovery endpoint to postpone OAuth flow

Crebs:
- Refactored domain discovery to use a helper for building the discovery URL.
- Added a result object for parsing the callback URL.
- Improved async handling so the continuation always gets called.
- Tried letting DomainDiscoveryCoordinator own its own webview, but it caused issues with showing/dismissing the right webview and broke some flows.  
- To keep things simple, the webview is now just shared with SFOAuthCoordinator.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
@@ -67,7 +67,7 @@ static NSString * const SFSDKLoginHostNameKey = @"SalesforceLoginHostNameKey";
         SFManagedPreferences *managedPreferences = [SFManagedPreferences sharedPreferences];
         SFSDKLoginHost *production = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_PRODUCTION"] host:@"login.salesforce.com" deletable:NO];
         SFSDKLoginHost *sandbox = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_SANDBOX"] host:@"test.salesforce.com" deletable:NO];
-        SFSDKLoginHost *welcome = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_WELCOME"] host:@"welcome.salesforce.com" deletable:NO];
+        SFSDKLoginHost *welcome = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_WELCOME"] host:@"welcome.salesforce.com/discovery" deletable:NO];
 
         // Add the Production, Sandbox and Welcome login hosts, unless an MDM policy explicitly forbids this.
         if (!(managedPreferences.hasManagedPreferences && managedPreferences.onlyShowAuthorizedHosts)) {
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
@@ -16,6 +16,10 @@ enum DomainDiscovery: String {
         }
     }
 
+    /* TODO: Keep this list of client ids up to date with those
+     * supported by Salesforce Welcome Discovery or remove it
+     * when no longer required.
+     */
     static let supportedClientIds: Set<String> = [
         "SfdcMobileChatterAndroid",
         "SfdcMobileChatteriOS"
@@ -39,7 +43,7 @@ public class DomainDiscoveryResult: NSObject {
 /// Coordinator for My Domain discovery via WKWebView before OAuth login.
 ///
 /// This class loads a discovery URL in a WKWebView and waits for a callback URL to be hit, returning the discovered domain and login hint.
-@objcMembers
+@objc(SFDomainDiscoveryCoordinator)
 public class DomainDiscoveryCoordinator: NSObject {
     /// Starts the domain discovery process by loading the discovery URL in the provided WKWebView.
     ///
@@ -111,13 +115,14 @@ extension DomainDiscoveryCoordinator {
     }
     
     private func parseDiscoveryCallbackURL(_ url: URL?) -> DomainDiscoveryResult? {
-        guard let url = url as URL? else { return nil }
+        guard let url = url else { return nil }
         let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
-        let loginHint = components?.queryItems?.first(where: { $0.name == "login_hint" })?.value
-        var myDomain = components?.queryItems?.first(where: { $0.name == "my_domain" })?.value
-        if let domain = myDomain, domain.hasPrefix("https://") {
-            myDomain = String(domain.dropFirst("https://".count))
+        guard let loginHint = components?.queryItems?.first(where: { $0.name == "login_hint" })?.value,
+              let myDomainRaw = components?.queryItems?.first(where: { $0.name == "my_domain" })?.value else {
+            SFSDKCoreLogger.e(classForCoder, message: "Domain discovery callback URL is missing required parameter(s): login_hint and/or my_domain.")
+            return nil
         }
-        return DomainDiscoveryResult(loginHint: loginHint ?? "", myDomain: myDomain ?? "")
+        let myDomain = myDomainRaw.hasPrefix("https://") ? String(myDomainRaw.dropFirst("https://".count)) : myDomainRaw
+        return DomainDiscoveryResult(loginHint: loginHint, myDomain: myDomain)
     }
 }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
@@ -1,0 +1,100 @@
+import Foundation
+import WebKit
+
+/// Coordinator for My Domain discovery via WebView before OAuth login.
+public enum DomainDiscoveryConstants {
+    public static let callbackURL = "sfdc://discocallback"
+}
+
+@objcMembers
+public class DomainDiscoveryCoordinator: NSObject {
+    private var webView: WKWebView
+    
+    @objc(init)
+    public convenience override init() {
+        let webView = WKWebView()
+        self.init(webView: webView)
+    }
+
+    public init(webView: WKWebView = WKWebView()) {
+        self.webView = webView
+        super.init()
+        self.webView.navigationDelegate = self
+    }
+
+    private var continuation: CheckedContinuation<(String, String), Error>?
+
+    @MainActor
+    @objc(runDomainDiscoveryWithCredentials:completion:)
+    public func runMyDomainDiscovery(
+        credentials: OAuthCredentials
+    ) async throws -> (String, String) {
+        
+        guard webView != nil else {
+            throw NSError(domain: "DomainDiscoveryCoordinator", code: 1000, userInfo: [NSLocalizedDescriptionKey: "Missing webView"])
+        }
+        // 1. Build the discovery URL
+        let discoveryHost = credentials.domain ?? "welcome.salesforce.com"
+        guard let clientId = credentials.clientId,
+              let clientVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String else {
+            throw NSError(domain: "DomainDiscoveryCoordinator", code: 1001, userInfo: [NSLocalizedDescriptionKey: "Missing required credentials"])
+        }
+
+        var components = URLComponents()
+        components.scheme = "https"
+        components.host = discoveryHost
+        components.path = ""
+        components.queryItems = [
+            URLQueryItem(name: "client_id", value: clientId),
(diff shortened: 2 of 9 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -0,0 +1,106 @@
...
+        components.path = "/discovery"
+        components.queryItems = [
+            URLQueryItem(name: "client_id", value: clientId),
+            URLQueryItem(name: "client_version", value: clientVersion),
+            URLQueryItem(name: "callback_url", value: callbackURL)
+        ]
+        return components.url as NSURL?
+    }
+    
+    private func isDomainDiscoveryCallbackURL(_ url: URL?) -> Bool {
```
wmathurin, line 115: That will not work with test environment.
@JohnsonEricAtSalesforce took a different approach and is checking for /discovery in the path of the login server. See https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2736/files#diff-25c822f0f7f00bf2d6a9fec0477b811f0264e0c9a22312378ba605ccfa194c17R1427

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -0,0 +1,106 @@
...
+    ///   - credentials: The OAuth credentials containing clientId, clientVersion, and domain.
+    ///
+    /// This method loads the discovery URL in the given webview. The result of the discovery should be handled by monitoring navigation actions and calling `handle(webAction:)`.
+    @MainActor
+    @objc public func runMyDomainsDiscovery(on webview: WKWebView,
+                                      with credentials: OAuthCredentials) {
+        guard let clientId = credentials.clientId,
+              let clientVersion = Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String,
+              let domain = credentials.domain else {
+            SFSDKCoreLogger().e(classForCoder, message: "Missing required credentials")
```
bbirman: for this and the one below
```suggestion
            SFSDKCoreLogger.e(classForCoder, message: "Missing required credentials")
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
Crebs: @wmathurin do you mean the "isDisoveryDomain" method, not the "isDomainDiscoveryCallbackURL" right?  I'm pretty sure that's what you mean there.  But want to make sure I'm not missing something.

Crebs:
@wmathurin ,
Quick follow-up on unsupported client apps using disco:

When I try this, the webview displays an error banner (“Something went wrong. Try again.”) inside itself (handled by the auth team). Since the error is displayed and swallowed by the web content, we don’t get an error back from WKWebView.

So the user sees this:

IMG_0025

Then then ahve to pick a different domain manually. Since we don't gent anything back from WKWebivew that it failed, I'm not sure if you had some other ideas.  Or is this good enought?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/SFSDKLoginHostStorage.m
```diff
@@ -67,7 +67,7 @@ - (id)init {
         SFManagedPreferences *managedPreferences = [SFManagedPreferences sharedPreferences];
         SFSDKLoginHost *production = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_PRODUCTION"] host:@"login.salesforce.com" deletable:NO];
         SFSDKLoginHost *sandbox = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_SANDBOX"] host:@"test.salesforce.com" deletable:NO];
-        SFSDKLoginHost *welcome = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_WELCOME"] host:@"welcome.salesforce.com" deletable:NO];
+        SFSDKLoginHost *welcome = [SFSDKLoginHost hostWithName:[SFSDKResourceUtils localizedString:@"LOGIN_SERVER_WELCOME"] host:@"welcome.salesforce.com/discovery" deletable:NO];
```
Crebs, line 70: @wmathurin This what you hand in mind for the disco comain?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -0,0 +1,116 @@
...
+            let result = parseDiscoveryCallbackURL(url)
+            return result
+        }
+        return nil
+    }
+    
+    @objc
+    public func isDisoveryDomain(_ domain: String?) -> Bool {
+        guard let domain = domain else { return false }
+        return domain.lowercased().contains(DomainDiscovery.URLComponent.path.rawValue)
```
Crebs: @wmathurin Now just lookd for the discovery path.

wmathurin: LGTM.
You should also check that the client id is one of the supported ones.
Here is the check on Android: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2736/files#diff-25c822f0f7f00bf2d6a9fec0477b811f0264e0c9a22312378ba605ccfa194c17R1438
Here is the list in the server disco code: https://git.soma.salesforce.com/globalIdentity/LoginService/blob/domain-disco-2/ui/lwr/src/modules/login/constants/myDomainDiscoveryConstants.ts#L13

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -15,6 +15,11 @@ enum DomainDiscovery: String {
             case callbackURL = "callback_url"
         }
     }
+
+    static let supportedClientIds: Set<String> = [
```
wmathurin, line 23: Maybe a TODO like here: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2736/files#diff-25c822f0f7f00bf2d6a9fec0477b811f0264e0c9a22312378ba605ccfa194c17R1438

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -0,0 +1,123 @@
...
+    init(loginHint: String, myDomain: String) {
+        self.loginHint = loginHint
+        self.myDomain = myDomain
+    }
+}
+
+/// Coordinator for My Domain discovery via WKWebView before OAuth login.
+///
+/// This class loads a discovery URL in a WKWebView and waits for a callback URL to be hit, returning the discovered domain and login hint.
+@objcMembers
```
bbirman: Can you add a prefix for the objective-c name?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -0,0 +1,123 @@
...
+    
+    private func parseDiscoveryCallbackURL(_ url: URL?) -> DomainDiscoveryResult? {
+        guard let url = url as URL? else { return nil }
+        let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        let loginHint = components?.queryItems?.first(where: { $0.name == "login_hint" })?.value
+        var myDomain = components?.queryItems?.first(where: { $0.name == "my_domain" })?.value
+        if let domain = myDomain, domain.hasPrefix("https://") {
+            myDomain = String(domain.dropFirst("https://".count))
+        }
+        return DomainDiscoveryResult(loginHint: loginHint ?? "", myDomain: myDomain ?? "")
```
bbirman: if `myDomain` is nil, what happens with the empty string in the result? should the result be nil?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
Crebs: yes.  Thanks for keeping honest here. 👍

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
Crebs: Yeah, I think it coud just retun nil in that case instead of an empty value.  I'll update that.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
Crebs: yeah, I'll that as well.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -0,0 +1,131 @@
...
+        }
+        if isDomainDiscoveryCallbackURL(url) {
+            let result = parseDiscoveryCallbackURL(url)
+            return result
+        }
+        return nil
+    }
+    
+    @objc
+    public func isDisoveryDomain(_ domain: String?, clientId: String?) -> Bool {
```
wmathurin: Typo isDisovery should be isDiscovery

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -811,8 +823,26 @@ - (NSURLSession*)session {
...
+    self.loginHint = loginHint;
+    self.credentials.domain = myDomain;
+    [[SFUserAccountManager sharedInstance] setLoginHost:myDomain];
+    [self authenticate];
+}
+
 #pragma mark - WKNavigationDelegate (User-Agent Token Flow)
 - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler {
+    
+    SFDomainDiscoveryResult *result = [self.domainDiscoveryCoordinator handleWithWebAction:navigationAction];
```
wmathurin: Maybe a little comment.
Should it be part of the `if() else if()` etc that follows?
```Objective-C
NSURL *url = navigationAction.request.URL;
NSString *requestUrl = [url absoluteString];
SFDomainDiscoveryResult *discoveryResult = [self.domainDiscoveryCoordinator handleWithWebAction:navigationAction];
if (discoveryResult) {
...
} else if (self isRedirectURL:requestUrl]) {
...
} else if ...
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -0,0 +1,131 @@
...
+/// Coordinator for My Domain discovery via WKWebView before OAuth login.
+///
+/// This class loads a discovery URL in a WKWebView and waits for a callback URL to be hit, returning the discovered domain and login hint.
+@objc(SFDomainDiscoveryCoordinator)
+public class DomainDiscoveryCoordinator: NSObject {
+    /// Starts the domain discovery process by loading the discovery URL in the provided WKWebView.
+    ///
+    /// - Parameters:
+    ///   - webview: The WKWebView in which to load the discovery URL for domain discovery.
+    ///   - credentials: The OAuth credentials containing clientId, clientVersion, and domain.
```
bbirman: nit: the credentials don't contain the clientVersion

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -835,13 +864,9 @@ - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigati
             [bioAuthManager presentBiometricWithScene:self.view.window.windowScene];
         }
     } else if ([self shouldUpdateDomain:url]) {
-        // To support case where my domain is entered through "Use Custom Domain"
```
bbirman, line 838: If the scenario hasn't changed, can the comment stay?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
Crebs: 👍 Thank you.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
Crebs: Yeah, I can add this comment back.

PR #3895, opened 2025-07-28
@W-18992574 fixing flapping test on websockets callback

Crebs:


libs/SalesforceSDKCore/SalesforceSDKCoreTests/WebSocketClientTests.swift
```diff
@@ -136,18 +136,22 @@ final class WebSocketClientTests: XCTestCase {
...
         let expectation = self.expectation(description: "Message received")
-        
+        var fulfilled = false
         // When
         client?.listen { result in
             switch result {
             case .success(_):
                 // Then
-                expectation.fulfill()
+                if !fulfilled {
```
bbirman: I think we should investigate why it's being fulfilled more than once.

It looks like listen has a while loop to keep calling receive and the mock socket will always have a value, should it be cancelled after the first success?

Crebs: yeah, I'll look more into it.  Good call.

Crebs: yeah, it's just my crappy mock task.  It'll just keep acting like it's getting messages until I call self.mockTask?.keepReceivingMessages.toggle().

bbirman (approved):

PR #3897, opened 2025-08-06
@W-19112860 fix: implement streamError for EncryptStream and DecryptStream

Crebs:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/EncryptStream.swift
```diff
@@ -131,6 +141,7 @@ public class EncryptStream: OutputStream {
 
     override public func close() {
         defer { stream.close() }
+        _streamError = nil
```
bbirman, line 147: Why is it being set to nil vs setting it to the error in the catch block below?

Crebs: From my understanding: 

The close() method is meant to close the stream and reset its state. Setting _streamError = nil here clears any previous error, so that if the stream is reopened or reused, it starts with a clean error state.

If I'm incorrect in my understanding, how the close works, I can rethink this again.

Crebs: Ah, I see what your saying.  I've got some unit tests now to make sure we cover these cases now too.  So I'll now set the error code to nil, but will reset it if the catch block below gets an error.

bbirman (approved):

PR #3898, opened 2025-08-16
Moving to SQLCipher 4.10.0

wmathurin:
Important, only merge after:
- https://github.com/forcedotcom/fmdb/pull/15

Notable changes:
- The SmartStore project now gets FMDB and SQLCipher via SPM
- We no longer have a submodule for FMDB
- We no longer (build and) check in SQLCipher.xcframework in this repo

I also modified update_podspecs_headers.sh:
- in the past, when there was a exclude_files in the pod spec, then the public headers were sorted
- now, we always sort the public headers - as a result in this PR, all the pod specs have changed

NB: we are getting FMDB Swift package from our (updated) fork - for more details see: https://github.com/forcedotcom/fmdb/pull/15


libs/SmartStore/SmartStore.xcodeproj/project.pbxproj
```diff
@@ -1385,6 +1348,30 @@
 			defaultConfigurationName = Release;
 		};
 /* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
+		4FC72E662E500C7100A8AA8D /* XCRemoteSwiftPackageReference "fmdb" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/wmathurin/fmdb";
```
wmathurin: NB: will switch that to forcedotcom/fmdb once https://github.com/forcedotcom/fmdb/pull/15 gets merged

.github/workflows/pr.yaml
```diff
@@ -6,7 +6,7 @@ on:
 
 jobs:
   static-analysis:
-    runs-on: macos-15
+    runs-on: macos-16
```
wmathurin, line 9: I thought that might fixed the static analysis but it did not help

wmathurin:
Tests are passing but static analysis is failing (already in previous PRs).

PR #3901, opened 2025-08-19
@W-19345684: [iOS][QRCode] Shows Unsupported URL on FreshLogin using QRCode

JohnsonEricAtSalesforce:
🥁 _*Ready For Review*_ 🎸

  This simple fix makes the iOS side of QR Code Log In override the user-agent or web server flows just as Android now does.  The error in the video was caused by the web server flow oriented QR code getting handled like a user agent one.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -851,13 +851,12 @@
                                            myDomain:discoveryResult.myDomain];
         decisionHandler(WKNavigationActionPolicyCancel);
     } else if ([self isRedirectURL:requestUrl]) {
-        // First, choose user agent or web server authentication from Salesforce SDK's properties.
-        BOOL isUsingWebServerAuthentication = [[SalesforceSDKManager sharedManager] useWebServerAuthentication];
-        // Second, switch authentication based on the in-use Salesforce Identity API UI Bridge front-door URL and code verifier.
-        if (self.frontdoorBridgeLoginOverride.frontdoorBridgeUrl) {
-            isUsingWebServerAuthentication = self.frontdoorBridgeLoginOverride.codeVerifier;
-        }
-        if (isUsingWebServerAuthentication) {
+        // If a front door bridge URL override is present, use its code verifier to choose between user agent or web server authentication.
+        if (self.frontdoorBridgeLoginOverride.frontdoorBridgeUrl // Check if an override is provided
+            ? self.frontdoorBridgeLoginOverride.codeVerifier != nil // If yes, only proceed if it's a web server flow as indicated by a code verifier.
+            : [[SalesforceSDKManager sharedManager] useWebServerAuthentication] // If there's no override use the default SDK setting.
+            )
+        {
             [self handleWebServerResponse:url]; // Web server flow/URLs with query string parameters.
         } else {
             [self handleUserAgentResponse:url]; // User agent flow/URLs with the fragment component.
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -851,9 +851,13 @@
                                            myDomain:discoveryResult.myDomain];
         decisionHandler(WKNavigationActionPolicyCancel);
     } else if ([self isRedirectURL:requestUrl]) {
-        // Determine if presence of override parameters require the user agent flow.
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -851,9 +851,13 @@ - (void)webView:(WKWebView *)webView decidePolicyForNavigationAction:(WKNavigati
...
         decisionHandler(WKNavigationActionPolicyCancel);
     } else if ([self isRedirectURL:requestUrl]) {
-        // Determine if presence of override parameters require the user agent flow.
-        BOOL overrideWithUserAgentFlow = self.frontdoorBridgeLoginOverride.frontdoorBridgeUrl && !self.frontdoorBridgeLoginOverride.codeVerifier;
-        if ( [[SalesforceSDKManager sharedManager] useWebServerAuthentication] && !overrideWithUserAgentFlow) {
+        // First, choose user agent or web server authentication from Salesforce SDK's properties.
+        BOOL isUsingWebServerAuthentication = [[SalesforceSDKManager sharedManager] useWebServerAuthentication];
+        // Second, switch authentication based on the in-use Salesforce Identity API UI Bridge front-door URL and code verifier.
+        if (self.frontdoorBridgeLoginOverride.frontdoorBridgeUrl) {
+            isUsingWebServerAuthentication = self.frontdoorBridgeLoginOverride.codeVerifier;
```
brandonpage: `codeVerifier` isn't a BOOL?  The previous `!self.frontdoorBridgeLoginOverride.codeVerifier` call just checked if it was nil.  

So if `self.frontdoorBridgeLoginOverride.codeVerifier` assigned to a BOOL just checks if it's nil what is the change in logic here?  It looks the same to me.

JohnsonEricAtSalesforce: Since this is Obj-C, any pointer value (for a string in this case) can scale as a BOOL. nil = NO and any value = YES.

The key difference is that the non-nil (YES) value for the object in frontdoorBridgeLoginOverride.frontdoorBridgeUrl is checked _first_.  That prevents the case I was seeing in the app at runtime where setting `useWebServerAuthentication` = `NO` would accidentally force user-agent handling for web-server-flow QR log in URLs.  We vetted really similar logic in the Android app and this also works live in the iOS app as well, so we have those results to guide the runtime review.

wmathurin: Would it be more readable with

```objective-c
// If frontdoorBridgeUrl is present, use codeVerifier to decide; otherwise, fall back to useWebServerAuthentication
if (self.frontdoorBridgeLoginOverride.frontdoorBridgeUrl  // Check if an override is provided
        ? self.frontdoorBridgeLoginOverride.codeVerifier   // If yes, only proceed if it's a web server flow (has codeVerifier)
        : [[SalesforceSDKManager sharedManager] useWebServerAuthentication]  // If no override, use default SDK setting
    ) 
{
    [self handleWebServerResponse:url]; // Web server flow/URLs with query string parameters.
} else {
    [self handleUserAgentResponse:url]; // User agent flow/URLs with the fragment component.
}
```

JohnsonEricAtSalesforce: That wouldn't compile, but this slight change will.  Note the ternary operator needs both results to be of the same type, so it can't scale the Int as a BOOL like the previous version did or the compiler produces:

`Incompatible operand types ('NSString * _Nullable' and 'int')`

```
if (self.frontdoorBridgeLoginOverride.frontdoorBridgeUrl  // Check if an override is provided
            ? self.frontdoorBridgeLoginOverride.codeVerifier != nil   // If yes, only proceed if it's a web server flow (has codeVerifier)
            : [[SalesforceSDKManager sharedManager] useWebServerAuthentication]  // If no override, use default SDK setting
            )
        {
            [self handleWebServerResponse:url]; // Web server flow/URLs with query string parameters.
        } else {
            [self handleUserAgentResponse:url]; // User agent flow/URLs with the fragment component.
        }
```

JohnsonEricAtSalesforce: I do like the syntax candy, but I found this version just a little harder to process just from the code since it's more terse.  It works though, so I tried massaging the comments a bit to see if that helped get the best of it all.  Take a look at this 👉🏻  https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3901/commits/81416398e8c8954e64cf93799acc4d5b34bbb4b6

brandonpage (approved):

bbirman (approved):

PR #3904, opened 2025-08-22
[Don't Merge] Swiftify AuthInfo

Crebs:
Converted AuthInfo from Objc to Swift. Thought?


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthInfo.swift
```diff
@@ -0,0 +1,104 @@
...
+// DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+// WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+// WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+
+import Foundation
+
+@objc(SFOAuthType)
+public enum AuthType: Int {
```
bbirman, line 31: This looks like a breaking change because a case in Swift would currently be `AuthInfo.AuthType.IDP` vs `AuthType.IDP`, but maybe we could do it in 14.0 as in-place change instead of having a deprecation path since it's more of a straight forward rename?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthInfo.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    // Default initializer for Swift and Objective-C
+    @objc public override init() {
+        self.authType = .unknown
+        super.init()
+    }
+
+    // Objective-C class factory method
+    @objc(infoWithAuthType:)
+    public class func info(withAuthType authType: AuthType) -> AuthInfo {
```
bbirman, line 90: What's the use case for this vs init at line 77? I got confused by the comments because I thought one was for Objective-C and one was for Swift but they look like they're both available for both languages

bbirman (commented):
nice candidate for modernizing

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthInfo.swift
Crebs: your right.  I think I tried to put this up before the weekend and missed a couple things.  I'll clean this up again.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthInfo.swift
Crebs: Yeah, that’s why we have the static constants below, from line 55 to 63. In Objective-C, they can still call SFOAuthTypeIDP.   But yeah, not a 13.x change for sure.

PR #3905, opened 2025-08-26
@W-19078443 Main Thread runtime warnings in MSDK

Crebs:
Addresses potential data races and thread safety issues in the Salesforce Mobile SDK Core and Analytics modules. The changes ensure that shared mutable state is accessed in a thread-safe manner and that account management operations are properly synchronized.


libs/SalesforceAnalytics/SalesforceAnalytics/Classes/Store/SFSDKEventStoreManager.m
```diff
@@ -69,10 +69,12 @@ - (instancetype) initWithStoreDirectory:(NSString *) storeDirectory dataEncrypto
 
         // Gets current number of events stored.
         self.eventCountMutex = [[NSObject alloc] init];
-        self.numStoredEvents = 0;
-        NSArray *files = [[NSFileManager defaultManager] contentsOfDirectoryAtPath:self.storeDirectory error:nil];
-        if (files) {
-            self.numStoredEvents = files.count;
+        @synchronized (self.eventCountMutex) {
```
wmathurin, line 72: You just created the object you synchronize on at line 71, so how will it keep other threads out?
Should you do @synchronized ([self class])  instead?

wmathurin: Or use a static member:
```
static NSObject *initMutex;
dispatch_once(&onceToken, ^{
    initMutex = [[NSObject alloc] init];
});

@synchronized (initMutex) {
    // initialization code
}
```

Crebs: ah yeah.  Good point.  Let me look at what the SApp did again.

PR #3908, opened 2025-09-02
@W-19461882: [M1][MSDK13.1] Enable Flexible Server Matching for QR Code Login (iOS)

JohnsonEricAtSalesforce:
🥁 _*Ready For Review.*_ 🎸

  This adds logic to front door bridge URL log in that can detect if the app has a matching login host and select it or automatically add a log in host for the front door bridge URL.

  This is being tested against several sets of orgs and users provided by Salesforce app as part of their QR Code Log In effort.

  Unit tests are included courtesy of Cursor.


wmathurin (commented):
It looks fine but maybe we could refactor it to make easy to test?
Maybe appLoginHostForFrontdoorBridgeUrl could be a method on SFSDKLoginHostStorage itself.
Then we could have tests involving SFSDKLoginHostStorage for a variety of scenarios.
You could go a step further and have the soft matching logic be in a method of its own taking the candidateHostUrl and the eligbleLoginHostUrls and have it return the one matching.  Most of the tests could focus on that method alone (which doesn't do persistence or anything like that).

JohnsonEricAtSalesforce:
> It looks fine but maybe we could refactor it to make easy to test? Maybe appLoginHostForFrontdoorBridgeUrl could be a method on SFSDKLoginHostStorage itself. Then we could have tests involving SFSDKLoginHostStorage for a variety of scenarios. You could go a step further and have the soft matching logic be in a method of its own taking the candidateHostUrl and the eligbleLoginHostUrls and have it return the one matching. Most of the tests could focus on that method alone (which doesn't do persistence or anything like that).

I was wondering about the scope of each class as I was breaking out all the new methods.  Does it seem like `SFSDKLoginHostStorage` is for storage of the app's login hosts as they're shown in the server picker, but would become cluttered if it also had to process the additional scope of front door URLs? Not every front door URL will translate to a new stored app login host because of the matching routine.  If we keep the scope of front-door URL logic in `AuthCoordinatorFrontdoorBridgeLoginOverride` we can keep `SFSDKLoginHostStorage` just for things that are actually stored persistently (since it's named "Storage" specifically).

Maybe a best of both options would be to create an object that helps `AuthCoordinatorFrontdoorBridgeLoginOverride` interact with `SFSDKLoginHostStorage` (And `UserAccountManager`).  Maybe a name like `FrontdoorBridgeUrlLoginHost`? For test and production use, we give it a front door URL and mocks for both `SFSDKLoginHostStorage` and `UserAccountManager` before comparing the immutable result of its initialization.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
```diff
@@ -73,12 +73,79 @@ public class AuthCoordinatorFrontdoorBridgeLoginOverride: NSObject {
...
         
         // Only set the properties if the front door URL host and the start URL consumer key match the app's current values.
         if self.matchesLoginHost && self.matchesConsumerKey {
             self.codeVerifier = codeVerifier
             self.frontdoorBridgeUrl = frontdoorBridgeUrl
         }
     }
+    
+    private var addingAndSwitchingLoginHostsAllowed: Bool {
+        !SFManagedPreferences.shared().onlyShowAuthorizedHosts && SFManagedPreferences.shared().loginHosts.count == 0
```
JohnsonEricAtSalesforce, line 109: @bbirman + @wmathurin, Does this expression best satisfy the semantics of the way this method is named? I saw this statement in some of the UI code, but wasn't sure it was definitively the best logic.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
```diff
@@ -73,12 +73,79 @@ public class AuthCoordinatorFrontdoorBridgeLoginOverride: NSObject {
...
+    private var loginHostStorage: SFSDKLoginHostStorage {
+        SFSDKLoginHostStorage.sharedInstance()
+    }
+    
+    private func addAppLoginHostForFrontdoorBridgeUrl(_ frontdoorBridgeUrl: URL) -> String? {
+        guard let frontdoorBridgeUrlHost = frontdoorBridgeUrl.host() else {
+            return nil
+        }
+        
+        loginHostStorage.add(SalesforceLoginHost(
```
JohnsonEricAtSalesforce: @bbirman + @wmathurin, just triple checking that this is the correct way to add a login host just like the user would do in the picker.  How's it look?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
```diff
@@ -73,12 +73,79 @@ public class AuthCoordinatorFrontdoorBridgeLoginOverride: NSObject {
...
+        else {
+            eligibleAppLoginHosts = [UserAccountManager.shared.loginHost]
+        }
+        
+        if (frontdoorBridgeUrlIsMyDomain) {
+            guard let startIndex = frontdoorBridgeUrlHost.range(of: ".my.")?.upperBound else { return nil }
+            let frontdoorBridgeUrlMyDomainSuffix = frontdoorBridgeUrlHost.suffix(from: startIndex)
+            
+            for eligibleAppLoginHost in eligibleAppLoginHosts {
+                if (eligibleAppLoginHost.hasSuffix(frontdoorBridgeUrlMyDomainSuffix)) {
```
JohnsonEricAtSalesforce: @wmathurin - Just matching the domain after `.my.` isn't working for my test user.  
The app ships with login host `https://login.test1.pc-rnd.salesforce.com`

The QR Code has `https://konsf.test1.my.pc-rnd.salesforce.com`

The logic we have here will soft-match the login host, but this user gets `To access this page, you have to log in to Salesforce.` when trying to use the login host.  They have to log in directly via `konsf`.

Changing this line to be strict, `if (eligibleAppLoginHost == frontdoorBridgeUrlHost) {`, works.  Thoughts?  What is the line between users that can use a login host and ones that cannot?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
```diff
@@ -73,12 +73,79 @@ public class AuthCoordinatorFrontdoorBridgeLoginOverride: NSObject {
...
         
         // Check if the front door URL host matches the app's selected login host
-        self.matchesLoginHost = frontdoorBridgeUrl.host() == UserAccountManager.shared.loginHost
+        var appLoginHost = appLoginHostForFrontdoorBridgeUrl(frontdoorBridgeUrl)
+        if (appLoginHost == nil && addingAndSwitchingLoginHostsAllowed) {
+            appLoginHost = addAppLoginHostForFrontdoorBridgeUrl(frontdoorBridgeUrl)
+        }
+        if let appLoginHost = appLoginHost {
+            self.matchesLoginHost = true
+            UserAccountManager.shared.loginHost = appLoginHost
```
JohnsonEricAtSalesforce, line 98: Interestingly, setting the loginHost here isn't reflecting that it was already added with a call to `loginHostStorage.add`.  That results in two entries.  So, I believe we can remove the call to `loginHostStorage.add`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
wmathurin: We should use the soft matching to allow QR login to proceed, but you keep using whatever came with the QR code.

JohnsonEricAtSalesforce: The behavior is actually that soft matching _blocks_ the test user I was trying.  They can log in using the specific My Domain host konsf but not the soft matched `login.test1` host.  Are you aware of any logic that separates users who _can_ log in using a login host and others who _cannot_ and must use their specific My Domain host as the login server?

Could my test user be misconfigured?

wmathurin: Let's say you just let the QR through without any validation, does it work?

JohnsonEricAtSalesforce: It does work without any validation, yes.  It's only when we "soft match" to the app's existing login server that we have issues.

When I was testing, I disabled soft-matching entirely and either strict matched to an existing app login host with the exact same host and the QR Code or I added the QR Code's exact login host and used it.

wmathurin: I don’t understand. If the soft match says “ok to proceed”, it should be the same as not doing any validation. No?

JohnsonEricAtSalesforce:
@wmathurin - With the latest logic we have in plus one more commit in https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3908/commits/a06d74794cf623855e6cf84491cd2dfa911c0cdd, I can successfully log in via Salesforce app with both my current test users and orgs where one has browser-based authentication on the other does not.  The fix for [W-19526686]([iOS][QRCode] When Auth Config "Use the native browser for user authentication on iOS" is enabled, users cannot login via QR code.) is actually really simple and in logic that is only active during front-door bridge URL authentication.  Since I cannot test the soft-matching for my second user without it and it makes for a complete solution, I'd recommend we include it in this pull request.  Thoughts? We could revert and defer https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3908/commits/a06d74794cf623855e6cf84491cd2dfa911c0cdd but that might actually complicate the testing effort.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AuthCoordinatorFrontdoorBridgeLoginOverride.swift
wmathurin: I think so.

wmathurin (approved):

PR #3909, opened 2025-09-04
Building/testing with iOS 26 also

wmathurin:


wmathurin:
I'm going to merge (can't see my changes to github actions until then).

PR #3914, opened 2025-09-06
Fix for failing testRestCalls

wmathurin:
I debugged and verified the same cache was being used across NSURLSession but there seems to be NSURLSession session-level cache partitioning that causes the second round of calls to miss. I decided to rewrite testRestCalls and use a test subclass of SFSDKEncryptedURLCache. testRestCalls sends actual requests to the server twice. After the first set of calls, we make sure the cache was written to. After the second set of calls, we make sure the cache was read from.


wmathurin:
All the SalesforceSDKCore tests are passing again :-)

wmathurin:
There are a bunch of new failures on the last run due to - I'll try again in a while:
> We can't save this record because the Account Flow process failed. Give your Salesforce admin these details. You've reached the limit of 10,000 notification actions per hour. Notification actions will start again in the next hour.

I've deactivated the flows in the org. I guess we should not setup flows that fire notifications on Account save in the same org we are running tests that create and update many Accounts!

brandonpage (approved):

PR #3916, opened 2025-09-12
iOS 26 CI

brandonpage:
- Use MacOS 26 for everything.  
- PRs now use Xcode 26 + iOS 26 and Xcode 16 + iOS 18.
- Fix version matrix so iOS 26, iOS 18 and iOS 17 run on "nightly" builds.

`xcresulttool` currently throws an error processing iOS 26 test results so it is skipped.  The report will not be in included in the summary, but the job itself will still pass/fail based on if all tests pass.

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -16,19 +16,15 @@ jobs:
         include:
           - ios: ^26
             xcode: ^26
-            macos: macos-26
           - ios: ^18
             xcode: ^16
-            macos: macos-26
           - ios: ^17
-            xcode: ^15
-            macos: macos-14
+            xcode: ^16
     uses: ./.github/workflows/reusable-workflow.yaml
     with:
       lib: ${{ matrix.lib }}
       ios: ${{ matrix.ios }}
       xcode: ${{ matrix.xcode }}
-      macos: ${{ matrix.macos }}
     secrets: inherit
 
   native-samples-nightly:
@@ -40,18 +36,14 @@ jobs:
         include:
           - ios: ^26
             xcode: ^26
-            macos: macos-26
           - ios: ^18
             xcode: ^16
-            macos: macos-26
           - ios: ^17
-            xcode: ^15
-            macos: macos-14
+            xcode: ^16
     uses: ./.github/workflows/reusable-workflow.yaml
     with:
       lib: ${{ matrix.app }}
       ios: ${{ matrix.ios }}
       xcode: ${{ matrix.xcode }}
-      macos: ${{ matrix.macos }}
       build_only: true
     secrets: inherit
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yaml
```diff
@@ -12,14 +12,17 @@ jobs:
...
+            macos: macos-26
+          - ios: ^18
             xcode: ^16
-          # - macos: macos-15
-          #   ios: ^26
-          #   xcode: ^26
-          # Note: iOS 26 simulator is not yet available on GitHub runners
+            macos: macos-26
+          - ios: ^17
+            xcode: ^15
```
wmathurin: If we run install.sh, won't [this](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/install.sh#L26) throw an error for Xcode 15?

.github/workflows/nightly.yaml
```diff
@@ -12,14 +12,17 @@ jobs:
       fail-fast: false
       matrix:
         lib: [SalesforceSDKCommon, SalesforceAnalytics, SalesforceSDKCore, SmartStore, MobileSync]
+        ios: [^26, ^18, ^17]
         include:
-          - macos: macos-15
-            ios: ^18
+          - ios: ^26
+            xcode: ^26
+            macos: macos-26
```
wmathurin: Cool! They just added [it](https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md), right?

.github/workflows/nightly.yaml
brandonpage: Good catch!  I will test that on my fork and see.  Unfortunately, I don't see a way around it since the macos-15 image does not have the iOS 17 runtime.

.github/workflows/nightly.yaml
brandonpage: Yup, it went live yesterday.

.github/workflows/nightly.yaml
wmathurin: I have not seen iOS 17 with xcode 16 on any of the mac-os runners.
We could stop calling install.sh [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/.github/workflows/reusable-workflow.yaml#L42). It really does not do much anymore (except for that xcode version check).

brandonpage: @wmathurin [You were right](https://github.com/brandonpage/SalesforceMobileSDK-iOS/actions/runs/17686550884/job/50272504636?pr=16).  I will look for an alternative approach.

wmathurin: Kill the call to install.sh - we don't have submodules anymore, so it's not really necessary. It's only helpful for devs to check if they have Xcode installed etc at this point.

brandonpage: I sort of wish I saw this comment 2 hours ago... but I was able to work around the issue.  It is pretty trivial to install the iOS 17 runtime, I spent far more time trying to get Github Actions terrible syntax working to make it conditional.

wmathurin: Maybe it's better this way since we don't officially support Xcode 15 anymore.

PR #3920, opened 2025-09-17
@W-19669793: [iOS] App Hangs When Submitting OTP And New Password In Native Reset Password Flow

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This bug fix resolves a hang in the complete step of the native reset password flow.  Only that flow was using a different REST client than all the others, which were recently updated in modernization efforts.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLoginManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLoginManagerInternal.swift
@@ -255,11 +255,11 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
                         recaptcha: reCaptchaParameterGenerationResult.nonEnterpriseReCaptchaToken,
                         recaptchaEvent: reCaptchaParameterGenerationResult.enterpriseReCaptchaEvent,
                         userData: UserData(
-                            email: email,
-                            username: username,
-                            password: newPassword,
+                            email: trimmedEmail,
+                            username: trimmedUsername,
                             firstName: firstName,
                             lastName: lastName),
+                        password: trimmedPassword,
                         otpVerificationMethod: otpVerificationMethodString)
                 ),
                 encoding: .utf8)
@@ -340,6 +340,9 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         /// The start registration request user data
         var userData: UserData
         
+        /// The user-entered new password
+        var password: String
+        
         /// The one-time-password's delivery method for verification in "email" or "sms"
         var otpVerificationMethod: String
         
@@ -347,6 +350,7 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
             case recaptcha = "recaptcha"
             case recaptchaEvent = "recaptchaevent"
             case userData = "userdata"
+            case password = "password"
             case otpVerificationMethod = "verificationMethod"
         }
     }
@@ -361,9 +365,6 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         /// A valid Salesforce username or email
         var username: String
         
-        /// The user-entered new password
-        var password: String
-        
         /// The user-entered first name
         var firstName: String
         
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLoginManagerInternal.swift
```diff
@@ -489,7 +489,7 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         )
         
         let completePasswordResetResponse = await handleResponseForRequest {
-            try await RestClient.shared.send(request: completePasswordResetRequest)
+            try await RestClient.sharedGlobal.send(request: completePasswordResetRequest)
```
JohnsonEricAtSalesforce, line 492: This fixes the hang in the work item when I tested it live in iOSNativeLoginTemplate.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLoginManagerInternal.swift
```diff
@@ -340,13 +340,17 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         /// The start registration request user data
         var userData: UserData
         
+        /// The user-entered new password
+        var password: String
```
JohnsonEricAtSalesforce, line 344: The location of the password parameter is incorrect, so I've fixed that as a rider change - if that's alright.  The structure of this data changed in the Identity API during development.  It was updated on Android since that was developed later.  It was missed on iOS.  The issue may not have been reported since a user could change their password successfully from the randomized default password the Identity API would create when it didn't find the password specified in the request.

I also tested this live in iOSNativeLoginTemplate plus I did a side-by-side comparison with the request data in the Android version at runtime.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLoginManagerInternal.swift
```diff
@@ -340,13 +340,17 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
...
+        
         /// The one-time-password's delivery method for verification in "email" or "sms"
         var otpVerificationMethod: String
         
         enum CodingKeys: String, CodingKey {
             case recaptcha = "recaptcha"
             case recaptchaEvent = "recaptchaevent"
             case userData = "userdata"
+            case password = "password"
             case otpVerificationMethod = "verificationMethod"
```
bbirman: does case matter for the keys? I was looking at Android to see the password difference you mentioned and it looks like it has "[verificationmethod](https://github.com/forcedotcom/SalesforceMobileSDK-Android/blob/b302878d7ea6c145794325ab015c7d7e2b89fa97/libs/SalesforceSDK/src/com/salesforce/androidsdk/auth/NativeLoginManager.kt#L360)" as all lower case

bbirman (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLoginManagerInternal.swift
JohnsonEricAtSalesforce: Yes, the case does indeed matter.  The default verification is email and that's all we have the ability to test with so this is another great cleanup.

We can't test with SMS, but I was able to verify that with the camel-cased name it always sends the email even when SMS is selected.  Android does not send the email when SMS is selected, so it seems logical that's the correct implementation.

We may be able to work with the Identity team to see how SMS could be tested.

JohnsonEricAtSalesforce: I got this fixed as well.

JohnsonEricAtSalesforce: We'd also have to work with the Identity team to add a phone number field to the registration process and to our template as well, just while I'm getting all my thoughts for a full test on paper.  The registration process is editable in Experience Builder.

bbirman: sounds good!

PR #3933, opened 2025-09-19
Deprecating sharedProcessPool in SFSDKWebViewStateManager

wmathurin:
`sharedProcessPool` will be removed in MSDK 14 from `SFSDKWebViewStateManager` (at which points the deprecation warnings will disappear).

`WKProcessPool` creation has no effect since iOS 15, so we are no longer using it.

Also removed `SFSDKWebViewStateManager.h` and `SFSDKWebViewStateManager.m` which should have been removed when the class was converted to Swift a while back.


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
```diff
@@ -68,26 +65,20 @@ public class SFSDKWebViewStateManager: NSObject {
     @objc
     @MainActor
     public static func removeSessionForcefully() async {
-        sharedProcessPool = nil
         await removeWKWebViewCookies()
     }
 
     
+    @available(*, deprecated, message: "Deprecated in Salesforce Mobile SDK 13.1 and will be removed in Salesforce Mobile SDK 14.0. WKProcessPool creation has no effect on iOS 15+ and this property will be removed.")
```
bbirman, line 72: should this be 13.2 now?

wmathurin: You're right. I'll send another PR.

wmathurin: Here it is: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3939

PR #3937, opened 2025-09-26
Fix related to flapping test

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/SFSDKPushNotificationDecryption.m
```diff
@@ -199,7 +199,7 @@ + (SFEncryptionKey *)getAESKeyFromSecret:(NSString *)secret error:(NSError **)er
     }
 
     CFRelease(privateKeyRef);
-    if (decryptedData == nil) {
+    if (decryptedData == nil || [decryptedData length] != 32) {
```
bbirman, line 202: From the comment in the flapping test (below), starting in 17.4 decrypting with a bad key returns data instead of nil so at the time I just changed our expected error in the test. But it looks like that data is also variable length (in my runs locally it was 72, 54, and 15). When the data length is less than 32, lines 209 or 210 crash so adding a length check before that

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKEncryptedPushNotificationTests.m
```diff
@@ -190,13 +190,9 @@ - (void)testNotificationTransformNonRSASecret {
...
-    if (@available(iOS 17.4, *)) {
-        // As of 17.4, decrypting a bad key with PKCS1 returns data instead of nil, so the secret decryption doesn't fail
-        // at the same point as before but using it later to decrypt the content still fails
-        XCTAssertEqual(nonRSASecretError.code, SFSDKPushNotificationErrorContentDecryptionFailed);
-    } else {
-        XCTAssertEqual(nonRSASecretError.code, SFSDKPushNotificationErrorSecretDecryptionFailed);
-    }
+    // As of 17.4, decrypting a bad key with PKCS1 returns data instead of nil, so the secret decryption doesn't fail
+    // at the same point as before but using it later to decrypt the content still fails
+    XCTAssert(nonRSASecretError.code == SFSDKPushNotificationErrorContentDecryptionFailed || nonRSASecretError.code == SFSDKPushNotificationErrorSecretDecryptionFailed);
```
bbirman, line 195: I kept the check for both errors in case there's a chance that the bad data is 32 long and passes the check but still fails to decrypt later

PR #3940, opened 2025-10-08
Add settings menu to login screen

bbirman:
https://github.com/user-attachments/assets/a407af67-df22-4cf5-b044-2dc494d2a694

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/WebViewStateManagerTests.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/WebViewStateManagerTests.swift
@@ -35,4 +35,38 @@ final class WebViewStateManagerTests: XCTestCase {
         SFSDKWebViewStateManager.sessionCookieManagementDisabled = false
         XCTAssertFalse(SFSDKWebViewStateManager.sessionCookieManagementDisabled)
     }
+    
+    @MainActor
+    func testClearCache() async throws {
+        // Add some test data
+        let html = "<html><body><script>localStorage.setItem('test', 'value');</script></body></html>"
+        let webView = WKWebView(frame: CGRect(x: 0, y: 0, width: 100, height: 100), configuration: WKWebViewConfiguration())
+        
+        // Load HTML and wait for JavaScript to execute
+        webView.loadHTMLString(html, baseURL: URL(string: "https://test.salesforce.com"))
+        try await Task.sleep(for: .seconds(1))
+        
+        // Define data types to check
+        let dataTypes: Set<String> = [WKWebsiteDataTypeDiskCache,
+                                      WKWebsiteDataTypeMemoryCache,
+                                      WKWebsiteDataTypeFetchCache,
+                                      WKWebsiteDataTypeLocalStorage,
+                                      WKWebsiteDataTypeSessionStorage,
+                                      WKWebsiteDataTypeIndexedDBDatabases,
+                                      WKWebsiteDataTypeWebSQLDatabases,
+                                      WKWebsiteDataTypeOfflineWebApplicationCache,
+                                      WKWebsiteDataTypeServiceWorkerRegistrations]
+        
+        let dataStore = WKWebsiteDataStore.default()
+        // Verify data exists before clearing
+        let initialRecords = await dataStore.dataRecords(ofTypes: dataTypes)
+        XCTAssertFalse(initialRecords.isEmpty, "Expected data to exist before clearing")
+        
+        // Clear the cache
+        await SFSDKWebViewStateManager.clearCache()
+        
+        // Verify data was cleared
+        let records = await dataStore.dataRecords(ofTypes: dataTypes)
+        XCTAssertTrue(records.isEmpty, "Expected data to be cleared")
+    }
 }


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -265,9 +255,53 @@ - (UIBarButtonItem *)createBackButton {
...
 - (UIBarButtonItem *)createSettingsButton {
     UIImage *image = [[SFSDKResourceUtils imageNamed:@"login-window-gear"] imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
-    UIBarButtonItem *settingsButton = [[UIBarButtonItem alloc] initWithImage:image style:UIBarButtonItemStylePlain target:self action:@selector(showLoginHost:)];
-    settingsButton.accessibilityLabel = [SFSDKResourceUtils localizedString:@"LOGIN_CHOOSE_SERVER"];
-    settingsButton.accessibilityIdentifier = @"choose connection button";
+    
+    NSMutableArray  *menuActions = [[NSMutableArray alloc] initWithCapacity:3];
+
+    // Don't show the change server option if there are no hosts to switch to.
+    SFManagedPreferences *managedPreferences = [SFManagedPreferences sharedPreferences];
```
bbirman, line 270: I moved this check to just be around the “Choose Connection” option instead of the settings icon but that would just be for MDM. @wmathurin what do you think about having another property like `showServerPicker` on the login config in case an app wants to show the settings icon but hide the change server option? Here’s the existing configuration doc: https://developer.salesforce.com/docs/platform/mobile-sdk/guide/oauth-hide-gear-icon.html

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
```diff
@@ -62,6 +63,14 @@ public class SFSDKWebViewStateManager: NSObject {
         }
     }
     
+    @objc
+    @MainActor
+    public static func clearCache() async {
+        let dataStore = WKWebsiteDataStore.default()
+        let websiteDataTypes: Set<String> = [WKWebsiteDataTypeFetchCache, WKWebsiteDataTypeMemoryCache, WKWebsiteDataTypeDiskCache, WKWebsiteDataTypeOfflineWebApplicationCache]
```
bbirman: I included any type that had cache in the name

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1044,6 +1044,25 @@ - (void)loginViewController:(SFLoginViewController *)loginViewController didChan
     [[NSNotificationCenter defaultCenter] postNotification:loginHostChangedNotification];
     NSString *sceneId = loginViewController.view.window.windowScene.session.persistentIdentifier;
     self.authSessions[sceneId].oauthRequest.loginHost = newLoginHost.host;
+    [self restartAuthenticationForViewController:loginViewController];
+}
+
+- (void)loginViewControllerDidClearCache:(SFLoginViewController *)loginViewController {
```
bbirman, line 1050: From [this comment about reloading](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2752#discussion_r2311540285), I used the same restart method as the one that happens after the login host is changed

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
wmathurin: Should we use: `WKWebsiteDataStore.allWebsiteDataTypes()` ?

bbirman: Is it okay that it would also include the cookies that we have the other method for? Having a way to get rid of everything sounds good to me

wmathurin: In my mind, it was about starting clean. Let me check the Android code.

wmathurin: On Android we are doing `webView.clearCache(true)` (see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2752/files#diff-2b4d6c9697f160bfbefe121309b911622ea530d287a7fbc780a872cfd75cd82aR247) which does not clear the cookies.

To do everything but the cookies we could do
```swift
let nonCookieDataTypes: Set<String> = [
    WKWebsiteDataTypeDiskCache,
    WKWebsiteDataTypeMemoryCache,
    WKWebsiteDataTypeLocalStorage,
    WKWebsiteDataTypeSessionStorage,
    WKWebsiteDataTypeIndexedDBDatabases,
    WKWebsiteDataTypeWebSQLDatabases,
    WKWebsiteDataTypeOfflineWebApplicationCache,
    WKWebsiteDataTypeServiceWorkerRegistrations
]```

bbirman: Sounds good, updated! I also changed the "Clear Cookies" option to use the method that clears the cookies regardless of the cookie management flag, I was thinking that would be better to make sure it starts clean like you said (let me know if you think I should change it back)

wmathurin (approved):

PR #3941, opened 2025-10-09
Saving scopes with rest of credentials

wmathurin:
- token end point response contains scopes
- this value is now captured and saved with credentials
- tests were updated
- also showing scopes in dev info

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
@@ -146,7 +146,6 @@ NS_SWIFT_NAME(OAuthCredentials)
  
  This property is available after authentication has successfully completed.
  
- @warning The setter for this property is exposed publicly only for unit tests. Client code should not set this property.
  @exception NSInternalInconsistencyException If accessed while the identifier property is `nil`.
  */
 @property (nonatomic, readonly, nullable) NSString *organizationId;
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
```diff
@@ -172,6 +172,15 @@ NS_SWIFT_NAME(OAuthCredentials)
  */
 @property (nonatomic, readonly, nullable) NSURL *apiInstanceUrl;
 
+/** The OAuth scopes granted for this session.
+ 
+ This property contains the list of OAuth scopes that were granted during authentication.
+ This property is set by the `SFOAuthCoordinator` after authentication has successfully completed.
+ 
+ @warning The setter for this property is exposed publicly only for unit tests. Client code should not set this property.
```
bbirman: Could it go into SFOAuthCredentials+Internal.h for unit tests? 
edit: I got confused by the comment, it's readonly?

wmathurin: We have the same comments all over the file.
They are very old (seems to date from https://github.com/forcedotcom/SalesforceMobileSDK-iOS/commit/61a9e26e7cdf1a2c7cede8417774121e165fdd6b#diff-b278b72952b2a17bb9191443c53295ad580cf65c013cfacc1bc44d06e59acb82R100-R101).
I'll remove it since it is readonly in the public header and only readwrite in the internal header.

PR #3943, opened 2025-10-13
Allowing no scopes to be provided

wmathurin:
- Not sending scope parameter in authorization url if no scopes defined
- Added test for scope parameter building method in SalesforceOAuthUnitTests.m
- Removed check for scopes in SFSDKAppConfig (class responsible for parsing bootconfig)
- Added new test suite BootconfigTests.swift
- Added check for id scope in SFIdentityCoordinator.m
- Added new helper class ScopeParser.swift
- Added new test suite ScopeParserTests.swift
- Moved some tests files that were in SalesforceSDKCore instead of SalesforceSDKCoreTests

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKAppConfig.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKAppConfig.h
@@ -27,7 +27,6 @@
 typedef NS_ENUM(NSInteger, SFSDKAppConfigErrorCode) {
     SFSDKAppConfigErrorCodeNoConsumerKey = 966,
     SFSDKAppConfigErrorCodeNoRedirectURI,
-    SFSDKAppConfigErrorCodeNoOAuthScopes
 } NS_SWIFT_NAME(BootConfig.ErrorCode);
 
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SFSDKAppConfig.m
```diff
@@ -100,10 +100,6 @@ - (BOOL)validate:(NSError **)error {
         [[self class] createError:error withCode:SFSDKAppConfigErrorCodeNoRedirectURI message:[SFSDKResourceUtils localizedString:@"appConfigValidationErrorNoRedirectURI"]];
         return NO;
     }
-    if (self.oauthScopes.count == 0) {
-        [[self class] createError:error withCode:SFSDKAppConfigErrorCodeNoOAuthScopes message:[SFSDKResourceUtils localizedString:@"appConfigValidationErrorNoOAuthScopes"]];
```
bbirman, line 104: Could `appConfigValidationErrorNoOAuthScopes` be removed from the strings file?

wmathurin: Will do.

PR #3944, opened 2025-10-16
@W-19368408: [MSDK] Local dev instant login (iOS)

JohnsonEricAtSalesforce:
🎸 _*Ready For Review*_ 🥁

  This is the iOS counterpart to https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2785.

  This adds a new process argument which may be used for debug app builds only to authenticate a user via test credentials in automated tests. The new code is wrapped in a `#ifdef DEBUG` preprocessor condition so that it will not be present for release builds.

  This new means of automating authentication for debug testing does not expose production builds to risk and also minimizes access since usernames and passwords are not used. The test credentials JSON provided by Salesforce Mobile SDK's REST Explorer sample app uses refresh tokens which can be remotely expired and cannot be used for privilege elevation beyond the public APIs.

  Here's an example of testing the new parameter interactively in Xcode by adding it to the app's scheme:
Screenshot 2025-10-15 at 18 56 39

  Here's an example for XCUITest:
```
super.launchArguments = ["creds: \(STRING_PACKAGE)"]
super.launch()
```

  This is an Appium example:
```
capabilities = {
     "platformName": "iOS",
     "deviceName": "iPhone 17 Pro Max",
     "platformVersion": "26.0",
     "app": "/path/to/your/app.app",
     "automationName": "XCUITest",
     "processArguments": {
       "args": ["creds", "<STRING PACKAGE>"]
     }
   }
```

And, finally, an example at the shell:
```
xcrun simctl launch booted <YOUR APP> -creds "<STRING PACKAGE>"
```

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -80,6 +80,7 @@ ignore:
   - "shared"
   - "native"
   - "libs/**/*Test*/*"
+  - "**/TestSetupUtils*"
 
 flag_management:
   default_rules: # the rules that will be followed for any flag added, generally


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -231,6 +231,18 @@ + (instancetype)sharedManager {
...
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
+
+#ifdef DEBUG
+        // For debug app builds only, use test instant log in if applicable.
+        NSArray<NSString *> *arguments = [[NSProcessInfo processInfo] arguments];
+        if ([arguments containsObject:@"-creds"]) {
+            NSString *creds = arguments[[arguments indexOfObject:@"-creds"] + 1];
+            
+            [TestSetupUtils populateAuthCredentialsFromString:creds];
```
JohnsonEricAtSalesforce: The crux is retrieving the new parameter and passing it to mostly existing test harness logic.  @bbirman, is this a logical place to accomplish this? I tried few other ideas, but ran into issues and circular dependencies on the manager singleton.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/TestSetupUtils.h
```diff
@@ -52,6 +52,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 + (SFSDKTestCredentialsData *)populateAuthCredentialsFromConfigFileForClass:(Class)testClass;
 
+/**
+ Loads a set of auth credentials from the provided JSON string, and configures
+ SFUserAccountManager and the current account with the data from that JSON.
+ @param testCredentialsJsonString The test credentials JSON as a string.
+ @return The configuration data used to configure SFUserAccountManager (useful e.g. for hybrid
+ apps which need the data to bootstrap SFHybridViewController).
+ */
++ (SFSDKTestCredentialsData *)populateAuthCredentialsFromString:(NSString *)testCredentialsJsonString;
```
JohnsonEricAtSalesforce, line 63: This new method lets us reuse the non-file logic around ingesting the JSON.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/TestSetupUtils.m
```diff
@@ -98,41 +147,4 @@ + (SFSDKTestCredentialsData *)populateAuthCredentialsFromConfigFileForClass:(Cla
     return credsData;
 }
 
-+ (void)synchronousAuthRefresh
```
JohnsonEricAtSalesforce, line 101: This method didn't change, though Github marked it as a diff due to other logic moving.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/TestSetupUtils.m
```diff
@@ -55,6 +55,55 @@ + (SFSDKTestCredentialsData *)populateAuthCredentialsFromConfigFileForClass:(Cla
...
+         authListener.returnStatus = kTestRequestStatusDidFail;
+     }];
+    [authListener waitForCompletion];
+    [[SFUserAccountManager sharedInstance] setCurrentUserInternal:user];
+    NSAssert([authListener.returnStatus isEqualToString:kTestRequestStatusDidLoad], @"After auth attempt, expected status '%@', got '%@'",
+             kTestRequestStatusDidLoad,
+             authListener.returnStatus);
+}
+
++ (SFOAuthCredentials *)newClientCredentials {
```
JohnsonEricAtSalesforce, line 111: This method did not change.  This method didn't change, though Github marked it as a diff due to other logic moving.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/TestSetupUtils.m
```diff
@@ -55,6 +55,55 @@ + (SFSDKTestCredentialsData *)populateAuthCredentialsFromConfigFileForClass:(Cla
...
+    return [TestSetupUtils authCredentialsFromJson:jsonResponse];
+    
+}
+
++ (SFSDKTestCredentialsData *)populateAuthCredentialsFromString:(NSString *)testCredentialsJsonString {
+    
+    return [TestSetupUtils authCredentialsFromJson:[SFJsonUtils objectFromJSONString:testCredentialsJsonString]];
+}
+
++ (void)synchronousAuthRefresh
```
JohnsonEricAtSalesforce: This method didn't change, though Github marked it as a diff due to other logic moving.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/TestSetupUtils.m
```diff
@@ -98,41 +147,4 @@ + (SFSDKTestCredentialsData *)populateAuthCredentialsFromConfigFileForClass:(Cla
...
-         authListener.returnStatus = kTestRequestStatusDidFail;
-     }];
-    [authListener waitForCompletion];
-    [[SFUserAccountManager sharedInstance] setCurrentUserInternal:user];
-    NSAssert([authListener.returnStatus isEqualToString:kTestRequestStatusDidLoad], @"After auth attempt, expected status '%@', got '%@'",
-             kTestRequestStatusDidLoad,
-             authListener.returnStatus);
-}
-
-+ (SFOAuthCredentials *)newClientCredentials {
```
JohnsonEricAtSalesforce, line 128: This method didn't change, though Github marked it as a diff due to other logic moving.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
bbirman: Is SFUserAccountManager a possibility? If not, this class sounds good to me, though wondering about using `load`, does it have something to do with the runtime?

JohnsonEricAtSalesforce: I looked through SFUserAccountManager just now.  It doesn't seem to have any logical place to add this.  I tried a couple, but they also trigger some circular initialization overflows a lot like some of the entry points in the manager I tried.  TestSetupUtils is dependent on the SDK and user account manager objects.

`load` is nice in this instance since it's the class level initializer for SalesforceSDKManager and is outside the other initialization methods.  You're correct that it's part of the Objective-C spec.  I recall there is a way to do something like that in Swift, but it has been a few years since I exercised that.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -195,6 +195,16 @@ + (void)initialize {
 
 + (void)initializeSDK {
     [self initializeSDKWithClass:InstanceClass];
+#ifdef DEBUG
+    // For debug app builds only, use test instant log in if applicable.
+    NSArray<NSString *> *arguments = [[NSProcessInfo processInfo] arguments];
+    if ([arguments containsObject:@"-creds"]) {
+        NSString *creds = arguments[[arguments indexOfObject:@"-creds"] + 1];
+        
+        [TestSetupUtils populateAuthCredentialsFromString:creds initializeSdk:NO];
```
JohnsonEricAtSalesforce, line 204: @bbirman - Now we have Local Dev Instant Login hosted in `SalesforceSDKManager.init` so we can avoid using `load`.  Notice the two new boolean parameters on these two statements which allow `TestSetupUtils` to execute correctly here in `initializeSDK` while still running exactly the same in existing tests.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/TestSetupUtils.m
```diff
@@ -55,6 +55,71 @@ + (SFSDKTestCredentialsData *)populateAuthCredentialsFromConfigFileForClass:(Cla
...
+        authListener.returnStatus = kTestRequestStatusDidLoad;
+        user = userAccount;
+        // Ensure tests don't change/corrupt the current user credentials.
+        if (user.credentials.refreshToken == nil) {
+            user.credentials = credentials;
+        }
+        if (postUserDidLogIn) {
+            NSDictionary *userInfo = @{kSFNotificationUserInfoAccountKey: userAccount,
+                                       kSFNotificationUserInfoAuthTypeKey: authInfo};
+            [[NSNotificationCenter defaultCenter] postNotificationName:kSFNotificationUserDidLogIn
```
JohnsonEricAtSalesforce, line 95: These two lines allow `TestSetupUtils` to stop the default login from being shown and deliver the test app directly to its authenticated view.

JohnsonEricAtSalesforce: That matches the behavior for Local Dev Instant Login for Android.

JohnsonEricAtSalesforce: There _is_ similar notification login in the user account manager, but its only a few lines and not visible to this scope, so I thought to leave that alone and replicate the notification here so we minimize authentication regression risk.  All the changes in this pull request are non-production code so far.

JohnsonEricAtSalesforce:
I ran this using the RestAPIExplorer sample and giving it the new parameter shown in the screenshot above.  After a fresh install, it brought my test user directly to the app's authenticated view in all my test runs ⭐️

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
bbirman: nice!

bbirman (approved):

wmathurin (approved):

PR #3949, opened 2025-11-04
Method to migrate refresh token

wmathurin:
Introducing a new method that allows migrating a user’s refresh token.
#### Scenario
- A user currently has a refresh token associated with consumer key K and scopes S.
- The new method should replace it with a refresh token associated with consumer key K′ and scopes S′.
#### Examples
- K = K′: used when requesting additional scopes.
- K ≠ K′: used when migrating from a connected app to a beacon app.
- S′ = ∅: used to request all available scopes offered by the app.
#### Constraints
- The process is not headless and should not run silently in the background.
- User interaction may be required (e.g., approval prompt).

Migration

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
@@ -80,6 +80,8 @@
 		4F2410B1282DCA4B00E5EFE3 /* SFSDKCollectionResponse.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F2410B0282DCA4B00E5EFE3 /* SFSDKCollectionResponse.m */; };
 		4F3139652331C5A1007B3705 /* SFSDKAuthRootController.m in Sources */ = {isa = PBXBuildFile; fileRef = 4F3139642331C5A1007B3705 /* SFSDKAuthRootController.m */; };
 		4F3139682331C5C7007B3705 /* SFSDKAuthRootController.h in Headers */ = {isa = PBXBuildFile; fileRef = 4F3139672331C5B9007B3705 /* SFSDKAuthRootController.h */; };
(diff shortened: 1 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -815,6 +825,45 @@ - (void)dismissAuthViewControllerIfPresentForScene:(UIScene *)scene completion:(
...
+    
+    // Store current user credentials to revoke them once migration completes
+    SFOAuthCredentials *preMigrationCredentials = self.currentUser.credentials;
+
+    // Creating a SFSDKAuthRequest and SFSDKAuthSession
+    SFSDKAuthRequest *request = [self migrateRefreshAuthRequest:newAppConfig];
+    SFSDKAuthSession *authSession = [[SFSDKAuthSession alloc] initWith:request credentials:nil];
+    authSession.isAuthenticating = YES;
+    authSession.authSuccessCallback = ^(SFOAuthInfo *authInfo, SFUserAccount *newUserAccount) {
+        if (preMigrationCredentials != nil && ![preMigrationCredentials.refreshToken isEqualToString:newUserAccount.credentials.refreshToken]) {
```
wmathurin, line 841: Similar to what we do during biometric auth login - the old refresh token needs to be revoked on the server.

wmathurin:
There are no new unit tests :-(
migrateRefreshToken will revoke the old refresh token - that means also that testing it with instant login will be challenging.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
wmathurin: That means also that testing it with instant login will be challenging.

bbirman (approved):

PR #3950, opened 2025-11-08
Add Login Options to Dev Menu

wmathurin:
Matching the Android work: https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2801

#### Overview
- Added a new button "Login Options" - only shown when on the login screen
- "Login Options" brings up an action sheet to configure the auth flow (toggles for web server flow, hybrid flow and supports welcome discovery), use of the static bootconfig or a dynamic bootconfig
- Also fixed dev info so it does not crash when invoked without a current user

#### Notes
- the code for the new screen was refactored out of AuthFlowTester into SalesforceSDKCore
- AuthFlowTester no longer does a "deferred" login
- Toggle for "supports welcome discovery" is new in this PR

LoginOptions

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/DomainDiscoveryCoordinatorTests.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/DomainDiscoveryCoordinatorTests.swift
@@ -25,7 +25,7 @@ final class DomainDiscoveryCoordinatorTests: XCTestCase {
         // Given
         let mockWebView = MockWKWebView()
         let coordinator = DomainDiscoveryCoordinator()
-        let credentials = OAuthCredentials(identifier: "test", clientId: "client123", encrypted: false)
+        let credentials = OAuthCredentials(identifier: "test", clientId: "client123", encrypted: false)!
         
         let expectedDomain = "foo.my.salesforce.com"
         let mockDomain = "https://\(expectedDomain)"
@@ -35,7 +35,7 @@ final class DomainDiscoveryCoordinatorTests: XCTestCase {
         mockWebView.simulatedCallbackURL = callbackURL
         
         // When
-        coordinator.runMyDomainsDiscovery(on: mockWebView, with: credentials!)
+        coordinator.runMyDomainsDiscovery(on: mockWebView, with: credentials)
         let results = coordinator.handle(action: mockWebView.mockAction!)
         
         // Then
@@ -47,14 +47,14 @@ final class DomainDiscoveryCoordinatorTests: XCTestCase {
         // Given
         let mockWebView = MockWKWebView()
         let coordinator = DomainDiscoveryCoordinator()
-        let credentials = OAuthCredentials(identifier: "test", clientId: "client123", encrypted: false)
+        let credentials = OAuthCredentials(identifier: "test", clientId: "client123", encrypted: false)!
         let expectedLoginHint = "testuser@example.com"
         let callbackURLString = "sfdc://discocallback?login_hint=\(expectedLoginHint.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!)"
         let callbackURL = URL(string: callbackURLString)!
         mockWebView.simulatedCallbackURL = callbackURL
         
         // When
-        coordinator.runMyDomainsDiscovery(on: mockWebView, with: credentials!)
+        coordinator.runMyDomainsDiscovery(on: mockWebView, with: credentials)
         let results = coordinator.handle(action: mockWebView.mockAction!)
         
         // Then
@@ -65,7 +65,7 @@ final class DomainDiscoveryCoordinatorTests: XCTestCase {
         // Given
         let mockWebView = MockWKWebView()
         let coordinator = DomainDiscoveryCoordinator()
-        let credentials = OAuthCredentials(identifier: "test", clientId: "client123", encrypted: false)
+        let credentials = OAuthCredentials(identifier: "test", clientId: "client123", encrypted: false)!
         let expectedDomain = "foo.my.salesforce.com"
         let mockDomain = "https://\(expectedDomain)"
         let callbackURLString = "sfdc://discocallback?my_domain=\(mockDomain.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed)!)"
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -471,25 +472,59 @@ - (NSString *)devInfoTitleString
...
-                 [presentedViewController presentViewController:keyValueStoreInspector animated:YES completion:nil];
-             }]
-    ];
+    NSMutableArray<SFSDKDevAction *> *actions = [NSMutableArray array];
+    SFUserAccountManager *userAccountManager = [SFUserAccountManager sharedInstance];
+    SFUserAccount *currentUser = userAccountManager.currentUser;
+    
+    // Check if we're showing the login screen
+    BOOL isShowingLogin = [presentedViewController isKindOfClass:[SFLoginViewController class]] ||
+                          [presentedViewController.presentingViewController isKindOfClass:[SFSDKAuthRootController class]];
```
wmathurin: Is there a cleaner way to do that?

wmathurin:
Also, I am not restarting the login flow so the new settings chosen in the action sheet won't be applied. I'll look at it on Monday.

wmathurin:
> Also, I am not restarting the login flow so the new settings chosen in the action sheet won't be applied. I'll look at it on Monday.

Done with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3950/commits/0330aa98d65e23e11d5840ea639b2e02bf6d344f

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
bbirman: Could potentially call isEnabled on the authWindow[ here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/b812b4886c8e1bbfed68d95478aad6e80ab99791/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Views/SFSDKWindowManager.h#L92), but from a different issue I've been wanting to add an `autoCreate` param to that method so that it doesn't always create the window if one doesn't already exist (could be separate work)

wmathurin: For now we could only support showing login options with regular login. So I would remove line 481.
For advanced auth, I'm missing the code to restart the auth flow once login options have been changed (see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3950/files#diff-1334e2942ea2928b6db7bca76fdebd960b717bc0463c404244b998a86ee08121R497).

wmathurin: Done here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3950/commits/a21a187107a0426e294df6adb61c29a5548cba80

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -471,25 +471,77 @@ - (NSString *)devInfoTitleString
...
-             }]
-    ];
+    NSMutableArray<SFSDKDevAction *> *actions = [NSMutableArray array];
+    SFUserAccountManager *userAccountManager = [SFUserAccountManager sharedInstance];
+    SFUserAccount *currentUser = userAccountManager.currentUser;
+    
+    // Check if we're showing the login screen
+    BOOL isShowingLogin = [presentedViewController isKindOfClass:[SFLoginViewController class]];
+    // TODO uncomment to support advanced auth case (once we add code to restart auth in that case below)
+    // || [presentedViewController.presentingViewController isKindOfClass:[SFSDKAuthRootController class]];
```
brandonpage, line 481: Restarting Auth is included now, right?

wmathurin: Only if it's regular auth - for now I commented out the case it's advanced auth.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -471,25 +471,77 @@ - (NSString *)devInfoTitleString
...
+    
+    // Show dev info - always available
+    [actions addObject:[[SFSDKDevAction alloc]initWith:@"Show dev info" handler:^{
+        UIViewController *devInfo = [SFSDKDevInfoViewController makeViewController];
+        [presentedViewController presentViewController:devInfo animated:YES completion:nil];
+    }]];
+    
+    // Login Options - only show on login screen
+    if (isShowingLogin) {
```
brandonpage, line 490: NIT: Is there a reason to _only_ show it on the login screen?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -471,25 +471,77 @@ - (NSString *)devInfoTitleString
...
+                    }
+                    // TODO support advanced auth case
+                }];
+            }];
+            [presentedViewController presentViewController:configPicker animated:YES completion:nil];
+        }]];
+    }
+    
+    // Logout - only show if there's a current user and not on login screen
+    if (currentUser && !isShowingLogin) {
```
brandonpage, line 506: I was thinking about exactly this the other day.  Android doesn't take deferred auth into account, I need to also check for a user.

brandonpage:
@wmathurin Still reviewing, but isn't moving `NativeLoginManager` a SemVer issue?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
wmathurin: It's only useful if you do a login right after - and if you do a login (let's say after a logout or when adding a new user), then you will get a chance to bring up the screen when the login screen is shown. So it feels like login is the right time to show the login options screen. The oauth settings should be inspectable through dev infos at any time though.

PR #3954, opened 2025-11-23
Fix actionable notifications options

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/PushNotificationManager+ActionableNotifications.swift
```diff
@@ -163,34 +163,7 @@ internal extension PushNotificationManager {
 }
 
 private extension PushNotificationManager {
-    
-    func createNotificationCategories(from type: NotificationType) -> [UNNotificationCategory] {
```
bbirman, line 167: It looks like the deleted code is duplicative of the methods in  [NotificationCategoryFactory.swift ](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/NotificationCategoryFactory.swift)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/PushNotificationManager+ActionableNotifications.swift
```diff
@@ -163,34 +163,7 @@ internal extension PushNotificationManager {
...
-    func createActions(from actionGroup: ActionGroup?) -> [UNNotificationAction] {
-        guard let actionGroup = actionGroup else {
-            return []
-        }
-        
-        return actionGroup.actions.compactMap { action in
-            UNNotificationAction(
-                identifier: action.identifier,
-                title: action.label,
-                options: [.foreground] // Ensures the app opens if needed
```
bbirman, line 189: This is the line causing the issue and the updated logic is already in the other class [here](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/ee6a97052f7f33be0513c674fd8e22c9249ad042/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/NotificationCategoryFactory.swift#L37)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/PushNotificationManager+ActionableNotifications.swift
wmathurin: Does `NotificationCategoryFactory.shared.createCategories` behave differently from the duplicate?

bbirman: It returns a set instead of an array but otherwise I didn't see a behavior difference until it calls into `createActions`. I was wondering if we should keep this and delete `NotificationCategoryFactory` but that one is public

wmathurin: Because of semver, we can't do that (in 13.2 at least).

PR #3955, opened 2025-11-25
Enhancements for AuthFlowTester

wmathurin:
- Showing all the user credential fields / grouped in collapsible sections
- User credentials do no collapse if revoke / refresh (i.e. make request)
- Make request shows an alert (like revoke) and response is a collapsible view (initially collapsed)

Enh2 Enhancements


wmathurin:
I wanted to highlight values that changed e.g. access token after revoke/refresh but it's proving tricky since the nested views are recreated when the parent view's state changes.

PR #3958, opened 2025-12-02
Fix callbacks for RestAPI

bbirman:
Manually tested for now, I still want to see if we can add a unit/integration test for it

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -20,7 +20,7 @@ jobs:
             xcode: ^16
           - ios: ^17
             xcode: ^16
-    uses: ./.github/workflows/reusable-workflow.yaml
+    uses: ./.github/workflows/reusable-test-workflow.yaml
     with:
       lib: ${{ matrix.lib }}
       ios: ${{ matrix.ios }}
@@ -40,10 +40,9 @@ jobs:
             xcode: ^16
           - ios: ^17
             xcode: ^16
-    uses: ./.github/workflows/reusable-workflow.yaml
+    uses: ./.github/workflows/reusable-build-workflow.yaml
     with:
-      lib: ${{ matrix.app }}
+      app: ${{ matrix.app }}
       ios: ${{ matrix.ios }}
       xcode: ${{ matrix.xcode }}
-      build_only: true
     secrets: inherit
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -108,12 +108,11 @@ jobs:
             xcode: ^26
           - ios: ^18
             xcode: ^16
-    uses: ./.github/workflows/reusable-workflow.yaml
+    uses: ./.github/workflows/reusable-test-workflow.yaml
     with:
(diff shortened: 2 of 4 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -353,18 +350,18 @@ - (void)enqueueRequest:(SFRestRequest *)request
             // 2xx indicates success.
             if ([SFRestAPI isStatusCodeSuccess:statusCode]) {
                 id dataForDelegate = [strongSelf prepareDataForDelegate:data request:request response:response];
-                if (successBlock) {
-                    successBlock(dataForDelegate, response);
+                if (request.successBlock) {
```
bbirman, line 353: This is the main change, use the callback on the individual request, otherwise the callback on one of the requests is used for all the requests

wmathurin: Was it a recent regression? Or was it always like that?

bbirman: I think it came in with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3862 in 13.1. That thread had a similar issue with replaying a request which I thought was fixed but the sample app doesn't exercise having different callbacks for each request so I think we missed this part of it

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -467,15 +467,7 @@ - (void)replayRequest:(SFRestRequest *)request response:(NSURLResponse *)respons
             } error:^(NSError *refreshError) {
                 __strong typeof(weakSelf) strongSelf = weakSelf;
                 [SFSDKCoreLogger e:[strongSelf class] format:@"Failed to refresh expired session. Error: %@", refreshError];
-                
-                // Call the failure block for the triggering request first
-                if (request.failureBlock) {
```
bbirman: For line 478, I saw an issue where the same request is being removed multiple times along with the callback being executed more than once (if the active queue is at least two requests), I'm thinking it's because `flushPendingRequestQueue` is synchronized to self but the `removeObject` call isn't.

The part I don't know the history on is why the failure block for first request was triggered separately, it looks like the same values are passed to the failure block whether it's called here or through `flushPendingRequestQueue`

bbirman:
- Unit tests are passing locally (core, common, analytics, SmartStore & MobileSync)
- Added a test for scenario where refresh is successful and queue is replayed to make sure success block is only called once per request, also manually tested in RestAPIExplorer
- Manually tested scenario where refresh fails for issue where failure block was being called multiple times

wmathurin (approved):

PR #3961, opened 2025-12-16
Fix CI build issues and archive test results

brandonpage:
Since CI won't run with the latest here, you can [check the nightly run I triggered on my fork](https://github.com/brandonpage/SalesforceMobileSDK-iOS/actions/runs/20256100729) to see the results.

### DIFF
--- a/.github/workflows/reusable-workflow.yaml
+++ b/.github/workflows/reusable-workflow.yaml
@@ -54,14 +54,15 @@ jobs:
           action: ${{ inputs.build_only && 'build' || 'test' }}
           code-coverage: ${{ ! inputs.build_only }}
           verbosity: xcbeautify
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-workflow.yaml
```diff
@@ -63,6 +63,12 @@ jobs:
           show-passed-tests: false
         # xcresulttool currently fails for iOS 26
         if: ${{ ! inputs.build_only && (success() || failure()) && inputs.ios != '^26' }}
```
brandonpage, line 65: @wmathurin @bbirman I looked for a solution to this and even tried a fix someone had sent the author but it didn't work.  I have to parse the test results for my PTOn, so maybe my _next_ PTOn (or just a side project) will be replacing `xcresulttool` with something home grown.

PR #3964, opened 2025-12-17
Add access token revoke request to SDK

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/RestClient.swift
```diff
@@ -329,3 +329,28 @@ extension RestClient {
...
+// MARK: Requests
+extension RestClient {
+    
+    /// Creates a RestRequest object that revokes the access token for a user
+    /// - Parameters:
+    ///   - user: User to revoken the token for, current user will be used if none provided
+    /// - Returns: A `RestRequest` object to revoke the access token if access token is present, nil otherwise
+    @objc
+    public func requestForRevokeAccessToken(user: UserAccount? = UserAccountManager.shared.currentUserAccount) -> RestRequest? {
+        guard let user = user ?? UserAccountManager.shared.currentUserAccount,
```
bbirman, line 342: The currentUser default is here and in the signature so Swift can call it without providing an argument like:  `RestClient.shared.requestForRevokeAccessToken()` 

and obj-c can call with nil:
`[[SFRestAPI sharedInstance] requestForRevokeAccessTokenWithUser:nil]`

PR #3972, opened 2026-01-20
Remove welcome discovery flag

bbirman:


bbirman:
For the test failures, one of them was fixed in `dev` (https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3937) and the other is a known flapper (need to fix but unrelated to welcome)

PR #3975, opened 2026-01-22
Add user agent to token calls

bbirman:
Confirmed with Charles that Android includes the MSDK user agent on the equivalent requests for `accessTokenForApprovalCode` and `accessTokenForRefresh`, updating iOS to match


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
```diff
@@ -145,11 +145,15 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)revokeRefreshToken:(SFOAuthCredentials *)credentials reason:(SFLogoutReason)reason;
 @end
 
+SFSDK_DEPRECATED(13.2, 14.0, "Will be removed.")
 @protocol SFSDKOAuthSessionManaging<NSObject>
```
bbirman, line 149: I think there used to be more in this protocol but I don't see it used anywhere (even for mocking) so I don't think it needs to be public, the one method was used as a helper method for SFNetwork and with this change it's not needed

PR #3980, opened 2026-02-05
Add deprecation markers for unused classes

bbirman:


libs/SalesforceSDKCommon/SalesforceSDKCommon/Classes/Util/SFTestContext.h
```diff
@@ -32,6 +32,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**Helps determine whether we're currently running tests or not.
  */
+__attribute__((deprecated("Deprecated in Salesforce Mobile SDK 13.2 and will be removed in Salesforce Mobile SDK 14.0.")));
```
bbirman, line 35: Didn't use the macro here because it's in core and common can't access it

wmathurin (approved):

PR #3984, opened 2026-02-10
Update default REST API version to v66.0

bbirman:


native/SampleApps/AuthFlowTester/AuthFlowTester/Views/RestApiTestView.swift
```diff
@@ -143,7 +143,7 @@ struct RestApiTestView: View {
         isResultExpanded = false // Start collapsed
         
         do {
-            let request = RestClient.shared.cheapRequest("v63.0")
```
bbirman, line 146: @wmathurin I changed this to pick up the default, let me know if it should stay pinned?

wmathurin: Good catch, it never needed to be pinned actually.

wmathurin (approved):

PR #3990, opened 2026-02-16
Ability to simulate welcome discovery flow (for testing)

wmathurin:
Overview
- SalesforceSDKManager has a new field simulatedDomainDiscoveryResult. When set (DEBUG builds only; setter is a no-op in Release), a callback URL is built from simulatedDomainDiscoveryResult to trigger the domain discovery callback handler. 
- Login options screen provides a way to type in a login host and user name to be used as simulatedDomainDiscoveryResult.
- Added new test suite WelcomeLoginTests
- Also moved existing tests into more test suites (to keep them smaller and make parallel runs faster)

WelcomeDiscoSimulation

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/DevConfig/DiscoveryResultEditor.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/DevConfig/DiscoveryResultEditor.swift
@@ -111,7 +111,7 @@ public struct DiscoveryResultEditor: View {
             }
 
             Button(action: applySimulatedResult) {
-                Text("Use for domain discovery simulation")
+                Text("Save simulated result")
                     .font(.headline)
                     .foregroundColor(.white)
                     .frame(maxWidth: .infinity)
@@ -119,7 +119,7 @@ public struct DiscoveryResultEditor: View {
                     .background(Color.orange)
                     .cornerRadius(8)
             }
-            .accessibilityIdentifier("useForDomainDiscoverySimulationButton")
+            .accessibilityIdentifier("saveSimulatedResultButton")
             .padding(.horizontal)
         }
         .padding(.vertical)
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/DevConfig/DiscoveryResultEditor.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/DevConfig/DiscoveryResultEditor.swift
@@ -1,0 +1,159 @@
+/*
+ DiscoveryResultEditor.swift
+ SalesforceSDKCore
+
+ Copyright (c) 2026-present, salesforce.com, inc. All rights reserved.
+
+ Redistribution and use of this software in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this list of conditions
+ and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice, this list of
+ conditions and the following disclaimer in the documentation and/or other materials provided
+ with the distribution.
+ * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+ endorse or promote products derived from this software without specific prior written
+ permission of salesforce.com, inc.
+
+ THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+ IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+ FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+ CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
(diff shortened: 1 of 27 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/DevConfig/DiscoveryResultEditor.swift
```diff
@@ -0,0 +1,164 @@
...
+    public var body: some View {
+        VStack(alignment: .leading, spacing: 12) {
+            HStack {
+                Button(action: {
+                    withAnimation {
+                        isExpanded.toggle()
+                    }
+                }) {
+                    HStack {
+                        Text("Simulate Domain Discovery")
```
bbirman: Not high priority but wondering if we eventually want to make these localizable since the dev menu has localizable strings?

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift
```diff
@@ -91,8 +92,10 @@ class BaseAuthFlowTesterTest: XCTestCase {
         dynamicScopeSelection: ScopeSelection = .empty,
         useWebServerFlow: Bool = true,
         useHybridFlow: Bool = true,
+        useWelcomeDiscovery: Bool = false,
     ) {
         let userConfig = getUser(loginHost: loginHost, user: user)
+        let hostConfig = try! testConfig.getLoginHost(loginHost)
```
bbirman: could the containing method throw instead of the force-try so the error could bubble up to the calling test?

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/WelcomeLoginTests.swift
```diff
@@ -0,0 +1,71 @@
...
+ DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+ WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import XCTest
+
+/// Tests for welcome discovery login flows.
+class WelcomeLoginTests: BaseAuthFlowTester {
```
bbirman, line 31: Do you think we could retire some of the unit tests in `DomainDiscoveryCoordinatorTests`? A handful of them have been flapping and the new tests seem like they exercise more of the flow. I haven't searched through old test results for all of the flappers, but `DomainDiscoveryCoordinatorTests.testCallbackSuccess` was one of them

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/DevConfig/DiscoveryResultEditor.swift
wmathurin: Done with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3990/commits/7c253f6ac3ed22b119511988c7710eddaee23734

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift
wmathurin: Done with https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3990/commits/a1c3d8015a64bd507bbbb498e91fd95a190a18be

bbirman (approved):

brandonpage:
@wmathurin Is there a reason the "Use for domain discovery simulation" button does not set the current server to Welcome before returning to the login page?  Not a big deal, just seems like it could be convenient.

wmathurin:
> @wmathurin Is there a reason the "Use for domain discovery simulation" button does not set the current server to Welcome before returning to the login page? Not a big deal, just seems like it could be convenient.

It would have to be added to the login server list if not present then the authentication restart that happens after exiting the login options screen would need to re-examine what login server is selected (<- that is probably already the case but I haven't checked). I wanted to keep it simple and only change the login server the normal way so that we don't end up following a path through the code that diverges too much from real-life situations.

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/WelcomeLoginTests.swift
wmathurin: I thought mocking the navigation action instead of the web view (see [489a1a8](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3990/commits/489a1a852e18691a6573fb7f7a788700e8656c9e)) would solve the flapping but it hasn't . I'll dig deeper today.

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/PageObjects/LoginPageObject.swift
```diff
@@ -227,16 +263,16 @@ class LoginPageObject {
...
 
-    private func useDynamicConfigButton() -> XCUIElement {
-        return app.buttons["Use dynamic config"]
+    private func saveDynamicConfigButton() -> XCUIElement {
+        return app.buttons["Save dynamic config"]
     }
     
     /// Returns the import button for either the static or dynamic configuration section.
-    /// The BootConfigPickerView has two BootConfigEditor sections - the first for static config, the second for dynamic config.
+    /// The LoginOptionsView has two BootConfigEditor sections - the first for static config, the second for dynamic config.
```
brandonpage: NIT: I would maybe consider creating a separate PageObject for LoginOptions.  

On Android I am utilizing the LoginPageObject from the UITests repo and have created a LoginOptionsPageObject in the main repo.  Kotlin makes this very clean since I can define extension functions like `fun LoginPageObject.openLoginOptions()` in the main repo without needing to modify the class in UITests.

brandonpage (approved):

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/PageObjects/LoginPageObject.swift
wmathurin: Good idea. Done here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/3990/commits/d028bf1eedc16a712b1500069ea488bb782fb193

wmathurin:
There are some unrelated flappers. I'll go ahead and merge.

PR #3991, opened 2026-02-20
Simpler UI test config

wmathurin:
We were only really using half the CAs / ECAs / beacon apps from ui_test_config.json. So simplifying the test config and renamed tests etc accordingly.


wmathurin:
@brandonpage what do you think?

brandonpage (approved):
LGTM

wmathurin:
The tests that pr.yaml is trying to run have been renamed - it should work once the pr is merged.

PR #3996, opened 2026-02-27
Re-register for push notifications when migrating refresh token

wmathurin:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -380,7 +380,6 @@ - (void)notifyDelegateOfSuccess:(SFOAuthInfo *)authInfo
     if ([self.delegate respondsToSelector:@selector(oauthCoordinatorDidAuthenticate:authInfo:)]) {
         [self.delegate oauthCoordinatorDidAuthenticate:self authInfo:authInfo];
     }
-    _authInfo = nil;
```
wmathurin, line 383: We should NOT clear the authInfo there - because the call line 381 will do some async work (like retrieve identity data) and after that it will make decision based on the authInfo type.
We got lucky with post-login notifications because when no authInfo is found, a new one is created and its type is "unknown" and the notifyLoginCompletion sends notifications in all cases.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/PushNotification/PushNotificationManager.swift
```diff
@@ -452,6 +456,12 @@ private extension PushNotificationManager {
             object: nil,
             queue: .main
         ) { [weak self] in self?.onAppWillEnterForeground($0) }
+        
+        migrateRefreshTokenObserver = NotificationCenter.default.addObserver(
```
wmathurin, line 460: Observing "refresh token migration" notification and re-registering for notifications if when observed.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/PushNotificationManagerTests.swift
```diff
@@ -528,8 +528,87 @@ class PushNotificationManagerTests: XCTestCase {
         XCTAssertNotEqual(error1, error3)
     }
 
+    // MARK: - Refresh Token Migration Tests
+
+    func testOnUserMigratedRefreshToken_WithDeviceToken_TriggersRegistration() {
```
wmathurin, line 533: Test that makes sure that a notification indicating that the refresh token was migration cause the registration rest request call to go out.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerTests.m
```diff
@@ -803,4 +809,34 @@ - (void)testMigrateRefreshToken {
     [self.uam.authSessions removeObject:sceneId];
 }
 
+- (void)testNotifyLoginCompletion_PostsMigrateRefreshTokenNotification {
```
wmathurin, line 812: Test that makes sure that notifyLoginCompletion sends a notification when the auth info is of type SFOAuthTypeRefreshTokenMigration.

bbirman (approved):

PR #4001, opened 2026-03-05
Fixing retry race condition

bbirman:
Still testing and would need to test against the the fork as well but want to get a CI run going for 13.2


wmathurin (approved):
I guess the scenario that caused the issue is where multiple requests are issued, one gets a 401 triggering a refresh and the refresh actually completes before the dataResponseBlock of some of the other requests already sent have run (e.g. if they are slow for whatever reasons e.g. if they have large payloads). 

When refresh completes, pending requests are replayed, causing new data tasks to be created for each one. Not cancelling or preventing the "old" dataResponseBlock from running caused a re-refresh and a unnecessary replay of the request and as a result a double callback leading to a double continuation resume error.

bbirman:
The merge commit had the wrong email from switching to a new laptop, updated so the CLA bot message shouldn't be an issue

PR #4003, opened 2026-03-06
SFRestAPI stress test

bbirman:


wmathurin (commented):
Are you able to repro the race condition from last week with the new test?

bbirman:
I think so 👍 it fails without https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4001 and passes with it. 

One issue I'd need to fix is that the Swift tests run in parallel and this one should be isolated so it doesn't interfere with other tests

wmathurin:
> I think so 👍 it fails without #4001 and passes with it.
> 
> One issue I'd need to fix is that the Swift tests run in parallel and this one should be isolated so it doesn't interfere with other tests

Maybe with another [test plan](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/native/SampleApps/AuthFlowTester/AuthFlowTester.xctestplan) ??

PR #4006, opened 2026-03-22
Add CLAUDE.md

bbirman:
Thinking this we'll keep updating as we go, so far I took the original file and then:
- trimmed references to other repos since that will be at the workspace level
- updated build commands
- tweaked some definitions and the framework hierarchy

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,6 +2,12 @@
 
 ---
 
+## About This Project
+
+The Salesforce Mobile SDK is a public, open-source SDK that enables developers to build mobile apps that integrate with the Salesforce platform. It is consumed by ISVs, SI partners, and internal Salesforce teams.
+
+**Key constraint**: This is a **public SDK**. Every change is visible to external developers. Backward compatibility, deprecation cycles, and semver discipline are non-negotiable.
+
 ## iOS Library Architecture
 
 Workspace: `SalesforceMobileSDK.xcworkspace`
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -2,13 +2,6 @@
 
 ---
 
-## About This Project
-
-The Salesforce Mobile SDK is a public, open-source SDK that enables developers to build native, hybrid, and React Native mobile apps that integrate with the Salesforce platform. It is consumed by ISVs, SI partners, and internal Salesforce teams.
-
-**Key constraint**: This is a **public SDK**. Every change is visible to external developers. Backward compatibility, deprecation cycles, and semver discipline are non-negotiable.
-
-
 ## iOS Library Architecture
 
 Workspace: `SalesforceMobileSDK.xcworkspace`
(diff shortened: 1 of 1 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,241 @@
+# CLAUDE.md — Salesforce Mobile SDK
+
+---
+
+## About This Project
+
+The Salesforce Mobile SDK is a public, open-source SDK that enables developers to build native, hybrid, and React Native mobile apps that integrate with the Salesforce platform. It is consumed by ISVs, SI partners, and internal Salesforce teams.
```
wmathurin: Maybe that's a good paragraph to have in the workspace claude.md. For the iOS, however, maybe we shouldn't mention hybrid and react native (since they live elsewhere)?

CLAUDE.md
```diff
@@ -0,0 +1,241 @@
...
+## Code Review Checklist
+
+When reviewing PRs (or acting as an AI reviewer), verify:
+
+- [ ] **Backward compatibility**: Does this change break any existing public API? If an API is being changed, is the old version properly deprecated with a migration path?
+- [ ] **Both platforms considered**: If this is a feature/fix that applies to both iOS and Android, is there a corresponding PR for the other platform?
+- [ ] **Tests included**: New/changed behavior has corresponding test coverage in the correct test target.
+- [ ] **No regressions**: Full test suite for the affected library passes.
+- [ ] **Multi-user**: Does this work correctly with multiple logged-in accounts?
+- [ ] **Encryption**: If touching SmartStore, does this work across all SQLCipher editions (Community, Commercial, Enterprise, FIPS)?
```
wmathurin: Not sure how Claude can do that (testing across editions)

CLAUDE.md
```diff
@@ -0,0 +1,241 @@
...
+- [ ] **Sample apps**: If changing public API, are sample and template apps (RestAPIExplorer, MobileSyncExplorerSwift, etc.) updated?
+
+---
+
+## Agent Behavior Guidelines
+
+These rules apply when Claude Code operates as an agent in these repos:
+
+### Do
+- Always run the test suite for the affected library before creating a commit.
```
wmathurin: Should we ask it to check that there is a test_credentials.json in shared/test (or a ui_test_config.json for ui tests) before starting? Many tests won't run without it.

CLAUDE.md
bbirman: Good catch, removing

CLAUDE.md
```diff
@@ -0,0 +1,232 @@
+# CLAUDE.md — Salesforce Mobile SDK
+
+---
+
```
wmathurin, line 4: Maybe you should bring back what you had earlier (without the reference to hybrid and react native)

## About This Project

The Salesforce Mobile SDK is a public, open-source SDK that enables developers to build mobile apps that integrate with the Salesforce platform. It is consumed by ISVs, SI partners, and internal Salesforce teams.

**Key constraint**: This is a **public SDK**. Every change is visible to external developers. Backward compatibility, deprecation cycles, and semver discipline are non-negotiable.

wmathurin (approved):

PR #4007, opened 2026-03-23
Fix: Choose Connection Screen Crash on Swipe Dismissal 

sfdctaka:
**What Changed**

**Before**: The code reused the same cached loginHostListViewController instance via the lazy property getter, causing a race condition when the view controller was still attached to a previous navigation controller during swipe dismissal.

**After**: Each time the "Choose Connection" screen is opened, a fresh SFSDKLoginHostListViewController instance is created, eliminating the possibility of the view controller being embedded in multiple navigation controllers simultaneously.

**Why This Fixes the Crash**

- No more reuse: Each presentation gets a brand new VC instance
- Clean state: No leftover navigation controller relationships
- Race-safe: Even rapid swipe gestures won't trigger the "nested navigation controllers" error
- Simple: No complex state management needed

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -356,6 +356,7 @@
 - (SFSDKLoginHostListViewController *)createLoginHostListViewController {
     SFSDKLoginHostListViewController *loginHostListViewController = [[SFSDKLoginHostListViewController alloc] initWithStyle:UITableViewStylePlain];
     loginHostListViewController.config = self.config;
+    loginHostListViewController.delegate = self;
     return loginHostListViewController;
 }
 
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -451,9 +451,13 @@ - (void)handleLoginHostSelectedAction:(SFSDKLoginHost *)newLoginHost {
 #pragma mark - Login Host
 
 - (void)showHostListView {
-    SFSDKNavigationController *navController = [[SFSDKNavigationController alloc] initWithRootViewController:self.loginHostListViewController];
+    // Create a FRESH instance each time instead of reusing the cached one
+    SFSDKLoginHostListViewController *hostListVC = [self createLoginHostListViewController];
+    hostListVC.delegate = self;
```
bbirman: Could setting the delegate be moved into `createLoginHostListViewController` too? and then could delete the other line here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/4b84a93c78a4e3949f6d3272ffcfaef4f0a37ab4/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m#L403

bbirman (approved):

wmathurin (approved):

PR #4008, opened 2026-03-24
@W-21717256 Adding completing callback to resetSessionCookie()

wmathurin:
That will allow us to fix the race condition causing infinite login loop in hybrid auth (where we call resetSessionCookie before setting cookies)

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
@@ -53,13 +53,21 @@ public class SFSDKWebViewStateManager: NSObject {
     @objc
     @MainActor
     public static func resetSessionCookie() {
+        resetSessionCookie(completion: {})
+    }
+
+    @objc
+    @MainActor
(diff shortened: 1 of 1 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/WebViewStateManager.swift
```diff
@@ -53,13 +53,21 @@ public class SFSDKWebViewStateManager: NSObject {
     @objc
     @MainActor
     public static func resetSessionCookie() {
```
wmathurin, line 55: In 13.0.0 and earlier this method was implemented as follow
```objective-c
+ (void)resetSessionCookie {
    if (_sessionCookieManagementDisabled) {
        [SFSDKCoreLogger d:self format:@"[%@ %@]: Cookie Management disabled. Will do nothing.", NSStringFromClass(self), NSStringFromSelector(_cmd)];
        return;
    }
    [self removeWKWebViewCookies];
}
```
As a result, there was no race conditions calling `resetSessionCookie` and then calling `setCookies`.

bbirman (approved):

PR #4009, opened 2026-03-26
Fix advanced auth login keyboard interaction

wmathurin:
- Fixes `BeaconAdvancedAuthLoginTests` failure in ASWebAuthenticationSession
- Password field was hidden by keyboard, making it untappable
- Now fills password first, then username (visible above keyboard), then submits with Return key


brandonpage (approved):
LGTM.  If it ends up being less consistent than you would like you can always [hide the keyboard](https://github.com/forcedotcom/SalesforceMobileSDK-UITests/blob/e805caa0cbe29e522447bd136069e2817292e892/iOS/PageObjects/LoginPageObjects/LoginPageObject.swift#L77).

bbirman (approved):

PR #4012, opened 2026-03-30
Expand README / Update documentation to reference External Client App

wmathurin:
Significantly expands the iOS SDK README with comprehensive documentation to help developers understand and use the SDK more effectively.     

Also updated CLAUDE.md to reference External Client App.


readme.md
```diff
@@ -38,6 +38,304 @@ See [release notes](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/relea
...
+
+## Usage
+
+### Authentication (OAuth2)
+
+```swift
+import SalesforceSDKCore
+
+// Configure SDK in AppDelegate
+SalesforceManager.shared.configure { (config: SalesforceSDKManager.Configuration) in
```
bbirman: What's the `configure` method?

wmathurin: Hmm - AFAIK it does not exist - not a bad idea though ;-)
Will send fix

readme.md
```diff
@@ -38,6 +38,304 @@ See [release notes](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/relea
 ### Native Applications
 The Salesforce Mobile SDK provides the essential libraries for quickly building native mobile apps that interact with the Salesforce cloud platform. The OAuth2 library abstracts away the complexity of securely storing the refresh token or fetching a new session ID when it expires. The SDK also provides wrappers for the Salesforce REST API that you can use from both Swift and Objective-C.
 
+## Libraries
```
bbirman, line 41: This part and the test/build commands section seems similar to what's in claude.md, do you think it could be removed there and a part of the instructions there would be to consume the readme? Or is it better to keep it in claude.md too?

readme.md
```diff
@@ -38,6 +38,304 @@ See [release notes](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/relea
...
+### MobileSync (Data Synchronization)
+
+```swift
+import MobileSync
+
+// Get sync manager
+let syncManager = SyncManager.shared(store: store)
+
+// Sync down from Salesforce
+let target = SoqlSyncDownTarget(
```
bbirman: I don't think this exact one exists either, I think it would be:
```
let target = SoqlSyncDownTarget.newSyncTarget("SELECT Id, Name, Industry FROM Account WHERE LastModifiedDate > {LastModifiedDate}")
```

readme.md
wmathurin: I took out code examples (for now??) to keep the readme short

readme.md
wmathurin: Good point - let me see if we can reduce redundancy.

readme.md
```diff
@@ -38,6 +38,42 @@ See [release notes](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/relea
...
+
+# Create app from template
+forceios create --appname MyApp --packagename com.mycompany.myapp --organization "My Company"
+```
+
+See templates for complete usage examples.
+
+### Building from Source
+
+```bash
```
bbirman: These look like the same steps from `## Setting up the repo`?

bbirman (approved):

PR #4016, opened 2026-04-04
Bump to iOS 18 and add corresponding Claude skill

bbirman:


.github/workflows/nightly.yaml
```diff
@@ -12,14 +12,12 @@ jobs:
       fail-fast: false
       matrix:
         lib: [SalesforceSDKCommon, SalesforceAnalytics, SalesforceSDKCore, SmartStore, MobileSync]
-        ios: [^26, ^18, ^17]
```
wmathurin, line 15: github action updates should probably be mention somewhere in the skill??

bbirman: I had it update the skill based on the CI commit and then tweaked it a little: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4016/changes/54fb2f437d9ad13fd3091e54f6b020477435a219

wmathurin (approved):

PR #4026, opened 2026-05-01
Add public API to trigger Login for Admin flow.

brandonpage:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.h
```diff
@@ -625,6 +626,19 @@ Use this method to stop/clear any authentication which is has already been start
...
+ login view controller. Forces browser-based (advanced) authentication via ASWebAuthenticationSession,
+ regardless of org configuration, to support phishing-resistant MFA.
+
+ This performs the same action as selecting "Login for Admin" from the login screen's settings
+ menu. Apps that hide the settings menu (e.g. by setting `showSettingsIcon = NO`) can call this
+ to expose the feature from their own UI.
+ @param loginViewController The login view controller whose scene's active auth session should
+ switch to "Login for Admin". Its window's scene is used to locate the session.
+ */
+- (void)loginViewControllerDidSelectLoginForAdmin:(SFLoginViewController *)loginViewController NS_SWIFT_NAME(loginViewControllerDidSelectLoginForAdmin(_:)) SFSDK_DEPRECATED(13.2.1, 14.0, "Will be removed in 14.0 when a permanent solution if provided.");
```
wmathurin: Typo: "when a permanent solution if provided" should be "when a permanent solution is provided"

wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.h
bbirman: Where would the app get the reference to the login view controller from? the window?

brandonpage: @bbirman Good callout.  As I see it they have two options:
1. Retain a reference from `UserAccountManager.shared.loginViewControllerConfig.loginViewControllerCreationBlock`.
2. Walk the view hierarchy. 

I assume option 1 is preferred?  Should I add that to the comment?

bbirman: Were you able to test with option 1? I think the reference would be to the block, not the view controller created by the block?

Otherwise I was wondering if option 2 could use SFSDKWindowManager's authWindow's topViewController, or if there's a way to get around the app providing the reference, but do we know if anyone is hiding the settings icon? Might not be worth it to invest more if not

brandonpage: I was thinking something like this:
```
UserAccountManager.shared.loginViewControllerConfig.loginViewControllerCreationBlock = { [weak self] in
  let vc = SalesforceLoginViewController()
  self?.currentLoginVC = vc   // retain weakly for later
  return vc
}
```

bbirman: ah gotchya! no preference for me, whatever you think looks cleanest

bbirman (approved):

PR #4040, opened 2026-05-19
Hide navigation bar on native login view presentation

sfdctaka:
## Summary
- Hide the `SFSDKNavigationController` navigation bar when presenting custom native login view controllers in `presentLoginView:`
- Previously, the wrapping nav controller's bar (styled with Salesforce blue) was visible by default. `SFLoginViewController` hides it internally, but custom native login VCs had no such handling — resulting in an unwanted nav bar appearing on top of the app's login UI, especially on re-presentation after fallback-to-web-auth → back
- Add unit test verifying the navigation bar is hidden when a native login VC is presented

## Test plan
- [x] New unit test `testNativeLoginPresentationHidesNavigationBar` passes
- [x] Test correctly fails when the fix is reverted
- [x] All existing `NativeLoginManagerTests` pass
- [x] Manual verification: use an app with custom native login, trigger fallback to web auth, press back — confirm no blue nav bar appears

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/NativeLoginManagerTests.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/NativeLoginManagerTests.swift
@@ -178,3 +178,14 @@ final class NativeLoginManagerTests: XCTestCase {
         return user
     }
 }
+
+// MARK: - Test Helpers
+
+private class MockPresenterViewController: UIViewController {
+    var onPresent: ((UIViewController) -> Void)?
+
+    override func present(_ viewControllerToPresent: UIViewController, animated flag: Bool, completion: (() -> Void)? = nil) {
+        onPresent?(viewControllerToPresent)
+        completion?()
+    }
+}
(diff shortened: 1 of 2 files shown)


libs/SalesforceSDKCore/SalesforceSDKCoreTests/NativeLoginManagerTests.swift
```diff
@@ -125,6 +125,47 @@ final class NativeLoginManagerTests: XCTestCase {
...
+        // Native login is already configured via `nativeLoginManager` property.
+        XCTAssertTrue(accountManager.nativeLoginEnabled, "Native login should be enabled.")
+        accountManager.shouldFallbackToWebAuthentication = false
+
+        let expectation = self.expectation(description: "Auth view presented")
+        var presentedController: UIViewController?
+
+        // Set up auth window with a mock VC that captures presentViewController: calls.
+        // This lets the real presentLoginView: execute end-to-end.
+        let mockPresenter = MockPresenterViewController()
```
bbirman, line 145: Is there a file missing from the commit? It looks like building the tests fails with `Cannot find 'MockPresenterViewController' in scope`

sfdctaka: Added back test helper which I manually mistakenly removed when fixing whitespace vs tab after Claude had formatted the code.

bbirman (approved):

wmathurin (approved):

PR #4041, opened 2026-05-21
Add option to automatically present biometric opt-in and lock 

sfdctaka:
## Summary

- Add `automaticPresentation` property to `BiometricAuthenticationManager` protocol that, when enabled:
  - Automatically presents the biometric opt-in dialog after login (if the user hasn't opted in yet)
  - Automatically triggers biometric unlock (Face ID/Touch ID) when the app locks (if the user has opted in)
- Defaults to `false` to preserve existing behavior — apps must explicitly opt in
- Ported @bbirman's changes

## Changes

- **BiometricAuthenticationManager.swift**: Added `automaticPresentation` property to the public protocol
- **BiometricAuthenticationManagerInternal.swift**: Implemented stored property (default `false`) and added auto-present logic at the end of `lock()` that iterates connected scenes and calls `presentBiometric(scene:)` when conditions are met
- **SFUserAccountManager.m**: After storing biometric policy on login, checks if opt-in dialog should be auto-presented (`!hasBiometricOptedIn && automaticPresentation`)
- **BiometricAuthenticationManagerTests.swift**: Added 7 tests covering all combinations of `automaticPresentation` × `hasBiometricOptedIn` for both lock and opt-in dialog scenarios
- **URLSessionTask+RetryPolicyTests.swift**: Added `automaticPresentation` to mock for protocol conformance

### Scenario A: Auto opt-in dialog after login

| # | automaticPresentation | hasBiometricOptedIn | Auth type | Expected |
|---|---|---|---|---|
| A1 | true | false | Fresh login | Opt-in dialog presented |
| A2 | true | true | Fresh login | No dialog (already opted in) |
| A3 | false | false | Fresh login | No dialog |
| A4 | true | false | Token refresh | No dialog (only on fresh login) |

### Scenario B: Auto biometric prompt on lock

| # | automaticPresentation | hasBiometricOptedIn | Lock triggered | Expected |
|---|---|---|---|---|
| B1 | true | true | Yes | presentBiometric called for each scene |
| B2 | true | false | Yes | No auto-present (not opted in) |
| B3 | false | true | Yes | No auto-present (feature off) |
| B4 | false | false | Yes | No auto-present |

### Scenario C: Interaction with native login

| # | Login type | automaticPresentation | Action | Expected |
|---|---|---|---|---|
| C1 | Native login | true | Lock then auto biometric | presentBiometric works correctly (auth window dismisses, native login VC doesn't interfere) |
| C2 | Native login | true | Fresh login then opt-in dialog | Dialog presented on correct VC (mainWindow, not native login VC) |
| C3 | Webview login | true | Lock then auto biometric | Works as before |
| C4 | Native login | true | Lock then biometric fails then fallback | User falls back to native login screen correctly |

### Scenario D: Existing behavior (regression tests)

| # | Scenario | Expected |
|---|---|---|
| D1 | `automaticPresentation = false` (default) | All existing behavior unchanged |
| D2 | Lock/unlock without `automaticPresentation` | User must manually tap biometric button |
| D3 | Opt-in without `automaticPresentation` | App must explicitly call `presentOptInDialog` |

## Test plan

- [x] Unit tests: Scenarios A1–A3, B1–B4, D1–D3 (7 new tests in BiometricAuthenticationManagerTests)
- [x] Manual test (C3): Standard template app with ECA + biometric custom attributes → opt-in dialog appears after first login
- [x] Manual test (B1): After opting in, background app past timeout → Face ID auto-presents on foreground
- [x] Manual test (D1): `automaticPresentation = false` (default) → no change in existing behavior
- [x] Manual test (C1–C4): Native login template + configured org with customer attributes: ENABLE_BIOMETRIC_AUTHENTICATION and BIOMETRIC_AUTHENTICATION_TIMEOUT

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
@@ -45,7 +45,9 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     }
     
     public var locked = false
-    
+
+    public var automaticPresentation = false
+
     internal var backgroundTimestamp: Double = 0
     // This is a local var so it can be stubbed for tests
     internal var laContext = LAContext()
@@ -134,7 +136,7 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     public func lock() {
         locked = true
         NotificationCenter.default.post(name: Notification.Name(rawValue: kSFBiometricAuthenticationFlowWillBegin), object: nil)
-        
+
         // Open the Login Screen
         _ = UserAccountManager.shared.login { result in
             switch result {
@@ -144,6 +146,12 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
                 break
             case .failure(let error):
                 SFSDKCoreLogger.e(BiometricAuthenticationManagerInternal.self, message: "Biometric authentication failed: \(error)")
+            }
+        }
+
+        if hasBiometricOptedIn() && automaticPresentation {
+            SFApplicationHelper.sharedApplication()?.connectedScenes.forEach() { scene in
+                presentBiometric(scene: scene)
             }
         }
     }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
@@ -46,7 +46,7 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     
     public var locked = false
 
-    public var automaticPresentation = false
+    public var automaticPresentation = true
 
     internal var backgroundTimestamp: Double = 0
     // This is a local var so it can be stubbed for tests
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
@@ -136,7 +136,7 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     public func lock() {
         locked = true
         NotificationCenter.default.post(name: Notification.Name(rawValue: kSFBiometricAuthenticationFlowWillBegin), object: nil)
-
+        
         // Open the Login Screen
         _ = UserAccountManager.shared.login { result in
             switch result {
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManager.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManager.swift
@@ -36,6 +36,11 @@ public protocol BiometricAuthenticationManager {
     /// If the device is currently locked.  Authenticated rest requests will fail if true.
     var locked: Bool { get }
 
+    /// If enabled, the SDK will automatically present the biometric opt-in dialog after login
+    /// (if the user has not yet opted in) and automatically trigger biometric unlock when the
+    /// app is locked (if the user has opted in). Defaults to false.
+    var automaticPresentation: Bool { get set }
+
     /// Locks the device immediately.  Authenticated rest requests will fail until the user unlocks the app.
     func lock()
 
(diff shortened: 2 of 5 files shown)


wmathurin (approved):
LGTM

bbirman:
Just double checking, there wasn't any issue with having the new logic on top of the [logic here in SFLoginViewController](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/a2d6db8ae0c42f3c48488bb771220f87a10597e4/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m#L95-L97)? Do they both get called but one ends up being a no-op?

sfdctaka:
@bbirman good catch. Both can fire:

The new logic in lock() fires first (synchronously after lock), before SFLoginViewController is presented (since login is async).
The existing logic in SFLoginViewController.viewDidLoad fires once the login VC loads.
`presentBiometric(scene:)` creates a fresh LAContext each time, so if the first call already resolved (or is in-progress), the second should be a no-op from the user's perspective — the system won't show two biometric prompts.

That said, the existing logic at line 95 (`bioAuthManager.locked && bioAuthManager.hasBiometricOptedIn`) already auto-presents biometric without checking automaticPresentation. So the existing behavior in SFLoginViewController was already doing automatic presentation for the standard login flow — the new code only adds it for the native login path (where SFLoginViewController isn't used). Should we gate the existing line 95 behind automaticPresentation as well for consistency?

bbirman:
> @bbirman good catch. Both can fire:
> 
> The new logic in lock() fires first (synchronously after lock), before SFLoginViewController is presented (since login is async). The existing logic in SFLoginViewController.viewDidLoad fires once the login VC loads. `presentBiometric(scene:)` creates a fresh LAContext each time, so if the first call already resolved (or is in-progress), the second should be a no-op from the user's perspective — the system won't show two biometric prompts.
> 
> That said, the existing logic at line 95 (`bioAuthManager.locked && bioAuthManager.hasBiometricOptedIn`) already auto-presents biometric without checking automaticPresentation. So the existing behavior in SFLoginViewController was already doing automatic presentation for the standard login flow — the new code only adds it for the native login path (where SFLoginViewController isn't used). Should we gate the existing line 95 behind automaticPresentation as well for consistency?

That sounds good! The only downside is that whatever the default is would be a change for one of the flows but 14.0 would be a good window for it, @brandonpage  @wmathurin any thoughts on making automatic presentation default to `true` so there's less enablement for biometric by default?

sfdctaka:
@bbirman @wmathurin @brandonpage Is it okay to automatically present the biometric the first time?

brandonpage:
> @bbirman @wmathurin @brandonpage Is it okay to automatically present the biometric the first time?

Yes, and I would honestly prefer if it was opt-out.  That was my original intention, but the client who we built this for (and never used it) wanted to show their own dialog.  UI styling are very consistent on iOS so I really don't see an issue with showing our prompt.  Perhaps on Android we may want to allow the app to override the dialog.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
```diff
@@ -45,7 +45,9 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     }
     
     public var locked = false
-    
+
+    public var automaticPresentation = false
```
brandonpage: Since this would go out in a major release, I see no problem with defaulting to `true`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
```diff
@@ -146,6 +148,12 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
                 SFSDKCoreLogger.e(BiometricAuthenticationManagerInternal.self, message: "Biometric authentication failed: \(error)")
             }
         }
+
+        if hasBiometricOptedIn() && automaticPresentation {
+            SFApplicationHelper.sharedApplication()?.connectedScenes.forEach() { scene in
+                presentBiometric(scene: scene)
+            }
+        }
```
brandonpage, line 156: Why is this needed?  The Login view controller should automatically present biometric upon lock if the user is opted in?

brandonpage (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
sfdctaka: Image

This is for the native login path. When useNativeLogin is enabled, SFLoginViewController isn't used — the app's custom VC is presented instead, so the existing auto-present at line 95 of SFLoginViewController.m never fires. This ensures native login apps get the same auto-present behavior when automaticPresentation is enabled.

That said, for the standard (webview) login path, you're right — SFLoginViewController already handles it. Should we gate line 95 behind automaticPresentation too for consistency, or leave it as-is since it's existing behavior?

bbirman:
Should the code in SFLoginViewController be removed now that there's presentation logic for both paths in BiometricAuthenticationManagerInternal?

sfdctaka:
> Should the code in SFLoginViewController be removed now that there's presentation logic for both paths in BiometricAuthenticationManagerInternal?

@bbirman  The SFLoginViewController code predates this and doesn't check automaticPresentation, so they're not fully equivalent. Consolidating them would be a good follow-up (we'd want SFLoginViewController to respect automaticPresentation too), but it's a behavioral change for existing webview-login apps that should be its own PR. Want me to file a story for it?

bbirman:
Yes, that was the gap identified from the original proof of concept work on the internal branch, now that "automaticPresentation" defaults to true there shouldn't be a behavioral difference to the consuming app/user

sfdctaka:
@bbirman I'll push another PR.

PR #4042, opened 2026-05-22
Fix thread-safety race in SFUserAccount encodeWithCoder

brandonpage:
## Summary
- `encodeWithCoder:` accessed ivars (`_accessScopes`, `_idData`, `_customData`) directly without synchronizing on the concurrent queue, while their setters use `dispatch_barrier_async`. This caused a race where encoding immediately after setting these properties could capture stale (nil/empty) values.
- Wraps the encoder calls in `dispatch_sync` on `_syncQueue` so pending barrier writes complete before the ivars are read.

## Test plan
- [ ] Verify `testUserAccountEncoding` passes on iOS 26 (Xcode 26) nightly CI
- [ ] Verify no deadlocks when archiving `SFUserAccount` from different threads
- [ ] Run full SalesforceSDKCore test suite


sfdctaka (approved):
LGTM!

wmathurin (approved):

bbirman (approved):

PR #4051, opened 2026-06-03
Add DPoP proof JWTs to iOS token exchanges

sfdctaka:
Adds opt-in DPoP sender-constrained proof JWTs to all Phase 1 OAuth token-endpoint requests on iOS: auth-code exchange, refresh-token, and the hybrid JWT-bearer swap. Behind a new `SalesforceSDKManager.useDPoP` flag (defaults to `NO`), so existing integrators are byte-for-byteunaffected.

When enabled, every outbound request to `/services/oauth2/token` carries a fresh `DPoP` header — an ES256-signed JWS whose payload binds the request to the URL, method, time, a unique JTI, and (when known) the server-issued
`DPoP-Nonce`. The signing key is a per-account P-256 keypair generated on first use, stored in the iOS Keychain, and
backed by the Secure Enclave when available.

- **What's added (iOS only — Android out of scope per spec A-12):**
  - `SFSDKDPoPKeyStore` — per-account P-256 keypair lifecycle (Secure Enclave with software-backed fallback).
  - `SFSDKDPoPProofBuilder` — JWS construction (ES256, JWK header, canonical `htu`,
`iat`/`jti`/optional `nonce`).
  - `SFSDKDPoPNonceCache` — process-lifetime `(htu, scope)`-keyed nonce cache fed by both reactive 400/401 challenges and proactive 200 OK rotations.
  - `SFSDKDPoPRequestDecorator` — gates on `useDPoP`, attaches `DPoP` header, harvests responses, and detects nonce challenges.
  - `SFSDKCryptoUtils` — two new helpers: `signDataES256:withKeyRef:`and `jwkExportFromPublicKeyRef:`.

- **Wiring:**
  - `SFSDKOAuth2.prepareBasicRequest:` attaches the proof for both auth-code and refresh paths.
  - A new `sendTokenEndpointRequest:` helper centralizes one-shot retry on the `use_dpop_nonce` /
401-with-`DPoP-Nonce` challenge.
  - `SFOAuthCoordinator.swapJWTWithCompletionHandler:` gets the same attach + retry inline (this path bypasses
`SFSDKOAuth2`/`SFNetwork`).
  - `SFUserAccountManager.postPushUnregistration:logoutReason:` clears the keystore + nonce cache for the logged-out account.

- **Public API surface:** Just one new property — `SalesforceSDKManager.useDPoP` (Swift: `usesDPoP`). All other types
are `@objc`-exposed for ObjC interop but used internally by the SDK.

 ## Design Notes

- **Per-account scope:** The keystore and nonce cache are scoped on `SFOAuthCredentials.identifier` rather than
`<orgId>_<userId>`. The auth-code exchange runs *before* an `SFUserAccount` exists, so org/user IDs aren't available at first proof. `credentials.identifier` is stable from credential construction and is what the SDK already uses to namespace per-account keychain entries.
- **Backward compatibility:** With `useDPoP=NO` (the default), `prepareBasicRequest:` produces a byte-identical
 request to before — no `DPoP` header, same URL, method, content-type, and cookie policy. Asserted by a regression test.
- **Secure Enclave:** Used when available; falls back to software-backed keychain storage on policy-locked devices, with one retry before surfacing failure.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/SFSDKDPoPKeyStore.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/SFSDKDPoPKeyStore.swift
@@ -1,0 +1,145 @@
+//
+//  SFSDKDPoPKeyStore.swift
+//  SalesforceSDKCore
+//
+//  Copyright (c) 2026-present, salesforce.com, inc. All rights reserved.
+//
+//  Redistribution and use of this software in source and binary forms, with or without modification,
+//  are permitted provided that the following conditions are met:
+//  * Redistributions of source code must retain the above copyright notice, this list of conditions
+//  and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright notice, this list of
+//  conditions and the following disclaimer in the documentation and/or other materials provided
+//  with the distribution.
+//  * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+//  endorse or promote products derived from this software without specific prior written
+//  permission of salesforce.com, inc.
+//
+//  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+//  IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+//  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+//  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+//  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+//  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+//  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import Foundation
+import Security
+
+@objc(SFSDKDPoPKeyPair)
+public final class SFSDKDPoPKeyPair: NSObject {
+    @objc public let publicKey: SecKey
+    @objc public let privateKey: SecKey
+
+    init(publicKey: SecKey, privateKey: SecKey) {
+        self.publicKey = publicKey
+        self.privateKey = privateKey
+    }
+}
+
+@objc(SFSDKDPoPKeyStoreError)
+public enum SFSDKDPoPKeyStoreError: Int, Error {
+    case missingScopeIdentifier = 1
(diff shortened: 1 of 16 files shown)


wmathurin:
Should we use the SFSDK prefix only for Objective-c and have "regular" names in swift.

For instance:
```swift
@objc(SFSDKDPoPKeyPair)
public final class SFSDKDPoPKeyPair: NSObject {
```
Could be:
```swift
@objc(SFSDKDPoPKeyPair)
public final class DPoPKeyPair: NSObject {
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/SFSDKDPoPKeyStore.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    private func fetchOrCreate(name: String) throws -> SFSDKDPoPKeyPair {
+        if let pair = lookup(name: name) {
+            return pair
+        }
+        let useSecureEnclave = SFSDKCryptoUtils.isSecureEnclaveAvailable()
+        if !useSecureEnclave {
+            SFSDKCoreLogger.i(Self.self, message: "DPoP keypair using software-backed storage; Secure Enclave unavailable")
+        }
+        let created = SFSDKCryptoUtils.createECKeyPair(withName: name,
```
bbirman, line 109: We've been slowly moving off of SFSDKCryptoUtils when possible, could any of the [EC methods in Encryptor.swift](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/7a20ddd567643abef6345cd335866c5de681dc57/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/Encryptor.swift#L223-L335) be used if they were public?

sfdctaka: @bbirman `KeyGenerator.ecKeyPair(name:)` covers the keypair lifecycle case and would let me drop the `SFSDKCryptoUtils.createECKeyPair` / `getECPrivateKeyRef` / `getECPublicKeyRef` calls entirely. It even handles the simulator-vs-Secure-Enclave split better than the manual fallback I wrote, so this is awesome to know.

The thing the DPoP layer needs is ES256 signing of arbitrary bytes (not encrypt/decrypt) and a JWK export of the public key. Those are the two new methods I added to `SFSDKCryptoUtils`(`signDataES256:withKeyRef:` and `jwkExportFromPublicKeyRef:`). `Encryptor.swift`'s methods are encrypt/decrypt-only, so they don't cover these. Two options:
1. Smaller scope (my preference for this PR): keep `signDataES256` / `jwkExport` in `SFSDKCryptoUtils` and only migrate the keypair lifecycle to `KeyGenerator`. Aligned with the "when possible" framing.
2. Bigger scope: move signing + JWK export into `Encryptor.swift` as part of this PR so DPoP touches `SFSDKCryptoUtils` zero times. Larger diff, but fully consistent.
 
To make option 1 work I'd need `KeyGenerator.ecKeyPair(name:)` and a `removeECKeyPair(name:)` to be public (today they're internal), and the KeyPair struct exposed (or the method returns a tuple). Happy to go either way, which do you prefer?

bbirman: yay! 

I would lean towards #2 or putting the new code in the SFSDKCryptoUtils swift extension just so we're not adding more Objective-C (obviously sometimes we need to for API boundaries but this one looks more straight forward). If you want to be able to break it up in multiple PRs, we could also potentially have a DPoP feature branch and then merge to dev at the end

sfdctaka: @bbirman Sounds like a plan. I'll do 2 and let's make a dev branch as this feature will need series of PRs.

sfdctaka:
Instead of making change to the `dev` branch this feature development will get its own feature branch.

PR #4052, opened 2026-06-03
[auto-generated] Fix CI: signal semaphore in SFSDKTestRequestListener setter

brandonpage:
## Summary
- Adds custom `setReturnStatus:` setter in `SFSDKTestRequestListener.m` that signals the completion semaphore when status transitions away from `kTestRequestStatusWaiting`
- Removes redundant explicit `dispatch_semaphore_signal` calls from 4 delegate methods (the setter handles it)
- Fixes MobileSync test timeouts (BriefcaseSyncDownTests, SyncManagerTests, BatchSyncUpTargetTests) caused by commit 7118b559c

## Root Cause
Commit 7118b559c replaced run-loop polling in `waitForCompletion` with `dispatch_semaphore_wait`, but only signaled the semaphore from OAuth/Identity delegate callback methods. External code that sets `returnStatus` directly via block-based callbacks (`TestSetupUtils.synchronousAuthRefresh`, `SyncManagerTestCase.sendSyncRequest`) never signaled the semaphore, causing every REST API call to timeout at 30 seconds instead of completing in 1-2 seconds.

## Test plan
- [ ] Verify MobileSync Nightly Tests pass on both iOS 18 and iOS 26
- [ ] Verify BriefcaseSyncDownTests, SyncManagerTests, BatchSyncUpTargetTests pass
- [ ] Verify SalesforceSDKCore tests still pass (no double-signal issues)
- [ ] Confirm this is test-infrastructure only — no production code changes

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/SFSDKTestRequestListener.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/SFSDKTestRequestListener.m
@@ -56,7 +56,14 @@ NSString* const kTestRequestStatusDidFail = @"didFail";
 - (void)dealloc {
     self.dataResponse = nil;
     self.lastError = nil;
-    self.returnStatus = nil;
+    _returnStatus = nil;
+}
+
+- (void)setReturnStatus:(NSString *)returnStatus {
+    _returnStatus = returnStatus;
+    if (![returnStatus isEqualToString:kTestRequestStatusWaiting]) {
+        dispatch_semaphore_signal(_completionSemaphore);
+    }
 }
 
 - (NSString *)waitForCompletion {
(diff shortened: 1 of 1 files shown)


wmathurin (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Test/SFSDKTestRequestListener.m
```diff
@@ -56,7 +56,14 @@ - (id)init
 - (void)dealloc {
     self.dataResponse = nil;
     self.lastError = nil;
-    self.returnStatus = nil;
+    _returnStatus = nil;
+}
+
+- (void)setReturnStatus:(NSString *)returnStatus {
+    _returnStatus = returnStatus;
```
sfdctaka, line 63: Should we check if `_returnSatus` is same as `returnStatus`?

sfdctaka: I guess it doesn't matter since it's just a string.

sfdctaka (approved):
LGTM!

PR #4054, opened 2026-06-04
Add ES256 signing + P-256 JWK to Encryptor; expose KeyGenerator EC API

sfdctaka:
## Summary
Pure-Swift crypto prerequisites for upcoming DPoP work. **No behavior changes; no DPoP code lands here.**

This PR is a direct response to PR-review feedback on the DPoP work ([review thread](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4051#discussion_r3346039293)) noting that the codebase is migrating off `SFSDKCryptoUtils` toward Swift. By landing the signing + JWK primitives in `Encryptor.swift` and reusing `KeyGenerator` for keypair lifecycle, this follow-up DPoP PR introduces zero new Objective-C and stays consistent with the codebase's Swift direction.

- `Encryptor.signES256(_:with:)` — signs `Data` with a P-256 private key and returns a 64-byte raw R‖S signature suitable for the JWS compact signature segment per RFC 7515 §3.4. Uses `SecKeyCreateSignature(.ecdsaSignatureMessageX962SHA256)` and `CryptoKit.P256.Signing.ECDSASignature` to strip the DER envelope.
- `Encryptor.jwkP256(from:)` — exports a P-256 public key as the RFC 7517 /RFC 7518 §6.2 JWK fields (`kty`, `crv`, `x`, `y`), with `x` and `y` as base64url-encoded 32-byte affine coordinates.
- Widens `KeyGenerator.KeyPair`, `KeyGenerator.ecKeyPair(name:)`, and a new name-based `KeyGenerator.removeECKeyPair(name:)` to `public` so external callers (and the upcoming DPoP keystore) can consume the existing keypair lifecycle without re-implementing the keychain dance.
- Three new `EncryptorError` cases for the new methods: `signingFailed`, `publicKeyExportFailed`, `unsupportedPublicKeyFormat`.

> ⚠️  **Public API addition — please review carefully.** Per `CLAUDE.md`,
> changes to public API surface require human review. The new symbols are:
> `Encryptor.signES256(_:with:)`, `Encryptor.jwkP256(from:)`,
> `KeyGenerator.KeyPair` (struct + properties), `KeyGenerator.ecKeyPair(name:)`,
> `KeyGenerator.removeECKeyPair(name:)`. Once these ship in a release, any
> signature change requires a deprecation cycle.

  ## Test plan
[...]

PR #4055, opened 2026-06-05
DPoP proof JWT at token exchange

sfdctaka:
Adds DPoP proof JWT generation to the OAuth2 token endpoint (auth-code exchange + refresh), so access tokens issued by the server can be sender-constrained to this device. Off by default; opt in via `SalesforceSDKManager.useDPoP`.

Builds on #4054, which moved the ES256 signing + P-256 JWK export into `Encryptor.swift` / `KeyGenerator`. This PR consumes those Swift APIs — no new Objective-C.

## What's in here
- `SFSDKDPoPKeyStore` — per-credentials P-256 keypair lifecycle, scoped on `SFOAuthCredentials.identifier`. Backed by `KeyGenerator` (Secure Enclave when available; software-backed keychain otherwise).
- `SFSDKDPoPProofBuilder` — builds the compact-serialized DPoP JWS for a token-endpoint request.
- `SFSDKDPoPNonceCache` — in-memory cache of server-issued `DPoP-Nonce` values, keyed `(htu, scope)`.
- `SFSDKDPoPRequestDecorator` — stamps the `DPoP` header on outgoing token requests.
- `SFSDKDPoPURL` — single source of truth for the `htu` form, shared between the proof builder and the nonce-cache
key.
- `SFSDKOAuth2` — wires the above into the approval-code and refresh flows: attaches the proof, harvests `DPoP-Nonce` from responses, retries once on `use_dpop_nonce` challenges.

## Highlights for review

- **New public API: `SalesforceSDKManager.useDPoP`** — opt-in flag (escalation per `CLAUDE.md`; please confirm).
- **OAuth2 flow change** — escalation per `CLAUDE.md`. Default behavior is unchanged when `useDPoP` is `NO`; the
helper is a thin pass-through and no DPoP code is reached.
- **Network-instance lifecycle in `SFSDKOAuth2`** was extracted into a single `sendTokenEndpointRequest:...` helper to
 avoid duplicating the nonce-retry logic across `accessTokenForApprovalCode:` and `accessTokenForRefresh:`. Mechanics are byte-for-byte the same; just centralized.
- **Nonce cache is in-memory by design.** Nonces are short-lived server-managed hints; cold-cache cost is one extra
round-trip with the standard challenge/retry handshake. Keypairs (the load-bearing secret) *are* persisted — in keychain.
- **Same htu canonicalization in two places** (proof payload + nonce-cache key) — extracted to `DPoPURL.htu(_:)` so
they can't drift.
- **Single retry per RFC** — the `retryOnNonceChallenge:` parameter is a recursion guard, not a feature flag. Hardcoded `YES` at the call sites and `NO` on the recursive call is intentional.
- **Xcode project edits.** `SalesforceSDKCore.xcodeproj` registers four new DPoP source files plus `SFSDKDPoPURL.swift`, `SFSDKDPoPTests.swift`, and `EncryptorTests.swift` (the last carries over from #4054 and could land separately if reviewers prefer).

## Test plan
- [x] `xcodebuild build` for `SalesforceSDKCore` (iphonesimulator)
- [x] `SFSDKDPoPTests` and `EncryptorTests` pass locally
- [x] `SalesforceSDKCore` test passes


wmathurin (commented):
NIT should the swift files match the name of the Swift class they contain?

wmathurin (commented):
Were you able to try this out against a live server (by using a ECA with DPOP required) ?

sfdctaka:
@wmathurin  @bbirman 

### Scope: this PR = iOS DPoP at the token endpoint only

A few things you might expect to see — and where they live instead:

**Deferred to follow-up work:**
- `SFRestRequest.m` `Authorization: Bearer` → `DPoP` switch → lands with the API-call DPoP work (switching it now
would break every authenticated REST call until per-request proof generation is in place).
- `ath` claim on the proof builder → API-call work (no incoming access token at code exchange, so not needed here).
- API-call nonce challenge / harvest / retry → nonce-support work. Token-endpoint nonce handling **is** in this PR.

**In this PR but worth flagging:**
- `SalesforceSDKManager.useDPoP` opt-in flag — defaults to `NO`; tests verify byte-identical request shape when off.
- Per-user keystore scoped on `SFOAuthCredentials.identifier` (chosen over `<orgId>_<userId>` because user/org IDs
aren't known until the token-endpoint *response*).
- Proactive `DPoP-Nonce` harvest from 200 OK responses at the token endpoint only — required for Phase-1 grants;
serial token-exchange path, so concurrency isn't a concern here.

PR #4054 added the Swift `Encryptor.signES256` / `Encryptor.jwkP256` helpers and widened `KeyGenerator` per earlier review feedback. This PR uses those — zero new Objective-C.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/SFSDKDPoPKeyStore.swift
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /// Removes any existing keypair for the scope. Idempotent.
+    @objc(deleteForScope:)
+    public func delete(forScope scope: String) {
+        guard !scope.isEmpty else { return }
+        let name = Self.keyName(for: scope)
+        queue.sync(flags: .barrier) {
+            do {
+                _ = try KeyGenerator.removeECKeyPair(name: name)
```
bbirman: nit: since it has `@discardableResult`, can it just be `try KeyGenerator.removeECKeyPair(name: name)`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/SFSDKDPoPProofBuilder.swift
```diff
@@ -0,0 +1,107 @@
...
+    ///   - httpMethod: HTTP verb (`"POST"` for token endpoint).
+    ///   - htu: Full request URL with no query and no fragment, per RFC 9449 §4.2.
+    ///   - nonce: Optional `nonce` claim from a prior `DPoP-Nonce` server hint.
+    ///   - keyPair: Public key is embedded in the JWS header `jwk`; private key signs.
+    ///   - now: Injected clock for deterministic tests; defaults to `Date()`.
+    @objc public static func buildProof(httpMethod: String,
+                                        htu: URL,
+                                        nonce: String?,
+                                        keyPair: DPoPKeyPair,
+                                        now: Date = Date()) throws -> String {
```
bbirman: Could the throw be typed to `DPoPProofBuilderError`?

bbirman (commented):
For the classes with queues, could you add tests for concurrent operations? We have a few existing tests using `DispatchQueue.concurrentPerform(iterations: X)` that helped me at least :)

wmathurin:
Maybe you should add dpop to the list of branches for which we run the PR github action (see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/.github/workflows/pr.yaml#L7) - at least until we merge it back to dev. That way, we could catch build/test regressions.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPNonceCache.swift
```diff
@@ -0,0 +1,88 @@
...
+
+    private let queue = DispatchQueue(label: "com.salesforce.dpop.nonceCache", attributes: .concurrent)
+    
+    // Nonces are ephemeral by spec.
+    private var storage: [String: String] = [:]
+
+    private override init() { super.init() }
+
+    @objc(nonceForHtu:scope:)
+    public func nonce(htu: URL, scope: String?) -> String? {
```
wmathurin, line 68: Shouldn't reading the nonce remove it from the cache or are you planning to have the consumer of the nonce make a second call to the cache to invalidate/remove the nonce?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPRequestDecorator.swift
```diff
@@ -0,0 +1,94 @@
...
+    public static func harvestNonce(from response: URLResponse?,
+                                    requestURL: URL?,
+                                    scope: String?) {
+        guard let http = response as? HTTPURLResponse,
+              let url = requestURL,
+              let nonce = http.value(forHTTPHeaderField: dpopNonceHeaderName),
+              !nonce.isEmpty else {
+            return
+        }
+        DPoPNonceCache.shared.setNonce(nonce, htu: url, scope: scope)
```
wmathurin, line 83: I know nonce handling is the not the focus of this PR - but here we will have to think about how to handle concurrent calls.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPNonceCache.swift
sfdctaka: Intentional, but you're right that it deserves a comment. Updated the class doc on DPoPNonceCache to spell out the rationale: the backend rotates the nonce on every response, so a successful request's harvestNonce brings the next nonce — staleness is bounded to one outbound request in steady state, and the server is the source of truth (a stale nonce produces a use_dpop_nonce challenge that the caller already retries). Read-and-remove would force concurrent callers to race for a single nonce. Today's caller is the token endpoint, which is serial, so the choice is invisible. I commented in the code that this policy will be revisited when DPoP extends to API calls.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPRequestDecorator.swift
sfdctaka: Agreed — and you're right it'll bite once REST API calls are in scope. Added a concurrency note at harvestNonce(...) flagging that today's caller pattern is serial (token endpoint only), and that when DPoP extends to REST, in-flight concurrent calls will all carry the same nonce and only one will rotate cleanly; the others will see use_dpop_nonce and retry. The follow-up will need to choose between accepting the extra round-trip, per-htu serialization, or pre-fetching. Out of scope here, but commented in the code so it doesn't get lost.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
```diff
@@ -360,9 +361,56 @@ - (NSMutableURLRequest *)prepareBasicRequest:(SFSDKOAuthTokenEndpointRequest *)e
...
+                                                     requestURL:request.URL
+                                                          scope:endpointReq.credentialsIdentifier];
+            NSInteger statusCode = 0;
+            if ([urlResponse isKindOfClass:[NSHTTPURLResponse class]]) {
+                statusCode = ((NSHTTPURLResponse *)urlResponse).statusCode;
+            }
+            if (retryOnNonceChallenge &&
+                [SFSDKDPoPRequestDecorator isNonceChallengeWithStatusCode:statusCode body:data response:urlResponse]) {
+                [SFSDKCoreLogger i:[strongSelf class] format:@"DPoP nonce challenge received; retrying token-endpoint request once."];
+                [request setValue:nil forHTTPHeaderField:@"DPoP"];
```
bbirman: nit: could @"DPoP" live in `SFSDKOAuthConstants`?

wmathurin (approved):

bbirman (approved):

PR #4057, opened 2026-06-10
W-22917328: Rename beacon_child_consumer_{key,secret} to auto_installed_app_org_consumer_{key,secret}

wmathurin:
## Summary

Updates the token endpoint wire-format keys and adds a legacy fallback for servers that haven't yet rolled out version 264.

When parsing the token response, the SDK now:
1. Checks for `auto_installed_app_org_consumer_{key,secret}` (new field name)
2. Falls back to `beacon_child_consumer_{key,secret}` if the new field is absent

A `TODO` comment marks both fallback constants and the fallback logic for removal once server version 264 has rolled out everywhere.

All code symbol names (`kSFOAuthBeaconChildConsumerKey`, `beaconChildConsumerKey` property, etc.) are **unchanged**.

## Files changed
- `SFSDKOAuthConstants.h` — new constant values + legacy fallback constants added
- `SFOAuthCredentials.m` — parsing updated with new-then-old fallback
- `SFSDKOAuthTokenEndpointResponseTests.m` — test key strings updated
- `SFOAuthCredentialsTests.m` — test key strings updated

## Test plan
- [x] `SalesforceSDKCore` unit test suite passes (iPhone 16 / iOS 18.6)
- [ ] AuthFlowTester UI tests pass (running externally)

## Related
- Android counterpart: forcedotcom/SalesforceMobileSDK-Android#2923
- GUS: W-22917328

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
@@ -469,11 +469,16 @@ NSException * SFOAuthInvalidIdentifierException(void) {
     if (params[kSFOAuthTokenFormat]) {
         self.tokenFormat = params[kSFOAuthTokenFormat];
     }
+    // TODO: Remove kSFOAuthLegacyBeaconChildConsumer* fallback once server version 264 has rolled out everywhere.
     if (params[kSFOAuthBeaconChildConsumerKey]) {
         self.beaconChildConsumerKey = params[kSFOAuthBeaconChildConsumerKey];
+    } else if (params[kSFOAuthLegacyBeaconChildConsumerKey]) {
+        self.beaconChildConsumerKey = params[kSFOAuthLegacyBeaconChildConsumerKey];
     }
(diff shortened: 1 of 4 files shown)


bbirman (approved):

wmathurin:
I decided to only change the value of the fields we look for in the response and keep beacon in our code and tests. The changes were getting large and would make cherry picking harder. Further more AutoInstalledApp is a mouthful compared to Beacon.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -469,11 +469,16 @@ - (void)updateCredentials:(NSDictionary *) params {
     if (params[kSFOAuthTokenFormat]) {
         self.tokenFormat = params[kSFOAuthTokenFormat];
     }
+    // TODO: Remove kSFOAuthLegacyBeaconChildConsumer* fallback once server version 264 has rolled out everywhere.
```
wmathurin, line 472: That way we can go against older instances and it should still work.

JohnsonEricAtSalesforce (approved):
👍🏻

PR #4060, opened 2026-06-11
DPoP proof JWT on API calls (with nonce retry per RFC 9449 §8/§9)

sfdctaka:
This PR extends DPoP coverage from the token endpoint to the SDK's authenticated REST and identity calls, plus adds the nonce-retry machinery the resource server requires per RFC 9449 §8/§9.

### DPoP on outbound API calls
- `SFRestRequest` / `SFRestAPI`: when `tokenType == "DPoP"`, requests are stamped with `Authorization: DPoP <access_token>` and a `DPoP` proof header bound to the request's method (`htm`) and URL (`htu`), with the access-token hash (`ath`) per RFC 9449 §4.3.
- `SFIdentityCoordinator`: same stamping for the identity endpoint.
- Bearer requests are untouched — `Authorization: Bearer …` is byte-identical to pre-DPoP output for non-DPoP token responses.

### Nonce challenge + retry (RFC 9449 §8 / §9)
- `DPoPNonceCache`: per-account, per-`htu` cache of the most recent server-issued nonce. Harvested from any 200, 400, or 401 response carrying a `DPoP-Nonce` header.
- Outbound proofs include `nonce` when the cache has one for the target `htu`.
- On a 401 with `DPoP-Nonce` (or a 400 body containing `error=use_dpop_nonce`), the originating request is retried once with the freshly harvested nonce. Retry is one-shot to avoid loops.

### Key material
- `DPoPKeyStore`: per-account ES256 / P-256 keypair, generated on first DPoP use and persisted in the keychain. The private key never leaves secure storage. cnf.jkt thumbprint computed via JWK SHA-256.

### Token-type wiring
- `SFOAuthCredentials.tokenType` (added in earlier PR) drives scheme selection. `"DPoP"` ⇒ DPoP path; anything else (including nil) ⇒ Bearer path.

## Backward compatibility

Purely additive. Non-DPoP orgs see no behavior change. The new headers are only emitted when the token endpoint returns `token_type=DPoP`.

## Test coverage

- `SFSDKDPoPTests` (988 lines): proof builder, nonce cache (per-account isolation, per-`htu` keying, eviction), request decorator (Bearer vs DPoP branch, nonce attachment, ath computation), end-to-end request stamping, retry on nonce challenge.
- Existing `SFOAuthCredentialsTests` and token-endpoint response tests updated.

## Notes for reviewers

- Concurrency for REST callers: today's caller patterns are largely serial after refresh; concurrent in-flight callers may each hit a \`use_dpop_nonce\` challenge and only one will rotate cleanly. A TODO marks the spot to revisit if/when concurrent DPoP REST calls become common (per-\`htu\` lock vs. pre-fetched nonce vs. accept-extra-round-trip).
- This PR targets the \`dpop\` integration branch; final merge to \`dev\` will follow once the backend resource-server rollout is complete.


wmathurin (commented):
Looks good to me.

bbirman (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
```diff
@@ -150,15 +150,32 @@ - (void)sendRequest
...
+                                              accessToken:self.credentials.accessToken
+                                                tokenType:self.credentials.tokenType
+                                                    error:&authError];
+    if (!ok) {
+        [SFSDKCoreLogger e:[self class] format:@"SFIdentityCoordinator: Failed to stamp authorization headers: %@", authError.localizedDescription];
+    }
     [request setTimeoutInterval:self.timeout];
     [request setHTTPShouldHandleCookies:NO];
-    [SFSDKCoreLogger d:[self class] format:@"SFIdentityCoordinator:Starting identity request at %@", self.credentials.identityUrl.absoluteString];
+    [SFSDKCoreLogger d:[self class] format:@"SFIdentityCoordinator:Starting identity request"];
```
bbirman: Did the URL logging get moved somewhere else?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
```diff
@@ -150,15 +150,32 @@ - (void)sendRequest
...
+
     __weak __typeof(self) weakSelf = self;
     self.networkIdentifier = [SFNetwork uniqueInstanceIdentifier];
     SFNetwork *network = [SFNetwork sharedEphemeralInstanceWithIdentifier:self.networkIdentifier];
     self.session = network.activeSession;
-    [network sendRequest:request dataResponseBlock:^(NSData *data, NSURLResponse *response, NSError *error) {
+    [SFSDKDPoPRequestDecorator sendRequestWithNonceRetry:request
+                                                   scope:self.credentials.identifier ?: @""
```
bbirman: What's the scenario for not having credentials?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
```diff
@@ -150,15 +150,32 @@ - (void)sendRequest
...
+                                     accessTokenProvider:^NSString * _Nullable {
```
bbirman: more of a style one, but it took me a second to parse this call, could the `accessTokenProvider` parameter be moved towards the end with the other blocks?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
```diff
@@ -150,15 +150,32 @@ - (void)sendRequest
...
     [request setHTTPMethod:@"GET"];
-    [request setValue:[NSString stringWithFormat:kHttpAuthHeaderFormatString, self.credentials.accessToken] forHTTPHeaderField:kHttpHeaderAuthorization];
+    NSError *authError = nil;
+    BOOL ok = [SFSDKDPoPRequestDecorator applyAuthHeaders:request
+                                                    scope:self.credentials.identifier
+                                              accessToken:self.credentials.accessToken
+                                                tokenType:self.credentials.tokenType
+                                                    error:&authError];
+    if (!ok) {
+        [SFSDKCoreLogger e:[self class] format:@"SFIdentityCoordinator: Failed to stamp authorization headers: %@", authError.localizedDescription];
```
bbirman, line 160: If there's no auth header at this point, can sending the request do anything? or should it notify the delegate of the error and return?

wmathurin (commented):
LGTM - but if nonce are sent by the token end point, we might be able to simplify the code a bit? Harvest the nonce whenever a refresh call happens. And also treat a nonce failure like 401s (to force a refresh and therefore to get a new nonce).

sfdctaka:
> LGTM - but if nonce are sent by the token end point, we might be able to simplify the code a bit? Harvest the nonce whenever a refresh call happens. And also treat a nonce failure like 401s (to force a refresh and therefore to get a new nonce).

Let me look into this now.

sfdctaka:
@wmathurin Dropped the resource-server reactive-retry layer and folded nonce failures into the existing 401→refresh→replay path, per your suggestion.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
bbirman: sorry if I missed it, what was the resolution for this one?

sfdctaka: Fixed: if applyAuthHeaders returns an error, SFIdentityCoordinator now notifies the delegate via notifyDelegateOfFailure: and returns. Did the same in the two SFUserAccountManager call sites (user-photo path bails silently after logging; shouldBlockUser: calls the errorBlock).

Not sure why I resolved your comment before. But now it should be 👍

wmathurin:
> @wmathurin Dropped the resource-server reactive-retry layer and folded nonce failures into the existing 401→refresh→replay path, per your suggestion.

I love the simplification. 

I see that the nonce cache is still in memory only so an application restart will force a refresh (because of missing the nonce for creating the JWT proof). Should we persist the nonce to disk (alongside the access token and other user credentials) to avoid that situation? Or maybe we could store the JWT proof that we generate from the nonce (as long as there is a 1-1 mapping from nonce to JWT proof)?

sfdctaka:
> > @wmathurin Dropped the resource-server reactive-retry layer and folded nonce failures into the existing 401→refresh→replay path, per your suggestion.
> 
> I love the simplification.
> 
> I see that the nonce cache is still in memory only so an application restart will force a refresh (because of missing the nonce for creating the JWT proof). Should we persist the nonce to disk (alongside the access token and other user credentials) to avoid that situation? Or maybe we could store the JWT proof that we generate from the nonce (as long as there is a 1-1 mapping from nonce to JWT proof)?

@wmathurin  Persisting the JWT proof unfortunately won't work: DPoP proofs carry `iat` and `jti` claims and servers reject them outside a small clock window (typically 60s–5min) and dedupe `jti` to prevent replay, so a proof persisted across a restart would be rejected on those grounds.

On persisting the nonce itself, here's why I'd hold off:

What it buys us:
* exactly one avoided HTTP round-trip on the first DPoP call after a cold start. That round-trip is the existing 401→refresh→replay path, which works correctly and is transparent to app code. It's not a UX-visible cost.

What it costs us:
* A new piece of persisted auth state — either a new keychain item (which then needs to participate in account-deletion, log-out, account-switch, keychain migration, and access-group rules), or a tag-along field on `SFOAuthCredentials` (which conflates stable identity material with a volatile server-issued bouncer pass).
* Invalidation logic for every staleness path: log-out, token refresh, 401, INVALID_NONCE, account switch, app upgrade, time-based expiry. Today we get all of these for free because process exit clears the cache.
* Security/privacy review surface: data classification, encryption at rest, Privacy Manifest entry, threat-model update. Modest but non-zero, and it has to be repeatable for external auditors since this is OSS.
* And it's not a guarantee: RFC 9449 §8 puts nonce lifetime fully under the server's control. A persisted nonce only maybe skips the refresh — when it's stale we pay the round-trip anyway. We'd be carrying all the cost above for a probabilistic win.

The current design is also future-proof in a way persistence isn't: if Salesforce ever rotates nonces more aggressively or starts emitting them from resource servers, the SDK needs zero code changes. Persistence would lock in assumptions about server behavior we can't enforce.

wmathurin (approved):

PR #4069, opened 2026-06-13
@W-23007797: [iOS] Stabilize flaky RestClientPublisherTests

JohnsonEricAtSalesforce:
## Summary
Stabilizes 4 flaky Combine publisher tests that failed intermittently — on iOS 18 in isolation and on both platforms under full-suite CI load.

## Root Causes
1. **Thread delivery race**: `RestClient.publisher(for:)` resolves its `Future` on the cooperative thread pool. XCTestExpectation fulfillment from a background thread races with `wait(for:timeout:)`.
2. **Timeout too short under load**: 5-10s timeouts insufficient when ~200 concurrent API calls hit the same org during full-suite execution.

## Fix
- Added `.receive(on: DispatchQueue.main)` before `.sink` in the shared `evaluateResults()` helper — ensures expectation fulfillment on the main thread where the waiter is synchronized
- Increased timeouts from 5/10s to 60s — accommodates CI load where the org responds slowly (safety net only; expectations fulfill immediately on response)
- Added missing `super.setUp()` call in instance setUp

## Tests Fixed
- `testQueryPublisher`
- `testRecordsPublisher`
- `testCompositePublisher`
- `testBatchPublisher`

## Verification
- [x] All 4 tests pass locally (3 consecutive runs)
- [x] 6 CI runs with no target test failures
- [x] Validated under full-suite load (combined branch)

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientPublisherTests.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientPublisherTests.swift
@@ -27,7 +27,7 @@ import Combine
 @testable import SalesforceSDKCore
 
 class RestClientPublisherTests: XCTestCase {
-    
+
     var currentUser: UserAccount?
     
     override class func setUp() {
@@ -36,29 +36,30 @@ class RestClientPublisherTests: XCTestCase {
         TestSetupUtils.populateAuthCredentialsFromConfigFile(for: SFSDKAuthUtilTests.self)
         TestSetupUtils.synchronousAuthRefresh()
     }
-    
+
     override func setUp() {
+        super.setUp()
         currentUser = UserAccountManager.shared.currentUserAccount
     }
-    
+
     func testQueryPublisher() {
         let request = RestClient.shared.request(forQuery: "select name from CONTACT", apiVersion: nil)
         let publisher = RestClient.shared.publisher(for: request)
-        
+
         let validTest = evaluateResults(publisher: publisher)
-        wait(for: validTest.expectations, timeout: 5)
+        wait(for: validTest.expectations, timeout: 60)
         validTest.cancellable?.cancel()
     }
-    
+
     func testRecordsPublisher() {
         let request = RestClient.shared.request(forQuery: "select name from CONTACT", apiVersion: nil)
         let publisher: AnyPublisher<RestClient.QueryResponse<TestContact>, Never> = RestClient.shared.records(forRequest: request)
-        
+
         let validTest = evaluateResults(publisher: publisher)
-        wait(for: validTest.expectations, timeout: 5)
+        wait(for: validTest.expectations, timeout: 60)
         validTest.cancellable?.cancel()
     }
-  
+
     func testCompositePublisher() {
         let accountName = self.generateRecordName()
         let contactName = self.generateRecordName()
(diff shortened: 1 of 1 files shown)


JohnsonEricAtSalesforce:
re-run ci

libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientPublisherTests.swift
```diff
@@ -36,29 +36,30 @@ class RestClientPublisherTests: XCTestCase {
         TestSetupUtils.populateAuthCredentialsFromConfigFile(for: SFSDKAuthUtilTests.self)
         TestSetupUtils.synchronousAuthRefresh()
     }
-    
+
     override func setUp() {
+        super.setUp()
```
JohnsonEricAtSalesforce, line 41: Added missing `super.setUp()` — correct XCTest practice. Not directly related to the flaky fix but fixes a latent lifecycle issue.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientPublisherTests.swift
```diff
@@ -36,29 +36,30 @@ class RestClientPublisherTests: XCTestCase {
...
-    
+
     func testQueryPublisher() {
         let request = RestClient.shared.request(forQuery: "select name from CONTACT", apiVersion: nil)
         let publisher = RestClient.shared.publisher(for: request)
-        
+
         let validTest = evaluateResults(publisher: publisher)
-        wait(for: validTest.expectations, timeout: 5)
+        wait(for: validTest.expectations, timeout: 60)
```
JohnsonEricAtSalesforce, line 50: Timeouts increased from 5/10s → 60s. These tests hit a live Salesforce org. Under full-suite CI load (~200 concurrent API calls across two platforms), the original timeouts were insufficient. 60s is a safety net — the expectation fulfills immediately when the response arrives; the timeout only fires on genuine org outage.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientPublisherTests.swift
```diff
@@ -68,64 +69,65 @@ class RestClientPublisherTests: XCTestCase {
...
-            case .finished:
-                finished.fulfill()
-            }
-        }, receiveValue: { response in
-            XCTAssertNotNil(response)
-            received.fulfill()
-        })
+
+        let cancellable = publisher?
+            .receive(on: DispatchQueue.main)
```
JohnsonEricAtSalesforce, line 113: Core fix: `.receive(on: DispatchQueue.main)` before `.sink`. RestClient's publisher resolves its Future on the cooperative thread pool via an unstructured `Task {}`. Without this, the sink closure fires on whatever background thread completes the Future — racing with `wait(for:timeout:)` on the main thread. This ensures all expectation fulfillments are serialized on main.

The sink closure logic (switch on completion, fulfill expectations) is unchanged — only the delivery scheduling is new.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientPublisherTests.swift
```diff
@@ -68,64 +69,65 @@ class RestClientPublisherTests: XCTestCase {
...
+                XCTAssertNotNil(response)
+                received.fulfill()
+            })
         return (expectations: [finished, received, failed], cancellable: cancellable)
     }
     
     private func generateRecordName() -> String {
-        let timecode = Date.timeIntervalSinceReferenceDate
-        return "SwiftPublishersTestsiOS\(timecode)"
+        return "SwiftPublishersTestsiOS\(Date.timeIntervalSinceReferenceDate)"
```
JohnsonEricAtSalesforce, line 130: Simplified from two lines to one. Pure style — no behavioral change.

JohnsonEricAtSalesforce (commented):
**Risk analysis:**

This fix preserves the original test intent — each test still verifies that a Combine publisher emits a value and completes (or fails) when a REST API request is made against a live Salesforce org.

The assertions in `evaluateResults()` are identical: same 3 expectations (finished, received, failed), same inverted-expectation logic for valid/invalid results, same `XCTAssertNotNil` on responses and errors. Only two things changed:

1. **`.receive(on: DispatchQueue.main)`** — ensures the sink closure runs on the main thread. This is strictly tighter: a test that passed by racing (fulfillment happened to land before the waiter checked) would now deterministically pass because delivery is serialized on the same thread as `wait(for:)`.

2. **60s timeout** — generous safety net for CI load. Does NOT mask failures: if the publisher never emits (real bug), the test still fails after 60s. If it emits normally (~1-2s), the test passes immediately. The timeout only prevents false negatives during org outages.

No production code is modified.

sfdctaka (approved):
LGTM!

PR #4072, opened 2026-06-14
@W-23038161: [iOS] Fix nil crash in SFSDKBatchResponse for batch requests with haltOnError

JohnsonEricAtSalesforce:
## Summary
Fixes an intermittent crash (`NSArrayM insertObject:nil`) in `testAsyncBatchRequestStopOnFailure` caused by nil-unsafe parsing in `SFSDKBatchResponse`.

## Root Cause
`SFSDKBatchResponse.initWith:` stored `[dict objectForKey:@"results"]` directly into `_results`. When the server returns a response where "results" is nil or not an NSArray (which can happen during error conditions with `haltOnError=true`), the nil value is stored. When Swift code accesses `batchResponse.results` (bridged as a non-optional NSArray) and iterates it, the nil triggers `NSArrayM insertObject:nil`.

## Fix
- **SFSDKBatchResponse.m**: Type-check "results" value before assignment; default to empty array if nil or non-array. Empty results with `hasErrors=true` is the correct representation of a halted batch.
- **RestClientTest.swift**: Restructured test for crash safety — replaced force-unwraps (`as!`) with `try XCTUnwrap`, added guard on results count, used `XCTAssertEqual` for better failure diagnostics.

## Tests Fixed
- `testAsyncBatchRequestStopOnFailure`

## Verification
- [x] Build passes locally
- [x] CI: 6 consecutive clean runs (overnight validation pending)

## Note
An earlier iteration of this PR included `OnceGuard` — a thread-safe continuation wrapper to prevent double-resume crashes in `RestClient.swift`. After review, we determined the nil-safety fix alone addresses the proven crash (NSArrayM insertObject:nil), and OnceGuard was speculative defense for an unproven double-callback scenario. OnceGuard was removed in a separate commit (preserved in branch history for reference). If double-resume is ever proven to occur, it can be reintroduced with logging rather than silent drop.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFSDKBatchResponse.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFSDKBatchResponse.m
@@ -29,7 +29,8 @@ WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH 
 -(instancetype)initWith:(NSDictionary *)dict {
     if (self=[super init]) {
        _hasErrors =  [[dict objectForKey:@"hasErrors"] boolValue];
-       _results =  [dict objectForKey:@"results"];
+       NSArray *results = [dict objectForKey:@"results"];
+       _results = [results isKindOfClass:[NSArray class]] ? results : @[];
     }
     return self;
 }
--- a/libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
@@ -264,46 +264,42 @@ class RestClientTests: XCTestCase {
     }
     
     func testAsyncBatchRequestStopOnFailure() async throws {
+        let accountName = self.generateRecordName()
+        let contactName = self.generateRecordName()
+        let apiVersion = RestClient.shared.apiVersion
+
+        let requestBuilder = BatchRequestBuilder()
+            .add(RestClient.shared.requestForCreate(withObjectType: "Account", fields: ["Name": accountName], apiVersion: apiVersion))
+            .add(RestClient.shared.requestForCreate(withObjectType: "Contact", fields: ["LastName": contactName], apiVersion: apiVersion))
+            .add(RestClient.shared.request(forQuery: "select Id from Account where Name ", apiVersion:  apiVersion)) // bad query
+            .add(RestClient.shared.request(forQuery: "select Id from Contact where Name = '\(contactName)'", apiVersion: apiVersion))
+            .setHaltOnError(true)
+
+        let batchRequest = requestBuilder.buildBatchRequest(apiVersion)
+        XCTAssertNotNil(batchRequest, "Batch Request should not be nil")
+        XCTAssertEqual(batchRequest.batchRequests.count, 4, "Batch Requests should have 4 requests")
+
+        let batchResponse: BatchResponse
         do {
-            // Create account
-            let accountName = self.generateRecordName()
-            let contactName = self.generateRecordName()
-            let apiVersion = RestClient.shared.apiVersion
-            
-            let requestBuilder = BatchRequestBuilder()
-                .add(RestClient.shared.requestForCreate(withObjectType: "Account", fields: ["Name": accountName], apiVersion: apiVersion))
-                .add(RestClient.shared.requestForCreate(withObjectType: "Contact", fields: ["LastName": contactName], apiVersion: apiVersion))
-                .add(RestClient.shared.request(forQuery: "select Id from Account where Name ", apiVersion:  apiVersion)) // bad query
-                .add(RestClient.shared.request(forQuery: "select Id from Contact where Name = '\(contactName)'", apiVersion: apiVersion))
-                .setHaltOnError(true)
-            
-            let batchRequest = requestBuilder.buildBatchRequest(apiVersion)
-            XCTAssertNotNil(batchRequest, "Batch Request should not be nil")
-            XCTAssertTrue(batchRequest.batchRequests.count == 4, "Batch Requests should have 4 requests")
-            
-            let batchResponse = try await RestClient.shared.send(batchRequest: batchRequest)
-            XCTAssertTrue(batchResponse.hasErrors, "BatchResponse results should not have any errors")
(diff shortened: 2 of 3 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFSDKBatchResponse.m
```diff
@@ -29,7 +29,8 @@ @implementation SFSDKBatchResponse
 -(instancetype)initWith:(NSDictionary *)dict {
     if (self=[super init]) {
        _hasErrors =  [[dict objectForKey:@"hasErrors"] boolValue];
-       _results =  [dict objectForKey:@"results"];
+       NSArray *results = [dict objectForKey:@"results"];
+       _results = [results isKindOfClass:[NSArray class]] ? results : @[];
```
JohnsonEricAtSalesforce, line 33: Nil-safety fix (1 production line changed). Original stored `[dict objectForKey:@"results"]` directly — nil if key missing or value is not an NSArray. When Swift iterates the bridged non-optional `NSArray`, nil triggers `NSArrayM insertObject:nil` crash.

Fix: type-check + default to `@[]`. Empty results with `hasErrors=true` is the correct representation of a halted batch — callers already check `hasErrors` before accessing individual results.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
```diff
@@ -264,46 +264,42 @@ class RestClientTests: XCTestCase {
...
+            .add(RestClient.shared.request(forQuery: "select Id from Account where Name ", apiVersion:  apiVersion)) // bad query
+            .add(RestClient.shared.request(forQuery: "select Id from Contact where Name = '\(contactName)'", apiVersion: apiVersion))
+            .setHaltOnError(true)
+
+        let batchRequest = requestBuilder.buildBatchRequest(apiVersion)
+        XCTAssertNotNil(batchRequest, "Batch Request should not be nil")
+        XCTAssertEqual(batchRequest.batchRequests.count, 4, "Batch Requests should have 4 requests")
+
+        let batchResponse: BatchResponse
         do {
```
JohnsonEricAtSalesforce, line 283: Narrowed the do/catch to only the `send` call. The old version wrapped the entire test body — if any assertion threw, the catch swallowed it with a generic "should not throw" message, hiding which line failed and what the actual error was. Now unexpected throws propagate naturally (the function is `async throws`) and XCTest reports the exact line.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
```diff
@@ -264,46 +264,42 @@ class RestClientTests: XCTestCase {
...
-            XCTAssertTrue(resp3["statusCode"] as? Int == 400, "Wrong status for first request")
-            XCTAssertTrue(resp4["statusCode"] as? Int == 412, "Request processing should have stopped on error")
+            batchResponse = try await RestClient.shared.send(batchRequest: batchRequest)
         } catch {
-            XCTFail("Send Batch Request should not throw an error")
+            XCTFail("Send Batch Request should not throw an error: \(error)")
+            return
         }
+
+        XCTAssertTrue(batchResponse.hasErrors, "BatchResponse should have errors")
```
JohnsonEricAtSalesforce, line 290: Guard on count prevents array-out-of-bounds crash if the server returns fewer results (possible with `haltOnError=true`). The `XCTAssertEqual` on the line above already reports the mismatch — this guard just prevents the subsequent indexing from crashing the test runner.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/RestClientTest.swift
```diff
@@ -264,46 +264,42 @@ class RestClientTests: XCTestCase {
...
+        XCTAssertEqual(batchResponse.results.count, 4, "Wrong number of results")
+        guard batchResponse.results.count == 4 else { return }
```
JohnsonEricAtSalesforce, line 292: `try XCTUnwrap` replaces `as!` force-unwraps. Same assertion (value must be a dictionary) but instead of `EXC_BAD_ACCESS` crash that kills the entire test suite, it records a test failure with a descriptive message and throws — other tests continue running. Strictly better for CI reliability.

JohnsonEricAtSalesforce (commented):
**Risk analysis:**

This PR makes one production code change: a nil-safety guard in `SFSDKBatchResponse.m` (ObjC). The property `results` is declared as `NSArray *` (non-nullable by convention). The original code could store nil; the fix ensures it always stores a valid NSArray. No downstream code should have relied on nil as a sentinel — the property's contract is non-nil.

The test changes are strictly safer: `as!` → `try XCTUnwrap` (assertion instead of crash), `guard` on count (prevents array-out-of-bounds), `XCTAssertEqual` (shows expected vs actual). Same assertions, same logic, no crashes.

**OnceGuard history:**

An earlier commit on this branch introduced `OnceGuard` — a `OSAllocatedUnfairLock`-based wrapper to ensure `CheckedContinuation` is resumed exactly once. The hypothesis was that ObjC callback machinery could invoke both `failureBlock` and `successBlock`, causing a fatal double-resume.

After operator review, we identified concerns:
1. The proven crash (`NSArrayM insertObject:nil`) is caused by nil results, not double-resume
2. OnceGuard silently drops the second resume — if a valid success arrives after a spurious failure, it would be lost
3. The double-callback scenario was never reproduced with evidence

Decision: remove OnceGuard (commit preserved in branch history), ship only the proven fix. If double-resume is ever observed, OnceGuard should be reintroduced with logging so the root cause is visible rather than silently suppressed.

wmathurin (approved):

wmathurin (approved):

JohnsonEricAtSalesforce:
Merging with `--admin`: CI shows UI test ^26 failure (56m timeout — simulator hang) and a codecov/project/MobileSync threshold miss.

Both are unrelated to this PR's changes (which modify `SFSDKBatchResponse.m` nil-safety and `RestClientTest.swift` test assertions). Unit tests pass on both iOS 18 and iOS 26. UI test ^18 passes. The ^26 UI test hang is a known simulator infrastructure issue (Apple FB21079497).

Target test `testAsyncBatchRequestStopOnFailure` passes on both platforms. This PR was validated with 6 consecutive clean runs during overnight CI validation (June 16-17).

PR #4074, opened 2026-06-17
Use instanceUrl host when refreshing OAuth tokens

sfdctaka:
Token refresh previously POSTed to the login-pool host (login.<env>.salesforce.com), which then redirected (302) to the user's My Domain. The redirect breaks DPoP htu binding (RFC 9449 §4.3) and adds an unnecessary round-trip on every refresh for Bearer flows.

`overrideDomainIfNeeded` now prefers credentials.instanceUrl when populated and falls back to domain otherwise. instanceUrl is only set after the first token response, so the code-exchange path (where instanceUrl is nil) keeps targeting the login pool unchanged. Community URL still takes precedence and is now nil-guarded.

The refresh log line in `accessTokenForRefresh:` now prints the target host so support can see at a glance whether a refresh hit My Domain or the login pool. Token values remain unlogged.

Tests: 6 new in `SFSDKOAuth2RefreshInstanceUrlTests` covering precedence, fallback, community precedence, code-exchange neutrality, and end-to-end URL composition via `prepareBasicRequest:`.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
@@ -372,8 +372,17 @@ NSException * SFOAuthInvalidIdentifierException(void) {
 }
 
 - (NSURL *)overrideDomainIfNeeded {
-    NSString *domain = self.communityId ? self.communityUrl.absoluteString : self.domain;
-    NSString *protocolHost = self.communityId ? domain : [NSString stringWithFormat:@"%@://%@", self.protocol, domain];
+    // Precedence: communityUrl > instanceUrl > domain. instanceUrl is populated only after the first
+    // token response, so a nil instanceUrl naturally identifies the code-exchange path.
+    if (self.communityId && self.communityUrl) {
+        return self.communityUrl;
+    }
+
+    if (self.instanceUrl) {
+        return self.instanceUrl;
+    }
+
+    NSString *protocolHost = [NSString stringWithFormat:@"%@://%@", self.protocol, self.domain];
     return [NSURL URLWithString:protocolHost];
 }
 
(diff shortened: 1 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.m
```diff
@@ -372,8 +372,17 @@ - (BOOL)hasPropertyValueChangedForKey:(NSString *) key {
 }
 
 - (NSURL *)overrideDomainIfNeeded {
-    NSString *domain = self.communityId ? self.communityUrl.absoluteString : self.domain;
-    NSString *protocolHost = self.communityId ? domain : [NSString stringWithFormat:@"%@://%@", self.protocol, domain];
+    // Precedence: communityUrl > instanceUrl > domain. instanceUrl is populated only after the first
+    // token response, so a nil instanceUrl naturally identifies the code-exchange path.
+    if (self.communityId && self.communityUrl) {
```
wmathurin, line 377: 👍 
We should set up a community/community users in our test org and have our automated auth flow UI tests exercise login / refresh with them.

bbirman, line 377: `instanceURL` and `communityURL` don't need the protocol anymore?

sfdctaka: Right — because `instanceUrl` and `communityUrl` are typed `NSURL *` in the public header (which already encodes scheme), whereas domain is `NSString *` (just a hostname like "login.salesforce.com", no scheme):
- `SFOAuthCredentials.h`:78 — `domain` is `NSString *` → still needs `protocol://` prepended
- `SFOAuthCredentials.h`:160 — `instanceUrl` is `NSURL *` → scheme already baked in
- `SFOAuthCredentials.h`:189 — `communityUrl` is `NSURL *` → scheme already baked in

The old code only had a `protocolHost` prepend for the domain fallback, and read `communityUrl.absoluteString` (already including "https://"). Same logic now, just expressed via direct `NSURL` returns instead of round-tripping through strings.

sfdctaka:
Manually smoke-tested on iOS Simulator against a My Domain dev org. Logged in, forced an access-token refresh, and confirmed the OAuth refresh `POST` targets the user's My Domain host directly — no 302 through the login pool. The new log line `Initiating refresh token flow to host: <my-domain>` printed the My Domain host as expected.

bbirman (approved):

wmathurin (approved):

PR #4076, opened 2026-06-19
Update SalesforceSDKCoreTestApp to use scene lifecycle 

bbirman:
Modernize to avoid this warning in the logs:
"`UIScene` lifecycle will soon be required. Failure to adopt will result in an assert in the future."


bbirman:
Updated to use the same structure as the common test app. If this sounds good I'll do the same thing to the test apps for analytics, mobile sync and smart store

wmathurin (approved):
👍

bbirman:
Will update the other test apps in other PRs to avoid having a large PR

PR #4077, opened 2026-06-19
Add Swift version to common config & consolidate settings

bbirman:


native/SampleApps/RestAPIExplorer/RestAPIExplorer.xcodeproj/project.pbxproj
```diff
@@ -742,7 +742,6 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.salesforce.mobilesdk.RestAPIExplorer;
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				SUPPORTS_MACCATALYST = YES;
-				SWIFT_VERSION = 5.0;
```
wmathurin, line 745: why is the SWIFT_VERSION dropped here but kept line 724

bbirman: The sample apps don't use config files so for those if the version was set at the target level I moved it to the project level (doesn't do much for RestAPIExplorer because it only has one target but MobileSyncExplorer and AuthFlowTester have two). I could also add config files if you have a preference, was on the fence about how wide to go with the change

wmathurin: I don't really have a preference.

wmathurin (approved):

PR #4078, opened 2026-06-19
[iOS] Fix Login for Admin and Welcome Discovery incompatibility

brandonpage:
## Summary

The 13.2.1 **Login for Admin** (LFA) settings menu action fundamentally did not work with the 2-phase **Welcome Discovery** login flow. Tapping LFA on the discovered My Domain (phase 2) launched `ASWebAuthenticationSession` against `welcome.salesforce.com` instead of the resolved My Domain — a non-functional UX. Tapping it on the discovery page itself (phase 1) made things worse.

This PR is a narrow fix that:

1. **Hides the LFA menu entry during phase 1 of Welcome Discovery.** A `UIDeferredMenuElement.elementWithUncachedProvider:` re-evaluates visibility every time the menu opens, keyed off `+[SFLoginViewController shouldShowLoginForAdminForSession:]`. Phase-1 detection: `SFDomainDiscoveryCoordinator.isDiscoveryDomain` AND `!coordinator.domainUpdated`. Per-scene scoping via `view.window.windowScene.session.persistentIdentifier`.
2. **Also no-ops the public `loginViewControllerDidSelectLoginForAdmin:` API in phase 1**, so external hosts that call it directly cannot bypass the visibility gate. Logs a warning. Documented on the public header.
3. **Routes the LFA browser session to the resolved My Domain in phase 2 with the captured `login_hint`** by introducing two **LFA-scoped overrides** on `SFSDKAuthRequest`:
   - `loginAsAdminMyDomain` — the resolved My Domain
   - `loginAsAdminLoginHint` — the username surfaced by the discovery page

   Consulted in `authenticateWithRequest:loginHint:` to set `credentials.domain`, the coordinator's `loginHint`, and the `appConfigForLoginHost:` lookup. The request's permanent `loginHost` is **left unchanged**.
4. **Clears all three LFA fields on cancel** (`loginAsAdmin`, `loginAsAdminMyDomain`, `loginAsAdminLoginHint`) so the post-cancel `restartAuthentication:` re-runs against the originally configured login host.

### Why LFA-scoped overrides instead of mutating `oauthRequest.loginHost`

An earlier version of this fix copied `coordinator.credentials.domain` onto `session.oauthRequest.loginHost` in-memory. That looked clean but produced an asymmetry: after a cancelled LFA, **Reload** and **Clear Cache** kept the WebView pinned to the My Domain instead of returning to `welcome.salesforce.com/discovery`. Putting the override in dedicated LFA-scoped fields keeps `loginHost` reflecting "the configured login server" at all times, so every other settings action stays correct by definition.

### No persistence — Welcome Discovery returns after logout

The discovered My Domain is held only in memory on `SFSDKAuthRequest`. It is **never** written to `setLoginHost:`, `SFSDKLoginHostStorage`, or `NSUserDefaults`. After logout, the app correctly reloads the Welcome Discovery host and the user goes through phase 1 again — preserving the existing intentional behavior.

## Behavior change: phase-2 LFA cancel

After tapping **Cancel** in the system browser sheet, the embedded WebView **returns to phase 1 of Welcome Discovery** (`welcome.salesforce.com/discovery`) and the user re-picks the account. This is **intentionally different from Android**, where the equivalent flow keeps the WebView on the resolved My Domain. The iOS choice matches **existing iOS cancel behavior on every other login flow** (cancel routes through `restartAuthentication:` against the configured login host) and avoids the Reload/Clear-Cache asymmetry described above. The cost is one extra account re-pick after a cancel; the benefit is symmetric, easy-to-reason behavior across all settings actions.

## Backport / cherry-pick notes

Welcome Discovery ships only to a few internal apps, so **this fix is not being patched into 13.2.x**. However, those apps may want to cherry-pick this change onto a pre-14.0 base before 14.0 ships, so the commit is deliberately structured to apply with minimal conflict:

- It reuses the existing `loginAsAdmin` property name (no rename), so it touches none of the ~70 `loginAsAdmin` references that pre-14.0 branches carry.
- It does **not** modify `SFSDKAuthSession.m` (the `useBrowserAuth || loginAsAdmin` line already exists on those branches), avoiding the deprecated-`userAgentForAuth` block that differs across releases.
- Verified against `v13.2.1`: 5 of 6 files apply cleanly; the only conflict is a one-line placement nudge in the test bridging header (`SalesforceSDKCoreTests-Bridging-Header.h`), where adjacent unrelated test categories that exist on `dev` are absent on 13.2.1. Trivial to resolve by keeping the new `SFLoginViewController (LoginForAdminTesting)` category.

### Automated tests

- [x] `SalesforceSDKCoreTests/LoginForAdminTests` — 34/34 pass
- [x] `SalesforceSDKCoreTests/WelcomeDiscoveryLoginHostTests` — 6/6 pass
- [x] `SalesforceSDKCoreTests/DomainDiscoveryCoordinatorTests` — pass

### Manual verification (AuthFlowTester)

- [x] **MTP-01** — Phase 1 of Welcome Discovery: Login for Admin entry hidden in the gear menu.
- [x] **MTP-02** — Phase 2 of Welcome Discovery: Login for Admin entry present after picking an account.
- [x] **MTP-03** — Phase 2 LFA tap: ASWebAuthenticationSession opens against the resolved My Domain with `login_hint=<username>`. Does NOT open against `welcome.salesforce.com`.
- [x] **MTP-04** — Phase 2 LFA cancel: embedded WebView returns to `welcome.salesforce.com/discovery` (phase 1) — **intentionally different from Android** (see Behavior change section above).
- [x] **MTP-05** — Phase 2 LFA: full happy-path login through to RestClient succeeds.
- [x] **MTP-06** — Logout after phase-2 LFA login: login screen returns to Welcome Discovery, NOT the My Domain.
- [x] **MTP-07** — Non-discovery server (`login.salesforce.com`): LFA still present and works as before.
- [x] **MTP-08** — Non-discovery server LFA cancel: pre-fix behavior preserved (no regression).
- [x] **MTP-09** — Opaque (PKCE) consumer key, `useWebServerFlow=NO`, phase-2 LFA: authorize URL still uses `response_type=code` with PKCE.
- [x] **MTP-10** — Multi-scene (iPad Split View): phase 1 in scene A hides LFA; phase 2 in scene B shows LFA. Per-scene scoping verified.
- [x] **MTP-11** — Public `loginViewControllerDidSelectLoginForAdmin:` called externally during phase 1: no-op, warning logged, no browser opens.
- [x] **MTP-12** — Public API called during phase 2: works, browser opens against the resolved My Domain.
- [x] **MTP-13** — `SFSDKLoginHostStorage` / `NSUserDefaults` inspection after phase-2 LFA + logout: My Domain is NOT persisted.
- [x] **MTP-14** — Reload / Clear Cache from gear menu during phase 2 (and after a cancelled LFA): menu predicate re-evaluates correctly; both Reload paths produce the same URL regardless of whether LFA was previously cancelled.

## Review checklist

- [x] No public API change — `SFSDKAuthRequest.h` is internal (not in `public_header_files`); the two new properties are not exposed. No symbols renamed.
- [x] No `Localizable.strings` change.
- [x] No persistence of the discovered My Domain — verified via MTP-13.
- [x] No new third-party dependencies, no `Podfile` / `.podspec` / `.xcodeproj` / `.xcworkspace` change.
- [x] Multi-user / multi-scene: verified per-scene scoping (manual MTP-10 pending).
- [x] No new compiler warnings.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -376,8 +376,7 @@
     }
 
     NSString *loginHost = session.oauthRequest.loginHost;
-    SFDomainDiscoveryCoordinator *discoveryCoordinator = [[SFDomainDiscoveryCoordinator alloc] init];
-    BOOL isDiscoveryHost = [discoveryCoordinator isDiscoveryDomain:loginHost];
+    BOOL isDiscoveryHost = [SFDomainDiscoveryCoordinator isDiscoveryDomain:loginHost];
     // Hide only when we are mid-discovery (no My Domain selected yet).
     if (isDiscoveryHost && !coordinator.domainUpdated) {
         return NO;
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
@@ -42,6 +42,9 @@
 #import "SalesforceSDKManager+Internal.h"
 #import <LocalAuthentication/LocalAuthentication.h>
 #import "SFRestAPI+Internal.h"
+#import "SFOAuthCoordinator+Internal.h"
+#import "SFSDKAuthSession.h"
+#import "SFSDKAuthRequest.h"
 #import <SalesforceSDKCore/SalesforceSDKCore-Swift.h>
 @interface SFLoginViewController () <SFSDKLoginHostDelegate, SFUserAccountManagerDelegate>
 
@@ -323,14 +326,29 @@
     }
 
     // Login for Admin - forces browser-based (advanced) authentication to support phishing-resistant MFA.
-    [menuActions addObject:[UIAction actionWithTitle:[SFSDKResourceUtils localizedString:@"LOGIN_FOR_ADMIN"]
+    // Wrapped in an uncached UIDeferredMenuElement so the show/hide predicate is re-evaluated each
+    // time the menu opens. The entry is hidden during phase 1 of Welcome Discovery (i.e. before the
+    // user has selected an account on welcome.salesforce.com/discovery), where we have no resolved
+    // My Domain to launch the browser session against.
+    __weak typeof(self) weakSelf = self;
+    UIDeferredMenuElement *loginForAdminElement = [UIDeferredMenuElement elementWithUncachedProvider:^(void (^ _Nonnull completion)(NSArray<UIMenuElement *> * _Nonnull)) {
+        __strong typeof(weakSelf) strongSelf = weakSelf;
+        if (![SFLoginViewController shouldShowLoginForAdminForSession:[strongSelf currentAuthSessionForMenu]]) {
+            completion(@[]);
+            return;
+        }
+        UIAction *action = [UIAction actionWithTitle:[SFSDKResourceUtils localizedString:@"LOGIN_FOR_ADMIN"]
                                                image:nil
                                           identifier:nil
                                              handler:^(__kindof UIAction* _Nonnull action) {
-        if ([self.delegate respondsToSelector:@selector(loginViewControllerDidSelectLoginForAdmin:)]) {
-            [self.delegate loginViewControllerDidSelectLoginForAdmin:self];
-        }
-    }]];
+            __strong typeof(weakSelf) handlerSelf = weakSelf;
+            if ([handlerSelf.delegate respondsToSelector:@selector(loginViewControllerDidSelectLoginForAdmin:)]) {
+                [handlerSelf.delegate loginViewControllerDidSelectLoginForAdmin:handlerSelf];
+            }
+        }];
+        completion(@[action]);
+    }];
+    [menuActions addObject:loginForAdminElement];
 
     UIMenu *menu = [UIMenu menuWithTitle:@"" // No title
                                 children:menuActions];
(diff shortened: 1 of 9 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -323,14 +326,29 @@ - (UIBarButtonItem *)createSettingsButton {
     }
 
     // Login for Admin - forces browser-based (advanced) authentication to support phishing-resistant MFA.
-    [menuActions addObject:[UIAction actionWithTitle:[SFSDKResourceUtils localizedString:@"LOGIN_FOR_ADMIN"]
+    // Wrapped in an uncached UIDeferredMenuElement so the show/hide predicate is re-evaluated each
+    // time the menu opens. The entry is hidden during phase 1 of Welcome Discovery (i.e. before the
```
wmathurin, line 330: So we are hiding the menu during phase 1, do we expect admin to look for that menu in phase 2 or do we expect that to configure the correct login server manually if they need to login as admin?

brandonpage: Hide in phase 1 so we don't have to deal with the discovery callback and the experience can be the same per platform (easier to document).

When the customer selects the org they want either the My Domain is configured to trigger Adv Auth for everyone or it loads in the WebView and the option to "Login for Admin" is in the menu.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
```diff
@@ -340,6 +358,33 @@ - (UIBarButtonItem *)createSettingsButton {
...
+    // Default to showing the entry when we have no session/coordinator info.
+    // This matches the previous (always-shown) behavior on non-discovery flows
+    // and during early SDK lifecycle before the auth session is wired up.
+    SFOAuthCoordinator *coordinator = session.oauthCoordinator;
+    if (session == nil || coordinator == nil) {
+        return YES;
+    }
+
+    NSString *loginHost = session.oauthRequest.loginHost;
+    SFDomainDiscoveryCoordinator *discoveryCoordinator = [[SFDomainDiscoveryCoordinator alloc] init];
```
wmathurin: A fresh `SFDomainDiscoveryCoordinator` is allocated here purely to call `isDiscoveryDomain:`, and the same pattern is repeated in `SFUserAccountManager.m` at the `loginViewControllerDidSelectLoginForAdmin:` guard. If `isDiscoveryDomain:` is stateless (just a string predicate, no mutable coordinator state needed), consider extracting it to a class method — e.g. `+[SFDomainDiscoveryCoordinator isDiscoveryDomain:]` — so callers don't need to alloc a throw-away instance. Minor, but two independent allocs for the same check at independent call sites looks slightly odd to a reader.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/LoginForAdminTests.swift
```diff
@@ -46,12 +46,12 @@ class LoginForAdminTests: XCTestCase {
 
     // MARK: - SFSDKAuthRequest loginAsAdmin Property
 
-    func testGivenNewAuthRequest_whenCreated_thenLoginAsAdminIsFalse() {
+    func testGivenNewAuthRequest_whenCreated_thenloginAsAdminIsFalse() {
```
wmathurin: Nit: the renamed test methods have a lowercase `l` in the middle of the method name (e.g. `testGivenloginAsAdmin_...`, `testGivenNologinAsAdmin_...`), which reads as a typo. The intent seems to be matching the ObjC property name casing, but `LoginAsAdmin` with a capital `L` would be more legible in a test name. Up to you, but worth a pass to recapitalize if you agree.

brandonpage: Good catch.  I originally replaced `loginAsAdmin` to `loginForAdmin` in a bunch of places but later changed it back to make cherry-picking cleaner.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DomainDiscoveryCoordinator.swift
```diff
@@ -100,6 +100,11 @@ public class DomainDiscoveryCoordinator: NSObject {
     
     @objc
     public func isDiscoveryDomain(_ domain: String?) -> Bool {
```
bbirman, line 103: I like the class one! Could we deprecate this one too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/SFLoginViewController.m
brandonpage: Done.  This adds new public API, but I don't think that is too big of a deal.

wmathurin (approved):

PR #4047, opened 2026-05-29
Update nightly schedules to weekdays and fix security issues

brandonpage:
## Summary

Implements [W-22712082 — Cleanup CI for all Repos](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002b3viMYAQ/view).

**Schedule changes**:
- **Unit nightly** (`nightly.yaml`) → **every weekday at 9 PM PT** (`0 5 * * 2-6` UTC), up from Tue/Thu-only.
- **UI nightly** (`ui-test-nightly.yaml`) → **every weekday at 11 PM PT** (`0 7 * * 2-6` UTC), up from Mon/Wed-only.

Both schedules now catch regressions within 1 business day instead of up to 4.

**Security hardening**: All workflows updated to follow the [GitHub Actions injection-prevention best practices](https://github.blog/security/vulnerability-research/how-to-catch-github-actions-workflow-injections-before-attackers-do/):

- **All third-party actions SHA-pinned** with the resolved tag in a comment.
- **Top-level `permissions:` block** added to every workflow: `contents: read` baseline, `pull-requests: write` only where Danger / `mikepenz/action-junit-report` posts PR comments.
- **`secrets: inherit` replaced with explicit secret pass-through**; reusable workflows now declare expected secrets in `on: workflow_call: secrets:`.
- **All `${{ ... }}` shell interpolation** in `run:` blocks refactored to `env:` variables with quoted shell expansions.
- **`pull_request_target`** retained with `# zizmor: ignore[dangerous-triggers]` and an inline comment documenting the Member Check mitigation.
- **`actions/checkout` steps** set `with: persist-credentials: false`.
- **xcodebuild logs archived on build failure** (xcactivitylog) for diagnostic purposes.

After this change, `zizmor --offline` reports 0 High-confidence findings across all six CI workflows.

## Test plan

- [x] Verified locally with `python3 yaml.safe_load`, `actionlint -shellcheck=`, `zizmor --offline`. All clean.
- [x] CI verification: opened test PR on personal fork (`brandonpage/SalesforceMobileSDK-iOS#23`) targeting the same `cleanup-ci-w22712082` branch as this PR. The test PR triggers the new workflows against a real source change in `SalesforceSDKCommon`. **No regressions observed** — see linked PR for run details.
- [ ] Reviewer to confirm the `permissions:` blocks match the team's expected privilege levels for each workflow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4079, opened 2026-06-20
[iOS] Grant statuses:write so Danger can post required PR commit statuses

sfdctaka:
## Summary
- Add `statuses: write` to the `static-analysis` and `test-orchestrator` job-level permission blocks in `.github/workflows/pr.yaml`.
- Both jobs run Danger to post the `danger/StaticAnalysis` and `danger/TestOrchestrator` commit statuses required by `dev`'s branch protection. Without `statuses: write`, Danger's status calls are rejected (`Danger does not have write access to the PR to set a PR status.`), so the required statuses never appear and PRs cannot merge even when every check completes successfully.
- Two-line diff, one new permission scope per job.

## Note (the irony)
This very PR will hit the same blocker until it merges. The required `danger/StaticAnalysis` and `danger/TestOrchestrator` commit statuses will not appear on this PR's head commit, because the workflow run evaluating it is still running with the old (insufficient) permissions. The merge button will stay disabled even after approval. Someone with admin rights will likely need to bypass branch protection once to land this. After that, every subsequent PR run picks up the new permission and unblocks automatically.

## Test plan
- [ ] After merge, observe that the next PR run posts both `danger/StaticAnalysis` and `danger/TestOrchestrator` commit statuses on its head commit.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -20,6 +20,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+      statuses: write
     env:
       BUNDLE_GEMFILE: ${{ github.workspace }}/.github/DangerFiles/Gemfile
     steps:
@@ -77,6 +78,7 @@ jobs:
     permissions:
       contents: read
       pull-requests: write
+      statuses: write
     env:
       BUNDLE_GEMFILE: ${{ github.workspace }}/.github/DangerFiles/Gemfile
     outputs:


bbirman:
What changed for it to stop working?

sfdctaka:
> What changed for it to stop working?

In #4047 (2026-05-29, "fix security issues") explicit per-job permissions: blocks were added to harden the pull_request_target workflow:
```
  permissions:
    contents: read
    pull-requests: write
```
The intent was correct, but in GitHub Actions any job-level permissions: block replaces the default GITHUB_TOKEN scopes rather than adding to them — anything not listed is dropped. The default token had implicit statuses: write; the new block doesn't, so Danger silently lost the ability to post commit statuses. Its log line is:
  
  ▎ Danger does not have write access to the PR to set a PR status.

The two required contexts (danger/StaticAnalysis, danger/TestOrchestrator) have not been posted on PR head commits since that merge. We didn't notice because dev branch protection has enforce_admins: false, so admin merges have been bypassing the missing required statuses — masking the regression for ~3 weeks. Non-admin PRs
(and admin PRs not bypassed) sit blocked.

  This PR adds statuses: write back to the two job permission blocks. Two-line change, no other behavior affected.

bbirman (approved):
lgtm but I'd want @brandonpage's approval too

sfdctaka:
@brandonpage Can you check this PR please?

wmathurin (approved):

brandonpage (approved):

PR #4087, opened 2026-06-26
Add token refresh coordinator

bbirman:
- Adds a layer to coordinate token refreshes so that only a single refresh is requested at a time across `SFRestAPI`,  `SFIdentityCoordinator` and `SFUserAccountManager`
- Consumers calling refresh directly should go through `SFUserAccountManager`
- Deprecates `SFOAuthSessionRefresher` so that it will be internal only in the future

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/UserAccountManager.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/UserAccountManager.swift
@@ -71,15 +71,27 @@ extension UserAccountManager: UserAccountManaging {
         }
    }
 
-    /// Kick off the login process for jwt token with credentials previously configured.
+    /// Refresh the session for the given credentials.
     /// - Parameters:
-    ///   - credentials: the OAuthCredentials object
+    ///   - credentials: The credentials to refresh.
     ///   - completionBlock: completion block to invoke with a success tuple (UserAccount, AuthInfo) or   UserAccountManagerError for failure wrapped in a Result type.
    public func refresh(credentials: OAuthCredentials, _ completionBlock: @escaping (Result<(UserAccount, AuthInfo), UserAccountManagerError>) -> Void) -> Bool {
         return __refreshCredentials(credentials, completion: { (authInfo, userAccount) in
             completionBlock(Result.success((userAccount,authInfo)))
         }) { (authInfo, error) in
             completionBlock(Result.failure(.refreshFailed(underlyingError: error, authInfo: authInfo)))
+        }
+    }
+
+    /// Refresh the session for the given credentials.
+    /// - Parameter credentials: The credentials to refresh.
+    /// - Returns: A tuple of the updated `UserAccount` and `AuthInfo` on success.
+    /// - Throws: `UserAccountManagerError.refreshFailed` if the refresh fails.
+    public func refresh(credentials: OAuthCredentials) async throws -> (UserAccount, AuthInfo) {
+        try await withCheckedThrowingContinuation { continuation in
+            _ = self.refresh(credentials: credentials) { result in
+                continuation.resume(with: result.mapError { $0 })
+            }
         }
     }
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
@@ -25,7 +25,7 @@
 #import <SalesforceSDKCommon/SFJsonUtils.h>
 #import "SFIdentityCoordinator+Internal.h"
 #import "SFOAuthCredentials.h"
-#import "SFOAuthSessionRefresher.h"
+#import "SFSDKTokenRefreshCoordinator.h"
 #import "SFUserAccountManager.h"
 #import "SFNetwork.h"
 #import "SFSDKAuthSession.h"
@@ -66,7 +66,6 @@ static NSString * const kSFIdentityDataPropertyKey            = @"com.salesforce
 @synthesize timeout = _timeout;
 @synthesize retrievingData = _retrievingData;
 @synthesize session = _session;
-@synthesize oauthSessionRefresher = _oauthSessionRefresher;
 
 #pragma mark - init / dealloc
 
@@ -171,8 +170,7 @@ static NSString * const kSFIdentityDataPropertyKey            = @"com.salesforce
         if (statusCode == 401 || statusCode == 403) {
             // The session timed out.  Identity service tends to send 403s for session timeouts.  Try to refresh.
             [SFSDKCoreLogger i:[self class] format:@"%@: Identity request failed due to expired credentials.  Attempting to refresh credentials.", NSStringFromSelector(_cmd)];
-            strongSelf.oauthSessionRefresher = [[SFOAuthSessionRefresher alloc] initWithCredentials:strongSelf.credentials];
-            [strongSelf.oauthSessionRefresher refreshSessionWithCompletion:^(SFOAuthCredentials *updatedCredentials) {
+            [[SFSDKTokenRefreshCoordinator sharedInstance] refreshSessionForCredentials:strongSelf.credentials completion:^(SFOAuthCredentials *updatedCredentials) {
                 [SFSDKCoreLogger d:[strongSelf class] format:@"%@: Credentials refresh successful.  Replaying original identity request.", NSStringFromSelector(_cmd)];
                 strongSelf.credentials = updatedCredentials;
                 dispatch_async(dispatch_get_main_queue(), ^{
@@ -221,7 +219,6 @@ static NSString * const kSFIdentityDataPropertyKey            = @"com.salesforce
     self.session = nil;
     self.credentials = nil;
     self.idData = nil;
-    self.oauthSessionRefresher = nil;
 }
 
 - (void)cleanupData
@@ -229,7 +226,6 @@ static NSString * const kSFIdentityDataPropertyKey            = @"com.salesforce
     [SFNetwork removeSharedInstanceForIdentifier:self.networkIdentifier];
     self.networkIdentifier = nil;
     self.session = nil;
-    self.oauthSessionRefresher = nil;
     self.retrievingData = NO;
 }
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
@@ -118,7 +118,7 @@ SFSDK_USE_DEPRECATED_BEGIN
         SFOAuthInfo *authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeRefresh];
         [userInfo setValue:authInfo forKey:kSFNotificationUserInfoAuthTypeKey];
         [[NSNotificationCenter defaultCenter] postNotificationName:kSFNotificationUserDidRefreshToken
-                                                            object:self
+                                                            object:[SFUserAccountManager sharedInstance]
                                                           userInfo:userInfo];
         self.completionBlock(self.credentials);
     }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
@@ -22,6 +22,12 @@
  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+
+#import "SalesforceSDKConstants.h"
+
+// TODO: Remove when the class is internal in Mobile SDK 15.0
+SFSDK_USE_DEPRECATED_BEGIN
+
 #import "SFOAuthSessionRefresher+Internal.h"
 #import "SFUserAccountManager.h"
 #import "SFOAuthCredentials+Internal.h"
@@ -102,19 +108,19 @@
 - (void)completeWithSuccess {
     [SFSDKCoreLogger i:[self class] format:@"%@ Session was successfully refreshed.", NSStringFromSelector(_cmd)];
     if (self.completionBlock) {
-        dispatch_async(dispatch_get_main_queue(), ^{
-            SFUserAccount *account = [[SFUserAccountManager sharedInstance] accountForCredentials:self.credentials];
-            NSMutableDictionary *userInfo = [NSMutableDictionary new];
-            if (account) {
-                [userInfo setValue:account forKey:kSFNotificationUserInfoAccountKey];
-            } else {
-                [SFSDKCoreLogger e:[self class] format:@"%@ No account for credentials", NSStringFromSelector(_cmd)];
-            }
-            [[NSNotificationCenter defaultCenter] postNotificationName:kSFNotificationUserDidRefreshToken
-                                                                object:self
-                                                              userInfo:userInfo];
-            self.completionBlock(self.credentials);
-        });
+        SFUserAccount *account = [[SFUserAccountManager sharedInstance] accountForCredentials:self.credentials];
+        NSMutableDictionary *userInfo = [NSMutableDictionary new];
+        if (account) {
+            [userInfo setValue:account forKey:kSFNotificationUserInfoAccountKey];
+        } else {
+            [SFSDKCoreLogger e:[self class] format:@"%@ No account for credentials", NSStringFromSelector(_cmd)];
+        }
+        SFOAuthInfo *authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeRefresh];
+        [userInfo setValue:authInfo forKey:kSFNotificationUserInfoAuthTypeKey];
+        [[NSNotificationCenter defaultCenter] postNotificationName:kSFNotificationUserDidRefreshToken
+                                                            object:self
+                                                          userInfo:userInfo];
+        self.completionBlock(self.credentials);
     }
 }
 
@@ -122,9 +128,7 @@
     [SFSDKCoreLogger e:[self class] format:@"%@ Refresh failed with error: %@", NSStringFromSelector(_cmd), error];
 
     if (self.errorBlock) {
-        dispatch_async(dispatch_get_main_queue(), ^{
-            self.errorBlock(error);
-        });
+        self.errorBlock(error);
     }
 }
 - (void)oauthCoordinator:(SFOAuthCoordinator *)coordinator didBeginAuthenticationWithSession:(ASWebAuthenticationSession *)session {
@@ -148,3 +152,5 @@
 }
 
 @end
+
+SFSDK_USE_DEPRECATED_END
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKTokenRefreshCoordinator.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKTokenRefreshCoordinator.h
@@ -1,0 +1,72 @@
+/*
+ Copyright (c) 2025-present, salesforce.com, inc. All rights reserved.
+
+ Redistribution and use of this software in source and binary forms, with or without modification,
+ are permitted provided that the following conditions are met:
+ * Redistributions of source code must retain the above copyright notice, this list of conditions
+ and the following disclaimer.
+ * Redistributions in binary form must reproduce the above copyright notice, this list of
+ conditions and the following disclaimer in the documentation and/or other materials provided
+ with the distribution.
+ * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+ endorse or promote products derived from this software without specific prior written
+ permission of salesforce.com, inc.
+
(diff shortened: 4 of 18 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/UserAccountManager.swift
```diff
@@ -83,6 +83,18 @@ extension UserAccountManager: UserAccountManaging {
         }
     }
 
+    /// Refresh the session for the given credentials.
+    /// - Parameter credentials: The credentials to refresh.
+    /// - Returns: A tuple of the updated `UserAccount` and `AuthInfo` on success.
+    /// - Throws: `UserAccountManagerError.refreshFailed` if the refresh fails.
+    public func refresh(credentials: OAuthCredentials) async throws -> (UserAccount, AuthInfo) {
```
bbirman, line 90: Added async wrapper for refreshing

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
```diff
@@ -102,29 +108,27 @@ - (void)refreshSessionWithCompletion:(void (^)(SFOAuthCredentials *))completionB
 - (void)completeWithSuccess {
     [SFSDKCoreLogger i:[self class] format:@"%@ Session was successfully refreshed.", NSStringFromSelector(_cmd)];
     if (self.completionBlock) {
-        dispatch_async(dispatch_get_main_queue(), ^{
```
bbirman, line 105: Note: I had the original spec keeping the main queue to match existing code but the old code actually has a mix of main and non-main dispatches. Main was causing a deadlock in the tests when everything was unified so I changed to be thread agnostic

wmathurin: {'reply': 'Thanks for the context — that explains the removal. If the contract is now "thread-agnostic", it\'s worth updating the header doc on `SFSDKTokenRefreshCoordinator` to drop the *"dispatched on the main queue"* claim, and doing a quick scan of call sites that may be implicitly relying on main-thread delivery (UI updates, notification observers, etc.).'}

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Extensions/UserAccountManager.swift
wmathurin: {'reply': None}

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKTokenRefreshCoordinator.h
```diff
@@ -0,0 +1,72 @@
...
+@property (class, nonatomic, readonly) SFSDKTokenRefreshCoordinator *sharedInstance;
+
+/**
+ * Request a token refresh for the given credentials.
+ *
+ * If a refresh is already in-flight for these credentials (keyed by `credentials.identifier`),
+ * the callbacks are appended to the waiting list and no new network request is made.
+ * When the single in-flight refresh completes, all registered callbacks receive the same result.
+ *
+ * Completion and error callbacks are dispatched on the main queue.
```
wmathurin, line 54: **Thread contract mismatch:** This says callbacks are dispatched on the main queue, but `completeRefreshForKey:credentials:error:` calls them directly on the serial queue — no main-queue hop. The old `SFOAuthSessionRefresher.completeWithSuccess` wrapped callbacks in `dispatch_async(dispatch_get_main_queue(), ...)`, which this PR removes.

Callers updating UI or expecting main-thread delivery will now run on a background thread. Either dispatch in `completeRefreshForKey:`, or drop this claim from the header and audit call sites.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
```diff
@@ -171,8 +170,7 @@ - (void)sendRequest
         if (statusCode == 401 || statusCode == 403) {
             // The session timed out.  Identity service tends to send 403s for session timeouts.  Try to refresh.
             [SFSDKCoreLogger i:[self class] format:@"%@: Identity request failed due to expired credentials.  Attempting to refresh credentials.", NSStringFromSelector(_cmd)];
-            strongSelf.oauthSessionRefresher = [[SFOAuthSessionRefresher alloc] initWithCredentials:strongSelf.credentials];
-            [strongSelf.oauthSessionRefresher refreshSessionWithCompletion:^(SFOAuthCredentials *updatedCredentials) {
+            [[SFSDKTokenRefreshCoordinator sharedInstance] refreshSessionForCredentials:strongSelf.credentials completion:^(SFOAuthCredentials *updatedCredentials) {
```
wmathurin, line 173: **Refresh errors are silently swallowed.** No `error:` block is provided, so a failed credential refresh (network error, `invalid_grant`, background expiry) produces no log and no recovery. At minimum:
```objc
error:^(NSError *error) {
    [SFSDKCoreLogger e:[self class] format:@"%@: Credential refresh failed: %@",
     NSStringFromSelector(_cmd), error];
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -449,47 +447,43 @@ - (NSError*) prepareErrorForDelegate:(NSData *)data response:(NSURLResponse *)re
...
-                if ([refreshError.domain isEqualToString:kSFOAuthErrorDomain] && refreshError.code == kSFOAuthErrorInvalidGrant) {
-                    [SFSDKCoreLogger i:[strongSelf class] format:@"%@ Invalid grant error received, triggering logout.", NSStringFromSelector(_cmd)];
-                    
-                    // Make sure we call logout on the main thread.
-                    dispatch_async(dispatch_get_main_queue(), ^{
-                        [[SFUserAccountManager sharedInstance] logoutUser:strongSelf.user reason:SFLogoutReasonTokenExpired];
-                    });
-                }
-            }];
+        if (self.refreshCycleActive) {
```
wmathurin, line 457: **`refreshCycleActive` is not reset by `cleanup`, creating a timing gap.** If `cleanup` runs while a refresh is in-flight, `refreshCycleActive` stays `YES`. Any new request enqueued after cleanup but before the coordinator callback resets the flag will hit this guard and silently skip triggering a refresh.

`testNewRefreshCyclePossibleAfterCleanup` avoids this by manually setting `refreshCycleActive = NO` rather than going through the real timing path. Consider resetting the flag inside `cleanup` after clearing `activeRequests`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
```diff
@@ -252,11 +252,9 @@ - (void)accessTokenForApprovalCode:(SFSDKOAuthTokenEndpointRequest *)endpointReq
                 endpointResponse = [[SFSDKOAuthTokenEndpointResponse alloc] initWithError:error];
             }
             [SFSDKCoreLogger d:[strongSelf class] format:@"SFOAuth2 session failed with error: error code: %ld, description: %@, URL: %@", (long)error.code, [error localizedDescription], errorUrlString];
-            dispatch_async(dispatch_get_main_queue(), ^{
-                if (completionBlock) {
-                    completionBlock(endpointResponse);
-                }
-            });
+            if (completionBlock) {
+                completionBlock(endpointResponse);
```
wmathurin, line 256: **Main-queue removal affects all callers of `SFSDKOAuth2`, not just the coordinator path.** The `dispatch_async(dispatch_get_main_queue(), ...)` wrapper was removed from error-path completions in both `accessTokenForApprovalCode:` and `accessTokenForRefresh:`. Any direct caller of these methods that touches UI in the completion block will now run on a background URL session thread.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKAuthUtilTests.swift
```diff
@@ -170,4 +171,127 @@ class SFSDKAuthUtilTests: XCTestCase {
...
+        let request = try XCTUnwrap(RestClient.shared.requestForRevokeAccessToken())
+        _ = try await RestClient.shared.send(request: request)
+
+        let (account, _) = try await UserAccountManager.shared.refresh(credentials: credentials)
+
+        XCTAssertNotNil(account.credentials.accessToken)
+        XCTAssertNotEqual(account.credentials.accessToken, originalAccessToken,
+                          "Access token should be rotated after refresh")
+        XCTAssertNotNil(account.credentials.refreshToken)
+        print("original refresh token: \(originalRefreshToken)")
```
wmathurin: Debug `print()` statements — looks unintentional.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKAuthUtilTests.swift
```diff
@@ -170,4 +171,127 @@ class SFSDKAuthUtilTests: XCTestCase {
...
+        }
+    }
+
+    func testSingleNotificationForConcurrentRefreshes() async throws {
+        let credentials = try XCTUnwrap(currentUser?.credentials)
+
+        let notificationCount = Mutex<Int>(0)
+
+        let observer = NotificationCenter.default.addObserver(
+            forName: .init(rawValue: "SFNotificationOAuthUserDidRefreshToken"),
```
wmathurin: **Fragile raw notification name string.** Observing `"SFNotificationOAuthUserDidRefreshToken"` instead of `kSFNotificationUserDidRefreshToken` means if the constant value ever changes this test silently vacuously passes. Use the typed constant.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKTokenRefreshCoordinatorTests.m
```diff
@@ -0,0 +1,599 @@
...
+        XCTAssertNotNil(error);
+        [errorExp fulfill];
+    }];
+
+    [self waitForExpectationsWithTimeout:2.0 handler:nil];
+}
+
+#pragma mark - Integration Test (Mock at authClient level)
+
+- (void)testIntegrationSingleNetworkCallForConcurrentRefreshes {
```
wmathurin, line 542: **Integration test assertion may be vacuously true.** `SFOAuthSessionRefresher` constructs its own `SFOAuthCoordinator` internally and calls `accessTokenForRefresh:` directly — it does not go through `SFUserAccountManager.authClient()`. With `realCoordinator.refresherFactory = nil` (real refresher path), `mockClient.accessTokenForRefreshCallCount` likely stays 0 regardless of how many network calls go out, making the key assertion (`== 1`) always pass.

Suggest adding `XCTAssertGreaterThan(mockClient.accessTokenForRefreshCallCount, 0, @"mock was never called")` before the coalescing assertion to confirm the mock is actually hit.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
```diff
@@ -102,29 +108,27 @@ - (void)refreshSessionWithCompletion:(void (^)(SFOAuthCredentials *))completionB
...
+        SFUserAccount *account = [[SFUserAccountManager sharedInstance] accountForCredentials:self.credentials];
+        NSMutableDictionary *userInfo = [NSMutableDictionary new];
+        if (account) {
+            [userInfo setValue:account forKey:kSFNotificationUserInfoAccountKey];
+        } else {
+            [SFSDKCoreLogger e:[self class] format:@"%@ No account for credentials", NSStringFromSelector(_cmd)];
+        }
+        SFOAuthInfo *authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeRefresh];
+        [userInfo setValue:authInfo forKey:kSFNotificationUserInfoAuthTypeKey];
+        [[NSNotificationCenter defaultCenter] postNotificationName:kSFNotificationUserDidRefreshToken
```
wmathurin, line 120: **Notification sender has changed.** The old `refreshCredentials:` in `SFUserAccountManager` posted `kSFNotificationUserDidRefreshToken` with `object:strongSelf` (the account manager). Now this posts with `object:self` (the refresher). Any observer filtering on `notification.object` — e.g. `addObserver:selector:name:object:accountManager` — will silently stop receiving notifications.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Identity/SFIdentityCoordinator.m
bbirman: The error block should be unchanged at line 182, unless I'm misunderstanding?

wmathurin: You are right, I should have reviewed better the Claude generated review.

bbirman:
Updated based on feedback and after some more deliberating, moved the completion blocks back to the main thread

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKTokenRefreshCoordinator.h
wmathurin: {'reply': "Fixed — `completeRefreshForKey:` now wraps all callbacks in `dispatch_async(dispatch_get_main_queue(), ...)`, matching the header's contract. Both the success and error paths are covered.", 'resolve': True}

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
wmathurin: {'reply': 'Fixed — `cleanup` now sets `self.refreshCycleActive = NO` inside the `@synchronized` block after clearing `activeRequests`, closing the timing gap.', 'resolve': True}

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
wmathurin: {'reply': 'The coordinator path is fully covered — `completeRefreshForKey:` dispatches all callbacks (success and error) to the main queue, so `SFRestAPI` and `SFUserAccountManager` callers are safe.\n\nTwo narrower gaps remain worth noting:\n\n1. **`accessTokenForApprovalCode:` success path** — this flows through `SFOAuthCoordinator` → `notifyDelegateOfSuccess:`, which calls the delegate directly on the current thread (no main-queue hop). `notifyDelegateOfFailure:` does dispatch to main, so only the success path is affected. The previously-removed `dispatch_async(main_queue)` in `SFSDKOAuth2` was the protection here.\n\n2. **Direct callers of `SFOAuthSessionRefresher`** — the refresher is still public (deprecated, not internal until 15.0). Both `completeWithSuccess` and `completeWithError:` had their main-queue dispatches removed. Any third-party consumer calling the refresher directly and updating UI in a callback block will silently get background-thread delivery. Worth a note in the deprecation comment or migration guide.\n\nResolving since neither gap is in the new coordinator path, but flagging for 15.0 cleanup.', 'resolve': True}

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKTokenRefreshCoordinatorTests.m
wmathurin: {'reply': 'Fixed — the guard assertion `XCTAssertGreaterThan(mockClient.accessTokenForRefreshCallCount, 0, ...)` was added. Also confirmed that `SFOAuthSessionRefresher` calls `[SFUserAccountManager sharedInstance].authClient()` directly, so the mock substitution is on the right seam and will be genuinely exercised.', 'resolve': True}

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
wmathurin: {'reply': 'Fixed — `completeWithSuccess` now posts with `object:[SFUserAccountManager sharedInstance]`, restoring the original sender. The test mock aligns with this as well.', 'resolve': True}

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -101,7 +98,22 @@ - (void)dealloc {
 #pragma mark - Cleanup / cancel all
 
 - (void)cleanup {
-    [self.activeRequests removeAllObjects];
+    @synchronized (self) {
+        NSSet *pendingRequests = [self.activeRequests asSet];
+        for (SFRestRequest *request in pendingRequests) {
+            NSURLSessionDataTask *oldTask = request.sessionDataTask;
+            request.sessionDataTask = nil;
+            [oldTask cancel];
```
sfdctaka, line 110: nit: Assuming these lines were done for this reason maybe we could put comment for them?
```
  // Nil sessionDataTask BEFORE cancelling: [cancel] schedules the URLSession
  // delegate callback asynchronously; clearing the property first ensures the
  // callback can't re-invoke failureBlock after we've already delivered it below.
```

bbirman: 👍 updated!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFSDKTokenRefreshCoordinator.m
```diff
@@ -0,0 +1,206 @@
...
+        if (!entry) return;
+
+        [SFSDKCoreLogger w:[self class] format:@"Background task expired during token refresh for credential %@. Delivering cancellation error.", key];
+
+        NSError *bgError = [NSError errorWithDomain:@"SFSDKTokenRefreshCoordinator"
+                                               code:-2
+                                           userInfo:@{NSLocalizedDescriptionKey: @"Token refresh interrupted: app background time expired"}];
+        [self completeRefreshForKey:key credentials:nil error:bgError];
+    });
+}
```
sfdctaka, line 204: Could this happen? 

When this handler fires, we tell every waiting caller "refresh failed" and remove the entry — but we don't cancel the underlying `SFOAuthSessionRefresher`. It has no cancel API, and its `NSURLSessionDataTask` is captured only inside the completion block, so we can't reach it.

What happens next:
1. iOS pulls background time → we fabricate the -2 error and dispatch it to waiters.
2. App foregrounds later, the token endpoint response finally arrives.
3. `SFOAuthSessionRefresher.completeWithSuccess` runs and:
  - overwrites `credentials.accessToken` / `refreshToken` in memory
  - posts `kSFNotificationUserDidRefreshToken` to every observer
  - calls back into the coordinator, which correctly no-ops because entry is gone

So the caller thinks the refresh failed, but the credentials silently rotated and a notification fired system-wide. Downstream observers (analytics, network layer, `SFRestAPI` retry) will act on a refresh nobody asked for.

Same shape via a different trigger: `SFRestAPI.cleanup` on logout clears its own `activeRequests` and cancels its `NSURLSessionDataTask`s, but doesn't touch the coordinator. If a refresh was in flight when the user logged out, it will still complete, mutate the (now-revoked) credentials object, and — because
`[SFUserAccountManager accountForCredentials:]` returns nil for a logged-out user — post the notification with userInfo missing the account key. Any observer that force-unwraps `userInfo[kSFNotificationUserInfoAccountKey]` will crash.

`testRefreshCallbackAfterCleanupIsHarmless` doesn't catch this because it asserts no double callback, not no side effects.

Suggested fix: give `SFOAuthSessionRefresher` a cancel (that cancels its `NSURLSessionDataTask`) and call it from both this handler and a new coordinator entry point invoked from `SFRestAPI.cleanup `/ user logout.

wmathurin: {'reply': 'The race you describe is real — the expiration handler removes the entry and notifies waiters of failure while the underlying `NSURLSessionDataTask` can still complete, leaving `completeWithSuccess` to mutate credentials and fire the notification after callers have already been told the refresh failed.\n\nThat said, a full cancel API on `SFOAuthSessionRefresher` is heavier than needed. The simplest targeted fix is a `cancelled` BOOL on `SFOAuthSessionRefresher` (or on `SFSDKTokenRefreshEntry`): the expiration handler sets it before removing the entry, and `completeWithSuccess`/`completeWithError:` guard on it as their first line. Zero cancellation plumbing required — the network call runs to completion but its side effects (credential mutation, notification post) are suppressed.\n\nFor the logout path: `SFRestAPI.cleanup` could call a new `cancelRefreshForCredentials:` on the coordinator, which sets the flag and delivers a cancellation error to any waiters. That covers the "credentials already revoked" crash scenario without requiring `SFRestAPI` to know about `NSURLSessionDataTask` internals.\n\nThe two-line flag check in `completeWithSuccess`/`completeWithError:` is the load-bearing part — the rest is bookkeeping.', 'resolve': False}

sfdctaka (approved):
LGTM!

wmathurin (approved):

PR #4088, opened 2026-06-26
[iOS] Validate login host input, recover to previous host on failure

sfdctaka:
## Summary

Three coordinated fixes in the SalesforceSDKCore iOS login flow:

1. **Validate host input.** `NewLoginHostView.save(host:hostLabel:)` rejects empty/whitespace/syntactically-invalid input (non-empty after trim, contains `.`, no internal whitespace, parses via `URL(string:)`) and shows an inline red error below the host field. No persistence, no dismiss, no `saveAction` call on rejection.
2. **Snapshot previous host.** `SFUserAccountManager.hostListViewController:didChangeLoginHost:` now records the prior `loginHost` to a new internal-only `previousLoginHost` property before applying the change.
3. **Smarter recovery + auto-cleanup.** When `hostConnectionErrorHandlerBlock`'s OK completion fires, the SDK now: (a) removes the failing host from `SFSDKLoginHostStorage` if it is `isDeletable == YES` (built-in/MDM hosts are never removed), (b) restarts auth on `previousLoginHost` if still in storage, falling back to index-0 otherwise. The fallback is guarded against an empty-storage edge case (MDM `onlyShowAuthorizedHosts == YES` with empty MDM list, or the deleted bad host being the only entry) — logs and bails rather than raising `NSRangeException`.

The alert UX is unchanged: single OK button, same title/message format keys.

## Files changed

- `Classes/Login/LoginHost/NewLoginHostView.swift` — validation predicate + `hostError` state + inline SwiftUI `Text` rendering.
- `Classes/UserAccount/SFUserAccountManager+Internal.h` — new `previousLoginHost` internal property.
- `Classes/UserAccount/SFUserAccountManager.m` — snapshot in `didChangeLoginHost:`; rewrite of `hostConnectionErrorHandlerBlock` OK completion.
- `SalesforceSDKCoreTests/NewLoginHostTests.swift` — 6 new test cases (empty, whitespace, no-dot, internal whitespace, invalid scheme, my-domain accept).
- `SalesforceSDKCoreTests/SFUserAccountManagerLoginHostRecoveryTests.m` — new XCTest fixture with 6 handler-recovery scenarios (snapshot, prev-in-storage, prev-nil, prev-missing, deletable-removed, non-deletable-kept).
- `SalesforceSDKCore.xcodeproj/project.pbxproj` — 4 entries wiring the new `.m` into the test target.
- `shared/resources/SalesforceSDKResources.bundle/en.lproj/Localizable.strings` — adds `LOGIN_INVALID_HOST`.

## Localization

New user-facing string added: `LOGIN_INVALID_HOST` = `"Please enter a valid host (for example, login.salesforce.com)."` — **translation needed for all non-English locales.**

## Public API

No public API changes. `previousLoginHost` lives in `SFUserAccountManager+Internal.h` only. No public header outside `+Internal.h` was touched.

## Test plan

- [x] `xcodebuild test -scheme SalesforceSDKCore -only-testing:SalesforceSDKCoreTests/NewLoginHostTests` — 10/10 pass
- [x] `xcodebuild test -scheme SalesforceSDKCore -only-testing:SalesforceSDKCoreTests/SFUserAccountManagerLoginHostRecoveryTests` — 6/6 pass
- [x] Manual simulator repro: working host → "+ Add Connection" → `foo` blocked inline with red error; `mobliesdk...salesforce.com` (typo) → OAuth fails → tap OK → typo host is gone from picker AND active host is back to the working host.

## Reviewer attention

- This touches the login UI flow + `Localizable.strings` — escalation triggers per `CLAUDE.md`. Please double-check the recovery branching in `hostConnectionErrorHandlerBlock`.
- Multi-scene safety: `previousLoginHost` is a singleton property. Theoretically races with a host change in a second scene while a first scene's failure alert is open. The alert OK completion runs on the main queue, so the practical risk is low, but flag if multi-window iPad is in scope.
- Follow-up worth considering: tighter validation to reject `user:pass@evil.com` / IDN homographs.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
@@ -210,6 +210,7 @@
 		B75233DE1F4DF0B700040B6E /* SFSDKAuthPreferences.h in Headers */ = {isa = PBXBuildFile; fileRef = B75233DC1F4DF0B700040B6E /* SFSDKAuthPreferences.h */; };
 		B75233E01F4DF0B700040B6E /* SFSDKAuthPreferences.m in Sources */ = {isa = PBXBuildFile; fileRef = B75233DD1F4DF0B700040B6E /* SFSDKAuthPreferences.m */; };
 		B759CD891F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m in Sources */ = {isa = PBXBuildFile; fileRef = B759CD881F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m */; };
+		B7FF237400000002 /* SFUserAccountManagerLoginHostRecoveryTests.m in Sources */ = {isa = PBXBuildFile; fileRef = B7FF237400000001 /* SFUserAccountManagerLoginHostRecoveryTests.m */; };
 		B759CD8B1F8C10DC0081AA87 /* SFSDKErrorManagerTests.m in Sources */ = {isa = PBXBuildFile; fileRef = B759CD8A1F8C10DC0081AA87 /* SFSDKErrorManagerTests.m */; };
 		B767369120A4AB0200F04103 /* SFSDKNavigationController.h in Headers */ = {isa = PBXBuildFile; fileRef = B767368F20A4AB0200F04103 /* SFSDKNavigationController.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		B767369320A4AB0200F04103 /* SFSDKNavigationController.m in Sources */ = {isa = PBXBuildFile; fileRef = B767369020A4AB0200F04103 /* SFSDKNavigationController.m */; };
@@ -838,6 +839,7 @@
 		B75233DC1F4DF0B700040B6E /* SFSDKAuthPreferences.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; path = SFSDKAuthPreferences.h; sourceTree = "<group>"; };
 		B75233DD1F4DF0B700040B6E /* SFSDKAuthPreferences.m */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.objc; path = SFSDKAuthPreferences.m; sourceTree = "<group>"; };
 		B759CD881F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; name = SDSDKAlertMessageTest.m; path = SalesforceSDKCoreTests/SDSDKAlertMessageTest.m; sourceTree = SOURCE_ROOT; };
+		B7FF237400000001 /* SFUserAccountManagerLoginHostRecoveryTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; name = SFUserAccountManagerLoginHostRecoveryTests.m; path = SalesforceSDKCoreTests/SFUserAccountManagerLoginHostRecoveryTests.m; sourceTree = SOURCE_ROOT; };
 		B759CD8A1F8C10DC0081AA87 /* SFSDKErrorManagerTests.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; name = SFSDKErrorManagerTests.m; path = SalesforceSDKCoreTests/SFSDKErrorManagerTests.m; sourceTree = SOURCE_ROOT; };
 		B767368F20A4AB0200F04103 /* SFSDKNavigationController.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = SFSDKNavigationController.h; sourceTree = "<group>"; };
 		B767369020A4AB0200F04103 /* SFSDKNavigationController.m */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.objc; path = SFSDKNavigationController.m; sourceTree = "<group>"; };
@@ -1072,6 +1074,7 @@
 				23A4C7452D0CAFA000DF55EB /* ScreenLockManagerTests.swift */,
 				4F7EB3F71BFFC87600768720 /* SDKCommonNSDataTests.m */,
 				B759CD881F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m */,
+				B7FF237400000001 /* SFUserAccountManagerLoginHostRecoveryTests.m */,
 				010A9B511CC1A131002AF4D3 /* SFCryptoStreamTestUtils.h */,
 				010A9B521CC1A131002AF4D3 /* SFCryptoStreamTestUtils.m */,
 				4F7EB3F81BFFC87600768720 /* SFEncryptionKeyTests.m */,
@@ -2198,6 +2201,7 @@
 				8214D955205316BA0007349E /* SFSDKSalesforceAnalyticsManagerTests.m in Sources */,
 				4F06AF891C49A18E00F70798 /* NSURL+SFStringUtilsTests.m in Sources */,
 				B759CD891F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m in Sources */,
+				B7FF237400000002 /* SFUserAccountManagerLoginHostRecoveryTests.m in Sources */,
 				4F06AF8C1C49A18E00F70798 /* SalesforceSDKIdentityTests.m in Sources */,
 				4F06AF921C49A18E00F70798 /* SFPreferencesTests.m in Sources */,
 				23EED8912E2ACF3900646B10 /* MockNavigationAction.swift in Sources */,
--- a/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
@@ -210,7 +210,7 @@
 		B75233DE1F4DF0B700040B6E /* SFSDKAuthPreferences.h in Headers */ = {isa = PBXBuildFile; fileRef = B75233DC1F4DF0B700040B6E /* SFSDKAuthPreferences.h */; };
 		B75233E01F4DF0B700040B6E /* SFSDKAuthPreferences.m in Sources */ = {isa = PBXBuildFile; fileRef = B75233DD1F4DF0B700040B6E /* SFSDKAuthPreferences.m */; };
 		B759CD891F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m in Sources */ = {isa = PBXBuildFile; fileRef = B759CD881F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m */; };
-		B7FF237400000002 /* SFUserAccountManagerLoginHostRecoveryTests.m in Sources */ = {isa = PBXBuildFile; fileRef = B7FF237400000001 /* SFUserAccountManagerLoginHostRecoveryTests.m */; };
+		12B74FEE769B03E0935DE852 /* SFUserAccountManagerLoginHostRecoveryTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 06FEB66C3E3DAC225240738C /* SFUserAccountManagerLoginHostRecoveryTests.m */; };
 		B759CD8B1F8C10DC0081AA87 /* SFSDKErrorManagerTests.m in Sources */ = {isa = PBXBuildFile; fileRef = B759CD8A1F8C10DC0081AA87 /* SFSDKErrorManagerTests.m */; };
 		B767369120A4AB0200F04103 /* SFSDKNavigationController.h in Headers */ = {isa = PBXBuildFile; fileRef = B767368F20A4AB0200F04103 /* SFSDKNavigationController.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		B767369320A4AB0200F04103 /* SFSDKNavigationController.m in Sources */ = {isa = PBXBuildFile; fileRef = B767369020A4AB0200F04103 /* SFSDKNavigationController.m */; };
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -101,6 +112,14 @@ struct NewLoginHostView: View {
                 .autocapitalization(.none)
                 .listRowSeparator(.hidden)
 
+            if let hostError {
+                Text(hostError)
+                    .foregroundStyle(.red)
```
bbirman: Could you see if `systemRedColor` would work here? iirc that one would adapt based on device settings (like adjust vibrancy in dark mode)

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -68,6 +68,7 @@ struct NewLoginHostField: View {
 struct NewLoginHostView: View {
     @State var host = ""
     @State var hostLabel = ""
+    @State var hostError: String?
```
wmathurin: Minor: `hostError` is internal view state — should be `@State private var hostError: String?` to prevent external mutation.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -101,6 +112,14 @@ struct NewLoginHostView: View {
                 .autocapitalization(.none)
                 .listRowSeparator(.hidden)
 
+            if let hostError {
```
wmathurin: The error persists while the user edits after a failed save. Add `.onChange(of: host) { _ in hostError = nil }` on the host `TextField` above so the inline error clears as soon as the user starts correcting input.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1902,10 +1903,38 @@ - (void)populateErrorHandlers {
...
+            SFSDKLoginHost *failing = [storage loginHostForHostAddress:failingHost];
+            if (failing && failing.isDeletable) {
+                for (NSUInteger i = 0; i < [storage numberOfLoginHosts]; i++) {
+                    if ([[storage loginHostAtIndex:i].host isEqualToString:failingHost]) {
+                        [storage removeLoginHostAtIndex:i];
+                        break;
+                    }
+                }
+            } else if (!failing) {
+                [SFSDKCoreLogger d:[strongSelf class] format:@"Failing host not found in storage; skipping removal."];
```
wmathurin: Log level: `.d` (debug) is too quiet here. A host missing from storage during a connection failure is unexpected — prefer `.w` so it surfaces in production log collection.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1902,10 +1903,38 @@ - (void)populateErrorHandlers {
...
+                recoveryHost = prev;
+            } else if ([storage numberOfLoginHosts] > 0) {
+                recoveryHost = [storage loginHostAtIndex:0].host;
+            }
+            if (recoveryHost) {
+                session.oauthRequest.loginHost = recoveryHost;
+                strongSelf.loginHost = recoveryHost;
+                [strongSelf restartAuthentication:session];
+            } else {
+                [SFSDKCoreLogger d:[strongSelf class] format:@"No recovery host available; skipping restart."];
```
wmathurin: Log level: `.d` (debug) is too quiet for this outcome. When there is no recovery host, authentication silently stops with no user feedback — this is a degraded UX path that warrants `.e` (error) or at minimum `.w` (warning) so it surfaces in crash/log collection.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/NewLoginHostTests.swift
```diff
@@ -52,4 +52,45 @@ final class NewLoginHostTests: XCTestCase {
         XCTAssertEqual(savedHost, expectedHost)
         XCTAssertEqual(savedLabel, expectedLabel)
     }
-} 
+
+    func testSaveRejectsEmpty() {
```
wmathurin, line 56: These rejection tests verify that `saveAction` was not called (`savedHost`/`savedLabel` are nil), but don't verify that an inline error was actually surfaced to the user. Since `@State` is struct-level in SwiftUI (not directly observable from outside), testing `hostError` requires a different approach — e.g. extracting the validation logic into a pure testable function, passing `hostError` as `@Binding`, or using ViewInspector. At minimum, consider a `testSaveErrorClearedOnSuccess` case to verify the happy path clears a prior error. Applies to all five rejection tests.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerLoginHostRecoveryTests.m
```diff
@@ -0,0 +1,205 @@
...
+                           userInfo:@{
+        @"_kCFStreamErrorCodeKey": @(-2103),
+        @"_kCFStreamErrorDomainKey": @(4)
+    }];
+}
+
+/// Drives the error handler block and waits until the alert OK completion has run.
+/// Replaces alertDisplayBlock with a fake that immediately fires actionOneCompletion,
+/// so we never present a real alert and the recovery logic runs deterministically.
+- (void)fireHandlerBlockForSession:(SFSDKAuthSession *)session {
```
wmathurin: `fireHandlerBlockForSession:` stubs `alertDisplayBlock` but lets `restartAuthentication:` call through to the real implementation. With a minimal `SFSDKAuthRequest` (test client-id, `test://callback` redirect), this may trigger async coordinator state changes that race with the `loginHost` assertions. Consider stubbing or swizzling `restartAuthentication:` so recovery tests isolate the decision logic from the actual OAuth restart.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerLoginHostRecoveryTests.m
```diff
@@ -0,0 +1,205 @@
...
+        if (message.actionOneCompletion) {
+            message.actionOneCompletion();
+        }
+        [completionRan fulfill];
+    };
+    mgr.errorManager.hostConnectionErrorHandlerBlock([self makeHostConnectionError], session, @{});
+    [self waitForExpectations:@[completionRan] timeout:5.0];
+}
+
+#pragma mark - Tests
```
wmathurin, line 190: Missing edge-case test: the PR description and the inline comment in `SFUserAccountManager.m` specifically call out the empty-storage guard (`numberOfLoginHosts == 0` → log and bail rather than raising `NSRangeException`). That guard path has no test here. Consider adding `test_givenEmptyStorage_when_handlerCompletionRuns_then_noRangeException` — drain storage, fire the handler, and assert no crash and no host assignment occurs.

libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
```diff
@@ -210,6 +210,7 @@
 		B75233DE1F4DF0B700040B6E /* SFSDKAuthPreferences.h in Headers */ = {isa = PBXBuildFile; fileRef = B75233DC1F4DF0B700040B6E /* SFSDKAuthPreferences.h */; };
 		B75233E01F4DF0B700040B6E /* SFSDKAuthPreferences.m in Sources */ = {isa = PBXBuildFile; fileRef = B75233DD1F4DF0B700040B6E /* SFSDKAuthPreferences.m */; };
 		B759CD891F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m in Sources */ = {isa = PBXBuildFile; fileRef = B759CD881F8BDBAC0081AA87 /* SDSDKAlertMessageTest.m */; };
+		B7FF237400000002 /* SFUserAccountManagerLoginHostRecoveryTests.m in Sources */ = {isa = PBXBuildFile; fileRef = B7FF237400000001 /* SFUserAccountManagerLoginHostRecoveryTests.m */; };
```
wmathurin: Hand-authored GUIDs (`B7FF237400000001` / `B7FF237400000002`). Real Xcode GUIDs are random 12-byte hex values (e.g. `B759CD891F8BDBAC0081AA87`). Sequential zero-padded IDs risk collision if Xcode or a future contributor auto-generates a GUID in the same range. Recommend letting Xcode add these entries via "Add Files to target" to get proper random UUIDs.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -87,6 +88,16 @@ struct NewLoginHostView: View {
...
+
+        let isValid = !hostToSave.isEmpty
+            && hostToSave.contains(".")
+            && hostToSave.rangeOfCharacter(from: .whitespaces) == nil
+            && URL(string: "https://\(hostToSave)") != nil
+        if !isValid {
+            self.hostError = SFSDKResourceUtils.localizedString("LOGIN_INVALID_HOST")
+            return
+        }
+        self.hostError = nil
```
bbirman, line 108: Could the error reset when the text changes too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -101,6 +112,14 @@ struct NewLoginHostView: View {
                 .autocapitalization(.none)
                 .listRowSeparator(.hidden)
 
+            if let hostError {
+                Text(hostError)
```
bbirman: nit: the spacing feels off to me (but I'm not UX :P), wondering if this could/should go into the same VStack that `NewLoginHostField` creates so that the error message seems more attached to the host field than in the middle of the two

Image

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1902,10 +1903,38 @@ - (void)populateErrorHandlers {
...
-            session.oauthRequest.loginHost = host;
-            strongSelf.loginHost = host;
-            [strongSelf restartAuthentication:session];
+            NSString *failingHost = session.oauthRequest.loginHost;
+            SFSDKLoginHostStorage *storage = [SFSDKLoginHostStorage sharedInstance];
+            SFSDKLoginHost *failing = [storage loginHostForHostAddress:failingHost];
+            if (failing && failing.isDeletable) {
+                for (NSUInteger i = 0; i < [storage numberOfLoginHosts]; i++) {
+                    if ([[storage loginHostAtIndex:i].host isEqualToString:failingHost]) {
+                        [storage removeLoginHostAtIndex:i];
```
bbirman: Is there any other info we could use to determine if deleting is the right path for the scenario? I'm worried it might be too broad for things like a time out or a connectivity issue vs having a bad URL

sfdctaka: You're right. The previous gate (`isDeletable`) was too broad. The dispatch hook upstream (`SFSDKAuthErrorManager.m:141`) routes anything with _kCFStreamError* userInfo into this handler, so transient codes like timeout (-1001), cannot-connect (-1004), and not-connected-to-internet (-1009) were also triggering removal. A user on flaky Wi-Fi could lose their legit custom host.

Tightened the auto-remove to fire only on strong "host is unusable" signals:
- kSFOAuthErrorDomain / kSFOAuthErrorInvalidURL
- NSURLErrorBadURL (-1000)
- NSURLErrorUnsupportedURL (-1002)
- NSURLErrorCannotFindHost (-1003, DNS NXDOMAIN)
- NSURLErrorDNSLookupFailed (-1006)
- NSURLErrorAppTransportSecurityRequiresSecureConnection (-1022)
 
All other codes that reach this handler now leave the host in storage but still recover (revert to previous host or fall back to index 0). Added a test (`test_givenDeletableFailingHostAndAmbiguousSignal...`) that fires a timeout and asserts the host stays.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
sfdctaka: image

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1902,10 +1903,62 @@ - (void)populateErrorHandlers {
...
+                    case NSURLErrorDNSLookupFailed:
+                    case NSURLErrorAppTransportSecurityRequiresSecureConnection:
+                        strongBadHostSignal = YES;
+                        break;
+                    default:
+                        break;
+                }
+            }
+            if (failing && failing.isDeletable && strongBadHostSignal) {
+                for (NSUInteger i = 0; i < [storage numberOfLoginHosts]; i++) {
```
bbirman: one more small one here, could you use `indexOfLoginHost:` instead of the for loop?

libs/SalesforceSDKCore/SalesforceSDKCoreTests/NewLoginHostTests.swift
sfdctaka: @wmathurin  Fair point — these tests cover the rejection contract (`saveAction` wasn't called) but don't observe `hostError` itself. Since it's `@State` private, surfacing it to a test needs one of: extracting validation into a pure function, lifting `hostError` to a `@Binding`, or pulling in ViewInspector. All bigger than this bugfix PR should carry, so I'd rather not do it inline. I'll file a follow-up to make `NewLoginHostView` properly unit testable (validation + error surfacing + clear-on-edit) and reference it here. OK to resolve?

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountManagerLoginHostRecoveryTests.m
sfdctaka: 1. `[storage removeAllLoginHosts]` doesn't actually empty the list — it preserves built-in production/sandbox at indices 0/1 unless MDM `onlyShowAuthorizedHosts` is set. Used KVC to snapshot and swap `loginHostList` for an empty array, then restore. Lighter than standing up a fake MDM singleton.
2. Can't assert on `mgr.loginHost` directly when storage is empty — `SFSDKAuthPreferences -loginHost` re-validates against storage and synthesizes a bundle-default fallback, so a read can't distinguish "handler didn't assign" from "getter resolved to default". Instead, the test counts invocations of the swizzled `restartAuthentication:` and asserts zero — proves the else branch fired and `recoveryHost`  stayed nil.
3. `XCTAssertNoThrow` covers the original concern (no `NSRangeException` from `loginHostAtIndex:0`).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/LoginHost/NewLoginHostView.swift
```diff
@@ -41,6 +41,8 @@ struct NewLoginHostField: View {
     let fieldPlaceholder: String
     let fieldInputAccessibilityID: String
     @Binding var fieldValue: String
+    var errorMessage: String? = nil
+    var errorAccessibilityID: String? = nil
```
bbirman: This doesn't look like it would be dynamic, could it be a non-optional `let`?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -1902,10 +1903,60 @@ - (void)populateErrorHandlers {
...
+            // is unusable (bad URL, DNS NXDOMAIN, ATS rejection, OAuth invalid URL). Ambiguous
+            // codes — timeout, cannot-connect, not-connected-to-internet, connection-lost,
+            // roaming-off, data-not-allowed — can fire for a perfectly valid host the user is
+            // temporarily unable to reach. Auto-removing on those would silently delete a
+            // legitimate custom host when the user is on flaky Wi-Fi or offline.
+            BOOL strongBadHostSignal = NO;
+            if ([error.domain isEqualToString:kSFOAuthErrorDomain] && error.code == kSFOAuthErrorInvalidURL) {
+                strongBadHostSignal = YES;
+            } else if ([error.domain isEqualToString:NSURLErrorDomain]) {
+                switch (error.code) {
```
bbirman, line 1929: This came up in a Claude review but I'm not as familiar with the error types, can you double check on it?

> NSURLErrorCannotFindHost and NSURLErrorDNSLookupFailed are classified as strong signals that the host is unusable, triggering permanent removal from SFSDKLoginHostStorage (persisted to UserDefaults). However, captive-portal networks routinely intercept DNS for unknown domains, producing these exact errors for perfectly valid hosts. An enterprise user connecting through hotel/airport WiFi would have their custom Salesforce instance host silently and permanently deleted. They'd need to manually re-enter it after reaching proper connectivity. Consider either requiring user confirmation before deletion, or downgrading DNS errors to the "ambiguous" category alongside timeouts.

sfdctaka: Pushed a follow-up that drops both `NSURLErrorCannotFindHost` and `NSURLErrorDNSLookupFailed` from the strong-signal list. Captive portals (hotel/airport WiFi) routinely hijack DNS and return those exact codes for perfectly valid enterprise hosts, so classifying them as strong would silently delete a user's custom org the first time they open the app behind one — with no recovery path. Both now fall through to the ambiguous bucket alongside timeouts and connection-lost. 
  
Strong signals are now URL-syntax / ATS / OAuth-side only: `kSFOAuthErrorInvalidURL`, `NSURLErrorBadURL`, `NSURLErrorUnsupportedURL`, `NSURLErrorAppTransportSecurityRequiresSecureConnection` — never anything that can be produced by network conditions. Added a regression test (`test_givenDeletableFailingHostAndDNSLookupFailed_failingHostKeptInStorage`) so this stays pinned. Comment in the block now spells out the captive-portal reasoning.

bbirman (approved):

wmathurin (approved):

PR #4091, opened 2026-06-30
Fix thread-safety race on SFUserAccount.notificationTypes

sfdctaka:
## Summary

`SFUserAccount.notificationTypes` is the only mutable property on `SFUserAccount` whose getter and setter touch the backing ivar directly without synchronization. Every other property on this class (`accessScopes`, `idData`, `credentials`, `accountIdentity`, `photo`, `customData`, `featureFlags`) routes through a private concurrent `_syncQueue` (`dispatch_sync` for reads, `dispatch_barrier_async` for writes), and `encodeWithCoder:` itself already reads `_notificationTypes` from inside that queue. The accessors were the gap.

When two callers write `account.notificationTypes` concurrently — or one writes while another reads — the previously assigned `NSArray` can be released twice or read mid-deallocation, producing `EXC_BAD_ACCESS (KERN_INVALID_ADDRESS)` inside `swift_arrayDestroy` / `_ContiguousArrayStorage.__deallocating_deinit`.

### Fix

In `libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccount.m`:

- `- (NSArray<NotificationType *> *)notificationTypes` now reads through `dispatch_sync(_syncQueue, ...)` with `__block` capture — matches the existing `accountIdentity` getter.
- `- (void)setNotificationTypes:` now writes through `dispatch_barrier_async(_syncQueue, ...)` with `[notificationTypes copy]` preserved inside the barrier block — matches the existing `setAccessScopes:` setter.

No public API change. `SFUserAccount.h` is unchanged. NSSecureCoding archive layout is unchanged. The fix is internal to `SFUserAccount.m`.

### Test coverage

Added `libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFUserAccountThreadSafetyTests.swift` with four XCTests:

- `test_notificationTypes_concurrent_reads_and_writes_do_not_crash` — 8 writer + 8 reader threads × 10,000 iterations each, 30 s timeout.
- `test_setNotificationTypes_takes_a_snapshot` — assigns an array, mutates the source, asserts snapshot unchanged (verifies the `[copy]` semantics survive).
- `test_notificationTypes_returns_assigned_value_after_drain` — sequential assign / read pairs confirm the barrier-async write drains in order.
- `test_encodeWithCoder_under_contention_does_not_throw` — 5,000 background writes interleaved with 500 `NSKeyedArchiver` encodes; decode must succeed and yield a non-empty array.

All four tests pass on `iPhone 16` simulator (iOS 18.6).

## Test plan

- [x] `xcodebuild test -workspace SalesforceMobileSDK.xcworkspace -scheme SalesforceSDKCore -only-testing:SalesforceSDKCoreTests/UserAccountThreadSafetyTests` — 4/4 PASS.
- [x] Targeted regression sweep on `SFUserAccount*` and `PushNotificationManager*` test classes — no new failures.
- [ ] Address Sanitizer pass on the new test class (recommended pre-merge).
- [ ] Thread Sanitizer pass on the new test class (recommended pre-merge).

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore.xcodeproj/project.pbxproj
@@ -399,6 +399,7 @@
 		CE7F662B1E556CA800DC3FBB /* SFNetwork.h in Headers */ = {isa = PBXBuildFile; fileRef = CE7F66291E556CA800DC3FBB /* SFNetwork.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		CE7F662C1E556CA800DC3FBB /* SFNetwork.m in Sources */ = {isa = PBXBuildFile; fileRef = CE7F662A1E556CA800DC3FBB /* SFNetwork.m */; };
 		CE81A9C81E9C26F900F3D0AD /* SFUserAccountManagerNotificationsTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CE81A9C61E9C26EF00F3D0AD /* SFUserAccountManagerNotificationsTests.m */; };
+		42996EA8CCEB33EF9E258033 /* SFUserAccountThreadSafetyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0DB0265069ECCAE8F59F5CD0 /* SFUserAccountThreadSafetyTests.swift */; };
 		CE88BD521D17065C00AE3BF7 /* SFSDKAILTNPublisher.h in Headers */ = {isa = PBXBuildFile; fileRef = CE88BD4F1D17065B00AE3BF7 /* SFSDKAILTNPublisher.h */; settings = {ATTRIBUTES = (Public, ); }; };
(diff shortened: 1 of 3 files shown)


sfdctaka:
The three Clang Static Analyzer findings (lines 272, 456, 458) are pre-existing in `SFUserAccount.m` on `dev` and are not introduced by this PR:

- **Line 272** — `_credentials = credentials;` in `setCredentialsInternal:`
- **Lines 456 / 458** — `identity.userId = newKey;` / `identity.orgId = newKey;` in `observeValueForKeyPath:`

Diff vs `forcedotcom/dev` only touches lines ~41 (a new string const), ~98–108 (wrapping `encodeWithCoder:` in `dispatch_sync`), ~125 (decoding the persisted feature-flags key), and ~147–159 (the `notificationTypes` accessors). None of the flagged lines are in the change set. Danger's Clang scan reports findings for the entire modified file, which is why they surface here.

Leaving these out of scope for this thread-safety fix to keep the PR focused. Happy to file a small follow-up to address the pre-existing nullability findings separately if that's preferred.

bbirman (approved):

wmathurin (commented):
👍

PR #4093, opened 2026-07-03
Make advanced authentication the default and deprecate forceAdvancedAuthentication

brandonpage:
## Summary

Makes Advanced Authentication (browser-based OAuth) the default for all interactive logins and deprecates the `forceAdvancedAuthentication` flag on the same release.

- Advanced Auth now drives interactive login regardless of the target server's My Domain auth-configuration (previously it was gated on the server opting in).
- `SalesforceSDKManager.forceAdvancedAuthentication` is **introduced already-deprecated in 14.0** and will be **removed in 15.0** (intentional single-release deprecation).
- Internal SDK code no longer touches the deprecated property directly. It reads/writes the same backing storage through a non-deprecated internal accessor (`sdk_forceAdvancedAuthentication`, ObjC) or via KVC (the DEBUG Swift dev menu, which can't see the internal header), so the SDK builds warning-free.
- Adds a dev-menu toggle for the flag and AuthFlowTester UI coverage for both the flag ON (external browser) and OFF (legacy WebView) paths.

## Deprecation pattern

The `deprecated` attribute lives on the public property's accessor declaration, so no separate backing field is needed (simpler than the Android equivalent). One auto-synthesized ivar is shared; a second, non-deprecated accessor over it is declared in `SalesforceSDKManager+Internal.h` for internal readers. Tests intentionally exercise the public property (including the "defaults to YES" contract) using the sanctioned `SFSDK_USE_DEPRECATED_BEGIN/END` (ObjC) and `@available(*, deprecated)` (Swift) suppression already used elsewhere in the repo.

## New UX

Unlike Android the Server Picker does not hover hover a backing screen, meaning when Advanced Auth is forced for all orgs there is actually no way to get back to the WebView login screen to access the back button or gear menu.  This is actually an existing issue with Advanced Auth, but the new default makes it worse so this seemed like the necessary time to improve the situation:

adv_auth_flag_ux

- The back button only shows in multi-user scenarios, just like on the WebView login screen.
- The only option under the gear icon is `Login Options` (since `Change Server`, `Clear Cookies`, `Clear Cache`, `Reload` and `Login for Admin` are all not applicable for Adv Auth) and is only shown on debug builds.

### Implementation note

In the forced-advanced-auth path `SFLoginViewController` is never created, so `SFSDKLoginHostListViewController` (the host list) is the screen the user lands on and has to carry the chrome that normally lives on the WebView login screen. It's presented in three roles — the forced-advanced-auth landing screen (standalone) and two sub-sheets (`SFLoginViewController`'s server picker, `SFSDKLoginFlowSelectionViewController`'s IdP flow) — and the chrome must appear only in the standalone role.

- A single `presentedAsLoginScreen` flag marks that standalone role (default `NO`, so the sub-sheets and IdP flow are unaffected).
- The back button and gear are gated **independently**: the back button by `-shouldShowBackButton` (logged-in user / idp / web-auth-fallback state, debug **or** release), and the gear by `isDevSupportEnabled` (debug only). The flag only decides whether this screen owns that chrome at all — it does not itself couple the two.

### Testing

`SFSDKLoginHostTests` covers the new chrome: gear visibility (flag on/off × dev-support on/off), `-shouldShowBackButton` across biometric-lock / web-auth-fallback / idp / account states, and that `viewDidLoad` installs the back button vs. Cancel accordingly.

The back-button-state tests read the shared, keychain-backed `SFUserAccountManager`, so they now clear/save/restore `currentUser` in setUp/tearDown and establish their no-account precondition explicitly. Without this, a `currentUser` left by an integration-test login persists across runs and flips `-shouldShowBackButton` to `YES`, failing `test_givenUnlockedNoFlowNoAccount_whenShouldShowBackButton_thenNo` in isolation. Verified on the iOS 18 and iOS 26 simulators (the CI matrix legs); full class 20/20 green.


wmathurin (commented):
**CI failure root cause — **

The failing test is a **new test added by this PR**, and it is correct — it caught a real logic error in the new  implementation in .

**Root cause**: The new config branch fires for *all* instances of the VC, not just the ones in the forced-advanced-auth path:

```objc
if ([self.config isKindOfClass:[SFSDKLoginViewControllerConfig class]]
    && ((SFSDKLoginViewControllerConfig *)self.config).shouldDisplayBackButton) {
    return YES;
}
```

`shouldDisplayBackButton` returns `![SalesforceSDKManager sharedManager].appConfig.shouldAuthenticate`. In the test environment `shouldAuthenticate` defaults to `NO` (no credentials configured), so `!NO = YES` → the branch returns `YES` for any bare VC instance, including the test's — even with no accounts and no auth flow in progress.

**Fix** — add the same `showsBackButtonAndLoginOptions` guard already used everywhere else in the file (lines 159, 316):

```objc
if (self.showsBackButtonAndLoginOptions
    && [self.config isKindOfClass:[SFSDKLoginViewControllerConfig class]]
    && ((SFSDKLoginViewControllerConfig *)self.config).shouldDisplayBackButton) {
    return YES;
}
```

**Tested locally** on iPhone 17 Pro simulator (iOS 18): all 20  pass, full  suite (492 tests) shows no regressions introduced by this change.

wmathurin (approved):

brandonpage:
Just like my Android PR for the same work, the UI Tests were failing before my changes so I will fix it after I merge this.

PR #4094, opened 2026-07-07
[W-23059473] Introduce SFOAuthErrorCode enum for token endpoint error responses

wmathurin:
## Summary
- Introduces `SFOAuthErrorCode.swift` — a typed `@objc` Swift enum mirroring the ~42 distinct wire values from the server-side `OauthErrorCode.java` (core/identity-common-api), plus an `.unknown` fallback
- Adds `errorCode: NSInteger` (typed as `SFOAuthErrorCode`) property to `SFSDKOAuthTokenEndpointErrorResponse`, populated at init time
- Updates `SFOAuthCoordinator -handleResponse:` to compare `errorCode` enum instead of string-comparing against `kSFOAuthErrorTypeUnsupportedGrantType`
- Adds `client_blocked` and `client_blocked_retry` to the `+errorWithType:description:` integer code mapping (previously fell through to `kSFOAuthErrorUnknown = 666`)
- Deprecates all `kSFOAuthErrorType*` constants in `SFSDKOAuthConstants.h` with a pointer to the new enum
- Groundwork for App Attestation error handling (W-22699715)

## Files Changed
- `SFOAuthErrorCode.swift` (**new**) — 42 enum cases + `.unknown` + `wireValue` + `from()` factory + `SFOAuthErrorCodeHelper` ObjC bridge class
- `SFSDKOAuth2.h` — `errorCode` property on `SFSDKOAuthTokenEndpointErrorResponse`
- `SFSDKOAuth2.m` — populate `_errorCode`; add `client_blocked`/`client_blocked_retry` int codes
- `SFOAuthCoordinator.m` — enum comparison for `isUnsupportedGrantType`
- `SFSDKOAuthConstants.h` — deprecate 18 `kSFOAuthErrorType*` constants
- `SFOAuthErrorCodeTests.swift` (**new**) — 6 unit tests (known values, nil, empty, unknown, round-trip)
- `SFOAuthCoordinatorLightningURLTests.swift` — derive test wire values from enum `wireValue`
- `SalesforceSDKCore.xcodeproj/project.pbxproj` — add new files to project/test target

## Test Plan
- [x] `xcodebuild build -scheme SalesforceSDKCore -sdk iphonesimulator` — succeeds
- [x] `SFOAuthErrorCodeTests` — 6/6 passed
- [x] `SFOAuthCoordinatorLightningURLTests` — 5/5 passed
- [x] Full `SalesforceSDKCore` suite — 706 tests, 0 failures

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -478,7 +478,7 @@
         id json = nil;
         json = [SFJsonUtils objectFromJSONData:data];
         if (json == nil) {
-            NSError *error = [SFSDKOAuth2 errorWithType:kSFOAuthErrorTypeJWTLaunchFailed
+            NSError *error = [SFSDKOAuth2 errorWithType:@"jwt_launch_failed"
                                                  description:@"Error parsing JWT token exchange response."
                                              underlyingError:[SFJsonUtils lastError]];
                 [self notifyDelegateOfFailure:error authInfo:self.authInfo];
@@ -486,7 +486,7 @@
         }
         if (![json isKindOfClass:[NSDictionary class]]) {
             NSString *errorDesc = [NSString stringWithFormat:@"Expected NSDictionary for JWT token response, received %@ instance.", NSStringFromClass([json class])];
-                NSError *error = [SFSDKOAuth2 errorWithType:kSFOAuthErrorTypeJWTLaunchFailed
+                NSError *error = [SFSDKOAuth2 errorWithType:@"jwt_launch_failed"
                                                  description:errorDesc];
             [self notifyDelegateOfFailure:error authInfo:self.authInfo];
             return;
@@ -670,7 +670,7 @@
                  [SFSDKCoreLogger d:[self class] format:@"Refresh attempt timed out after %f seconds.", self.timeout];
                  [self stopAuthentication];
              }
-             BOOL isUnsupportedGrantType = [response.error.tokenEndpointErrorCode isEqualToString:kSFOAuthErrorTypeUnsupportedGrantType];
+             BOOL isUnsupportedGrantType = (response.error.errorCode == SFOAuthErrorCodeUnsupportedGrantType);
              BOOL isLightningURL = [self.credentials.domain containsString:@".lightning."];
              if (isUnsupportedGrantType && isLightningURL) {
                  [SFSDKCoreLogger e:[self class] format:@"Code exchange failed with unsupported_grant_type against Lightning URL: %@. Lightning URLs do not support authorization_code grant type. Use a My Domain login server URL instead.", self.credentials.domain];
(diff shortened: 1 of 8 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
```diff
@@ -501,37 +506,37 @@ + (NSError *)errorWithType:(NSString *)type description:(NSString *)description
...
+    } else if ([type isEqualToString:@"invalid_client_credentials"]) {
         code = kSFOAuthErrorInvalidClientCredentials;
-    } else if ([type isEqualToString:kSFOAuthErrorTypeInvalidGrant]) {
+    } else if ([type isEqualToString:@"invalid_grant"]) {
         code = kSFOAuthErrorInvalidGrant;
-    } else if ([type isEqualToString:kSFOAuthErrorTypeInvalidRequest]) {
+    } else if ([type isEqualToString:@"invalid_request"]) {
         code = kSFOAuthErrorInvalidRequest;
-    } else if ([type isEqualToString:kSFOAuthErrorTypeInactiveUser]) {
+    } else if ([type isEqualToString:@"inactive_user"]) {
```
bbirman, line 523: I was wondering why inactive_user wasn't in the new code set but then I saw on the server side it falls under "invalid_grant". Is there a way to get the sub-category too?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -478,15 +478,15 @@ - (void)beginJwtTokenExchangeFlow {
...
+            NSError *error = [SFSDKOAuth2 errorWithType:@"jwt_launch_failed"
                                                  description:@"Error parsing JWT token exchange response."
                                              underlyingError:[SFJsonUtils lastError]];
                 [self notifyDelegateOfFailure:error authInfo:self.authInfo];
             return;
         }
         if (![json isKindOfClass:[NSDictionary class]]) {
             NSString *errorDesc = [NSString stringWithFormat:@"Expected NSDictionary for JWT token response, received %@ instance.", NSStringFromClass([json class])];
-                NSError *error = [SFSDKOAuth2 errorWithType:kSFOAuthErrorTypeJWTLaunchFailed
+                NSError *error = [SFSDKOAuth2 errorWithType:@"jwt_launch_failed"
```
bbirman, line 489: This is hardcoded now because the client-side is creating the error?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
wmathurin: Yes — sub-categories are accessible via `error_description`. The server populates that field from `OauthErrorCode.getDescription()`, which is the human-readable description string in the server enum (e.g. `"inactive user"`, `"inactive organization"`, `"expired access/refresh token"`, etc.). The SDK already parses and exposes this as `tokenEndpointErrorDescription` on `SFSDKOAuthTokenEndpointErrorResponse`, so callers can combine an `errorCode == SFOAuthErrorCodeInvalidGrant` check with a string comparison on `errorDescription` to distinguish sub-cases if needed.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
wmathurin: Correct — this is a client-side error, not a server wire value. The JWT response is malformed (wrong type), so the SDK generates the error locally rather than receiving it from the server. `jwt_launch_failed` was previously held in the `kSFOAuthErrorTypeJWTLaunchFailed` constant (now deprecated), but since this error string is not a server-defined wire value it isn't in `SFOAuthErrorCode`. The raw string is used directly here since the deprecated constant is no longer the preferred reference.

bbirman (approved):

PR #4097, opened 2026-07-08
App attestation for code exchange + refresh flow

bbirman:
Created new branch off of dev

This PR includes:
- Attestation structure for key generation & challenge generation
- Flag to enable / disable
- Using attestation on code exchange and refresh

Doesn't include:
- Enforcing web server if attestation is enabled (upcoming PR)
- Feature flag (upcoming PR)
- Native login
- Error handling / retryable errors

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
@@ -145,6 +145,7 @@ public class AppAttestation: NSObject {
         let keychainItem = AttestationKeychainItem(attestationId: attestationId, keyId: keyId)
         let keychainItemData = try JSONEncoder().encode(keychainItem)
         let keychainResult = KeychainHelper.write(service: appAttestKeyName, data: keychainItemData, account: nil)
+        // Intentional: prefer re-registration on next launch over blocking auth
         if !keychainResult.success {
             var message = "Unable to write attestation keyId to keychain"
             if let error = keychainResult.error {
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
@@ -160,7 +160,10 @@ public class AppAttestation: NSObject {
         do {
             return try await operation()
         } catch let error as DCError {
-            SFSDKCoreLogger.log(AppAttestation.self, level: .error, message: "App Attest error: \(error.code.rawValue)")
+            SFSDKCoreLogger.log(AppAttestation.self, level: .error, message: "App Attest Device Check error: \(error.code.rawValue)")
+            throw error
+        } catch {
+            SFSDKCoreLogger.log(AppAttestation.self, level: .error, message: "App Attest error: \(error.localizedDescription)")
             throw error
         }
     }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
@@ -1,0 +1,187 @@
+//
+//  AppAttestation.swift
+//  SalesforceSDKCore
+//
+//  Created by Brianna Birman on 1/5/26.
+//  Copyright (c) 2026-present, salesforce.com, inc. All rights reserved.
+// 
+//  Redistribution and use of this software in source and binary forms, with or without modification,
+//  are permitted provided that the following conditions are met:
+//  * Redistributions of source code must retain the above copyright notice, this list of conditions
+//  and the following disclaimer.
+//  * Redistributions in binary form must reproduce the above copyright notice, this list of
+//  conditions and the following disclaimer in the documentation and/or other materials provided
+//  with the distribution.
+//  * Neither the name of salesforce.com, inc. nor the names of its contributors may be used to
+//  endorse or promote products derived from this software without specific prior written
+//  permission of salesforce.com, inc.
+// 
+//  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+//  IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+//  FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR
+//  CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+//  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+//  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY
+//  WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import CryptoKit
+import DeviceCheck
+
+
+@objc(SFSDKAppAttestation)
+public class AppAttestation: NSObject {
+    enum AppAttestationError: Error {
+        case challengeEncodingFailed
+    }
+
+    struct AttestationObject: Codable {
+        let attestationId: String
+        let attestationData: String // Base-64 encoded
+    }
+    
+    struct AttestationKeychainItem: Codable {
+        let attestationId: String
+        let keyId: String
+    }
+    
+    static let appAttestKeyName = "com.salesforce.mobilesdk.attestation"
+    
+    /// Returns an attestation string if attestation is enabled and the domain is a My Domain,
+    /// otherwise returns nil. Callers do not need to check gating conditions themselves.
+    @objc
+    public static func attestationIfEnabled(for domain: String, consumerKey: String) async -> String? {
+        let isLoginPool = domain == kSFOAuthProductionLoginURL
+            || domain == kSFOAuthSandboxLoginURL
+            || domain == kSFOAuthWelcomeLoginURL
+
+        guard UserAccountManager.shared.appAttestationEnabled,
+              !isLoginPool,
+              !consumerKey.isEmpty else {
+            return nil
+        }
+
+        do {
+            return try await attestationObject(for: domain, consumerKey: consumerKey)
+        } catch {
+            SFSDKCoreLogger.log(AppAttestation.self, level: .error, message: "Attestation error: \(error.localizedDescription)")
+            // TODO: In coordination with error stories, if device error prevents attestation from being generated, should this throw and/or have a retry path like deleting old key?
+            return nil
+        }
+    }
+    
+    // Internal for unit tests
+    static func existingKeyId(for attestationId: String) throws -> String? {
+        let attestKeyQuery = KeychainHelper.read(service: appAttestKeyName, account: nil)
+        if let attestKeyData = attestKeyQuery.data,
+           let keychainItem = try? JSONDecoder().decode(AttestationKeychainItem.self, from: attestKeyData) {
+            if keychainItem.attestationId == attestationId {
+                return keychainItem.keyId
+            } else {
+                // The ID can change if app is deleted / reinstalled but an old key can remain, delete the old one
+                // and then generate a new key like normal
+                let removeResult = KeychainHelper.remove(service: appAttestKeyName, account: nil)
+                if !removeResult.success {
+                    SFSDKCoreLogger.log(AppAttestation.self, level: .error, message: "Unable to delete old attestation keyId from keychain: \(removeResult.error?.localizedDescription ?? "")")
+                }
+            }
+        }
+        return nil
+    }
+
+    private static func attestationObject(for domain: String, consumerKey: String) async throws -> String {
+        let attestationId = SalesforceManager.shared.deviceId()
+
+        let keyId = try await keyId(for: attestationId, domain: domain, consumerKey: consumerKey)
+
+        let challenge = try await requestChallengeFromSalesforce(domain: domain, consumerKey: consumerKey, attestationId: attestationId)
+        guard let challengeData = challenge.data(using: .utf8) else {
+            throw AppAttestationError.challengeEncodingFailed
+        }
+        let hash = Data(SHA256.hash(data: challengeData))
+
+        let assertion = try await performAppAttestOperation {
+            try await DCAppAttestService.shared.generateAssertion(keyId, clientDataHash: hash)
+        }
+        let assertionString = assertion.base64EncodedString()
+
+        let attestationObject = AttestationObject(attestationId: attestationId, attestationData: assertionString)
+        let jsonData = try JSONEncoder().encode(attestationObject)
+        return jsonData.base64EncodedString()
+    }
+
+
+    private static func generateAttestation(keyId: String, challenge: String) async throws -> String {
+        guard let challengeData = challenge.data(using: .utf8) else {
+            throw AppAttestationError.challengeEncodingFailed
+        }
+        let hash = Data(SHA256.hash(data: challengeData))
+
+        let attestation = try await performAppAttestOperation {
+            try await DCAppAttestService.shared.attestKey(keyId, clientDataHash: hash)
+        }
+        return attestation.base64EncodedString()
+    }
+
+    //    1. Client creates a key pair
+    //    2. Client makes request to get challenge  - /mobile/attest/challenge
+    //    3. Client creates an Apple attestation object with obtained challenge
+    //    4. Client makes request to pre-register public key - /mobile/attest/registerkey
+    private static func keyId(for attestationId: String, domain: String, consumerKey: String) async throws -> String {
+        if let existingKey = try existingKeyId(for: attestationId) {
+            return existingKey
+        }
+
+        let keyId = try await performAppAttestOperation {
+            try await DCAppAttestService.shared.generateKey()
+        }
+
+        let challenge = try await requestChallengeFromSalesforce(domain: domain, consumerKey: consumerKey, attestationId: attestationId)
+        let attestation = try await generateAttestation(keyId: keyId, challenge: challenge)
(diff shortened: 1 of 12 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
```diff
@@ -77,13 +78,22 @@ - (void)refreshSessionWithCompletion:(void (^)(SFOAuthCredentials *))completionB
         return;
     }
     
+    __weak typeof(self) weakSelf = self;
+    [SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain consumerKey:self.credentials.clientId completionHandler:^(NSString * _Nullable attestation) {
+        __strong typeof(weakSelf) strongSelf = weakSelf;
+        [strongSelf executeRefreshWithAttestation:attestation];
```
wmathurin: If `weakSelf` has been deallocated by the time the attestation callback fires, `strongSelf` is nil and `[strongSelf executeRefreshWithAttestation:]` is a silent no-op — neither `completionBlock` nor `errorBlock` is ever called. Before this change, `refreshSessionWithCompletion:` was synchronous, so owners may not hold a strong reference across the now-async gap. Callers waiting for a callback will hang indefinitely. Consider explicitly calling the error block if `strongSelf` is nil, or retaining `self` for the duration of the async operation.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -618,6 +618,12 @@ - (void)updateCredentials:(NSDictionary *) params {
 }
 
 - (void)beginTokenEndpointFlow {
+    [SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain consumerKey:self.credentials.clientId completionHandler:^(NSString * _Nullable attestation) {
+        [self executeTokenEndpointFlowWithAttestation:attestation];
```
wmathurin: This block captures `self` strongly, unlike every other block in this file which uses the `__weak`/`__strong` pattern. During the attestation network round-trip (challenge request), the coordinator is kept alive by the block even if it should have been released (e.g., login cancelled). Should use `__weak typeof(self) weakSelf = self` consistent with the rest of the file.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
wmathurin: If `weakSelf` has been deallocated by the time the attestation callback fires, `strongSelf` is nil and `[strongSelf executeRefreshWithAttestation:]` is a silent no-op — neither `completionBlock` nor `errorBlock` is ever called. Before this change, `refreshSessionWithCompletion:` was synchronous, so owners may not hold a strong reference across the now-async gap. Callers waiting for a callback will hang indefinitely. Consider explicitly calling the error block if `strongSelf` is nil, or retaining `self` for the duration of the async operation.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
```diff
@@ -0,0 +1,187 @@
...
+        let attestationId: String
+        let keyId: String
+    }
+    
+    static let appAttestKeyName = "com.salesforce.mobilesdk.attestation"
+    
+    /// Returns an attestation string if attestation is enabled and the domain is a My Domain,
+    /// otherwise returns nil. Callers do not need to check gating conditions themselves.
+    @objc
+    public static func attestationIfEnabled(for domain: String, consumerKey: String) async -> String? {
```
wmathurin: The generated ObjC completion handler for this `async` method (`attestationIfEnabledFor:consumerKey:completionHandler:`) fires on Swift's cooperative thread pool, not the main thread. Both `SFOAuthCoordinator` and `SFOAuthSessionRefresher` receive the callback on that background thread and immediately access `self.responseData`, `self.credentials`, `self.approvalCode`, etc. — properties that were only ever touched synchronously before this change. If those callers assume main-thread access (typical for OAuth coordinator flows), this is a data race. Consider dispatching to the expected thread inside the completion handler before the downstream methods are invoked.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
```diff
@@ -0,0 +1,187 @@
...
+        let challenge = try await requestChallengeFromSalesforce(domain: domain, consumerKey: consumerKey, attestationId: attestationId)
+        let attestation = try await generateAttestation(keyId: keyId, challenge: challenge)
+        try await registerKeyWithSalesforce(keyId: keyId, consumerKey: consumerKey, attestationId: attestationId, attestationObject: attestation, domain: domain)
+
+        // Only persist after successful registration — if any step above fails,
+        // the next attempt will start fresh with a new key.
+        let keychainItem = AttestationKeychainItem(attestationId: attestationId, keyId: keyId)
+        let keychainItemData = try JSONEncoder().encode(keychainItem)
+        let keychainResult = KeychainHelper.write(service: appAttestKeyName, data: keychainItemData, account: nil)
+        if !keychainResult.success {
```
wmathurin, line 164: When the keychain write fails, the error is logged but execution falls through and `keyId` is still returned as if everything succeeded. The key is registered with the server but not persisted locally. On the next launch, `existingKeyId()` returns nil, a new key is generated and registered again, and the previous server-side key becomes orphaned. This will cause a redundant registration on every launch if the keychain is systematically unavailable. Should a keychain write failure be propagated as a thrown error rather than silently continuing?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
```diff
@@ -0,0 +1,187 @@
...
+            if let error = keychainResult.error {
+                message += ": \(error.localizedDescription)"
+            }
+            SFSDKCoreLogger.log(AppAttestation.self, level: .error, message: message)
+        }
+
+        return keyId
+    }
+
+    private static func performAppAttestOperation<T>(_ operation: () async throws -> T) async throws -> T {
```
wmathurin, line 175: `performAppAttestOperation` only intercepts `DCError` — non-`DCError` throws from `DCAppAttestService` (network errors, cancellation, etc.) pass through here without any logging. The top-level `attestationIfEnabled` catches them and logs `error.localizedDescription`, but context about which specific operation failed is lost. The PR description notes error handling is still a work in progress, so flagging for tracking — consider either logging all errors here, or removing the wrapper in favour of inline logging at each call site.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
bbirman: updated!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
bbirman: updated!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: I was originally thinking that regenerating if there was a keychain issue would be less disruptive than blocking login/refresh, but good either way! Should I change it?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: added a line for logging other errors

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: I'm not sure I understand this one, it doesn't look like the token endpoint flow ran on main before, what assumes main thread access?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
```diff
@@ -0,0 +1,190 @@
...
+    /// otherwise returns nil. Callers do not need to check gating conditions themselves.
+    @objc
+    public static func attestationIfEnabled(for domain: String, consumerKey: String) async -> String? {
+        let isLoginPool = domain == kSFOAuthProductionLoginURL
+            || domain == kSFOAuthSandboxLoginURL
+            || domain == kSFOAuthWelcomeLoginURL
+
+        guard UserAccountManager.shared.appAttestationEnabled,
+              !isLoginPool,
+              !consumerKey.isEmpty else {
```
sfdctaka: https://developer.apple.com/documentation/devicecheck/dcappattestservice/issupported

On any device where App Attest is unavailable (simulator, older devices, some enterprise MDM configs), `DCAppAttestService.shared.generateKey()` throws `DCError.featureUnsupported` on every single token-endpoint call — you eat an unnecessary network round-trip (challenge request) and two log lines per login/refresh. Guarding on `DCAppAttestService.shared.isSupported` before touching the network would avoid the noise and the wasted request. 
```
  guard UserAccountManager.shared.appAttestationEnabled,
        !isLoginPool,
        !consumerKey.isEmpty,
        DCAppAttestService.shared.isSupported else {
      return nil
  }
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
wmathurin: Fair point — I was hedging with 'if those callers assume main-thread access.' The real concern is whether callers treat `beginTokenEndpointFlow`/`refreshSessionWithCompletion:` as serialized on a specific thread, because the continuation now lands on Swift's cooperative pool rather than the calling thread. If the ObjC callers already handle these flows on an arbitrary background thread (which sounds like your understanding), there's no issue. Worth a quick check: is there any `dispatch_async(dispatch_get_main_queue(), ...)` wrapping these call sites, or does any caller assume subsequent delegate callbacks arrive on the same thread it called from?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
wmathurin: Your reasoning is sound — failing silently and re-registering on the next launch is less disruptive than blocking auth for a transient keychain failure. No need to change the behavior. The only downside is orphaned server-side keys accumulating if the keychain is persistently broken, but that's an edge case. I'd suggest adding a brief comment at the keychain write failure branch (e.g. `// Intentional: prefer re-registration on next launch over blocking auth`) so future readers understand the choice rather than treating it as an oversight.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: Good idea! Definitely for the simulator. For older devices / MDM configs, wondering if it should still go through so there's logging around around the failure (especially if the ECA has strict enforcement enabled). What do you guys think?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
```diff
@@ -0,0 +1,190 @@
...
+        let attestationId: String
+        let keyId: String
+    }
+    
+    static let appAttestKeyName = "com.salesforce.mobilesdk.attestation"
+    
+    /// Returns an attestation string if attestation is enabled and the domain is a My Domain,
+    /// otherwise returns nil. Callers do not need to check gating conditions themselves.
+    @objc
+    public static func attestationIfEnabled(for domain: String, consumerKey: String) async -> String? {
```
sfdctaka: Objective-C: `SFOAuthCredentials.domain` is declared as `NSString * _Nullable`— it's legal for it to be
nil.

Swift: `AppAttestation.attestationIfEnabled(for domain: String, ...)` declares domain as a non-optional String — it must have a value.

When ObjC calls into Swift across this bridge, the compiler inserts an implicit "force-unwrap", basically `domain!`. If the ObjC side ever hands in nil, that force-unwrap crashes the app with something like:

`Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value`

Where this call happens:

Both ObjC call sites do this:
```
// SFOAuthCoordinator.m:622
[SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain
                                 consumerKey:self.credentials.clientId
                           completionHandler:...];
```

```
// SFOAuthSessionRefresher.m:81  (same shape)
[SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain consumerKey:self.credentials.clientId completionHandler:^(NSString * _Nullable attestation) //                       If it's nil → crash.
```

Why it's "unlikely but sharp"

In `SFOAuthCredentials.m`:187, domain gets defaulted to "login.salesforce.com" in the initializer. So
in the normal login flow it will never be nil. But:

- A caller could construct/mutate an `SFOAuthCredentials `with `domain = nil` (Swift setter of domain: 
String?).
- A serialized/decoded credentials object with a missing key could produce nil.
- Test doubles or edge-case identity data could set it to nil.

In any of those cases, an otherwise-recoverable auth failure becomes a hard crash inside attestation
gating.

---
Quick fix:
`libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift:53`

Before:
```
@objc
public static func attestationIfEnabled(for domain: String, consumerKey: String) async -> String? {
    let isLoginPool = domain == kSFOAuthProductionLoginURL
        || domain == kSFOAuthSandboxLoginURL
        || domain == kSFOAuthWelcomeLoginURL
        
    guard UserAccountManager.shared.appAttestationEnabled,
          !isLoginPool,
          !consumerKey.isEmpty else {
        return nil
    }
    ...
```

After:
```
@objc
public static func attestationIfEnabled(for domain: String?, consumerKey: String) async -> String? {
    guard let domain, !domain.isEmpty else {
        return nil
    }   
    
    let isLoginPool = domain == kSFOAuthProductionLoginURL
        || domain == kSFOAuthSandboxLoginURL
        || domain == kSFOAuthWelcomeLoginURL
        
    guard UserAccountManager.shared.appAttestationEnabled,
          !isLoginPool,
          !consumerKey.isEmpty else {
        return nil
    }
    ...
```
Two tiny changes:
1. for domain: String → for domain: String?
2. Add guard let domain, !domain.isEmpty else { return nil } at the top.

Now nil-domain simply returns nil (attestation skipped) instead of crashing. The ObjC signature stays
the same, callers don't change.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
sfdctaka: Good point. I'm going to defer that to @wmathurin . I'm good either way 👍

wmathurin: I'd split the guard in two:

1. **Simulator**: always guard on `isSupported` — no attestation will ever succeed there and the noise of network calls + log lines on every token refresh is not useful.
2. **Older devices / MDM enforcement**: agree with bbirman — if the admin has enabled strict ECA enforcement, we *want* the failure logged. For those, let it fall through to `performAppAttestOperation` so the `DCError.featureUnsupported` surfaces in the error log with context.

A simple way to handle both: check `isSupported` only in simulator builds (via `#if targetEnvironment(simulator)`), or alternatively check `isSupported && !isSimulator` — but actually the cleanest is to just always run through the normal path and let the logging in `performAppAttestOperation` do its job for real devices. The simulator case is the only one where we'd want a silent early return.

Or simplest of all: add `DCAppAttestService.shared.isSupported` to the existing guard so all unsupported environments are gated out silently, and document in the `appAttestationEnabled` property comment that apps should only enable it on builds with the App Attest entitlement. That makes the developer's intent explicit, and MDM/ECA enforcement scenarios should be handled by the server rejecting the request (not the client pre-screening). What do you think?

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
wmathurin: Good catch. sfdctaka's fix is exactly right — change `domain: String` to `domain: String?` and add a `guard let domain, !domain.isEmpty` at the top. The normal login flow won't hit this (domain defaults to `login.salesforce.com` in the credentials initializer), but test doubles, decoded credentials with a missing key, or direct construction with `domain = nil` would otherwise crash instead of skipping attestation gracefully. Please apply the fix.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
wmathurin: Clarifying one thing on the ECA concern: ECA (External Client Application) enforcement is configured server-side on the connected app. So if a device has `isSupported = false` and skips attestation, the server will reject the token request with a clear OAuth error — that surfaces through the normal auth error path with a meaningful message. The client-side `DCError.featureUnsupported` log isn't needed for diagnosing ECA misconfiguration; the server response handles that.

So I'd go with adding `DCAppAttestService.shared.isSupported` to the existing guard. The only thing the client-side check buys is skipping the wasted challenge network call on devices that can never produce a valid attestation (simulator, iOS < 14). No diagnostic value is lost.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: 👍 double checked, neither

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: updated!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: updated!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/AppAttestation.swift
bbirman: updated!

sfdctaka (approved):
LGTM!

wmathurin (approved):

PR #4098, opened 2026-07-09
Fix nil-sceneId crash on advanced-auth browser callback for pre-scene logins

brandonpage:
## What

Fixes a crash on the advanced-authentication browser callback for logins that start **before any `UIScene` has connected** — i.e. apps that begin login from `AppDelegate.didFinishLaunchingWithOptions` (hybrid and React Native).

## Root cause

When login starts pre-scene, `request.scene.session.persistentIdentifier` is `nil`, so `SFSDKAuthSession.sceneId` was `nil`. On the `ASWebAuthenticationSession` completion handler, the SDK built an options dictionary keyed by the scene id (`kSFIDPSceneIdKey`) and looked the session back up in the `authSessions` store by that id. A `nil` key both crashed the dictionary insert and dropped the session from the store.

## Fix

- **`SFSDKAuthSession.m`** — synthesize a unique per-session scene id (`com.salesforce.mobilesdk.unscopedAuthSession-<UUID>`) when no scene is connected, so each scene-less session gets its own `authSessions[]` key and the browser callback can key back to it.
- **`SFOAuthCoordinator.m`** — add a `nil` guard when building the callback options dictionary.
- **`SFOAuthCoordinatorTests.m`** — unit tests covering the scene-less login case.

## Testing

- New unit tests for the scene-less path.
- End-to-end proven on a generated hybrid app with `forceAdvancedAuthentication` enabled: browser login → token exchange → account + photo fetch all succeed with no crash (previously crashed on the browser callback).

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator+Internal.h
@@ -88,6 +88,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)migrateRefreshToken:(SFUserAccount *)user;
 
+/**
+ Builds the options dictionary handed to the URL handler on the advanced-auth browser callback.
+ Guards against a nil sceneId so nil is never inserted into the dictionary.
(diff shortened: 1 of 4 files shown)


sfdctaka (approved):
LGTM!

wmathurin (approved):

PR #4100, opened 2026-07-10
Use web server flow when app attestation is configured

bbirman:


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -151,7 +151,7 @@ - (void)authenticate {
         self.authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeRefresh];
     } else if (self.useBrowserAuth) {
         self.authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeAdvancedBrowser];
-    } else if ([[SalesforceSDKManager sharedManager] useWebServerAuthentication]) {
+    } else if ([[SalesforceSDKManager sharedManager] useWebServerAuthentication] || [SFUserAccountManager sharedInstance].appAttestationEnabled) {
```
wmathurin: Maybe we could have a method that tells us whether to use web server flow instead of having the same || expression in multiple places.
That's the approach that was taken on Android: see https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/2913/changes#diff-2b4d6c9697f160bfbefe121309b911622ea530d287a7fbc780a872cfd75cd82aR230

wmathurin (approved):

PR #4102, opened 2026-07-13
Improve error handling at token refresh (app attestation)

sfdctaka:
## Summary

Adds first-class handling for the server's App Attestation error responses during token refresh, so the SDK reacts appropriately instead of falling through to a generic refresh failure.

### Changes

- Add `kSFOAuthErrorAppAttestationFailed` and `kSFOAuthErrorAppAttestationFailedRetry` to the `SFOAuthErrorDomain` error codes
- Add `SFLogoutReasonAppAttestationFailed` to the `SFLogoutReason` enum
- Map wire values `client_blocked` / `client_blocked_retry` in `SFSDKOAuth2 errorWithType:description:underlyingError:`
- In `SFRestAPI replayRequest:response:`, branch on the new codes:
  - `client_blocked` → logout with `SFLogoutReasonAppAttestationFailed`
  - `client_blocked_retry` → propagate the error to pending requests without logout, so the app can retry
- Preserve existing `invalid_grant` → `logout(TokenExpired)` behavior

## Test plan

- [x] Unit test wire-string → error code mapping for both `client_blocked` and `client_blocked_retry` (Swift)
- [x] Unit test `replayRequest:response:` branching for `invalid_grant`, `client_blocked`, and `client_blocked_retry` (Objective-C)
- [ ] Manual verification against a server returning the new error strings

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
@@ -481,12 +481,22 @@ successBlock:(SFRestResponseBlock)successBlock
             [strongSelf flushPendingRequestQueue:refreshError rawResponse:response];
             strongSelf.refreshCycleActive = NO;
         }
-        if ([refreshError.domain isEqualToString:kSFOAuthErrorDomain] && refreshError.code == kSFOAuthErrorInvalidGrant) {
-            [SFSDKCoreLogger i:[strongSelf class] format:@"%@ Invalid grant error received, triggering logout.", NSStringFromSelector(_cmd)];
-            // Make sure we call logout on the main thread.
-            dispatch_async(dispatch_get_main_queue(), ^{
-                [[SFUserAccountManager sharedInstance] logoutUser:strongSelf.user reason:SFLogoutReasonTokenExpired];
-            });
+        if ([refreshError.domain isEqualToString:kSFOAuthErrorDomain]) {
+            SFUserAccount *user = strongSelf.user;
+            void (^triggerLogout)(SFLogoutReason, NSString *) = ^(SFLogoutReason reason, NSString *logMessage) {
+                [SFSDKCoreLogger i:[strongSelf class] format:@"%@ %@", NSStringFromSelector(_cmd), logMessage];
+                dispatch_async(dispatch_get_main_queue(), ^{
+                    [[SFUserAccountManager sharedInstance] logoutUser:user reason:reason];
+                });
+            };
+            NSInteger errorCode = [SFOAuthErrorCodeHelper from:refreshError.userInfo[kSFOAuthError]];
+            if (errorCode == SFOAuthErrorCodeInvalidGrant) {
+                triggerLogout(SFLogoutReasonTokenExpired, @"Invalid grant error received, triggering logout.");
+            } else if (errorCode == SFOAuthErrorCodeAppAttestationFailed) {
+                triggerLogout(SFLogoutReasonAppAttestationFailed, @"App attestation failed, triggering logout.");
+            } else if (errorCode == SFOAuthErrorCodeAppAttestationFailedRetry) {
+                [SFSDKCoreLogger i:[strongSelf class] format:@"%@ App attestation retry needed, no automatic logout.", NSStringFromSelector(_cmd)];
+            }
         }
     }];
 }
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
@@ -63,7 +63,9 @@ enum {
     kSFOAuthErrorJWTInvalidGrant,
     kSFOAuthErrorRequestCancelled,
     kSFOAuthErrorRefreshFailed, //generic error
-    kSFOAuthErrorInvalidURL
+    kSFOAuthErrorInvalidURL,
+    kSFOAuthErrorAppAttestationFailed,       // app attestation failed (terminal)
+    kSFOAuthErrorAppAttestationFailedRetry   // app attestation failed (client should retry)
 };
 
 typedef NS_ENUM(NSInteger, SFLogoutReason) {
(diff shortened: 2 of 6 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
```diff
@@ -63,7 +63,9 @@ enum {
     kSFOAuthErrorJWTInvalidGrant,
     kSFOAuthErrorRequestCancelled,
     kSFOAuthErrorRefreshFailed, //generic error
-    kSFOAuthErrorInvalidURL
+    kSFOAuthErrorInvalidURL,
```
wmathurin: **Spec contradiction — these two constants should not be added.**

The spec for this work item (`specs/W-23059472-W-23059473-oauth-error-code-enum/spec.md §3`) is explicit:

> The existing if-else chain is **not extended** for `client_blocked` or `client_blocked_retry` — any code that needs to distinguish those errors must use `response.error.errorCode` (the typed `SFOAuthErrorCode` property), not the `NSError` integer code. Adding new integer constants for these values would just perpetuate the fragile pattern the enum is designed to replace.

The correct approach: after refresh fails, read the typed `SFOAuthErrorCode` from the endpoint response (already available on `response.error.errorCode`) and branch on that instead of on `refreshError.code`.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
```diff
@@ -473,6 +473,8 @@ + (NSString *)stringValueForLogoutReason:(SFLogoutReason)reason {
             return @"unexpected_response";
         case SFLogoutReasonRefreshTokenRotated:
             return @"refresh_token_rotated";
+        case SFLogoutReasonAppAttestationFailed:
```
wmathurin, line 470: **Missing switch case — undefined behavior at runtime.**

`SFLogoutReasonAppAttestationFailed` is added to the enum but this `switch` has no `default:` and no case for the new value. When `logoutUser:reason:SFLogoutReasonAppAttestationFailed` fires, this method falls off the end of a non-void ObjC function (UB). In practice it returns `nil`, silently corrupting the `logout_reason` field in the token-revocation HTTP request.

Fix:
```objc
case SFLogoutReasonAppAttestationFailed:
    return @"app_attestation_failed";
```
Also worth adding `default: return @"unknown";` so future additions don't silently regress.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
```diff
@@ -479,13 +479,24 @@ - (void)replayRequest:(SFRestRequest *)request response:(NSURLResponse *)respons
...
+                        // Make sure we call logout on the main thread.
+                        dispatch_async(dispatch_get_main_queue(), ^{
+                            [[SFUserAccountManager sharedInstance] logoutUser:strongSelf.user reason:SFLogoutReasonTokenExpired];
+                        });
+                    } else if (refreshError.code == kSFOAuthErrorAppAttestationFailed) {
+                        [SFSDKCoreLogger i:[strongSelf class] format:@"%@ App attestation failed, triggering logout.", NSStringFromSelector(_cmd)];
+
+                        // Make sure we call logout on the main thread.
+                        dispatch_async(dispatch_get_main_queue(), ^{
+                            [[SFUserAccountManager sharedInstance] logoutUser:strongSelf.user reason:SFLogoutReasonAppAttestationFailed];
```
wmathurin: **Duplicate structure + unnecessary `strongSelf` retain (nit).**

The `kSFOAuthErrorInvalidGrant` and `kSFOAuthErrorAppAttestationFailed` branches are structurally identical (log → `dispatch_async` → `logoutUser:reason:`). A local block removes the duplication and makes future error codes a one-liner:

```objc
void (^triggerLogout)(SFLogoutReason, NSString *) = ^(SFLogoutReason reason, NSString *msg) {
    [SFSDKCoreLogger i:[strongSelf class] format:msg, NSStringFromSelector(_cmd)];
    SFUserAccount *user = strongSelf.user;
        [[SFUserAccountManager sharedInstance] logoutUser:user reason:reason];
    });
};
```

Note the `SFUserAccount *user` capture: each `dispatch_async` block currently captures `strongSelf` (the whole `SFRestAPI` instance) but only uses `strongSelf.user`. Capturing only `user` releases the `SFRestAPI` retain sooner.

wmathurin (commented):
Three findings from review — two blockers and one cleanup suggestion. Details are inline.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.h
sfdctaka: Good catch — reverted. The integer constants are removed from `SFSDKOAuth2.h` and the wire-value mappings are gone from `SFSDKOAuth2.m`. `SFRestAPI` now reads `refreshError.userInfo[kSFOAuthError]` and branches on the typed `SFOAuthErrorCode` (via `SFOAuthErrorCodeHelper.from(_:)`), matching the pattern in the enum spec. Fixed in 135fcb70e.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Util/SFSDKOAuth2.m
sfdctaka: This one was actually already addressed in the original commit — the `case SFLogoutReasonAppAttestationFailed: return @"app_attestation_failed";` was added right below `SFLogoutReasonRefreshTokenRotated` (SFSDKOAuth2.m:476–477). Line 476 in the diff view lands on the newly-added case itself, so I think this was a rendering artifact of the review UI rather than a missing case. Happy to double-check if you see it differently.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
sfdctaka: Applied. Extracted a local `triggerLogout(reason, message)` block used by both the `invalidGrant` and `appAttestationFailed` branches. The dispatched block now captures a local `SFUserAccount *user = strongSelf.user` instead of the whole `SFRestAPI` instance, releasing the `SFRestAPI` retain sooner. Fixed in 135fcb70e.

wmathurin (approved):

PR #4105, opened 2026-07-14
Fix login-host recovery on iOS 26: broaden classifier + surface prefetch errors

sfdctaka:
## What

On iOS 26 the login-host recovery path is never entered when a user enters an unreachable custom login host at the picker. Two independent defects contribute to the same user-visible symptom:

1. The `HostConnectionErrorHandler` classifier in `SFSDKAuthErrorManager` did not recognize modern iOS 26 error shapes.
2. `SFOAuthCoordinator` silently swallowed host-connection errors from the auth-configuration pre-flight, so the failure delegate was never invoked and the classifier never got a chance to fire.

Symptom before this PR: the app either loops on an alert (fix #1's regression) or, worse, hangs on a blank splash with no alert at all (fix #2's regression) and the bad host stays set as `loginHost` across app restarts.

## Root cause 1 — classifier misses iOS 26 error shapes

The classifier gated on:

```objc
if ((error.userInfo[@\"_kCFStreamErrorCodeKey\"] && error.userInfo[@\"_kCFStreamErrorDomainKey\"]) ||
    ([error.domain isEqualToString:kSFOAuthErrorDomain] && error.code == kSFOAuthErrorInvalidURL)) {
    ...
    return YES;
}
return NO;
```

Both iOS 18 and iOS 26 surface DNS failures at the top level as `NSURLErrorDomain / -1003 CannotFindHost`, but the two versions structure `userInfo` differently:

| | iOS 18 (works) | iOS 26 (misses) |
|---|---|---|
| Top-level CFStream keys | `_kCFStreamErrorDomainKey=12`, `_kCFStreamErrorCodeKey=8` **present** | **absent** |
| `NSUnderlyingError` | `kCFErrorDomainCFNetwork / -1003` | `kNWErrorDomainDNS / -65554 NoSuchRecord` |
| Classifier result | TRUE → host-connection recovery fires | FALSE → wrong path |

Apple moved DNS resolution from the legacy CFNetwork stream stack to Network.framework starting in iOS 26. The `_kCFStreamError*` sentinels moved to a nested `kNWErrorDomainDNS` underlying error and are no longer present at the top level of `userInfo`.

## Root cause 2 — coordinator swallows prefetch errors

`SFOAuthCoordinator.authenticate` calls `SFSDKAuthConfigUtil.getMyDomainAuthConfig` as a pre-flight probe. On failure the callback logged and fell straight through to `beginWebViewFlow`, dispatching a WKWebView load against the same unreachable host. The delegate's `oauthCoordinator:didFailWithError:authInfo:` was never called, so the error manager never saw the error, no alert was shown, and `loginHost` was never rolled back to `previousLoginHost` — hence the sticky-splash-hang across restarts.

### 1. Broaden the classifier

Extract host-connection classification into a shared class method `+[SFSDKAuthErrorManager errorIsHostConnectionFailure:]` and keep the two existing branches (`_kCFStreamError*` keys and `kSFOAuthErrorInvalidURL`), adding a third that switches on six `NSURLErrorDomain` codes:

- `NSURLErrorCannotFindHost` (-1003)
- `NSURLErrorDNSLookupFailed` (-1006)
- `NSURLErrorCannotConnectToHost` (-1004)
- `NSURLErrorTimedOut` (-1001)
- `NSURLErrorNotConnectedToInternet` (-1009)
- `NSURLErrorNetworkConnectionLost` (-1005)

Classification now keys off `NSURLErrorDomain` top-level codes, not off `userInfo` structure — future changes to how Apple structures the underlying error will not re-break the classifier.

Handler ordering is unchanged: `NetworkFailureErrorHandler` still claims network codes for Refresh flows with an existing access token **before** the host-connection handler sees them. A dedicated ordering test guards this.

### 2. Surface prefetch host-connection errors

In `SFOAuthCoordinator.m`'s `getMyDomainAuthConfig:` callback, gate on the same shared predicate. When prefetch returns a host-connection error, route through `notifyDelegateOfFailure:` so the error manager's classifier can present the alert and let `SFUserAccountManager`'s recovery handler roll `loginHost` back to `previousLoginHost`.

Non-host errors (parse failures, non-2xx responses, endpoint absent on standard orgs) still fall through — auth-config remains optional.

The predicate is intentionally kept internal to `SFSDKAuthErrorManager`. `SFOAuthCoordinator.m` and the test file both reach it via forward-declared categories, so no public header changes and no `.xcodeproj` file changes.

## Test plan

New tests in `SFSDKErrorManagerTests.m`:

- [x] Six positive tests — one per `NSURLErrorDomain` host-connection code — assert the host-connection handler is invoked with a bare `NSError` (no CFStream keys)
- [x] Legacy iOS <= 18 shape (`NSURLErrorDomain / -1003` **with** CFStream keys in `userInfo`) still claimed by the host-connection handler
- [x] `kSFOAuthErrorInvalidURL` still claimed by the host-connection handler
- [x] Three negative tests — `NSURLErrorCancelled`, `NSURLErrorUserAuthenticationRequired`, `kSFOAuthErrorInvalidGrant` — assert the host-connection handler is **not** invoked and the correct fallback handler claims each error
- [x] Ordering guard — on a Refresh flow with an existing access token, `NetworkFailureErrorHandler` still claims `NSURLErrorTimedOut` before the host-connection handler sees it
- [x] Five predicate tests directly on `+errorIsHostConnectionFailure:` (nil inputs, legacy CFStream keys, `kSFOAuthErrorInvalidURL`, six NSURL codes, unrelated codes)
- [x] Full `SFSDKErrorManagerTests` suite (22 methods) passes on iOS 26 simulator
- [x] Manual repro on iPhone 17 Pro / iOS 26.2 simulator: entering `https://a.z` at the picker triggers the \"Server Error / Can't connect to the server: The request timed out.\" alert (previously silent splash-hang); OK dismisses; previous host resumes
- [x] Manual regression on iOS 18 simulator: same repro continues to recover as before

## Blast radius

- Two production files:
  - `libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m` — one classifier block refactored to use a shared class method
  - `libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m` — one gate added in the auth-config callback
- One test file: `libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKErrorManagerTests.m` — appended
- No public header changes, no `.xcodeproj` changes, no `Localizable.strings` changes, no new dependencies

## Credit

Root-cause analysis and proposed classifier fix by Matt Wimmer.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -28,6 +28,7 @@
 #import "SFOAuthCoordinator+Internal.h"
 #import "SFOAuthInfo.h"
 #import "SFSDKAuthConfigUtil.h"
+#import "SFSDKAuthErrorManager.h"
 #import "SFSDKCryptoUtils.h"
 #import "NSData+SFSDKUtils.h"
 #import "NSString+SFAdditions.h"
@@ -55,6 +56,14 @@
 #import <SalesforceSDKCommon/SFSDKDatasharingHelper.h>
 #import <LocalAuthentication/LocalAuthentication.h>
 #import "SFSDKResourceUtils.h"
+// Forward declaration of the shared host-connection classifier so this file
+// stays in sync with the classifier block in SFSDKAuthErrorManager. Not in the
+// public header: this is an internal predicate, tested directly in
+// SFSDKErrorManagerTests.
+@interface SFSDKAuthErrorManager (SFOAuthCoordinatorInternalUse)
++ (BOOL)errorIsHostConnectionFailure:(NSError *)error;
+@end
+
 @interface SFOAuthCoordinator()
 
 @property (nonatomic) NSString *networkIdentifier;
@@ -197,8 +206,19 @@
             [SFSDKAuthConfigUtil getMyDomainAuthConfig:^(SFOAuthOrgAuthConfiguration *authConfig, NSError *error) {
                 __strong typeof(weakSelf) strongSelf = weakSelf;
                 dispatch_async(dispatch_get_main_queue(), ^{
-                    // Ignore any errors why retrieving authconfig. Default to WKWebView
-                    // Errors should have already been logged.
+                    // A prefetch error against an unreachable login host is the earliest
+                    // authoritative signal that the host itself is bad. Falling through to
+                    // beginWebViewFlow would dispatch a WKWebView load against the same host
+                    // and hang silently, stranding the user on a blank screen and leaving the
+                    // bad host set as the sticky loginHost across app restarts. Surface it to
+                    // the failure delegate so the error manager's hostConnectionErrorHandler
+                    // can present an alert and roll loginHost back to the previous good host.
+                    // Non-host errors (parse failures, non-2xx responses, endpoint absent on
+                    // standard orgs) continue to fall through — auth-config is optional.
+                    if ([SFSDKAuthErrorManager errorIsHostConnectionFailure:error]) {
+                        [strongSelf notifyDelegateOfFailure:error authInfo:strongSelf.authInfo];
+                        return;
+                    }
                     if (!self.frontdoorBridgeLoginOverride &&
                         (authConfig.useNativeBrowserForAuth ||
                          [SalesforceSDKManager sharedManager].sdk_forceAdvancedAuthentication)) {
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -56,13 +56,7 @@
 #import <SalesforceSDKCommon/SFSDKDatasharingHelper.h>
 #import <LocalAuthentication/LocalAuthentication.h>
 #import "SFSDKResourceUtils.h"
-// Forward declaration of the shared host-connection classifier so this file
-// stays in sync with the classifier block in SFSDKAuthErrorManager. Not in the
-// public header: this is an internal predicate, tested directly in
-// SFSDKErrorManagerTests.
-@interface SFSDKAuthErrorManager (SFOAuthCoordinatorInternalUse)
-+ (BOOL)errorIsHostConnectionFailure:(NSError *)error;
-@end
+#import "SFSDKAuthErrorManager+Internal.h"
 
 @interface SFOAuthCoordinator()
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
@@ -28,6 +28,7 @@
  */
 
 #import "SFSDKAuthErrorManager.h"
+#import "SFSDKAuthErrorManager+Internal.h"
 #import "SFAuthErrorHandlerList.h"
 #import "SFAuthErrorHandler.h"
 #import "SFOAuthCoordinator+Internal.h"
@@ -135,7 +136,13 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
                                            }];
     [authHandlerList addAuthErrorHandler:self.networkFailureAuthErrorHandler];
     
-    // Host connection error handler
+    // Host connection error handler.
+    //
+    // NSURLErrorTimedOut / CannotConnectToHost / NetworkConnectionLost / NotConnectedToInternet
+    // also appear in +errorIsNetworkFailure:. NetworkFailureErrorHandler runs first in the chain
+    // and claims those codes only on Refresh flows with an existing access token; all other
+    // contexts return NO there and fall through to this handler. Ordering is guarded by
+    // testNetworkFailureClaimsFirst_RefreshWithToken.
     self.hostConnectionErrorHandler = [[SFAuthErrorHandler alloc] initWithName:kSFHostConnectionErrorHandler
                                      authSessionBlock:^BOOL(NSError *error, SFSDKAuthSession *authSession, NSDictionary *options) {
                                         if ([[weakSelf class] errorIsHostConnectionFailure:error]) {
@@ -186,10 +193,11 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
 }
 
 /**
- * Evaluates an NSError object to see if it represents a network failure during
- * an attempted connection.
+ * Evaluates an NSError object to see if it represents a host-connection failure —
+ * an unreachable or non-existent login host — rather than a transient network
+ * failure on an otherwise reachable host.
  * @param error The NSError to evaluate.
- * @return YES if the error represents a network failure, NO otherwise.
+ * @return YES if the error should trigger the host-connection recovery path, NO otherwise.
  */
 + (BOOL)errorIsHostConnectionFailure:(NSError *)error
 {
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
@@ -138,23 +138,7 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
     // Host connection error handler
     self.hostConnectionErrorHandler = [[SFAuthErrorHandler alloc] initWithName:kSFHostConnectionErrorHandler
                                      authSessionBlock:^BOOL(NSError *error, SFSDKAuthSession *authSession, NSDictionary *options) {
-                                        BOOL cfStreamHostError = (error.userInfo[@"_kCFStreamErrorCodeKey"] && error.userInfo[@"_kCFStreamErrorDomainKey"]);
-                                        BOOL oauthInvalidURL   = ([error.domain isEqualToString:kSFOAuthErrorDomain] && error.code == kSFOAuthErrorInvalidURL);
-                                        BOOL urlHostError      = NO;
-                                        if ([error.domain isEqualToString:NSURLErrorDomain]) {
-                                            switch (error.code) {
-                                                case NSURLErrorCannotFindHost:            // -1003
-                                                case NSURLErrorDNSLookupFailed:           // -1006
-                                                case NSURLErrorCannotConnectToHost:       // -1004
-                                                case NSURLErrorTimedOut:                  // -1001
-                                                case NSURLErrorNotConnectedToInternet:    // -1009
-                                                case NSURLErrorNetworkConnectionLost:     // -1005
-                                                    urlHostError = YES;
-                                                    break;
-                                                default: break;
-                                            }
-                                        }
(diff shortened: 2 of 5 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
```diff
@@ -138,11 +138,26 @@ - (SFAuthErrorHandlerList *)populateDefaultAuthErrorHandlerList
...
-                                        if ((error.userInfo[@"_kCFStreamErrorCodeKey"] && error.userInfo[@"_kCFStreamErrorDomainKey"]) ||
-                                            ([error.domain isEqualToString:kSFOAuthErrorDomain] && error.code == kSFOAuthErrorInvalidURL)) {
+                                        BOOL cfStreamHostError = (error.userInfo[@"_kCFStreamErrorCodeKey"] && error.userInfo[@"_kCFStreamErrorDomainKey"]);
+                                        BOOL oauthInvalidURL   = ([error.domain isEqualToString:kSFOAuthErrorDomain] && error.code == kSFOAuthErrorInvalidURL);
+                                        BOOL urlHostError      = NO;
+                                        if ([error.domain isEqualToString:NSURLErrorDomain]) {
+                                            switch (error.code) {
+                                                case NSURLErrorCannotFindHost:            // -1003
+                                                case NSURLErrorDNSLookupFailed:           // -1006
+                                                case NSURLErrorCannotConnectToHost:       // -1004
```
wmathurin: The four codes below (`NSURLErrorCannotConnectToHost`, `NSURLErrorTimedOut`, `NSURLErrorNotConnectedToInternet`, `NSURLErrorNetworkConnectionLost`) overlap with `+errorIsNetworkFailure:`. `NetworkFailureErrorHandler` runs first in the chain but only claims those codes when `authType == Refresh && accessToken != nil`; for all other auth contexts it returns `NO` and they fall through here. The behavior is correct and is guarded by `testNetworkFailureClaimsFirst_RefreshWithToken`, but a future reader who extends `errorIsNetworkFailure:` won't know to check the downstream handler. Worth a brief comment here:

```objc
// NSURLErrorTimedOut / CannotConnectToHost / NetworkConnectionLost / NotConnectedToInternet
// also appear in +errorIsNetworkFailure:. NetworkFailureErrorHandler runs first and claims
// those codes only on Refresh+existing-token flows; all other contexts fall through to here.
```

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKErrorManagerTests.m
```diff
@@ -127,15 +127,209 @@ - (void)testGenericError {
...
+        [networkErrorExpectation fulfill];
+    };
+    errorManager.hostConnectionErrorHandlerBlock = ^(NSError *e, SFSDKAuthSession *s, NSDictionary *o) {
+        XCTFail(@"Host connection handler must not claim network errors on Refresh-with-token flows");
+    };
+
+    XCTAssertNotNil(errorManager.networkErrorHandlerBlock);
+    BOOL handled = [errorManager processAuthError:error authContext:session options:userInfo];
+    XCTAssertTrue(handled, @"Network Error Should have been handled by the ErrorManager on Refresh flow");
+    [[SFUserAccountManager sharedInstance] deleteAccountForUser:account error:nil];
```
wmathurin: `deleteAccountForUser:error:` is called before `waitForExpectationsWithTimeout:`. `processAuthError:` drives the callback synchronously so the expectation is already fulfilled by this point and the test passes today. But if the handler ever became async, cleanup would race ahead of the assertion and the test would pass vacuously. Safer to move cleanup after the wait:

```objc
[self waitForExpectationsWithTimeout:20.0 handler:nil];
[[SFUserAccountManager sharedInstance] deleteAccountForUser:account error:nil];
```

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKErrorManagerTests.m
```diff
@@ -127,15 +127,209 @@ - (void)testGenericError {
...
 
+#pragma mark - Host connection classifier tests
+
+// Shared helper: assert that the host connection handler claims a bare NSError
+// (no CFStream keys) constructed with the given domain and code, when the
+// session is not a Refresh-with-existing-token flow.
+- (void)assertHostConnectionClaimsDomain:(NSString *)domain code:(NSInteger)code description:(NSString *)description {
+    SFSDKAuthErrorManager *errorManager = [[SFSDKAuthErrorManager alloc] init];
+    SFSDKAuthSession *authSession = [[SFSDKAuthSession alloc] init];
+    XCTAssertNotNil(errorManager);
```
wmathurin: `[[SFSDKAuthErrorManager alloc] init]` cannot return `nil` — the designated initializer always returns `self`. This assertion doesn't add safety (and the same applies at line 316). Not blocking, but consider dropping these from new test code so they don't imply nil-alloc is a real failure mode here.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKErrorManagerTests.m
```diff
@@ -127,15 +127,209 @@ - (void)testGenericError {
...
+    XCTestExpectation *expectation = [self expectationWithDescription:description];
+    NSDictionary *userInfo = [[NSMutableDictionary alloc] init];
```
wmathurin: `[[NSMutableDictionary alloc] init]` for an empty `userInfo` — `@{}` is the idiomatic Objective-C literal for an empty immutable dictionary and is cleaner. The pattern is consistent with existing tests in the file, but worth tidying in new code:

```objc
NSDictionary *userInfo = @{};
```

wmathurin (commented):
## Updated review — both commits

**Commit 1** (`deeeb59`) broadened the host-connection classifier. **Commit 2** (`8666b33`) extracted it into `+errorIsHostConnectionFailure:` (good — eliminates the inline duplication I flagged in the first review) and used it to fail-fast in the `getMyDomainAuthConfig:` callback in `SFOAuthCoordinator`, preventing the silent WKWebView hang on a bad host.

Four inline comments follow.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFSDKAuthErrorManager.m
```diff
@@ -192,6 +191,37 @@ + (BOOL)errorIsInvalidAuthCredentials:(NSError *)error
  * @param error The NSError to evaluate.
  * @return YES if the error represents a network failure, NO otherwise.
  */
++ (BOOL)errorIsHostConnectionFailure:(NSError *)error
```
wmathurin, line 202: **Copy-paste docstring mismatch.** The docblock immediately above (lines 188–193) was copied from `+errorIsNetworkFailure:` and still reads _"represents a network failure during an attempted connection"_ / _"YES if the error represents a network failure"_. It now documents `+errorIsHostConnectionFailure:`. Suggested replacement:

```objc
/**
 * Returns YES if the error represents a host-connection failure — an unreachable
 * or non-existent login host — rather than a transient network failure on an
 * otherwise reachable host.
 * @param error The NSError to evaluate.
 * @return YES if the error should trigger the host-connection recovery path, NO otherwise.
 */
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -55,6 +56,14 @@
 #import <SalesforceSDKCommon/SFSDKDatasharingHelper.h>
 #import <LocalAuthentication/LocalAuthentication.h>
 #import "SFSDKResourceUtils.h"
+// Forward declaration of the shared host-connection classifier so this file
+// stays in sync with the classifier block in SFSDKAuthErrorManager. Not in the
+// public header: this is an internal predicate, tested directly in
+// SFSDKErrorManagerTests.
+@interface SFSDKAuthErrorManager (SFOAuthCoordinatorInternalUse)
++ (BOOL)errorIsHostConnectionFailure:(NSError *)error;
```
wmathurin: **Forward-declaration duplication risk.** This category declaration (and the matching one in the test file with a different category name) works today, but if `+errorIsHostConnectionFailure:` is ever renamed or its signature changes, the compiler will silently accept the stale declarations — informal category declarations bypass the "method not found" warning you'd get from a header import. Both call sites would then call the old selector at runtime.

Consider extracting a `SFSDKAuthErrorManager+Internal.h` with this single declaration, imported by both `SFOAuthCoordinator.m` and the test file. That keeps it off the public API surface while giving the compiler one source of truth to enforce.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKErrorManagerTests.m
```diff
@@ -127,15 +134,265 @@ - (void)testGenericError {
...
+#pragma mark - errorIsHostConnectionFailure predicate
+
+// The shared predicate is the single source of truth for what counts as a
+// host-connection failure. Both the classifier block and SFOAuthCoordinator's
+// auth-config prefetch callback consult it, so exercising it directly guards
+// both call sites against silent classification drift.
+
+- (void)testErrorIsHostConnectionFailure_NilInputs {
+    XCTAssertFalse([SFSDKAuthErrorManager errorIsHostConnectionFailure:nil]);
+    NSError *nilDomain = [NSError errorWithDomain:@"" code:0 userInfo:nil];
```
wmathurin: **Test name/coverage mismatch.** The variable is named `nilDomain` but it is constructed with `@""` — an empty string — not `nil`. `+errorWithDomain:code:userInfo:` raises `NSInvalidArgumentException` if domain is actually `nil`, so the `error.domain == nil` guard in `+errorIsHostConnectionFailure:` is untestable via this initializer and is not exercised here. The test is still useful (empty/foreign domain falls through to `return NO`), but the name implies nil-domain coverage that doesn't exist. Suggest renaming to `emptyDomain` and adding a brief comment explaining why nil-domain is not tested.

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKErrorManagerTests.m
```diff
@@ -127,15 +134,265 @@ - (void)testGenericError {
...
+        [networkErrorExpectation fulfill];
+    };
+    errorManager.hostConnectionErrorHandlerBlock = ^(NSError *e, SFSDKAuthSession *s, NSDictionary *o) {
+        XCTFail(@"Host connection handler must not claim network errors on Refresh-with-token flows");
+    };
+
+    XCTAssertNotNil(errorManager.networkErrorHandlerBlock);
+    BOOL handled = [errorManager processAuthError:error authContext:session options:userInfo];
+    XCTAssertTrue(handled, @"Network Error Should have been handled by the ErrorManager on Refresh flow");
+    [[SFUserAccountManager sharedInstance] deleteAccountForUser:account error:nil];
```
wmathurin: **Cleanup before wait.** `deleteAccountForUser:error:` runs before `waitForExpectationsWithTimeout:`. `processAuthError:` is synchronous today so the expectation is fulfilled by this point, but if the handler were ever made async the account would be torn down before the network block fires — the session state becomes a dangling reference and the test passes vacuously. Move cleanup after the wait:

```objc
[self waitForExpectationsWithTimeout:20.0 handler:nil];
[[SFUserAccountManager sharedInstance] deleteAccountForUser:account error:nil];
```

wmathurin (approved):

PR #4106, opened 2026-07-15
DPoP: send dpop_jkt on /authorize (RFC 9449 §10 code binding)

sfdctaka:
## Summary

Bind the authorization code returned by `/authorize` to the DPoP key pair that will later prove possession at `/token`, per **RFC 9449 §10**. Mirrors the Android implementation.

When `SalesforceSDKManager.usesDPoP` is `YES`, the login server is a my-domain host, and `credentials.identifier` is set, the approval URL now carries `dpop_jkt=<RFC 7638 thumbprint>`. Pool hosts (`login/test/welcome.salesforce.com`) are excluded — Salesforce blocks DPoP at the pool servers.

### Changes

- **`DPoPProofBuilder.jwkThumbprint(publicKey:)`** — new `@objc` static that returns the 43-char base64url RFC 7638 thumbprint. Composed from existing primitives (`Encryptor.jwkP256`, `JSONSerialization(.sortedKeys)`, `sfsdk_sha256`, `sfsdk_base64UrlString`). New `.thumbprintFailed` error case.
- **`SFSDKAuthConfigUtil.isPoolLoginHost:`** — new predicate. The three pool-host string constants are lifted from `.m` file-locals into the header as `FOUNDATION_EXTERN` so both the auth-config prefetch check and the dpop_jkt gate share one source of truth.
- **`SFOAuthCoordinator.approvalURLForEndpoint:…`** — after the existing scope/login-hint appends, invokes a new private helper `-appendDPoPJktIfNeededTo:domain:credentials:` that gates on `usesDPoP` + `isPoolLoginHost` + non-empty identifier, loads the key pair via `DPoPKeyStore.keyPair(forCredentials:)`, and appends `&dpop_jkt=…`. Crypto/keychain errors soft-fail (warn-log, URL untouched, login proceeds).
- **`kSFOAuthDPoPJktParamName`** — new constant next to `kSFOAuthCodeChallengeParamName`.

### Tests

- `DPoPProofBuilder.jwkThumbprint` — RFC 7638 fixture against RFC 6979 §A.2.5 P-256 test key, uniqueness across pairs, 43-char base64url shape.
- `SFOAuthCoordinator.generateApprovalUrlString`:
  - `dpop_jkt` present + shape correct under DPoP + my-domain + identifier
  - `dpop_jkt` value equals `jwkThumbprint(publicKey:)` of the pair `DPoPKeyStore` returns at `/token` time (authorize↔token binding invariant)
  - All three pool hosts (`login`, `test`, `welcome.../discovery`) — no `dpop_jkt`
  - `usesDPoP=NO` — byte-identical baseline; no `dpop_jkt`
  - Empty `credentials.identifier` — soft-fail path, no `dpop_jkt`, no exception
  - `migrateRefreshToken:` construction path carries `dpop_jkt`
  - Both entry points: user-agent (response_type=token/hybrid_token) and web-server (response_type=code + code_challenge)

## Test plan

- [x] Manual: log into a `Require DPoP` ECA on a my-domain org via RestAPIExplorer with `usesDPoP=YES`; confirm `dpop_jkt` present on the `/authorize` request and matches the token-endpoint JWK thumbprint.
- [x] Verify pool-host login still works (`login.salesforce.com`) — no `dpop_jkt` sent, flow unchanged.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPProofBuilder.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPProofBuilder.swift
@@ -32,6 +32,7 @@ public enum DPoPProofBuilderError: Int, Error {
     case jwkExportFailed = 1
     case serializationFailed = 2
     case signingFailed = 3
+    case thumbprintFailed = 4
 }
 
 /// Builds an RFC 9449 §4 DPoP proof JWS for a single token-endpoint request.
@@ -95,6 +96,36 @@ public final class DPoPProofBuilder: NSObject {
         }
         let signatureSegment = (signature as NSData).sfsdk_base64UrlString()
         return "\(signingInput).\(signatureSegment)"
+    }
+
+    /// RFC 7638 JWK thumbprint of a P-256 EC public key.
+    ///
+    /// Computes: `base64url(SHA-256(canonical_json({crv:"P-256", kty:"EC", x:<...>, y:<...>})))`
+    /// where `canonical_json` is UTF-8, no whitespace, keys in lexicographic order.
+    ///
+    /// Used at `/authorize` time to bind the authorization code to the same DPoP
+    /// key pair that will prove possession at `/token` (RFC 9449 §10 authorization
+    /// code binding via `dpop_jkt`).
+    ///
+    /// - Parameter publicKey: The P-256 `SecKey` whose thumbprint to compute.
+    ///   Same key that will later populate the DPoP proof header's `jwk` claim.
+    /// - Returns: 43-character base64url string (SHA-256 hash, base64url-encoded, no padding).
+    /// - Throws: `DPoPProofBuilderError.jwkExportFailed` if `Encryptor.jwkP256` fails;
+    ///   `DPoPProofBuilderError.thumbprintFailed` if canonicalization or hashing fails.
+    @objc public static func jwkThumbprint(publicKey: SecKey) throws -> String {
+        let jwk: [String: String]
+        do {
+            jwk = try Encryptor.jwkP256(from: publicKey)
+        } catch {
+            throw DPoPProofBuilderError.jwkExportFailed
+        }
+        // RFC 7638: canonical JSON with lexicographic key ordering, UTF-8, no whitespace.
+        guard let canonicalData = try? JSONSerialization.data(withJSONObject: jwk,
+                                                              options: [.sortedKeys, .withoutEscapingSlashes]),
+              let digest = (canonicalData as NSData).sfsdk_sha256() else {
+            throw DPoPProofBuilderError.thumbprintFailed
+        }
+        return (digest as NSData).sfsdk_base64UrlString()
     }
 
     // MARK: - Helpers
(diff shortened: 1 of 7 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPProofBuilder.swift
```diff
@@ -97,6 +98,36 @@ public final class DPoPProofBuilder: NSObject {
...
+    /// - Throws: `DPoPProofBuilderError.jwkExportFailed` if `Encryptor.jwkP256` fails;
+    ///   `DPoPProofBuilderError.thumbprintFailed` if canonicalization or hashing fails.
+    @objc public static func jwkThumbprint(publicKey: SecKey) throws -> String {
+        let jwk: [String: String]
+        do {
+            jwk = try Encryptor.jwkP256(from: publicKey)
+        } catch {
+            throw DPoPProofBuilderError.jwkExportFailed
+        }
+        // RFC 7638: canonical JSON with lexicographic key ordering, UTF-8, no whitespace.
```
wmathurin: Nit: RFC 7638 §3.2 requires the canonical JSON to contain **only** the required members for the key type — for P-256 that's exactly `{crv, kty, x, y}`, no optional fields like `kid`, `use`, or `key_ops`. `jwk` is passed directly to `JSONSerialization` here, so if `Encryptor.jwkP256` ever grows extra fields the thumbprint will silently diverge from what the server computes off the DPoP proof's `jwk` claim, breaking the authorize↔token binding. The fixture test would catch it, but worth a comment stating the invariant in-context so the dependency is visible to a future `jwkP256` author.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPProofBuilder.swift
```diff
@@ -97,6 +98,36 @@ public final class DPoPProofBuilder: NSObject {
...
+        guard let canonicalData = try? JSONSerialization.data(withJSONObject: jwk,
+                                                              options: [.sortedKeys, .withoutEscapingSlashes]),
+              let digest = (canonicalData as NSData).sfsdk_sha256() else {
```
wmathurin, line 130: Nit: `try?` on the `JSONSerialization` call above swallows any error before this throw, so if canonicalization ever fails the failure is silent. In practice a `[String: String]` dict won't cause a serialization error, but a log here (matching the style of `appendDPoPJktIfNeededTo:`) would make failures diagnosable: `SFSDKCoreLogger.w(DPoPProofBuilder.self, format: "DPoP jwkThumbprint: canonicalization or hash failed")`

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPProofBuilder.swift
sfdctaka: Addressed in f60327ee1 — added the RFC 7638 §3.2 minimality note tying the thumbprint's correctness to `jwkP256`'s output staying at exactly `{crv, kty, x, y}`, and a `SFSDKCoreLogger.w` line before the `thumbprintFailed` throw so canonicalization failures are diagnosable.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPProofBuilder.swift
sfdctaka: Addressed in f60327ee1 — added the RFC 7638 §3.2 minimality note tying the thumbprint's correctness to `jwkP256`'s output staying at exactly `{crv, kty, x, y}`, and a `SFSDKCoreLogger.w` line before the `thumbprintFailed` throw so canonicalization failures are diagnosable.

wmathurin (approved):

PR #4114, opened 2026-07-24
App Attestation support in SalesforceSDKCore

sfdctaka:
## Summary

Adds Apple App Attestation to the SalesforceSDKCore OAuth code exchange and refresh flows.

- Merging `app-attestation` feature branch into the `dev` branch. I branched `app-attestation` to `app-attestation-dev-merge` branch so I can first rebase it with `dev` locally. All commits in `app-attestation` have been previously reviewed and approved by the team. So there is no net new code in this PR.
- New `AppAttestation.swift` (production) and `AppAttestationTests.swift` (unit tests) under SalesforceSDKCore.
- `SFSDKAppAttestation.attestationIfEnabledFor:consumerKey:completionHandler:` produces the `attestation` value for the token endpoint; returns `nil` when the feature is disabled, unsupported on the device (simulator, older OS), or the domain is not attestation-enabled.
- Refresh (`SFOAuthSessionRefresher`) and code-exchange (`SFOAuthCoordinator`) paths were refactored to a small two-step form so the async attestation lookup can complete before the token request is dispatched. When attestation is off, both paths behave exactly as before (`attestation` param is skipped).
- Introduces `SFUserAccountManager.appAttestationEnabled` (public opt-in).
- Adds `kSFOAuthAttestation` constant and the `AA` per-user app-feature marker (registered per-user on the completed session, cleared from the global set on finalization).
- Switches to the web server flow when app attestation is configured.
- Consumer-facing wiring: `RestAPIExplorer/AppDelegate.swift` enables the flag.

## Test plan

- [x] `SalesforceSDKCore` scheme builds
- [x] Unit tests pass in Xcode (App Attestation, feature-flag markers, DPoP scene tests)
- [x] Existing DPoP UI automation continues to pass
- [x] Verify against an attestation-enabled Connected App (device only; simulator is expected to short-circuit to `nil`)

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -4,7 +4,7 @@ on:
   # pull_request_target is used so secrets are available to fork PRs.
   # Mitigated by per-job Member Check (see "Check Write Permission" / "Validate Write Permission").
   pull_request_target:  # zizmor: ignore[dangerous-triggers]
-    branches: [dev, master, app-attestation]
+    branches: [dev, master]
     paths-ignore:
       - '**/*.md'
       - 'LICENSE'
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -4,7 +4,7 @@ on:
   # pull_request_target is used so secrets are available to fork PRs.
   # Mitigated by per-job Member Check (see "Check Write Permission" / "Validate Write Permission").
   pull_request_target:  # zizmor: ignore[dangerous-triggers]
-    branches: [dev, master]
+    branches: [dev, master, app-attestation]
     paths-ignore:
       - '**/*.md'
       - 'LICENSE'
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -670,8 +670,11 @@
 }
 
 - (void)beginTokenEndpointFlow {
+    __weak typeof(self) weakSelf = self;
     [SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain consumerKey:self.credentials.clientId completionHandler:^(NSString * _Nullable attestation) {
-        [self executeTokenEndpointFlowWithAttestation:attestation];
+        __strong typeof(weakSelf) strongSelf = weakSelf;
+        if (!strongSelf) return;
+        [strongSelf executeTokenEndpointFlowWithAttestation:attestation];
     }];
 }
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -154,7 +154,7 @@
         self.authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeRefresh];
     } else if (self.useBrowserAuth) {
         self.authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeAdvancedBrowser];
-    } else if ([[SalesforceSDKManager sharedManager] useWebServerAuthentication]) {
+    } else if ([[SalesforceSDKManager sharedManager] useWebServerAuthentication] || [SFUserAccountManager sharedInstance].appAttestationEnabled) {
         self.authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeWebServer];
     } else {
         self.authInfo = [[SFOAuthInfo alloc] initWithAuthType:SFOAuthTypeUserAgent];
@@ -857,7 +857,7 @@
 - (NSString *)generateApprovalUrlString {
     return [self approvalURLForEndpoint:[self brandedAuthorizeURL]
                             credentials:self.credentials
-                          webServerFlow:(self.useBrowserAuth || [[SalesforceSDKManager sharedManager] useWebServerAuthentication])
+                          webServerFlow:(self.useBrowserAuth || [[SalesforceSDKManager sharedManager] useWebServerAuthentication] || [SFUserAccountManager sharedInstance].appAttestationEnabled)
                                protocol:nil
                                  domain:nil
                           codeChallenge:nil];
@@ -1013,7 +1013,7 @@
         // If a front door bridge URL override is present, use its code verifier to choose between user agent or web server authentication.
         if (self.frontdoorBridgeLoginOverride.frontdoorBridgeUrl // Check if an override is provided
             ? self.frontdoorBridgeLoginOverride.codeVerifier != nil // If yes, only proceed if it's a web server flow as indicated by a code verifier.
-            : (self.useBrowserAuth || [[SalesforceSDKManager sharedManager] useWebServerAuthentication]) // If there's no override use browser auth or the default SDK setting.
+            : (self.useBrowserAuth || [[SalesforceSDKManager sharedManager] useWebServerAuthentication] || [SFUserAccountManager sharedInstance].appAttestationEnabled) // If there's no override use browser auth or the default SDK setting.
             )
         {
             [self handleWebServerResponse:url]; // Web server flow/URLs with query string parameters.
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -670,6 +670,12 @@
 }
 
 - (void)beginTokenEndpointFlow {
+    [SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain consumerKey:self.credentials.clientId completionHandler:^(NSString * _Nullable attestation) {
+        [self executeTokenEndpointFlowWithAttestation:attestation];
+    }];
+}
+
+- (void)executeTokenEndpointFlowWithAttestation:(NSString * _Nullable)attestation {
     self.responseData = [NSMutableData dataWithLength:512];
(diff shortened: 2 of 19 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -670,6 +677,15 @@ - (void)updateCredentials:(NSDictionary *) params {
 }
 
 - (void)beginTokenEndpointFlow {
+    __weak typeof(self) weakSelf = self;
+    [SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain consumerKey:self.credentials.clientId completionHandler:^(NSString * _Nullable attestation) {
```
JohnsonEricAtSalesforce, line 681: **Requested change — nonnull-bridge crash hardening.**

`attestationIfEnabled(for domain: String?, consumerKey: String)` takes `consumerKey` as a **non-optional** Swift `String`, but `self.credentials.clientId` is declared `nullable` (`SFOAuthCredentials.h`). Because the attestation gating lives *inside* the Swift function, this call is crossed on **every** token-endpoint flow — code exchange and refresh — for **all** consumers, even when `appAttestationEnabled == NO`. If `clientId` is ever nil, the ObjC→Swift nonnull-`String` bridge traps (fatal error). The `NSAssert(clientId.length > 0)` in `authenticate` does not protect this in shipping builds — `ENABLE_NS_ASSERTIONS = NO` in Release strips it.

Reaching nil requires a malformed config (missing consumer key), so this is low-probability — but it is a new universal code path where the failure mode is a hard crash, and the fix is cheap: either make the Swift `consumerKey` a `String?` (the existing `!consumerKey.isEmpty` guard already handles the empty/nil-coalesced case), or guard `self.credentials.clientId.length > 0` at both call sites here and in `SFOAuthSessionRefresher.m`. Flagging as a requested change so it is addressed or consciously waived before merge rather than shipped as an implicit-crash path.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -2214,6 +2214,15 @@ - (void)finalizeAuthCompletion:(SFSDKAuthSession *)authSession {
...
     }
 
+    // AA: write per-user and clear the transient global flag
+    BOOL usedAppAttestation = [[SFSDKAppFeatureMarkers appFeatures] containsObject:kSFAppFeatureAppAttestation];
+    if (usedAppAttestation) {
+        [SFSDKAppFeatureMarkers registerAppFeature:kSFAppFeatureAppAttestation forUser:userAccount];
+    } else {
+        [SFSDKAppFeatureMarkers unregisterAppFeature:kSFAppFeatureAppAttestation forUser:userAccount];
+    }
+    [SFSDKAppFeatureMarkers unregisterAppFeature:kSFAppFeatureAppAttestation];
```
JohnsonEricAtSalesforce, line 2224: **Optional / non-blocking (telemetry-only).**

The `AA` marker is registered into the **global** feature set in `SFOAuthCoordinator.authenticate`, then read → written per-user → cleared from the global set here at finalization. The marker store is serial-queue-protected so there is no data race, but under genuinely concurrent multi-user logins (one user with attestation, one without) a second user's finalization could observe and clear the first user's transient global `AA` flag before the first finalizes — misattributing the marker. `DP` (DPoP) avoids this by registering per-user directly at the point of use. Since the impact is limited to feature-marker analytics, this is not a merge blocker — noting it as a small robustness improvement and for parity with the equivalent upcoming Android work.

.github/workflows/pr.yaml
```diff
@@ -4,7 +4,7 @@ on:
   # pull_request_target is used so secrets are available to fork PRs.
   # Mitigated by per-job Member Check (see "Check Write Permission" / "Validate Write Permission").
   pull_request_target:  # zizmor: ignore[dangerous-triggers]
-    branches: [dev, master]
+    branches: [dev, master, app-attestation]
```
JohnsonEricAtSalesforce: **Requested change — remove the feature branch from the CI trigger.**

`app-attestation` was added to the PR trigger branches to run CI on the feature branch. Once this merges to `dev` and the `app-attestation` branch is deleted, this entry becomes dead config that references a branch that no longer exists. Please remove `app-attestation` from this list before merge so the trigger only reflects the long-lived branches (`dev`, `master`).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthSessionRefresher.m
```diff
@@ -83,6 +84,12 @@ - (void)refreshSessionWithCompletion:(void (^)(SFOAuthCredentials *))completionB
         return;
     }
     
+    [SFSDKAppAttestation attestationIfEnabledFor:self.credentials.domain consumerKey:self.credentials.clientId completionHandler:^(NSString * _Nullable attestation) {
```
JohnsonEricAtSalesforce, line 88: **Optional / minor consistency.**

This attestation completion captures `self` strongly, while the sibling call site in `SFOAuthCoordinator.beginTokenEndpointFlow` uses the `weakSelf`/`strongSelf` + nil-check pattern. It is not a retain cycle or leak (the block is not stored on `self`; it is released after the one-shot completion), so behavior is fine — just flagging the inconsistency; consider mirroring the weak/strong pattern here for uniformity.

One more minor, unrelated point while in this area: in `AppAttestation.swift` the challenge→SHA256 step is duplicated between `attestationObject(for:)` and `generateAttestation(keyId:challenge:)` — a small DRY opportunity, not required.

JohnsonEricAtSalesforce (changes requested):
Reviewed as the Android-twin workstation — we developed App Attestation for MSDK Android, so this iOS PR was lensed against that implementation, the shared work epic, and the merged iOS feature PRs.

**Scope / carry-forward:** This PR is the approved `app-attestation` branch rebased onto current `dev`. A `git range-diff` confirms the feature commits map 1:1 to the previously-approved commits (the only deltas are context-line shifts from the newer base), and the net-new files (`AppAttestation.swift`, `AppAttestationTests.swift`) are byte-identical to the approved state. So there is no net-new unreviewed code — the prior approvals on the individual feature PRs carry forward.

**What I verified:**
- **OAuth/credential safety:** the `attestation` value is URL-encoded at both token-endpoint append sites (code exchange + refresh); no token/PII is logged (errors log only `localizedDescription` / `DCError.code` via `SFSDKCoreLogger`); challenge/registerkey are https + `requiresAuthentication = NO`; the App Attest keyId is persisted to the keychain only after successful server registration.
- **No behavior regression on the default (disabled) path:** when attestation is off, `attestationIfEnabled` returns `nil` and the token request is built with no `attestation` param — identical wire output to before. The refactor does introduce one async hop before the token request even when disabled, but the downstream token-endpoint call was already asynchronous, both callers of `beginTokenEndpointFlow` already `dispatch_async`, and the token-refresh path already re-dispatches its completion onto a serial queue — so there is no new ordering or thread-affinity assumption, and the completion is still invoked exactly once.
- **Public API:** `SFUserAccountManager.appAttestationEnabled` is additive and defaults to NO; appropriate for the current major-in-development cycle.
- **Tests:** `AppAttestationTests.swift` genuinely covers the testable seams (data transforms, all `shouldAttemptAttestation` gating branches, `existingKeyId` keychain match/mismatch/corrupt/partial paths). No vacuous assertions. The uncovered lines are the device-only `DCAppAttestService` / live-network paths that cannot execute on the simulator — which is exactly the `codecov/patch` red.

**CI (both red checks are non-code, non-blocking):**
- `ui-tests-pr (^18) / test-ui` — bootstrap crash ("Early unexpected exit … signal kill before starting test execution") on `LegacyLoginTests`, unrelated to this change. This job is currently red across unrelated branches (including a docs-only branch), so it is a harness flake, not a regression. Recommend re-running the UI job.
- `codecov/patch` — `AppAttestation.swift` is uncovered mainly on the `DCAppAttestService` / keychain / live-network device paths, which cannot execute on the simulator/CI (the feature short-circuits to `nil` there by design). The testable seam (`shouldAttemptAttestation`, `existingKeyId`) is covered.

**Requesting two changes** (both inline): (1) in `SFOAuthCoordinator.m`, a `nullable` `clientId` is passed into a non-optional Swift `String` parameter on a code path that now runs for every authentication — a small but real crash-hardening fix; and (2) in `.github/workflows/pr.yaml`, remove the `app-attestation` feature branch from the CI trigger list before merge. The remaining inline comments are optional cleanups.

Strong implementation overall and safe to merge once the two requested changes are addressed (or consciously waived).

sfdctaka:
@JohnsonEricAtSalesforce Updated after your feedback. Please take another look.

JohnsonEricAtSalesforce (approved):
Thanks @sfdctaka — I re-reviewed at `aa254763` and both requested changes are resolved:

- **Nonnull-bridge hardening (`SFOAuthCoordinator` / `AppAttestation.swift`):** `consumerKey` is now `String?` end-to-end, with `let consumerKey` guards in both `attestationIfEnabled` and `shouldAttemptAttestation`, so a nil `credentials.clientId` can no longer trap the ObjC→Swift bridge on the universal token-endpoint path. The added `test_givenNilConsumerKey…` case locks it in.
- **CI trigger (`pr.yaml`):** `app-attestation` removed; trigger list is back to `[dev, master]`.

You also picked up both optional cleanups — the weak/strong-self pattern in `SFOAuthSessionRefresher`, and the extracted `sha256(of:)` helper removing the duplicated hashing. Appreciated.

On the remaining non-blocking observation (the `AA` feature-marker being staged as a transient **global** flag and promoted per-user at finalization): I dug into it further and it's a real — but telemetry-only — accuracy issue that isn't specific to this PR (`WD` and `QR` use the same pattern). Rather than expand this PR's scope, I filed it as a standalone iOS bug covering the broader pattern: **W-23616382**. It does not block this merge.

CI: `codecov/…SalesforceSDKCore` red is the known device-only path (`DCAppAttestService` / live network) that can't execute on the simulator; `ui-tests-pr` is the pre-existing `LegacyLoginTests` harness flake. Everything else is green.

LGTM — approving.

PR #4116, opened 2026-07-29
Surface refresh token rotation state in dev info screen

sfdctaka:
## Summary

- Adds `lastTokenRotationDate` to `OAuthCredentials`, persisted via NSSecureCoding and preserved on copy.
- `SFOAuthSessionRefresher` stamps the timestamp when a refresh returns a new refresh token (same site that already registers the per-user RTR feature flag).
- `SalesforceSDKManager.getDevSupportInfos` emits a new `section:RTR` block with two rows: RTR active (per-user `RTR` feature flag) and the ISO-8601 UTC timestamp of the last observed rotation.

## Behavior

- Section is emitted unconditionally, so devs can distinguish "RTR not enabled" from "RTR enabled but no rotation observed yet".
- `Last Rotation` renders as `Never` when the credential has no timestamp.
- When no current user is signed in, RTR active shows `N/A` and last rotation shows `Never`.
- Timestamps use `en_US_POSIX` / UTC / `yyyy-MM-dd'T'HH:mm:ss'Z'` for a stable, locale-independent format.
- Additive keychain field: older archives decode `lastTokenRotationDate` as `nil`; archive version is unchanged.

Screenshot 2026-07-29 at 3 37 17 PM

## Test plan

- [x] `SalesforceOAuthUnitTests -testCredentialsCoding` — NSSecureCoding round-trip (populated and nil cases).
- [x] `SalesforceOAuthUnitTests -testCredentialsCopying` — `copyWithZone:` preserves the new field.
- [x] `SFOAuthSessionRefresherTests` — timestamp stamped on rotation, preserved when refresh token is unchanged.
- [x] `SalesforceSDKManagerTests` — four new tests covering the four dev-info states (active+timestamp, active+no-timestamp, inactive, no current user).
- [x] Full `SalesforceSDKCoreTests` suite: 904/904 passing.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
@@ -555,6 +555,19 @@ SFNativeLoginManagerInternal *nativeLogin;
     return [actions copy];
 }
 
+static NSString *SFSDKISO8601StringFromDate(NSDate *date) {
+    if (!date) return nil;
+    static NSDateFormatter *formatter = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        formatter = [[NSDateFormatter alloc] init];
+        formatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
+        formatter.timeZone = [NSTimeZone timeZoneWithName:@"UTC"];
+        formatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss'Z'";
+    });
+    return [formatter stringFromDate:date];
+}
+
 - (NSArray<NSString *>*) getDevSupportInfos
 {
     SFUserAccountManager* userAccountManager = [SFUserAccountManager sharedInstance];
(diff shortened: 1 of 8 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -640,7 +653,31 @@ - (NSString *)devInfoTitleString
...
+    [devInfos addObject:@"section:RTR"];
+    {
+        SFUserAccount *rtrUser = [SFUserAccountManager sharedInstance].currentUser;
+        NSString *rtrActive;
+        if (!rtrUser) {
+            rtrActive = @"N/A";
+        } else {
+            NSSet<NSString *> *features = [SFSDKAppFeatureMarkers appFeaturesForUser:rtrUser];
+            rtrActive = [features containsObject:kSFAppFeatureRTR] ? @"YES" : @"NO";
+        }
```
brandonpage, line 667: It looks like we don't actually know if the user is subject to RTR until the first token refresh.  Is that correct?

I am not pointing this out because I think it could be improved, just making sure I understand correctly.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
```diff
@@ -190,11 +190,20 @@ NS_SWIFT_NAME(OAuthCredentials)
...
 @property (nonatomic, readonly, nullable) NSDate *issuedAt;
 
+/** The timestamp of the most recent refresh-token rotation observed for this credential.
+
+ Set by `SFOAuthSessionRefresher` when a token refresh returns a `refresh_token`
+ that differs from the previously-stored one. `nil` if no rotation has been observed
+ for this credential (or if this credential predates the field's introduction — older
+ archives decode this field as `nil`).
+ */
+@property (nonatomic, readonly, nullable) NSDate *lastTokenRotationDate;
```
brandonpage, line 205: So if we just updated `issuedAt` we wouldn't actually know if the token came from a rotation or not.  👍

brandonpage (approved):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
sfdctaka: Yes, correct. `kSFAppFeatureRTR` is registered inside `SFOAuthSessionRefresher` only when a refresh returns a changed refresh_token, so pre-first-rotation the row reads NO even if the org has RTR enabled. The Last Rotation timestamp is the more precise signal ("did we observe a rotation, and when?"); RTR active is a stickier per-user roll-up of the same fact. If we ever want a true policy-level indicator we'd need a new server signal(that's worth a separate work ticket).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCredentials.h
sfdctaka: Exactly. `issuedAt` gets stamped on every access-token refresh, so it can't distinguish "same refresh_token, new access_token" from a true rotation. `lastTokenRotationDate` only advances when the refresh_token itself changes.

PR #4119, opened 2026-08-04
[iOS] Add ftr_ B/L telemetry markers for browser-login reason and login server type (W-23701450)

wmathurin:
## Summary

Adds 9 new per-user \`ftr_\` telemetry markers to the SDK user-agent string:

- **B1–B4**: explain *why* browser login (Advanced Auth) was used
- **L1–L5**: record *which login server type* was used on every non-refresh login

Follow-up to W-23126676 (Force Advanced Authentication flag). Spec and plan live in the Workspace repo under \`specs/W-23701450-W-23240736-browser-login-telemetry-markers/\`. Companion Android PR: forcedotcom/SalesforceMobileSDK-Android.

## B-markers — "why browser login was used"

Registered per-user alongside \`BW\`. Exactly one B-marker per user. Priority: **B3 > B2 > B4 > B1**.

| Code | Reason |
|------|--------|
| \`B1\` | Server auth-config opted in (\`UseiOSNativeBrowserForAuthentication\`) |
| \`B2\` | MDM policy required browser auth (\`SFManagedPreferences.requireCertificateAuthentication\`) |
| \`B3\` | "Login for Admin" flow |
| \`B4\` | \`forceAdvancedAuthentication\` SDK flag (W-23126676) |

## L-markers — "which login server type"

Registered per-user on every non-refresh login (browser or WebView). Exactly one L-marker per user.

| Code | Login server | Detection |
|------|-------------|-----------|
| \`L1\` | Production pool | host matches \`login.salesforce.com\` or \`login.*.salesforce.com\` (no \`.my.\`) — covers internal envs like \`login.test1.pc-rnd.salesforce.com\` |
| \`L2\` | Sandbox | exact match \`test.salesforce.com\` |
| \`L3\` | Welcome Discovery | WD global flag set at auth-completion time (checked before cleared) |
| \`L4\` | My Domain | host contains \`.my.\` and ends with \`.salesforce.com\` — covers \`*.my.salesforce.com\` and internal \`*.my.*.salesforce.com\` |
| \`L5\` | Other | fallthrough |

## Implementation

- **\`SFSDKAppFeatureMarkers.h/.m\`**: 9 new extern string constants
- **\`SFSDKAuthConfigUtil.h/.m\`**: new \`+isProductionLoginHost:\` and \`+isMyDomainHost:\` helpers; \`+isPoolLoginHost:\` updated to delegate to \`+isProductionLoginHost:\`
[...]

PR #4121, opened 2026-08-05
feat(W-23721537): speed up AuthFlowTester UI tests via launch-arg driven SDK reset

wmathurin:
## Summary

- Adds `SalesforceSDKManager.resetForUITesting()` (`#if DEBUG`) — logs out all users (including async server refresh-token revocation), clears per-user in-memory feature flags, resets the selected login host to `login.salesforce.com`, removes persisted custom login servers, and restores all auth flags to their `init` defaults.
- `AppDelegate` calls it when `--resetSDKForUITesting` is in launch arguments.
- `BaseAuthFlowTester.launch()` passes `--resetSDKForUITesting` so every test gets a clean slate in-process at startup, replacing 10–20 s of UI-driven tearDown/launch cleanup per test.
- `tearDown` is now a no-op (no UI logout, no DPoP toggle navigation).
- `restart()` creates a fresh `XCUIApplication` without the reset arg so session-persistence tests continue to work.

## SDK fix

**`SFUserAccountManager.m` — preserve BW flag through refresh token migration**

`finalizeAuthCompletion:` had an unconditional `else` that unregistered the BW (`kSFAppFeatureSafariBrowserForLogin`) per-user flag for any auth type other than `SFOAuthTypeAdvancedBrowser`. Refresh token migration completes with `SFOAuthTypeRefreshTokenMigration` — a silent token exchange that does not change how the user originally authenticated — so BW was being cleared even when the user had logged in via the external browser. Added an explicit no-op guard for `SFOAuthTypeRefreshTokenMigration` that leaves the existing per-user flag intact. Normal token refreshes and other non-browser auth types continue to clear BW as designed.

## Test fixes (found while running the suite after the main change)

Several pre-existing assertion bugs surfaced because the launch-arg reset makes every test start from a known clean state — previously these were masked by state leaking between tests.

| Fix | Root cause |
|---|---|
| `tokenFormat` in credentials JSON was `"Opaque"` instead of `""` | `UserCredentialsView.tokenFormat` applied a display substitution (`""` → `"Opaque"`) that was flowing into the exported JSON. Split into `tokenFormatRaw` (export) and `tokenFormat` (display). |
| `restartAndValidateUser` wiped the session on restart | `app` still carried `--resetSDKForUITesting` in `launchArguments`; fixed by creating a fresh `XCUIApplication` in `restart()`. |
| 2 s unnecessary wait per test when adding a login server | `hasHost()` used a 10 s timeout to probe for a custom host that `resetForUITesting` always removes. Reduced to 2 s (host list is synchronously in-memory). |
| `forceAdvancedAuthentication: nil` was overriding SDK default to off | `configureLoginOptions` was called with `forceAdvancedAuthentication ?? false`, actively writing `false` into the Login Options JSON and switching the app to WebView mode even when the caller meant "use the SDK default (browser on)". Fixed by making `forceAdvancedAuthentication` a non-nullable `Bool = true` across all helpers (matching Android's `AuthFlowTest`), removing the optional path entirely. |
| MU flag asserted absent while two users were still logged in | `switchToUserAndValidate` calls in `testFirstStatic/Dynamic_DifferentApps`, `testBothDynamic_DifferentApps`, and `testMigrateOneUserOnly` (via `migrateAndValidate`) were missing `isMultiUser: true`. Added `isMultiUser` param to `migrateAndValidate` and fixed all call sites. |
| RT flag asserted absent after RTR cycle survived restart | `validateUser` had no `isRtr` param, always calling `validateUserAgent` with `isRtr: false`. After `launchLoginAndValidate` runs an RTR cycle the RT flag persists through the non-resetting restart. Added `isRtr` to `validateUser` and `validateUserAgent`. |
| RT flag present at test start (e.g. `testECAOpaqueRtr_Hybrid`) | `SFSDKPerUserFeatureMarkersMap` was never cleared by `logoutAllUsers`. RT flag from a previous test's RTR cycle survived into the next test's initial UA check. Added `#if DEBUG resetPerUserFeaturesForUITesting` to `SFSDKAppFeatureMarkers` and called it from `resetForUITesting`. |
| `invalid_client_id` error on standard server in ForceAdvancedAuth disabled test | `setForceAdvancedAuthentication` was importing a My-Domain ECA consumer key (`ecaOpaque`) when testing against `login.salesforce.com`, triggering `invalid_client_id`. Made `staticAppConfigName` optional (defaults to `nil`); the app's default `bootconfig.plist` key is valid on the standard server. |

## Refactors

| Refactor | Detail |
|---|---|
| `forceAdvancedAuthentication: Bool? = nil` → `Bool = true` | Matches Android. `nil` was ambiguous and masked the `?? false` bug. Non-nullable default makes the intent explicit: browser on unless the test explicitly opts out with `false`. |
| `restartAndValidateUser` no longer has `isRtr` param | Restart-and-validate is session-persistence only. RTR check after restart is a separate concern; tests call `assertRevokeAndRefreshWorks(isRtr: true)` explicitly, consistent with DPoP tests and the skipped `testECAJwtRtr_Hybrid_WithRestart`. |
| Extract `resetAuthFlags` helper in `SalesforceSDKManager` | Both `-init` and `resetForUITesting` set the same auth flags. Extracted into a private instance method so the values are defined once and can't diverge. |

## New tests

- `LegacyLoginTests`: added `testCAOpaque_{Default,Subset,All}Scopes_WebServerFlow_InAppWebView` — mirrors the existing browser-path web server flow tests with `forceAdvancedAuthentication: false` to cover the in-app WebView path under the same scope variations.

## Files changed

| File | Change |
|---|---|
| `SalesforceSDKCore/Classes/Common/SalesforceSDKManager.h` | Declare `+ (void)resetForUITesting` under `#if DEBUG` |
| `SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m` | Implement `resetForUITesting`; extract `resetAuthFlags`; call `resetPerUserFeaturesForUITesting`; import `SFSDKLoginHostStorage.h` |
| `SalesforceSDKCore/Classes/Common/SFSDKAppFeatureMarkers.h` | Declare `+ (void)resetPerUserFeaturesForUITesting` under `#if DEBUG` |
| `SalesforceSDKCore/Classes/Common/SFSDKAppFeatureMarkers.m` | Implement `resetPerUserFeaturesForUITesting` — clears `SFSDKPerUserFeatureMarkersMap` |
| `SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m` | Preserve BW per-user flag for `SFOAuthTypeRefreshTokenMigration` in `finalizeAuthCompletion:` |
| `AuthFlowTester/Classes/AppDelegate.swift` | Call `resetForUITesting()` when launch arg present |
| `AuthFlowTester/Views/UserCredentialsView.swift` | Split `tokenFormat` into raw (export) and display variants |
| `AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift` | Add launch arg; simplify `tearDown`; fix `restart()`; make `forceAdvancedAuthentication` non-nullable (`Bool = true`); remove `isRtr` from `restartAndValidateUser`; add `isRtr`/`isMultiUser` params where needed; make `setForceAdvancedAuthentication` `staticAppConfigName` optional; restore `migrateAndValidate` default to `true` |
| `AuthFlowTesterUITests/PageObjects/LoginPageObject.swift` | Reduce `hasHost` timeout from 10 s to 2 s |
| `AuthFlowTesterUITests/Tests/ForceAdvancedAuthTests.swift` | Drop redundant `staticAppConfigName` from disabled-auth test |
| `AuthFlowTesterUITests/Tests/MultiUserLoginTests.swift` | Pass `isMultiUser: true` to `switchToUserAndValidate` while two users are logged in |
| `AuthFlowTesterUITests/Tests/RefreshTokenMigrationTests.swift` | Pass `isMultiUser: true` to `migrateAndValidate` / `switchToUserAndValidate`; pass `forceAdvancedAuthentication: false` for user-agent-flow migration tests |
| `AuthFlowTesterUITests/Tests/LegacyLoginTests.swift` | Add three `_InAppWebView` variants with `forceAdvancedAuthentication: false` |

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
@@ -229,6 +229,15 @@ SFNativeLoginManagerInternal *nativeLogin;
     [SalesforceSDKManager sharedManager];
 }
 
+- (void)resetAuthFlags {
+    self.useEphemeralSessionForAdvancedAuth = YES;
+    self.useWebServerAuthentication = YES;
+    self.useHybridAuthentication = YES;
+    self.useDPoP = NO;
+    self.sdk_forceAdvancedAuthentication = YES;
+    self.blockSalesforceIntegrationUser = NO;
+}
+
 #if DEBUG
 + (void)resetForUITesting {
     // 1. Log out all users — clears on-disk account data, DPoP keychain keys, in-memory maps.
@@ -249,14 +258,9 @@ SFNativeLoginManagerInternal *nativeLogin;
     [storage removeAllLoginHosts];
     [storage save];
 
-    // 5. Reset all auth flags to the values set in -init.
+    // 5. Reset all auth flags to their -init defaults.
     SalesforceSDKManager *mgr = [SalesforceSDKManager sharedManager];
-    mgr.useEphemeralSessionForAdvancedAuth = YES;
-    mgr.useWebServerAuthentication = YES;
-    mgr.useHybridAuthentication = YES;
-    mgr.useDPoP = NO;
-    mgr.sdk_forceAdvancedAuthentication = YES;
-    mgr.blockSalesforceIntegrationUser = NO;
+    [mgr resetAuthFlags];
     mgr.simulatedDomainDiscoveryResult = nil;
 }
 #endif
@@ -361,12 +365,7 @@ SFNativeLoginManagerInternal *nativeLogin;
         [self computeWebViewUserAgent]; // web view user agent is computed asynchronously so very first call to self.userAgentString(...) will be missing it
         self.userAgentString = [self defaultUserAgentString];
         self.URLCacheType = kSFURLCacheTypeEncrypted;
-        self.useEphemeralSessionForAdvancedAuth = YES;
-        self.useWebServerAuthentication = YES;
-        self.blockSalesforceIntegrationUser = NO;
-        self.useHybridAuthentication = YES;
-        self.useDPoP = NO;
-        self.sdk_forceAdvancedAuthentication = YES;
+        [self resetAuthFlags];
         [self setupServiceConfiguration];
         _snapshotViewControllers = [SFSDKSafeMutableDictionary new];
         _nativeLoginViewControllers = [SFSDKSafeMutableDictionary new];
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
@@ -43,6 +43,7 @@
 #import "SFSDKSalesforceSDKUpgradeManager.h"
 #import <SalesforceSDKCommon/NSUserDefaults+SFAdditions.h>
 #import "SFSDKWindowManager+Internal.h"
+#import "SFSDKLoginHostStorage.h"
 
 // Error constants
 NSString * const kSalesforceSDKManagerErrorDomain     = @"com.salesforce.sdkmanager.error";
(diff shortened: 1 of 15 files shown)


sfdctaka:
This file is not included in the PR.
image

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m
```diff
@@ -228,6 +229,38 @@ + (void)initializeSDKWithClass:(Class)className {
...
+
+    // 5. Reset all auth flags to the values set in -init.
+    SalesforceSDKManager *mgr = [SalesforceSDKManager sharedManager];
+    mgr.useEphemeralSessionForAdvancedAuth = YES;
+    mgr.useWebServerAuthentication = YES;
+    mgr.useHybridAuthentication = YES;
+    mgr.useDPoP = NO;
+    mgr.sdk_forceAdvancedAuthentication = YES;
+    mgr.blockSalesforceIntegrationUser = NO;
+    mgr.simulatedDomainDiscoveryResult = nil;
```
sfdctaka, line 264: nit: `SalesforceSDKManager` can have a helper method so both `init` in there and the code above can keep sharing the same values and possibly in the future prevent the values from diverging? 

Image

wmathurin: Done here: https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4121/commits/a8ceed745b82c43be0888cc59818d4a5f60f2198

wmathurin:
> This file is not included in the PR. image

Good catch. The assert was there before this PR. Fixed PR description.

sfdctaka (approved):
LGTM!

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -2182,6 +2182,9 @@ - (void)finalizeAuthCompletion:(SFSDKAuthSession *)authSession {
     SFOAuthType completedAuthType = authSession.oauthCoordinator.authInfo.authType;
     if (completedAuthType == SFOAuthTypeAdvancedBrowser) {
         [SFSDKAppFeatureMarkers registerAppFeature:kSFAppFeatureSafariBrowserForLogin forUser:userAccount];
+    } else if (completedAuthType == SFOAuthTypeRefreshTokenMigration) {
+        // Migration exchanges the token but does not change how the user originally
+        // authenticated. Preserve the existing per-user BW flag rather than clearing it.
```
brandonpage, line 2187: Good catch.

brandonpage (approved):

wmathurin:
A lot of tests fixed. Will further validate in the other [PR](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4119).

PR #4124, opened 2026-08-06
fix(auth): preserve per-user AA flag on token refresh (W-23616382)

wmathurin:
## Summary

- Moves the `AA` (App Attestation) per-user promotion block inside the `completedAuthType != SFOAuthTypeRefresh` guard in `finalizeAuthCompletion:` in `SFUserAccountManager.m`
- Silent token refreshes previously ran the promotion block unconditionally, calling `unregisterAppFeature:forUser:` when the transient global was absent — silently clearing the persisted `AA` flag on every refresh
- The global clear (`unregisterAppFeature:kSFAppFeatureAppAttestation`) fires unconditionally outside the guard, as intended
- `WD` and `QR` were already correctly placed inside the guard; this brings `AA` into alignment with them

## Test plan

- [ ] `SFSDKAppFeatureMarkersTests` — all 22 tests pass (verified locally)
- [ ] `test_givenAAFlagPersistedOnUser_whenTokenRefreshCompletes_thenAAFlagPreserved` — new regression test
- [ ] `test_givenAARegisteredGlobally_whenNonRefreshLoginCompletes_thenAAPromotedToUserAndGlobalCleared` — new regression test
- [ ] Verify AA appears correctly in user agent after app restart following attestation login
- [ ] Verify AA persists across silent token refreshes (remains in user agent)

GUS: W-23616382

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
@@ -2284,22 +2284,22 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
         } else {
             [SFSDKAppFeatureMarkers unregisterAppFeature:kSFAppFeatureQrCodeLogin forUser:userAccount];
         }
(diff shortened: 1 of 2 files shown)


sfdctaka (approved):
LGTM!

PR #4125, opened 2026-08-07
feat(telemetry): add A1-A5/TM/JT/OT/BN per-user feature flag markers (W-23762052)

wmathurin:
## Summary

- Add 9 new per-user feature flag markers: **A1–A5** (auth-flow type), **TM** (token migration), **JT/OT** (token format), **BN** (beacon child app)
- A-markers use the existing transient-global → per-user two-phase pattern (same as BW, L-markers): registered globally during auth, promoted to per-user in `finalizeAuthCompletion:` and global cleared
- **BW and A-markers are orthogonal**: BW captures *that* a browser was used; the A-marker captures the OAuth grant × hybrid dimension independently. `SFOAuthTypeAdvancedBrowser` always uses the auth-code (web-server) grant so it always maps to A1 (non-hybrid) or A2 (hybrid)
- TM, JT/OT, BN are per-user only — no global phase, like DP
- **Production fix**: `finalizeAuthCompletion:` now (1) preserves TM across `SFOAuthTypeRefresh` (only clears on full re-login) and (2) unconditionally clears TM from the global `SFSDKAppFeatureMarkersSet` after every auth completion, preventing it from bleeding into other users' UA strings
- Update `BaseAuthFlowTester.swift` `validateUserAgent` to assert all new flags; add multi-user flag-isolation and token-migration isolation UI tests

## GUS story
W-23762052: https://gus.my.salesforce.com/lightning/r/ADM_Work__c/a07EE00002grYFgYAM/view

## Key behavioral rules
| Flag | When written | Global phase? | Refresh path touches it? | Cleared by full re-login? |
|---|---|---|---|---|
| A1–A5 | Every fresh login | Yes (transient) | No | Yes — re-promoted from global on next login |
| TM | Migration only | No | No | Yes — explicitly unregistered |
| JT / OT | Full login and migration | No | No | Yes — re-written with new token format |
| BN | Full login and migration | No | No | Yes — re-written from credentials |

## Test plan
- [x] SFSDKAppFeatureMarkersTests: all new unit tests pass
- [x] AuthFlowTester UI tests: all tests pass locally (verified 2026-08-09)


sfdctaka (approved):
LGTM!

PR #4126, opened 2026-08-07
fix(dpop): gate proof attachment on per-credential state

sfdctaka:
## Summary

DPoP proof attachment was gated by the process-wide `SalesforceManager.usesDPoP` flag. When the flag was flipped off while a DPoP-bound credential was still in use, subsequent refresh calls and REST/identity requests silently dropped the proof, producing a 401 loop.

Gate on `credentials.tokenType == "DPoP"` **or** `DPoPKeyStore.hasKeyPair(scope)` instead. The global flag continues to govern *new* logins only; once a credential is bound, every request for it carries a proof regardless of flag state (belt-and-suspenders).

## Changes

**SDK**
- `DPoPKeyStore` — new `hasKeyPair(forScope:)` / `hasKeyPair(forCredentials:)` side-effect-free presence check under a reader lock (never generates a key on miss).
- `DPoPRequestDecorator` — new `shouldAttachDPoP(scope:tokenType:)` predicate; new 4-arg `decorate` overload threads `tokenType` through; existing 3-arg overload preserved for backwards compatibility.
- `SFSDKOAuthTokenEndpointRequest` — new `tokenType` property, carried through unchanged from `credentials.tokenType`.
- `SFOAuthCoordinator` / `SFOAuthSessionRefresher` — populate `tokenType` at refresh construction so the predicate honors bound credentials even with the global flag off.

**Sample app / UI test hooks (AuthFlowTester)**
- New `--disableDPoPAtStart` and `--invalidateCurrentUserAccessTokenAtStart` launch args to drive the flag-flip + refresh scenario from XCUITest.

**Tests**
- `SFSDKDPoPTests` — added `hasKeyPair` presence-check test; belt-and-suspenders predicate matrix (empty scope, nil / Bearer / DPoP × key-present / key-absent); flag-off + tokenType=DPoP round-trip test; Bearer-credential never-attach test.
- `SFOAuthSessionRefresherTests` — refresh path attaches proof when `usesDPoP=false` but credential is DPoP-bound.
- `MultiUserLoginTests` — UI test locking in flag-off-then-refresh end-to-end.

## Backwards compatibility

- Public API additions only. Existing `SFSDKDPoPRequestDecorator.decorate(_:scope:)` / `decorate(_:scope:accessToken:)` are preserved and delegate to the new 4-arg overload.
- The global `SalesforceManager.usesDPoP` flag continues to work as before for new logins.

## Test plan

- [x] `SFSDKDPoPTests` — 44 tests pass (iPhone 16 Pro sim, iOS 18.6).
- [x] `SFOAuthSessionRefresherTests` — pass.
- [x] `SalesforceSDKCore` full suite — 194/195 pass; the one pre-existing failure (`testPublicApiCalls`) is unrelated.
- [x] `MultiUserLoginTests.testDPoPProofAttachedOnRefreshAfterFlagOff` — GREEN on iPhone 16 Pro sim.
- [ ] Full-suite CI regression on merge.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPKeyStore.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPKeyStore.swift
@@ -89,11 +89,17 @@ public final class DPoPKeyStore: NSObject {
     /// Returns `true` iff a private key is already present in the Keychain for `scope`.
     /// Side-effect-free — never generates a key on miss. Used to gate DPoP proof attachment
     /// on the presence of previously-minted key material for the credential.
-    func hasKeyPair(forScope scope: String) -> Bool {
+    public func hasKeyPair(forScope scope: String) -> Bool {
         guard !scope.isEmpty else { return false }
         let name = Self.keyName(for: scope)
+        // Non-barrier read: safe to run concurrently with other reads.
+        // All mutating methods in this class use .barrier for exclusive access.
         return queue.sync {
             guard let privateTag = try? KeyGenerator.keyTag(name: name, prefix: KeyGenerator.ecPrivateKeyTagPrefix) else {
+                // A throw here isn't necessarily "key absent" — it can be a Keychain access
+                // denied, entitlement, or device-locked error. Returning `false` means
+                // "proceed without a proof," so surface the failure for diagnostics.
+                SFSDKCoreLogger.w(Self.self, message: "DPoPKeyStore.hasKeyPair: Keychain lookup failed for scope \(scope)")
                 return false
             }
             return (try? KeyGenerator.ecKey(tag: privateTag)) != nil
@@ -101,7 +107,7 @@ public final class DPoPKeyStore: NSObject {
     }
 
     /// Convenience: presence check scoped on `credentials.identifier`.
-    func hasKeyPair(forCredentials credentials: OAuthCredentials) -> Bool {
+    public func hasKeyPair(forCredentials credentials: OAuthCredentials) -> Bool {
         return hasKeyPair(forScope: credentials.identifier)
     }
 
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPKeyStore.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPKeyStore.swift
@@ -86,6 +86,25 @@ public final class DPoPKeyStore: NSObject {
         return try keyPair(forScope: credentials.identifier)
     }
 
+    /// Returns `true` iff a private key is already present in the Keychain for `scope`.
+    /// Side-effect-free — never generates a key on miss. Used to gate DPoP proof attachment
+    /// on the presence of previously-minted key material for the credential.
+    func hasKeyPair(forScope scope: String) -> Bool {
+        guard !scope.isEmpty else { return false }
+        let name = Self.keyName(for: scope)
+        return queue.sync {
+            guard let privateTag = try? KeyGenerator.keyTag(name: name, prefix: KeyGenerator.ecPrivateKeyTagPrefix) else {
+                return false
+            }
+            return (try? KeyGenerator.ecKey(tag: privateTag)) != nil
+        }
+    }
+
+    /// Convenience: presence check scoped on `credentials.identifier`.
+    func hasKeyPair(forCredentials credentials: OAuthCredentials) -> Bool {
+        return hasKeyPair(forScope: credentials.identifier)
+    }
+
     /// Removes any existing keypair for the scope. Idempotent.
     @objc(deleteForScope:)
     public func delete(forScope scope: String) {
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPRequestDecorator.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPRequestDecorator.swift
@@ -37,14 +37,19 @@ public final class DPoPRequestDecorator: NSObject {
     /// `token_type: "DPoP"` (RFC 6749 §5.1 / RFC 9449 §6.1).
     @objc public static let dpopTokenType = "DPoP"
 
-    /// Belt-and-suspenders gate: attach a DPoP proof for `scope` when *either*
-    /// signal says the credential is DPoP-bound —
+    /// Gate deciding whether to attach a DPoP proof for `scope`.
     ///
-    /// 1. `tokenType == "DPoP"` (per RFC 6749 §5.1 / RFC 9449 §6.1), or
-    /// 2. a DPoP keypair for this scope is already persisted in the Keychain
-    ///    (covers the pre-tokenType window during initial auth-code exchange,
-    ///    and any refresh where the persisted `tokenType` hasn't yet been rehydrated
-    ///    onto the request struct).
+    /// An explicit `tokenType` is authoritative:
+    /// - `tokenType == "DPoP"` (case-insensitive, per RFC 6749 §5.1 / RFC 9449 §6.1) → attach.
+    /// - any other known type (e.g. `"Bearer"`) → do **not** attach, and short-circuit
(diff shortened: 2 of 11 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPRequestDecorator.swift
```diff
@@ -37,27 +37,66 @@ public final class DPoPRequestDecorator: NSObject {
...
+    ///    and any refresh where the persisted `tokenType` hasn't yet been rehydrated
+    ///    onto the request struct).
+    ///
+    /// Empty scope always returns `false` — an empty `SFOAuthCredentials.identifier`
+    /// would collide with unrelated accounts' key material.
+    ///
+    /// Note: this gate is deliberately independent of `SalesforceManager.shared.usesDPoP`.
+    /// The global flag governs *new* logins (whether to request DPoP-bound tokens); once a
+    /// credential is bound, every request for it carries a proof regardless of the flag.
+    static func shouldAttachDPoP(scope: String, tokenType: String?) -> Bool {
```
wmathurin, line 60: **[MAJOR] False-positive DPoP on Bearer re-auth with stale key pair**

`shouldAttachDPoP` returns `true` whenever key material is present in the Keychain, regardless of `tokenType`. If a user re-authenticates as Bearer with the same scope/credentials identifier and the old DPoP key pair isn't cleaned up, this predicate returns `true` for a Bearer credential — attaching a DPoP proof the server doesn't expect.

The spec confirms this path exists: the `DP` per-user flag is "unregistered if user migrates from DPoP to non-DPoP session", which means the Bearer re-auth scenario is a supported flow.

Suggested fix — short-circuit on explicit non-DPoP `tokenType` before the Keychain check:

```swift
private func shouldAttachDPoP(scope: String, tokenType: String?) -> Bool {
    if let tokenType {
        return tokenType.caseInsensitiveCompare("DPoP") == .orderedSame
    }
    // tokenType is nil — transition window between /authorize and /token; fall back to key material
    return DPoPKeyStore.shared.hasKeyPair(forScope: scope)
}
```

This also means Bearer credentials pay zero Keychain overhead on every REST call (no I/O when `tokenType` is a known non-DPoP value).

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPKeyStore.swift
```diff
@@ -86,6 +86,25 @@ public final class DPoPKeyStore: NSObject {
...
+    /// Side-effect-free — never generates a key on miss. Used to gate DPoP proof attachment
+    /// on the presence of previously-minted key material for the credential.
+    func hasKeyPair(forScope scope: String) -> Bool {
+        guard !scope.isEmpty else { return false }
+        let name = Self.keyName(for: scope)
+        return queue.sync {
+            guard let privateTag = try? KeyGenerator.keyTag(name: name, prefix: KeyGenerator.ecPrivateKeyTagPrefix) else {
+                return false
+            }
+            return (try? KeyGenerator.ecKey(tag: privateTag)) != nil
```
wmathurin, line 105: **[MINOR] `queue.sync` differs from the barrier pattern used everywhere else in this class**

All other `DPoPKeyStore` methods use `queue.sync(flags: .barrier)`, making the concurrent queue behave serially. This method uses plain `queue.sync` (a non-barrier concurrent read), which is semantically correct for a read-only operation — but it silently changes the queue's concurrency model: concurrent reads can now interleave with in-flight barrier writes.

This is safe, but the deviation from the established pattern will surprise a future maintainer. Please add a comment documenting the intent:

```swift
// Non-barrier read: safe to run concurrently with other reads.
// All mutating methods in this class use .barrier for exclusive access.
return queue.sync {
    ...
}
```

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPKeyStore.swift
```diff
@@ -86,6 +86,25 @@ public final class DPoPKeyStore: NSObject {
         return try keyPair(forScope: credentials.identifier)
     }
 
+    /// Returns `true` iff a private key is already present in the Keychain for `scope`.
+    /// Side-effect-free — never generates a key on miss. Used to gate DPoP proof attachment
+    /// on the presence of previously-minted key material for the credential.
+    func hasKeyPair(forScope scope: String) -> Bool {
```
wmathurin: **[MINOR] `internal` access is inconsistent with the rest of `DPoPKeyStore`'s `public` API surface**

Every other substantive method on `DPoPKeyStore` (`keyPair(forScope:)`, `keyPair(forCredentials:)`, `delete(forScope:)`, etc.) is `public`. These two `hasKeyPair` methods are `internal`.

If this is intentional (implementation detail only), please add a doc comment saying so — the inconsistency is otherwise surprising. If it's an oversight, consider making them `public` so external consumers can inspect whether a credential is DPoP-bound without mutating state.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/DPoP/DPoPKeyStore.swift
```diff
@@ -86,6 +86,25 @@ public final class DPoPKeyStore: NSObject {
...
+        guard !scope.isEmpty else { return false }
+        let name = Self.keyName(for: scope)
+        return queue.sync {
+            guard let privateTag = try? KeyGenerator.keyTag(name: name, prefix: KeyGenerator.ecPrivateKeyTagPrefix) else {
+                return false
```
wmathurin, line 103: **[NIT] Silent swallow masks real Keychain errors**

If `KeyGenerator.keyTag(name:prefix:)` throws for a reason other than "key not present" (e.g., Keychain access denied, entitlement error, device lock), this returns `false` silently. `false` here means "no DPoP — proceed without proof," which is the same 401-loop failure this PR is fixing, but now with no diagnostic signal.

`keyPair(forScope:)` already logs errors. Please add at minimum a `warn`-level log here for parity:

```swift
} catch {
    SFSDKCoreLogger.w(Self.self, format: "DPoPKeyStore.hasKeyPair: Keychain lookup failed for scope \(scope): \(error)")
    return false
}
```

libs/SalesforceSDKCore/SalesforceSDKCoreTests/SFSDKDPoPTests.swift
```diff
@@ -259,6 +259,50 @@ class SFSDKDPoPTests: XCTestCase {
...
+        XCTAssertFalse(DPoPRequestDecorator.shouldAttachDPoP(scope: testScope, tokenType: "Bearer"))
+
+        // Token type "DPoP" alone (before key pair minted) short-circuits to true.
+        XCTAssertTrue(DPoPRequestDecorator.shouldAttachDPoP(scope: testScope, tokenType: "DPoP"))
+
+        // Mint the key pair. Now key material is enough on its own, even without tokenType.
+        _ = try DPoPKeyStore.shared.keyPair(forScope: testScope)
+        XCTAssertTrue(DPoPRequestDecorator.shouldAttachDPoP(scope: testScope, tokenType: nil))
+        XCTAssertTrue(DPoPRequestDecorator.shouldAttachDPoP(scope: testScope, tokenType: "Bearer"))
+        XCTAssertTrue(DPoPRequestDecorator.shouldAttachDPoP(scope: testScope, tokenType: "DPoP"))
```
wmathurin, line 278: **This assertion contradicts the `shouldAttachDPoP` implementation and is the root cause of the CI failure.**

The implementation correctly short-circuits on any explicit non-DPoP `tokenType` *before* doing any Keychain I/O — so `tokenType: "Bearer"` always returns `false`, regardless of whether key material is present. That is the whole point of the fix for the Bearer-re-auth false-positive.

Change `XCTAssertTrue` → `XCTAssertFalse` here (and update the comment on the preceding line accordingly):

```swift
// Key material present, but an explicit "Bearer" tokenType takes priority — no proof.
XCTAssertFalse(DPoPRequestDecorator.shouldAttachDPoP(scope: testScope, tokenType: "Bearer"))
```

The implementation is correct; only the test expectation needs to change.

wmathurin (approved):

PR #4128, opened 2026-08-12
Add auth_trigger and sdkInfo params to native browser authorize URL

brandonpage:
## Summary
- Appends `sdkInfo` (SDK user-agent string) and `auth_trigger` (why native browser login was chosen) query params to the OAuth authorize URL, native browser (ASWebAuthenticationSession) path only
- WebView path is unchanged — it already sends the real User-Agent header
- `auth_trigger` priority: `login_for_admin` > `mdm` > `force_advanced_auth` > `org_config`

## Test plan
- [ ] `SFOAuthCoordinatorTests` passes
- [ ] Manual login via native browser confirms both params present in the authorize request

GUS: W-23789310

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
@@ -457,13 +457,30 @@
     }
 }
 
-- (void)continueNativeBrowserFlowWithSharedBrowserSessionEnabled:(BOOL)shareBrowserSession {
-    if (![NSThread isMainThread]) {
-        dispatch_async(dispatch_get_main_queue(), ^{
-            [self continueNativeBrowserFlowWithSharedBrowserSessionEnabled:shareBrowserSession];
-        });
-        return;
+// Returns the single auth_trigger value that best describes why native browser login is being
+// used. Mirrors the priority logic (and underlying signals) of computeBMarkerForAuthSession:
+// completedAuthType: in SFUserAccountManager.m — login_for_admin > mdm > force_advanced_auth >
+// org_config. Computed lazily from state that is fixed for the duration of a single
+// authentication attempt, so no threading through the modality-decision call sites is needed.
+- (NSString *)computeAuthTrigger {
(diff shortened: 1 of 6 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
```diff
@@ -457,25 +457,60 @@ - (void)beginNativeBrowserFlowWithSharedBrowserSessionEnabled:(BOOL)shareBrowser
...
                                            codeChallenge:nil];
     approvalUrl = [NSString stringWithFormat:@"%@&state=%@", approvalUrl, self.credentials.identifier];
-    
+
     if (!shareBrowserSession) {
         approvalUrl = [NSString stringWithFormat:@"%@&prompt=login", approvalUrl];
     }
-    
+
+    NSString *sdkInfoValue = [[SalesforceSDKManager sharedManager] userAgentString:@"" forUser:nil];
```
wmathurin: Doesn't the userAgentString includes the web view user agent string too (see https://github.com/forcedotcom/SalesforceMobileSDK-iOS/blob/dev/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Common/SalesforceSDKManager.m#L1033)? We probably want just the first part. The actual user agent of the native browser is captured separately by the server.

wmathurin (changes requested):

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/OAuth/SFOAuthCoordinator.m
brandonpage: Good catch, we don't need need that part.

brandonpage: Fixed in bec1f5a — split out a new internal `sdkUserAgentString:forUser:` on `SalesforceSDKManager` that omits the trailing WebView user agent, and use that for `sdkInfo` instead of the public `userAgentString:forUser:`. Added a regression test asserting `sdkInfo` excludes the full (WebView-appended) UA string.

wmathurin (approved):

PR #4130, opened 2026-08-13
fix(W-23836436): allow dpop_jkt on login pool servers

wmathurin:
## Summary

- Removes the `isPoolLoginHost` guard from `appendDPoPJktIfNeededTo:domain:credentials:` in `SFOAuthCoordinator.m` — pool servers now support DPoP authorization code binding
- `welcome.salesforce.com/discovery` is unaffected (it is a discovery endpoint and never directly receives an `/authorize` request)

## Known limitation — server-side bug W-23864247

During testing against `login.test1.pc-rnd.salesforce.com`, the token exchange returns HTTP 400 `invalid_dpop_proof` even though the client's DPoP proof is cryptographically correct and its JWK thumbprint exactly matches the `dpop_jkt` sent in `/authorize`. This is a server-side issue filed as **W-23864247** (Auth Protocols team). `test_givenDPoP_whenLoginViaPoolServer_thenTokenTypeIsDPoP` throws `XCTSkip` until the server fix is confirmed.

## Changes

**Production code**
- `SFOAuthCoordinator.m` — remove `[SFSDKAuthConfigUtil isPoolLoginHost:domain]` early-exit; only `domain == nil` exits early

**Unit tests**
- `SFOAuthCoordinatorTests.swift` — invert two pool-host tests: `login.salesforce.com` and `test.salesforce.com` now assert `dpop_jkt` **present** (renamed `thenUrlHasDPoPJkt`)
- `welcome.salesforce.com/discovery` test unchanged (still asserts absent)

**UI tests (AuthFlowTester)**
- `ui_test_config.json.sample` — add top-level `loginPoolHost` field
- `UITestConfigUtils.swift` — add `loginPoolHost: String?` to `TestConfig`; add `getLoginPoolHost()` accessor
- `BaseAuthFlowTester.swift` — add `useLoginPoolHost: Bool = false` to `login()` and `launchLoginAndValidate()`; when true, configures the pool server URL as the login host while keeping credentials from the specified `loginHost`
- `DPoPLoginTests.swift` — add `test_givenDPoP_whenLoginViaPoolServer_thenTokenTypeIsDPoP` (`XCTSkip` pending W-23864247); rename `test_givenDPoPECA_whenAdminLogin_thenDPoPBindingWorksThroughSafariVC` → `ThroughBrowser`

## Test plan

- [x] `SFOAuthCoordinatorTests` — all 18 tests pass (verified locally)
- [ ] `DPoPLoginTests.test_givenDPoP_whenLoginViaPoolServer_thenTokenTypeIsDPoP` — blocked on server bug W-23864247

## GUS

W-23836436 — [iOS] Allow dpop_jkt on login pool servers (/authorize code binding)

### DIFF
--- a/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift
+++ b/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift
@@ -34,6 +34,7 @@ let kBrowserLoginForceFlag        = "B4"
 private let kAllBMarkers          = ["B1", "B2", "B3", "B4"]
 
 // L-marker codes (which login server type)
+let kLoginServerProduction        = "L1"
 let kLoginServerWelcomeDiscovery  = "L3"
 let kLoginServerMyDomain          = "L4"
 private let kAllLMarkers          = ["L1", "L2", "L3", "L4", "L5"]
@@ -189,10 +190,16 @@ class BaseAuthFlowTester: XCTestCase {
         let loginHostToUse: String
         if useWelcomeDiscovery {
             loginHostToUse = "welcome.salesforce.com/discovery"
-        } else if useLoginPoolHost, let poolHost = try? UITestConfigUtils.shared.getLoginPoolHost() {
-            loginHostToUse = poolHost
-                .replacingOccurrences(of: "https://", with: "")
-                .replacingOccurrences(of: "http://", with: "")
+        } else if useLoginPoolHost {
+            do {
+                let poolHost = try UITestConfigUtils.shared.getLoginPoolHost()
+                loginHostToUse = poolHost
+                    .replacingOccurrences(of: "https://", with: "")
+                    .replacingOccurrences(of: "http://", with: "")
+            } catch {
+                XCTFail("useLoginPoolHost is true but getLoginPoolHost() failed: \(error)")
+                return
+            }
         } else {
             loginHostToUse = hostConfig.urlNoProtocol
         }
@@ -454,7 +461,8 @@ class BaseAuthFlowTester: XCTestCase {
             isMultiUser: isMultiUser,
             usesWelcomeDiscovery: useWelcomeDiscovery,
             loginForAdmin: loginForAdmin,
-            useDPoP: useDPoP
+            useDPoP: useDPoP,
+            useLoginPoolHost: useLoginPoolHost
         )
     }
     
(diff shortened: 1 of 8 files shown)


native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift
```diff
@@ -182,8 +183,19 @@ class BaseAuthFlowTester: XCTestCase {
...
+        // so the auth code binding goes through the pool server while credentials come from loginHost.
         loginPage.returnToHostList(expectingBrowser: advancedAuthEnabled)
-        let loginHostToUse = useWelcomeDiscovery ? "welcome.salesforce.com/discovery" : hostConfig.urlNoProtocol
+        let loginHostToUse: String
+        if useWelcomeDiscovery {
+            loginHostToUse = "welcome.salesforce.com/discovery"
+        } else if useLoginPoolHost, let poolHost = try? UITestConfigUtils.shared.getLoginPoolHost() {
+            loginHostToUse = poolHost
+                .replacingOccurrences(of: "https://", with: "")
+                .replacingOccurrences(of: "http://", with: "")
```
sfdctaka: minor: `getLoginPoolHost()` is designed to `throw` when `loginPoolHost` is absent/empty, but `try?` swallows that error and silently falls through to the regular loginHost. So if someone runs `test_givenDPoP_whenLoginViaPoolServer_thenTokenTypeIsDPoP` against a `ui_test_config.json` that lacks `loginPoolHost`, the test routes through the regular login host, still passes, and reports a green "pool server" test that never exercised the pool server path — a false positive on exactly the thing this PR exists to verify.

Since the caller opted in with `useLoginPoolHost: true`, the missing config is a setup error that should be loud. Suggest letting it throw (the method already does) or asserting:
```
  } else if useLoginPoolHost {
      let poolHost = try UITestConfigUtils.shared.getLoginPoolHost()  // or XCTFail on nil
      loginHostToUse = poolHost
          .replacingOccurrences(of: "https://", with: "")
          .replacingOccurrences(of: "http://", with: "")
  }
```
`login()` isn't currently `throws`. Simplest is an `XCTFail` + return, matching how other required-config lookups surface in this file.

sfdctaka (approved):
LGTM with a comment.

wmathurin:
I was able to run all the UI tests locally successfully.
Screenshot 2026-08-14 at 5 49 47 PM

PR #4131, opened 2026-08-13
fix(W-23695429): escape single quotes in SmartSQL IN-clause record IDs

wmathurin:
## Summary
- Escapes single quotes in the ID lists passed to `deleteRecordsFromLocalStore` in `SFSyncTarget.m` and `SyncTarget.swift`
- The IDs are joined into a SmartSQL `IN ('id1','id2',...)` clause; without escaping a single quote in an ID could break the query
- Salesforce record IDs are 15/18-char alphanumeric so actual risk is zero, but this satisfies scanner rule BUG-03 (SQL injection, P3) filed as W-23695429

## Test plan
- [ ] Build MobileSync scheme: `xcodebuild build -workspace SalesforceMobileSDK.xcworkspace -scheme MobileSync -sdk iphonesimulator` — confirmed passing
- [ ] Run MobileSync test suite: `xcodebuild test -workspace SalesforceMobileSDK.xcworkspace -scheme MobileSync -sdk iphonesimulator`
- [ ] Verify sync-down + clean-ghosts still work end-to-end in MobileSyncExplorerSwift sample app

### DIFF
--- a/libs/MobileSync/MobileSync/Classes/SyncTarget.swift
+++ b/libs/MobileSync/MobileSync/Classes/SyncTarget.swift
@@ -47,7 +47,8 @@ extension SyncDownTarget {
 
     func deleteRecordsFromLocalStore(syncManager: SyncManager, soupName: String, ids: [String], idField: String) throws {
         if !ids.isEmpty {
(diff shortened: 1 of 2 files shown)


sfdctaka (approved):
LGTM!

PR #4133, opened 2026-08-15
In-place upgrade to DPoP + per-call DPoP intent on token migration

sfdctaka:
## What

Adds a public **`UserAccountManager.upgradeToDPoP(_:success:failure:)`** API that binds an existing Bearer (non-DPoP) session's refresh token to DPoP **in place** — re-authenticating with the user's own client id, redirect URI, and scopes, so no connected-app change is involved. This is the production path for migrating users to DPoP proactively, while the server still accepts the Bearer session, ahead of an ECA flipping to DPoP-enforced.

Supporting changes:

- **`migrateRefreshToken(for:newAppConfig:useDPoP:success:failure:)` overload** — a per-call `useDPoP: NSNumber?` (`@YES` binds the migration to DPoP, `@NO` migrates unbound, `nil` defers to the global flag). The existing `migrateRefreshToken(for:newAppConfig:success:failure:)` now delegates with `nil`, preserving its behavior exactly.
- The per-call intent is threaded **`SFSDKAuthRequest.useDPoP` → `SFSDKAuthSession` → `SFOAuthCoordinator.dpopOverride`**, and `appendDPoPJktIfNeededTo:` now resolves DPoP from `dpopOverride` when set, falling back to `SalesforceSDKManager.useDPoP` otherwise. **Normal login never sets `dpopOverride`, so its `/authorize` gating is unchanged.**
- **AuthFlowTester sample app**: "Upgrade to DPoP" affordance in the migration sheet; new UITest coverage; `SFOAuthCoordinatorTests` cover the per-call gate.

The global `SalesforceSDKManager.useDPoP` flag remains the default DPoP posture for **new logins** only; the migration API is an explicit action on an existing session and is deliberately flag-independent.

## Cross-platform note

This is the iOS half of a cross-platform pair; the **Android sibling is a separate PR in the Android repo** (parallel work). The two are behavior-matched: same per-call `useDPoP` (both directions), same flag-independent migration, same "global flag governs only new logins" split, same DPoP-bound post-upgrade assertions.

## Changes beyond the original spec/plan (added to make on-device tests pass)

On-device UITest runs surfaced two test-only issues that required changes not in the initial spec/plan; recording them here for review (all test-infra, no product code):

1. **`upgradeToDPoP` legitimately registers the `TM` (token-migration) user-agent flag.** The upgrade validation helper initially asserted `TM` *absent* (on the assumption that same-app re-auth isn't a "migration"). On device the UA was `ftr_A2.B4.BW.DP.JT.L4.TM.UA` — `TM` tracks the migration *mechanism*, not "migrated to a different app," so an in-place upgrade correctly sets it. Fixed the helper to assert `TM` present (`wasMigrated: true`) and to pass `isJwt: true` for the `.ecaJwt` token.

2. **SwiftUI custom-`Button` hittability after the sheet reorg.** Adding the "Upgrade to DPoP" section above the existing config editor wrapped the sheet body in a `ScrollView`, after which the pre-existing "Migrate refresh token" button reported a `{-1,-1}` hit point ("Not hittable") though fully visible. Fixed in the page object with a coordinate-based tap (`coordinate(withNormalizedOffset:).tap()`) when `!isHittable`; the spec-required layout is unchanged (no product/UI revert).

## Review flags

- Touches the OAuth `/authorize` DPoP gate, a new public `UserAccountManager` API, and login-UI-adjacent sample-app code — requires maintainer sign-off. The `migrateRefreshToken` overload and the internal `dpopOverride`/`useDPoP` properties are additive/backward-compatible; the existing migration entry point is unchanged.

## Testing

- `SalesforceSDKCore` + AuthFlowTester UITest targets `build-for-testing` succeed (`** TEST BUILD SUCCEEDED **`).
- `SFOAuthCoordinatorTests` cover the per-call gate (override on with global flag off ⇒ `dpop_jkt` present; normal login still gated on the flag).
- On-device: `test_givenBearerSession_whenUpgradeToDPoP_thenDPoPBound` and `testLogin_DPoP_ECA_Without_DPoP_Fails` pass, and the pre-existing migration tests remain green.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
@@ -836,8 +836,10 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
                     success:(SFUserAccountManagerSuccessCallbackBlock)completionBlock
                     failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
 
-    // Store current user credentials to revoke them once migration completes
-    SFOAuthCredentials *preMigrationCredentials = self.currentUser.credentials;
+    // Store the migrated user's credentials to revoke them once migration completes. Snapshot the
+    // passed-in `user`, not `self.currentUser`: the two differ when upgrading a background account,
+    // and revoking currentUser's token here would sign out the wrong user.
+    SFOAuthCredentials *preMigrationCredentials = user.credentials;
 
     // Creating a SFSDKAuthRequest and SFSDKAuthSession
     SFSDKAuthRequest *request = [self migrateRefreshAuthRequest:newAppConfig];
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
@@ -827,12 +827,21 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
                newAppConfig:(SFSDKAppConfig *)newAppConfig
                     success:(SFUserAccountManagerSuccessCallbackBlock)completionBlock
                     failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
-    
+    [self migrateRefreshToken:user newAppConfig:newAppConfig useDPoP:nil success:completionBlock failure:failureBlock];
+}
+
+- (void)migrateRefreshToken:(SFUserAccount *)user
+               newAppConfig:(SFSDKAppConfig *)newAppConfig
+                    useDPoP:(nullable NSNumber *)useDPoP
+                    success:(SFUserAccountManagerSuccessCallbackBlock)completionBlock
+                    failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
+
     // Store current user credentials to revoke them once migration completes
     SFOAuthCredentials *preMigrationCredentials = self.currentUser.credentials;
 
     // Creating a SFSDKAuthRequest and SFSDKAuthSession
     SFSDKAuthRequest *request = [self migrateRefreshAuthRequest:newAppConfig];
+    request.useDPoP = useDPoP;
     SFSDKAuthSession *authSession = [[SFSDKAuthSession alloc] initWith:request credentials:nil];
     authSession.isAuthenticating = YES;
     authSession.authSuccessCallback = ^(SFOAuthInfo *authInfo, SFUserAccount *newUserAccount) {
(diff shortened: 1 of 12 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -861,6 +870,18 @@ - (void)migrateRefreshToken:(SFUserAccount *)user
...
+               success:(SFUserAccountManagerSuccessCallbackBlock)completionBlock
+               failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
+    SFOAuthCredentials *credentials = user.credentials;
+    SFSDKAppConfig *sameAppConfig = [[SFSDKAppConfig alloc] initWithDict:@{
+        @"remoteAccessConsumerKey": credentials.clientId ?: @"",
+        @"oauthRedirectURI": credentials.redirectUri ?: @"",
+        @"oauthScopes": credentials.scopes ?: @[]
+    }];
+    [self migrateRefreshToken:user newAppConfig:sameAppConfig useDPoP:@YES success:completionBlock failure:failureBlock];
+}
```
wmathurin, line 885: This accepts an arbitrary `user`, but `migrateRefreshToken:` snapshots `self.currentUser.credentials` before the migration and later revokes that snapshot. If the caller upgrades a non-current account, the success path can revoke the current user’s refresh token instead of the upgraded user’s old token. Please snapshot `user.credentials` there and add a two-user regression: make B current, upgrade A, then verify B can still refresh.

sfdctaka:
@wmathurin Good catch — this was a real bug. `migrateRefreshToken:newAppConfig:useDPoP:success:failure:` snapshotted `self.currentUser.credentials` to decide which refresh token to revoke once migration completed. When migrating a **non-current** (background) account, that revoked the *current* user's token instead of the migrated user's — signing out the wrong user.

Fixed by snapshotting the passed-in `user.credentials`:

```objc
// Snapshot the passed-in `user`, not `self.currentUser`: the two differ when upgrading a
// background account, and revoking currentUser's token here would sign out the wrong user.
SFOAuthCredentials *preMigrationCredentials = user.credentials;
```

Added a regression test — `testMigrateRefreshToken_whenUserIsNotCurrentUser_revokesMigratedUsersToken` — where user B is current, we migrate non-current user A (rotated token), and assert the revoked credentials are A's old token and **not** B's. Uses a small `SFSDKRevokeCapturingOAuthClient` stub injected via `authClient`. Full `SFUserAccountManagerTests` suite passes (18 tests, 0 failures).

Note for cross-platform parity: this does **not** reproduce on Android (PR #2995) — that revocation path keys on the matched account's identity, so the wrong-user hazard can't occur. I added regression-insurance coverage there anyway.

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/BaseAuthFlowTester.swift
```diff
@@ -704,7 +704,103 @@ class BaseAuthFlowTester: XCTestCase {
...
+        )
+
+        loginPage.returnToHostList(expectingBrowser: advancedAuthEnabled)
+
+        // Selecting the login host triggers /authorize. For an enforced ECA with useDPoP: false,
+        // the server rejects before any login form renders, so there is nothing further to drive —
+        // assert on the outcome instead of attempting performLogin.
+        loginPage.configureLoginHost(host: hostConfig.urlNoProtocol)
+
+        assertMainPageNotLoaded()
```
wmathurin, line 810: Is there some error text in the web view or some alert for the error? Could we assert on that ?

wmathurin (changes requested):

sfdctaka:
Good question — I dug into this by running the test and capturing the rejection surface. Unlike the invalid-client-id / invalid-scope negative tests (which go through the in-app WebView and render an HTML error page we can assert on), this DPoP-enforced rejection happens in the **external browser (ASWebAuthenticationSession)**. The enforced ECA rejects the unbound `/authorize` with a short non-HTML body (~92 bytes), so the browser can't render it and falls back to a QuickLook file preview — a generic file icon labeled `authorize` / `data - 92 bytes` / `Open in…`. There's no `error=` / `error_description=` text or any rendered error element to hook.

The only assertable elements are iOS's own file-preview chrome, which isn't a stable SDK/server contract, so `assertMainPageNotLoaded()` — the app never reaches the authenticated credentials view — is the strongest reliable signal for this flow. I've already gone ahead and added an inline comment in `launchAndAttemptLoginExpectingFailure` documenting this so it's clear to the next reader.

wmathurin (approved):

PR #4134, opened 2026-08-15
Default DPoP on for new logins in MSDK 14

sfdctaka:
## What

Flips `SalesforceSDKManager.usesDPoP` from `NO` to `YES`. DPoP (RFC 9449) is now validated by the server, so new apps and new logins request DPoP-bound tokens by default.

The change governs **new logins only** — the per-credential attachment gate keeps existing Bearer credentials Bearer and existing DPoP credentials bound. Updates the property doc to reflect the new default and drops the old opt-in advisory.

## Changes

- `SalesforceSDKManager.h/.m` — default flipped to `YES`; doc updated.
- `LegacyLoginTests` — now passes `useDPoP: false` explicitly on all its login calls so it stays a dedicated Bearer class under the new default; `LegacyLoginTestsNotHybrid` inherits the override.

## Rebase note

This is a small, self-contained flip that overlaps the DPoP upgrade/migration work in #4133 (both touch DPoP defaults and the AuthFlowTester test classes). Whichever of the two merges first, the other will need a rebase. No preference on order.

## Testing

Tested and passing.

### DIFF
--- a/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/LegacyLoginTests.swift
+++ b/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/LegacyLoginTests.swift
@@ -51,34 +51,34 @@ class LegacyLoginTests: BaseAuthFlowTester {
 
     /// Login with CA opaque using default scopes and web server flow.
     func testCAOpaque_DefaultScopes_WebServerFlow() throws {
-        launchLoginAndValidate(staticAppConfigName: .caOpaque, useHybridFlow: useHybridFlow())
+        launchLoginAndValidate(staticAppConfigName: .caOpaque, useHybridFlow: useHybridFlow(), useDPoP: false)
     }
 
     /// Login with CA opaque using subset of scopes and web server flow.
     func testCAOpaque_SubsetScopes_WebServerFlow() throws {
-        launchLoginAndValidate(staticAppConfigName: .caOpaque, staticScopeSelection: .subset, useHybridFlow: useHybridFlow())
+        launchLoginAndValidate(staticAppConfigName: .caOpaque, staticScopeSelection: .subset, useHybridFlow: useHybridFlow(), useDPoP: false)
     }
 
     /// Login with CA opaque using all scopes and web server flow.
     func testCAOpaque_AllScopes_WebServerFlow() throws {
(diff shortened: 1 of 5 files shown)


wmathurin (commented):
Are there updates needed in native/SampleApps/AuthFlowTester/README.md ?

native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/LegacyLoginTests.swift
```diff
@@ -51,34 +51,34 @@ class LegacyLoginTests: BaseAuthFlowTester {
 
     /// Login with CA opaque using default scopes and web server flow.
     func testCAOpaque_DefaultScopes_WebServerFlow() throws {
-        launchLoginAndValidate(staticAppConfigName: .caOpaque, useHybridFlow: useHybridFlow())
+        launchLoginAndValidate(staticAppConfigName: .caOpaque, useHybridFlow: useHybridFlow(), useDPoP: false)
```
wmathurin, line 54: The AuthFlowTester README and `AuthFlowTesterUITests/overview.md` should call out the new intent here: Mobile SDK 14 defaults DPoP on, but `LegacyLoginTests` explicitly uses `useDPoP: false` so it remains the dedicated Bearer compatibility suite. Without that note, the docs describe it only as a legacy-flow suite and do not explain why it overrides the new default.

sfdctaka:
Good call on both — done in 6777990be.

Added a **DPoP default note** to both docs:
- `native/SampleApps/AuthFlowTester/README.md` (LegacyLoginTests section)
- `native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/overview.md` (LegacyLoginTests section)

Each note calls out that MSDK 14 defaults DPoP **on** for new logins (`SalesforceSDKManager.useDPoP` now defaults to `true`), but that LegacyLoginTests explicitly passes `useDPoP: false` on every case so it stays the dedicated **Bearer (non-DPoP) compatibility suite** — exercising the legacy CA flows as pre-14 apps did — and points readers to `DPoPLoginTests` for DPoP-bound coverage. That way the docs explain *why* the suite overrides the new default rather than reading only as a legacy-flow suite.

wmathurin (approved):

PR #4136, opened 2026-08-17
Support biometric authentication with advanced authentication

brandonpage:
## Summary

Suppress Advanced Authentication (`ASWebAuthenticationSession`)  to show OS Biometric Prompt if locked.  When a biometric-locked user declines or fails the Face ID / Touch ID prompt (e.g. taps "Use Password" or cancels), the login screen is automatically launched. 

## What changed

- **`BiometricAuthenticationManagerInternal.swift`**
  - Added `suppressInitialBrowserAuthentication`, a one-shot flag armed by `lock()` so the browser attempt it triggers doesn't race the biometric prompt. Consumed (cleared) by the gate on its next check.
  - On biometric cancel/failure, wait for the scene to return to `.foregroundActive` (dismissing the system biometric sheet briefly leaves it inactive, and `ASWebAuthenticationSession.start()` silently no-ops from a non-active scene), then resume the suppressed browser session in place — keeping its covering window up so the app is never exposed.
  - Guard `handleAppForeground()` against re-locking (and re-arming the flag) while already locked.
- **`SFUserAccountManager` (`.h`/`.m`)**
  - `willBeginBrowserAuthentication:` gate honors the suppression flag: suppresses the browser and presents the login-host picker as a fallback landing screen when biometric is locked; otherwise proceeds normally.
  - Added `resumeBrowserAuthentication:` to resume the suppressed session's held browser-launch callback (falls back to a fresh login if the callback is gone).
  - `isAlreadyPresentingLoginController:` now also recognizes the host-list picker so the Advanced Auth retry dismisses it before presenting the browser.
- **`SFSDKLoginHostListViewController.m`** — adds a "retry biometric" nav-bar button (picker-screen counterpart to `SFLoginViewController`'s existing biometric button), shown only when biometric is locked, opted-in, and available. Reuses the existing `biometricLoginButton` localized string — no new strings.

## Tests

- New `BiometricAdvancedAuthGateTests.swift` covering the gate / suppression-flag behavior.
- Extended `SFUserAccountManagerTests.m` and `SFSDKLoginHostTests.m`.
- `SalesforceSDKCore` builds clean; new + affected suites pass on the simulator. Also device-tested on a physical iPhone for the Face ID cancel → browser flow.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
@@ -53,6 +53,12 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     internal var laContext = LAContext()
     private let kBioAuthPolicyIdentifier = "com.salesforce.security.bioauthpolicy"
     private let kBioAuthEnabledIdentifier = "com.salesforce.security.bioauth"
+
+    /// One-shot flag telling `willBeginBrowserAuthentication:` to suppress the browser launch that
+    /// `lock()`'s `login()` triggers, so it doesn't race the biometric prompt. The gate clears it
+    /// on its next check, so a later attempt (fallback picker, gear-menu retry) is not suppressed.
+    /// Read/written from SFUserAccountManager.m, hence `@objc`.
+    @objc internal var suppressInitialBrowserAuthentication = false
     
     private override init() {
         super.init()
@@ -66,7 +72,10 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     
     /// Locks the screen if necessary
     @objc public func handleAppForeground() {
-        if shouldLock() {
+        // Skip if already locked: shouldLock() ignores `locked`, so a foreground while locked would
+        // call lock() again and re-arm suppressInitialBrowserAuthentication, wrongly suppressing a
+        // browser-auth attempt that already cleared it.
+        if !locked && shouldLock() {
             lock()
         }
     }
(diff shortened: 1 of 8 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
```diff
@@ -136,7 +145,11 @@ public class BiometricAuthenticationManagerInternal: NSObject, BiometricAuthenti
     public func lock() {
         locked = true
         NotificationCenter.default.post(name: Notification.Name(rawValue: kSFBiometricAuthenticationFlowWillBegin), object: nil)
-        
+
+        // Suppress the browser launch from the login() below when biometric will be shown, so it
+        // doesn't race the prompt. The gate in SFUserAccountManager clears this on its next check.
+        suppressInitialBrowserAuthentication = showNativeLoginButton()
```
wmathurin: This state is global, but `UserAccountManager.shared.login` starts authentication for every connected scene. The first scene that reaches `willBeginBrowserAuthentication:` consumes this one-shot flag; subsequent scenes can launch `ASWebAuthenticationSession` while the app is still biometrically locked. Please scope suppression to the auth session or scene identifier, and add a two-scene regression test.

wmathurin (commented):
Inline review comment.

libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Security/BiometricAuthentication/BiometricAuthenticationManagerInternal.swift
brandonpage: Good catch — fixed in 47c3da69b. Replaced the global one-shot `suppressInitialBrowserAuthentication` with a per-scene suppression set keyed by the scene's `persistentIdentifier`:

- `lock()` now arms suppression for **each** connected scene before triggering `login()` (which fans out per scene).
- The gate consumes only the current scene's id (`consumeBrowserAuthenticationSuppressionForSceneId:`), so one scene's consume can't drain another's — every locked scene suppresses its own browser launch.
- `presentBiometric(scene:)` disarms only its own scene on the capability race.

Added a two-scene regression test (`test_willBeginBrowserAuth_twoScenes_eachSuppressedIndependently`) asserting both scenes return `NO` / allocate no `ASWebAuthenticationSession`, plus a per-scene consume invariant test (`test_consumeSuppression_isScopedPerScene`). Full `BiometricAdvancedAuthGateTests` suite (16 tests) passes.

wmathurin (approved):

PR #4137, opened 2026-08-19
fix(setVersion.sh): support pre-release version formats (e.g. 14.0.0-rc.0)

wmathurin:
## Summary

Strip the pre-release suffix before computing the C preprocessor macro name and value in `update_salesforce_sdk_constants`. For `14.0.0-rc.0`, the old code produced `__SALESFORCE_SDK_14_0_0-rc_0` (invalid C identifier — hyphen is not allowed in macro names) and `14000-rc00` (not a valid integer). The fix derives both from `base_version=${version%%-*}` so the macro is always a valid C identifier.

GUS: W-23916129

## Test plan

- [ ] `./setVersion.sh -v 14.0.0-rc.0` completes without error; generated header contains valid `#define __SALESFORCE_SDK_14_0_0 140000`
- [ ] `./setVersion.sh -v 14.0.0` continues to work as before (no regression)

### DIFF
--- a/setversion.sh
+++ b/setversion.sh
@@ -54,8 +54,9 @@ update_salesforce_sdk_constants ()
     local file=$1
     local version=$2
     local isDev=$3
-    local defineNameForVersion="__SALESFORCE_SDK_${version//./_}"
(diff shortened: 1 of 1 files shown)


sfdctaka (approved):
LGTM!

PR #4142, opened 2026-08-24
Add downgradeFromDPoP for in-place DPoP→Bearer migration

sfdctaka:
## What

Adds `SFUserAccountManager.downgradeFromDPoP(_:success:failure:)`, the inverse of the existing `upgradeToDPoP`. It rolls a DPoP-bound session back to unbound Bearer **in place** — same connected app, redirect URI, and scopes — independent of the global `usesDPoP` flag. It is a **no-op** (success callback, unchanged account) when the session is already Bearer.

## Why

Completes the in-place DPoP migration pair so apps can move a user off DPoP without a full logout/login, mirroring `upgradeToDPoP`.

## Changes

- **`SFUserAccountManager`** — new `downgradeFromDPoP:` wrapper delegating to `migrateRefreshToken` with `useDPoP:@NO`; no-op guard when already Bearer. On success it deletes the obsolete DPoP key pair + nonce-cache entry keyed to the pre-migration credentials, and unregisters the `DP` user-agent feature marker for the now-Bearer completion.
- **`SFOAuthCoordinator` (`/authorize` binding)** — decides whether to append `dpop_jkt` by precedence: explicit per-call override → per-user `credentials.tokenType` → global flag. Previously an interactive re-login of a downgraded account fell through to the global flag (default on in 14.0) and silently re-bound the session to DPoP. **This is an OAuth2 `/authorize` behavior change and needs maintainer review.**
- **AuthFlowTester** — "Downgrade from DPoP" action (enabled only for a DPoP-bound session) + downgrade UI test.
- **Tests** — coordinator precedence unit tests (17/17 pass) + downgrade delegation / no-op-guard unit tests.

## Review notes

- Touches the OAuth2 credential path, a public API, and the `/authorize` binding decision — requesting maintainer (Wolf) review before merge; kept as **draft**.
- Device re-run of the downgrade UI test's post-downgrade revoke→refresh leg is the remaining verification.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
@@ -866,11 +866,27 @@ static NSString * const kSFGenericFailureAuthErrorHandler = @"GenericFailureErro
     request.useDPoP = useDPoP;
     SFSDKAuthSession *authSession = [[SFSDKAuthSession alloc] initWith:request credentials:nil];
     authSession.isAuthenticating = YES;
+    // Snapshot DPoP-ness before the revoke block below can mutate `preMigrationCredentials`:
+    // `revokeRefreshToken:` synchronously clears `tokenType` (and other fields) on the credentials
+    // object it's handed, so reading `tokenType` after that call would always see it as non-DPoP.
+    BOOL preMigrationWasDPoP = [preMigrationCredentials.tokenType isEqualToString:@"DPoP"];
     authSession.authSuccessCallback = ^(SFOAuthInfo *authInfo, SFUserAccount *newUserAccount) {
         if (preMigrationCredentials != nil && ![preMigrationCredentials.refreshToken isEqualToString:newUserAccount.credentials.refreshToken]) {
-            
+
             id<SFSDKOAuthProtocol> authClient = self.authClient();
             [authClient revokeRefreshToken:preMigrationCredentials reason:SFLogoutReasonRefreshTokenRotated];
+        }
+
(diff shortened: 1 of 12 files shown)


libs/SalesforceSDKCore/SalesforceSDKCore/Classes/UserAccount/SFUserAccountManager.m
```diff
@@ -904,6 +928,26 @@ - (void)upgradeToDPoP:(SFUserAccount *)user
     [self migrateRefreshToken:user newAppConfig:sameAppConfig useDPoP:@YES success:completionBlock failure:failureBlock];
 }
 
+- (void)downgradeFromDPoP:(SFUserAccount *)user
+               success:(SFUserAccountManagerSuccessCallbackBlock)completionBlock
+               failure:(SFUserAccountManagerFailureCallbackBlock)failureBlock {
+    // No-op if the session is already unbound (Bearer): there's nothing to migrate, so return
+    // the user unchanged rather than kicking off a needless re-authentication.
+    if (![user.credentials.tokenType isEqualToString:@"DPoP"]) {
```
wmathurin: All the `isEqualToString:@"DPoP"` comparisons in this file (this one, the `upgradeToDPoP` no-op guard at line 916, and the success-callback cleanup checks in `migrateRefreshToken`) are case-sensitive, while the new per-user check in `appendDPoPJktIfNeededTo` in `SFOAuthCoordinator` uses `caseInsensitiveCompare:[SFSDKDPoPRequestDecorator dpopTokenType]`. If the server ever sends a casing variant the two paths would disagree on whether a session is DPoP-bound.

Suggest using `[SFSDKDPoPRequestDecorator dpopTokenType]` with a consistent comparison strategy here and in the parallel guards (rather than inlining `@"DPoP"`) so there's a single source of truth for the token-type string.

wmathurin (approved):

PR #4143, opened 2026-08-24
Re-enable DPoP pool server test; fix L1 user-agent marker expectation

wmathurin:
## Summary

- **W-23864247 is fixed server-side.** The pool login server was previously rejecting valid `dpop_jkt` token exchanges with `invalid_dpop_proof`; the server fix is confirmed and the test is now re-enabled.
- **L1 marker fix.** Pool server login routes through production login infrastructure (`login.salesforce.com`), so the user-agent carries the L1 (production) marker, not L4 (My Domain). Added a `useLoginPoolHost: Bool = false` parameter to the public `assertRevokeAndRefreshWorks` overload that selects `kLoginServerProduction` instead of `kLoginServerMyDomain` when `true`.
- **README updated.** The test-table row for `test_givenDPoP_whenLoginViaPoolServer_thenTokenTypeIsDPoP` no longer notes a skip; updated to describe the expected behavior.

## Changes

- `DPoPLoginTests.swift` — removed `XCTSkip` block; passed `useLoginPoolHost: true` to `assertRevokeAndRefreshWorks`
- `BaseAuthFlowTester.swift` — added `useLoginPoolHost` parameter; uses `kLoginServerProduction` (L1) when `true`
- `README.md` — updated test-table row

## Test plan

- [ ] `test_givenDPoP_whenLoginViaPoolServer_thenTokenTypeIsDPoP` verified passing at 126.6s on iPhone 17 iOS 26.1 simulator
- [ ] No other DPoP tests regressed (all other `assertRevokeAndRefreshWorks` callers pass `useLoginPoolHost` as `false` by default, so L4 behavior is unchanged)

### DIFF
--- a/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/DPoPLoginTests.swift
+++ b/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Tests/DPoPLoginTests.swift
@@ -175,15 +175,10 @@ class DPoPLoginTests: BaseAuthFlowTester {
 
     // MARK: - Pool Server Login
 
-    /// Login via the pool server (login.test1.pc-rnd.salesforce.com) with DPoP enabled
-    /// and verify that dpop_jkt was accepted and DPoP binding holds after a revoke+refresh.
-    ///
-    /// Skipped: server-side bug W-23864247 — the pool login server returns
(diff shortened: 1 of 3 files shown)


sfdctaka (approved):
LGTM!

PR #4144, opened 2026-08-25
fix(auth): fix DPoP validation propagation in AuthFlowTester UI tests

wmathurin:
## Summary

- Fixes `BaseAuthFlowTester` test infrastructure so explicit `useDPoP` enablement correctly propagates through the full validation call chain (`launchAndLogin` → `validate()` → `validateUser` → `assertRevokeAndRefreshWorks`)
- Syncs `README.md` and `overview.md` with the actual test suite (missing tests, stale test names)
- All non-Beacon tests pass. Beacon tests (`BEACON_OPAQUE` / `BEACON_JWT`) fail pre-existing across all suites — same failures observed on Android ([forcedotcom/SalesforceMobileSDK-Android#3005](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/3005)), pointing to an environment or server-side issue with the Beacon app configuration in the test sandbox rather than a code regression.

## Root cause (DPoP validation)

SDK 14.0 made DPoP default-on: commit `a9d333f01` changed `self.useDPoP = NO` → `self.useDPoP = YES` in `resetAuthFlags`. The validation logic in `BaseAuthFlowTester` derived whether to expect DPoP credentials solely from the app config name (`isDPoP` returns `true` only when the config name contains `_dpop`). Tests that explicitly enable DPoP via Login Options (e.g. `DPoPLoginTests` operating on `ecaOpaque`) called `validate()` with `useDPoP: true`, but `validateUser` and `assertRevokeAndRefreshWorks` still consulted `userAppConfig.isDPoP` — always `false` for `ecaOpaque` — so DPoP credentials were validated against Bearer expectations, causing assertion failures.

## Root cause (WelcomeLoginTests / parallel flakiness)

`WelcomeLoginTests` non-beacon tests (`testWelcomeDiscovery_RegularAuthLoginHost`, `_DynamicConfig`) appeared to consistently fail all 3 retries when run in parallel with Beacon tests. Beacon tests each time out after 109s × 3 retries = ~327s per test, saturating the simulator pool. The same two tests pass reliably when run in isolation (117s, 118s). No code change was required; the failures were a parallelism artefact, not a product regression.

## Fix

Propagate `useDPoP` through the `BaseAuthFlowTester` call chain so explicit DPoP enablement overrides the config-name-derived default:

- `launchAndLogin` and `loginOtherUser` (no-validate variant) — added `useDPoP: Bool = false`, forwarded to `login()`
[...]

PR #4145, opened 2026-08-25
fix(dpop): handle 400 use_dpop_nonce on revoke path after cold restart (W-23501382)

wmathurin:
## Problem

After a cold app restart the in-memory `DPoPNonceCache` is empty. The first DPoP-decorated REST call — most commonly the access-token revoke that `assertRevokeAndRefreshWorks` issues — sends a nonce-less DPoP proof. Salesforce's authorization server replies with **HTTP 400 `use_dpop_nonce`** and the desired nonce in the `DPoP-Nonce` response header.

The existing retry path in `SFRestAPI` only handles **401** (expired credentials) and **403 Bad_OAuth_Token**; it does not handle 400, so the revoke fails and the test throws `XCTAssertTrue failed - Failed to revoke access token`.

This was the root cause of the prematurely-closed W-23501382.

## Fix

**`SFRestAPI.enqueueRequest`** — before the existing `shouldRetry` guard, detect a 400 response whose body contains `use_dpop_nonce`, harvest the server-issued nonce from the `DPoP-Nonce` response header into `DPoPNonceCache`, and re-enqueue the request once with the updated proof.

**`SFRestRequest+Internal.h`** — adds `dpopNonceRetried` BOOL flag so the retry fires at most once per request (prevents looping if the server keeps challenging).

Per RFC 9449 §8 the server SHOULD include the desired nonce in the `DPoP-Nonce` header of the challenge response; `harvestNonceFromResponse:requestURL:scope:` is a no-op if the header is absent, so the retry falls through to the normal error path if the server doesn't cooperate.

## Test

`DPoPLoginTests.test_givenDPoPUser_whenAppRestart_thenSessionAndKeypairSurvive` — previously `XCTSkip`'d — is now enabled and passing:

1. Login with ECA JWT DPoP (no hybrid), user `.third`
2. Cold restart (`restart()`)
3. `restartAndValidateUser` — confirms EC keypair and session reloaded from Keychain/disk
4. `assertRevokeAndRefreshWorks` — confirms revoke + refresh succeeds despite cold nonce cache

| Suite | Test | Result |
|---|---|---|
| `DPoPLoginTests` | `test_givenDPoPUser_whenAppRestart_thenSessionAndKeypairSurvive` | ✅ passed (150s) |

## Notes

- This PR is intended to be merged **after** #4144 (AuthFlowTester test harness fixes), which is being merged separately
- The fix is scoped to the REST client path — the token endpoint's own nonce-challenge retry (via the 401→refresh→replay path) is unchanged
- Android's equivalent test passes because its revoke sends to the login host over raw OkHttp, bypassing DPoP; iOS revokes go to the instance host through the DPoP-decorating REST stack, so this path was never exercised on Android

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/RestAPI/SFRestAPI.m
@@ -364,6 +364,22 @@ successBlock:(SFRestResponseBlock)successBlock
                     request.successBlock(dataForDelegate, response);
                 }
             } else {
+                // DPoP nonce challenge (HTTP 400 with use_dpop_nonce in body): harvest the
+                // server-issued nonce and retry the request once with the updated proof.
+                // This covers the post-restart case where the in-memory nonce cache is empty
+                // and the first outbound DPoP call (e.g. revoke) triggers a nonce challenge.
+                // RFC 9449 §8 — the server SHOULD return the desired nonce in DPoP-Nonce.
+                NSString *bodyStr = data ? [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding] : nil;
+                if (statusCode == 400
+                    && !request.dpopNonceRetried
+                    && [bodyStr containsString:SFSDKDPoPRequestDecorator.nonceErrorCode]) {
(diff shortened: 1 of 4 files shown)


sfdctaka (approved):
LGTM!

PR #4146, opened 2026-08-26
fix(auth): guard migrateRefreshToken dispatch_async against stale sessions (W-23992574)

wmathurin:
## Summary

- `testDowngradeFromDPoP_*` tests queue a `dispatch_async(dispatch_get_main_queue(), ...)` in `migrateRefreshToken:` then immediately clean up the auth session; the pending block fires during `testLogin`'s `waitForExpectations:timeout:20`, triggering `NSInternalInconsistencyException: Use SFRestAPI sharedInstance for authenticated requests` (seen in [Actions run 32898781351](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/actions/runs/32898781351))
- Add a session-alive guard (`authSessions[authSession.sceneId]`) in both the outer and inner `dispatch_async` block so stale sessions are silently skipped
- Fix `testLogin`'s empty failure block to surface errors via `XCTFail` — previously silent failures masked real timeouts

## Test plan

- [x] Ran `SFUserAccountManagerTests` locally: 24 tests, 0 failures (including `testLogin` passing in ~0.9 s with a real network token refresh)
- [x] Validated against CI failure scenario: `testDowngradeFromDPoP_*` tests run before `testLogin` with no crash


sfdctaka (approved):
LGTM!

Issue #3005, opened 2019-09-17
Sign In with Apple requirements


az-oolloow:
There has been a number of reports online saying that Apple will soon require all apps providing SSO to offer a Sign in with Apple support

The developer documentation is available here:
https://developer.apple.com/sign-in-with-apple/get-started/
https://developer.apple.com/documentation/signinwithapplerestapi

How can this be integrated with Salesforce when the App is built as a Native Swift app with Salesforce Mobile SDK?

az-oolloow:
This is the specific Apple Guideline requirement link
https://developer.apple.com/app-store/review/guidelines/#sign-in-with-apple

Looking at it, it may be argued that an app built with SF SDK falls under 
```
Sign in with Apple is not required if:
....
Your app is an education, enterprise, or business app that requires the user to sign in with an existing education or enterprise account.
```

But I guess it'd be good to hear confirmation from SF MobileSDK developers or SFDC directly

wmathurin:
That's our understanding too. But we can't speak for Apple though.

PR #4147, opened 2026-08-26
test(auth): add DPoP RTR pool-server safety-net test (W-23991713)

wmathurin:
## Summary

- Adds `test_givenDPoPRtr_whenLoginViaPoolServer_thenRefreshTokenSurvivesIdentityFetch` to `DPoPLoginTests` — mirrors the equivalent test on Android (PR #3005)
- Verifies that a pool-server login with DPoP + RTR does **not** consume the refresh token during the post-login identity fetch
- Safety net for the RTR-unsafe pattern in `SFIdentityCoordinator`: if the identity fetch were to trigger a credential refresh (consuming the RTR refresh token), `assertRevokeAndRefreshWorks(isRtr: true, ...)` would fail because the refresh token would already be spent
- Test passes today because `credentials.identityUrl` already holds the raw pool-server URL — `SFIdentityCoordinator` routes the identity call correctly without needing a credential refresh

## Related

- W-23991713 — iOS story: fix RTR-unsafe credential-refresh in SFIdentityCoordinator + add pool-server DPoP RTR test
- W-23992239 — Server bug: My Domain `/id/` endpoint returns `Bad_OAuth_Token` for pool-server-issued DPoP tokens instead of `401 use_dpop_nonce`
- Android PR #3005 — equivalent fix + test

## Test plan

- [x] `test_givenDPoPRtr_whenLoginViaPoolServer_thenRefreshTokenSurvivesIdentityFetch` — passes (129 s)
- [x] `test_givenDPoP_whenLoginViaPoolServer_thenTokenTypeIsDPoP` — passes, no regression (163 s)


sfdctaka (approved):
LGTM!

PR #4154, opened 2026-08-28
[iOS] Capture ui_sid from DPoP token response and use it in mainSid

wmathurin:
## Summary

- Adds `uiSid` property to `SFOAuthCredentials` (public readonly in `.h`, readwrite in `+Internal.h`)
- Persists `uiSid` to the keychain in `SFOAuthKeychainCredentials` under `com.salesforce.mobilesdk.oauth.uiSid`
- In `updateCredentials:`, reads `ui_sid` from the token endpoint response, guarded on `tokenType == "dpop"`
- Updates `mainSid` getter: returns `uiSid` if present, otherwise falls back to existing logic (`parentSid` for jwt, `accessToken` otherwise)
- Adds 4 unit tests in `SFOAuthCredentialsTests.m`

## Dependencies

**This PR depends on (already merged):**
- [SalesforceMobileSDK-iOS#4153](https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4153) — [iOS] Move mainSid calculation to SFOAuthCredentials ✅

**Related PRs (same feature):**
- [SalesforceMobileSDK-Android#3019](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/3019) — [Android] Capture ui_sid from DPoP token response and use it in mainSid
- [SalesforceMobileSDK-Android#3017](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/3017) — [Android] Move mainSid calculation to UserAccount
- [SalesforceMobileSDK-iOS-Hybrid#323](https://github.com/forcedotcom/SalesforceMobileSDK-iOS-Hybrid/pull/323) — [iOS-Hybrid] Use creds.mainSid in SalesforceWebViewCookieManager

## Tracking
- W-24024985 — [iOS] Capture ui_sid from token endpoint and use it in mainSid
- Fixes W-23984602

PR #4155, opened 2026-08-28
[iOS] Fix stale uiSid after DPoP-to-Bearer downgrade

wmathurin:
## Summary

Addresses review feedback on the merged PR #4154.

- **`SFOAuthCredentials.m`**: when `token_type` is present in the token endpoint response and is not `dpop`, explicitly sets `self.uiSid = nil`. This clears any stale DPoP `uiSid` from the keychain after a DPoP-to-Bearer downgrade; without this, `mainSid` would return the old session ID instead of the Bearer access token.
- **`SFOAuthCredentialsTests.m`**: adds a regression test (`test_givenExistingUiSid_whenUpdateCredentialsWithBearerTokenType_thenUiSidCleared`) that exercises the full DPoP-to-Bearer transition through `updateCredentials:`.

## Tracking
- W-24024985
- Fixes W-23984602


sfdctaka (approved):
LGTM!

PR #4160, opened 2026-08-31
[iOS] NativeLoginManager Basic-Auth header uses URL-safe Base64 instead of standard Base64, breaking login for some username/password combinations

JohnsonEricAtSalesforce:
## Summary

`NativeLoginManagerInternal`'s helper for building the Basic-Auth `Authorization` header (`generateColonConcatenatedBase64String()`) encoded the colon-concatenated username and password with the URL-safe Base64 alphabet and no padding (via `urlSafeBase64Encode()`). HTTP Basic Authentication (RFC 7617) requires the standard Base64 alphabet with padding. For most credential pairs the two alphabets produce identical output, which is why this went unnoticed — they only diverge when the encoded bytes would contain `+`, `/`, or need `=` padding under the standard alphabet. When that happens, login fails with a generic "check your username and password" error even though the credentials are correct (confirmed working via web login for the same user).

Same defect class as the already-fixed, closed Android bug [W-23986240](https://github.com/forcedotcom/SalesforceMobileSDK-Android/pull/3007) — the iOS Swift implementation independently had the identical bug in its own encoding helper, and the Android fix was never ported to iOS.

- Changed `generateColonConcatenatedBase64String()` to encode with the standard Base64 alphabet (plain `data.base64EncodedString()`) instead of `urlSafeBase64Encode()`.
- This helper is shared by `login()`, `submitAuthorizationRequest()` (used by the OTP/registration-completion flows), and `submitPasswordlessAuthorizationRequest()` (OTP/passwordless login), so all three call sites are fixed by the one change.

## Scope note

This does not touch `urlSafeBase64Encode()`'s other callers in the same file — `generateCodeVerifier()` and `generateChallenge()` — which correctly need URL-safe Base64 per RFC 7636 for PKCE `codeVerifier`/`codeChallenge` values and are unrelated to this HTTP Basic-Auth header.

## Test plan

- [x] Added a regression test covering `login()`, using a username/password pair whose combined UTF-8 bytes diverge between the standard and URL-safe alphabets (contains `/` and requires `=` padding under the standard alphabet); asserts the `Authorization` header matches standard Base64 output.
- [x] Added a regression test covering `submitPasswordlessAuthorizationRequest()` (a call site sharing the same helper), same divergent credential pair.
- [x] Ran `NativeLoginManagerTests` on the iOS simulator — both new tests pass. (Two pre-existing, unrelated failures in this suite — `testShouldShowBackButton`/`testShouldShowBackButtonWithBioAuth` — are keychain-entitlement issues in this local sandboxed environment, reproduced identically on unmodified `dev`.)
- [x] Found via a real reproduction during release validation: fresh self-registration on iOS 26 (Native Login template) followed by an immediate logout and login attempt with the identical, just-set password failed; the same password worked correctly via Salesforce web login for the same user, isolating the defect to this SDK's Basic-Auth header encoding.

### DIFF
--- a/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLogin/NativeLoginManagerInternal.swift
+++ b/libs/SalesforceSDKCore/SalesforceSDKCore/Classes/Login/NativeLogin/NativeLoginManagerInternal.swift
@@ -768,8 +768,10 @@ public class NativeLoginManagerInternal: NSObject, NativeLoginManager {
         guard let valuesUtf8EncodedData = "\(value1):\(value2)".data(
             using: .utf8
         ) else { return nil }
-        
-        return urlSafeBase64Encode(data: valuesUtf8EncodedData)
+
+        // HTTP Basic Authentication (RFC 7617) requires the standard base64 alphabet with
+        // padding, not the URL-safe alphabet used elsewhere in this file for PKCE values.
+        return valuesUtf8EncodedData.base64EncodedString()
     }
     
     /// Generates either the reCAPTCHA token parameter for non-enterprise reCAPTCHA configurations or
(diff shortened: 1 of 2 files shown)


JohnsonEricAtSalesforce:
Superseded by https://github.com/forcedotcom/SalesforceMobileSDK-iOS/pull/4161 (recreated from a correctly-named branch).

PR #4166, opened 2026-09-02
fix(AuthFlowTesterUITests): remove beacon RTR workaround (W-24028119)

wmathurin:
## Summary

Removes the `beacon_` clause from `expectsRefreshTokenRotation` in `UITestConfigUtils.swift`.

The server bug (W-23971480) that caused beacon apps to unconditionally rotate the refresh token on every refresh has been fixed in sdb38. The workaround and its comment are no longer needed — `beacon_opaque` and `beacon_jwt` tests should now pass without the special-case RTR expectation.

## Test plan

- [x] `BeaconLoginTests.testBeaconOpaque_DefaultScopes` — was failing with RTR mismatch, now passes
- [ ] Full beacon test suite on a connected org

### DIFF
--- a/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/UITestConfigUtils.swift
+++ b/native/SampleApps/AuthFlowTester/AuthFlowTesterUITests/Util/UITestConfigUtils.swift
@@ -139,10 +139,9 @@ struct AppConfig: Codable {
     }
 
     /// Whether a normal refresh is expected to rotate this configuration's refresh token.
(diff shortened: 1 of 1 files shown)


wmathurin:
The `AdvancedAuthBeaconLoginTests` failures ("Refresh token should not have changed (non-RTR app)") were caused by a stale auto-installed ECA for the beacon apps on `advauthflowtestingmsdksdb38.test1.my.pc-rnd.salesforce.com` — the server-side W-23971480 fix was deployed there but the old ECA was still in place, causing RTR to fire. The forgotten ECA has now been deleted; re-running the affected tests.

brandonpage (approved):

PR #4168, opened 2026-09-03
Fix REST retry completion race

wmathurin:
## Context

PR #4001 added a stale-task guard so a response from an attempt superseded by authentication replay would be ignored. The guard compared the callback task with `SFRestRequest.sessionDataTask`, but the comparison was outside the synchronization domain used by refresh replay.

That left this check-then-replay interleaving possible:

1. Task 1 completes and passes the stale-task check.
2. Before Task 1 invokes its terminal callback, token refresh replay sees the request as active and installs Task 2.
3. Task 1 invokes the original success or failure block.
4. Task 2 completes and invokes the same block again.

For the Swift async API, those two Objective-C callbacks attempt to resume the same checked continuation twice.

This PR also closes the admission-side interleaving: `send:` adds a new request to `activeRequests` before `enqueueRequest:` publishes its first `sessionDataTask`. A refresh replay snapshot can therefore observe a newly admitted request with no task yet. Replay must not create that request's first attempt while its original sender is still preparing the same request.

## Relationship to refresh-token rotation work

The stale-task guard was sufficient for the ordering covered by the original regression test: replay fully installs Task 2 before Task 1 enters its completion handler. It was not a general synchronization guarantee because Task 1 could already be between its identity check and terminal callback.

The refresh-token-rotation marker work in #4092 did not touch `SFRestAPI` or its callback scheduling. The rotation-motivated refresh coordinator in #4087 did change `SFRestAPI` to coalesce refreshes across SDK consumers, but its final implementation kept refresh completion callbacks on the main queue, matching the prior `SFOAuthSessionRefresher` behavior. REST completions were already delivered on the independent URL-session delegate queue.

The missing atomicity therefore predates the rotation work. Coordinator-driven coalescing and larger replay batches may make the latent timing window easier to encounter in production, but they did not create the underlying race.

## Timing summary

| When request B arrives relative to request A's refresh replay | What replay observes for B | Before this PR | After this PR |
|---|---|---|---|
| After replay has taken its snapshot or finished | B is absent from the snapshot | B's original send publishes one task | Same: B's original send publishes one task |
| Before the snapshot, with B's first task already published | B is active with a non-nil current task | Replay could replace B, but B's old completion could pass the stale check and still call back | Replay and terminal completion use one ownership lock; only the winning task can call back |
| Exactly across admission and publication | B is active but `sessionDataTask` is nil | Replay could create a task while B's original sender was still preparing, allowing two independently created attempts | Replay skips B; the original sender remains the sole owner of initial task publication |
| While B's published task is finishing | B is active with a current task whose completion is in progress | The check-then-callback window could allow both the old and replayed attempts to call back | Terminal claim and successor installation serialize; exactly one path wins |
| After B recognizes a DPoP nonce challenge but before it claims retry ownership | Authentication replay can replace B's task during the gap | The stale DPoP path could independently publish another successor | The DPoP path rechecks under the ownership lock, loses as stale, and cannot publish or mutate retry state |

## What changed

- Add a private current-task operation protected by the existing synchronized `SFRestAPI` state lock.
- Keep the early stale-task check to avoid unnecessary response parsing, then revalidate ownership immediately before every terminal callback.
- Make terminal completion an atomic state transition that:
  - confirms the callback belongs to the current data task;
  - clears `SFRestRequest.sessionDataTask`;
  - removes the request from `activeRequests`; and
  - copies the success or failure block for delivery.
- Invoke copied application callbacks only after releasing the state lock, allowing safe re-entry into `SFRestAPI`.
- Serialize authentication replay and DPoP nonce replay with the same current-task ownership check.
- Create/resume and publish successor tasks while holding the state lock. `SFNetwork` resumes a task before returning it, so this prevents an exceptionally fast completion from observing task ownership before publication.
- Skip active replay entries whose `sessionDataTask` is nil. Admission precedes initial task publication, so only the original sender may create the first attempt; replay may replace only an existing attempt.
- Document the required `SFNetwork` invariant: `sendRequest` must remain nonblocking and must invoke completion asynchronously.
- Update cleanup and refresh-failure flushing to invalidate and drain requests under the lock, then cancel tasks and invoke failure blocks outside the lock.
- Use consistent parse-then-terminal-claim ordering for success, HTTP failure, and network-error paths.
- Add a self-contained iOS token-lifecycle guide covering code exchange, refresh coordination, RTR, DPoP nonce handling, REST replay, request-attempt ownership, and the admission guard.

Authentication and DPoP retry responses remain nonterminal and keep the request active for their successor task. Public API signatures and retry policy are unchanged.

## Behavior note

Terminal requests are now removed from `activeRequests` before application success, failure, or delegate code is invoked; previously normal removal happened after application code returned.

This is an intentional callback-timing change on a public-facing behavior. It is required to make terminal delivery and replay eligibility one atomic transition. It also fixes a latent request-reuse issue: if application code sends the same `SFRestRequest` again from its callback, cleanup from the previous attempt can no longer remove the newly active attempt.

## Regression tests

The completion-race tests use a test-only `SFRestAPI` subclass to pause Task 1 during response preparation, after its early stale check but before terminal delivery. While Task 1 is paused, the test concurrently replays active requests and verifies Task 2 is installed. Task 1 is then released and Task 2 is completed.

The admission-race test uses a test-only `SFRestRequest` subclass to pause request B after it has entered `activeRequests` but before it publishes its first task. Replay takes its snapshot during that pause. The test verifies replay creates no task, releases the original sender, then verifies exactly one task and one terminal callback.

The DPoP race test pauses Task 1 after recognizing `use_dpop_nonce` but immediately before its retry ownership claim. Authentication replay installs Task 2 during the pause. When released, Task 1 must fail its ownership recheck without setting `dpopNonceRetried` or installing Task 3; Task 2 then delivers the only terminal callback.

The suite now covers:

- A new background-thread submission exactly concurrent with the replay snapshot.
- HTTP 200 terminal success racing replay.
- HTTP 500 terminal failure racing replay.
- Network-error terminal completion racing replay.
- A DPoP nonce retry claim racing authentication replay.
- A sequential DPoP nonce challenge installing and completing a successor task.
- A completion with neither response nor error.
- A response arriving after the owning `SFRestAPI` has been deallocated.
- Existing sequential replay, refresh-failure, cancellation, and cleanup cases.

The replay barriers invoke `resendActiveRequestsRequiringAuthentication` directly. This deterministically exercises the same ownership transition used by successful refresh completion without requiring a live token exchange.

## Test boundary

Barrier-forced concurrent races cover new-request admission during the replay snapshot; Task 1 terminal completion during replay for HTTP 200, HTTP 500, and network-error paths; and a DPoP nonce retry claim racing authentication replay. The no-response/no-error fallback and post-deallocation callback paths are covered by focused tests without a concurrent barrier.

The replay barriers deliberately call the successful-refresh replay transition directly rather than performing a live token exchange. Token refresh coordination and token-endpoint DPoP behavior remain covered by their dedicated suites.

## Coverage follow-up

The initial Codecov report identified 18 uncovered changed lines: the deallocated-API guard, stale network-error return, no-response fallback, and DPoP nonce-retry branch. Four focused tests exercise all of those groups. The admission and DPoP race tests additionally force the nil-task replay guard and the competing-retry ownership recheck.

A local coverage-enabled run confirmed that every one of the original 18 missed production lines executes at least once. CI will report coverage for the two additional race guards at the current head.

## Documentation

- `docs/auth/README.md` indexes the iOS authentication implementation documentation.
- `docs/auth/token-lifecycle.md` is self-contained for contributors to this public repository; it does not depend on internal Workspace documentation.
- The guide explains the separate responsibilities of the per-credential refresh coordinator and the per-request `SFRestAPI` ownership lock.
- The REST lifecycle explicitly documents that replay replaces published attempts but skips a newly admitted request until its original sender publishes the first task.
- The DPoP lifecycle explains how a nonce retry and authentication replay remain one current-task chain regardless of which ownership claim wins.

## Validation

- Focused `SFRestAPIDataTaskRaceTests` run after the DPoP race follow-up: 15 passed, 0 failed, 0 skipped.
- Fresh complete `SalesforceSDKCore` suite after the DPoP race follow-up: 1,034 passed, 0 failed, 0 skipped.
- Test environment: iPhone 16 simulator, iOS 18.6.
- Complete-suite coverage was disabled to avoid the previously observed Xcode coverage-finalization hang.
- Confirmed the diff introduces no public header or API changes.
- Documentation and source changes pass `git diff --check`.

The initial full-suite attempt completed test execution but Xcode stalled while merging coverage data and finalizing its result bundle. Complete-suite runs with code coverage disabled avoid that Xcode tooling issue and complete normally.

### DIFF
--- a/docs/auth/README.md
+++ b/docs/auth/README.md
@@ -1,0 +1,10 @@
+# Authentication engineering documentation
+
+This directory contains implementation guides for authentication behavior in the iOS Mobile
+SDK.
+
+## Documents
+
+| Document | Purpose |
+|---|---|
+| [`token-lifecycle.md`](token-lifecycle.md) | Code exchange, token refresh, DPoP nonce handling, Refresh Token Rotation, and REST replay concurrency. |
--- a/docs/auth/token-lifecycle.md
+++ b/docs/auth/token-lifecycle.md
@@ -332,6 +332,12 @@ task 1 -> HTTP 400 use_dpop_nonce + DPoP-Nonce
   -> any later task-1 completion is stale and ignored
 ```
 
+If authentication replay installs a successor after task 1 recognizes the nonce challenge but
+before it claims retry ownership, task 1 fails its ownership recheck and does not install another
+successor. Conversely, if the DPoP retry claim wins the lock first, authentication replay can only
+replace the successor it published. The two retry mechanisms therefore preserve one current-task
+chain rather than creating independent successors.
+
 ### 9.4 Completion racing authentication replay
 
 ```text
--- a/docs/auth/token-lifecycle.md
+++ b/docs/auth/token-lifecycle.md
@@ -205,8 +205,9 @@ When an attempt receives an authentication failure accepted by the retry policy:
 2. Under `@synchronized(self)`, `replayRequest` elects one refresh cycle for the API instance.
 3. The API asks the shared `SFSDKTokenRefreshCoordinator` to refresh the credential.
 4. Other failed requests remain in `activeRequests`; they do not start another refresh cycle.
-5. On refresh success, `resendActiveRequestsRequiringAuthentication` snapshots the active set and
-   installs a new task for each pending request with automatic auth retry disabled.
+5. On refresh success, `resendActiveRequestsRequiringAuthentication` snapshots the active set. It
+   skips any newly admitted request whose initial task has not been published yet, and installs a
+   successor for each request that already owns a task, with automatic auth retry disabled.
 6. Each old task is cancelled after its successor is installed.
 7. On refresh failure, the active set is atomically drained and each request receives one failure.
 
@@ -247,6 +248,12 @@ replace the task between parsing and delivery.
 The successor is created and assigned to `request.sessionDataTask` while holding the state lock.
 `SFNetwork.sendRequest` resumes its task before returning, so the lock prevents an exceptionally
 fast asynchronous completion from observing the request before task ownership is published.
+
+Admission to `activeRequests` intentionally happens before initial request preparation, so a newly
+submitted request can appear in a refresh replay snapshot while `sessionDataTask` is still nil.
+Replay skips that entry: only the original sender may publish the first task, while replay may only
+replace an existing task. This prevents replay and the original sender from independently creating
+two initial attempts for the same request.
 
 **Invariant:** `SFNetwork.sendRequest` must remain nonblocking and must not invoke its completion
 synchronously. Changing that contract could introduce a lock-ordering or reentrancy hazard.
@@ -339,6 +346,20 @@ Whichever acquires the ownership lock first wins:
 
 The request receives one terminal result in either ordering.
 
+### 9.5 New request racing the replay snapshot
+
+```text
+new send adds request B to activeRequests
+request B pauses before publishing its first task
+refresh replay snapshots request B and sees sessionDataTask == nil
+replay skips request B
+the original send publishes B's first task
+B's sole task claims and delivers its terminal callback
+```
+
+This is the admission-side counterpart to the completion race. Replay owns replacement of a
+published attempt; it never creates the initial attempt on behalf of an in-progress sender.
+
 ---
 
 ## 10. Concurrency Invariants
@@ -353,6 +374,8 @@ The request receives one terminal result in either ordering.
 7. Application and delegate callbacks execute outside the request-state lock.
 8. Cleanup invalidates request ownership before cancellation or client notification.
 9. A replayed REST request cannot automatically start a second auth refresh cycle.
+10. Refresh replay skips an active request with no published task; only its original sender may
+    publish the first attempt.
 
 The focused regression coverage lives in `SFRestAPIDataTaskRaceTests`, with coordinator behavior
 covered by `SFSDKTokenRefreshCoordinatorTests` and token-endpoint nonce handling covered by the
--- a/docs/auth/token-lifecycle.md
+++ b/docs/auth/token-lifecycle.md
@@ -139,19 +139,26 @@ look like a rotation.
 ```text
 applyAuthHeaders(request, scope, accessToken, tokenType):
   if accessToken is empty -> leave request unchanged
-  set Authorization to "DPoP <token>" when tokenType is DPoP,
-                       or "Bearer <token>" otherwise
-  if shouldAttachDPoP(scope, tokenType):
+  if isDPoPTokenType(tokenType):
+    set Authorization to "DPoP <token>"
     load or create the credential-scoped EC P-256 keypair
     canonicalize the request URL for htu
     read a cached nonce for the URL and credential scope
     build a fresh proof with htm, htu, iat, jti, nonce, and ath
     set the DPoP header
+  else:
+    set Authorization to "Bearer <token>"
 ```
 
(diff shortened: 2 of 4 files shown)


JohnsonEricAtSalesforce (approved):
👍🏻

brandonpage (approved):

JohnsonEricAtSalesforce:
No inline comments. The core fix — atomically claiming terminal ownership under `@synchronized(self)` before running client callbacks — held up against an adversarial trace; I couldn't construct a two-thread sequence that breaks it.

**Confirmed against the diff:**
- `performIfCurrentDataTask:forRequest:action:` gates every terminal path and both replay paths (auth-refresh, DPoP nonce). Claim, clear `sessionDataTask`, remove from `activeRequests`, and copy the block all happen under the lock; the callback runs after it releases.
- Task publication (`request.sessionDataTask = dataTask`) now happens inside that same lock, closing the fast-completion-before-publish window.
- Removal from `activeRequests` moved earlier (into the atomic claim), so a request resent synchronously from inside its own callback gets a clean slate rather than having its new attempt wiped by the old one's cleanup.
- `cleanup`/`flushPendingRequestQueue` collect tasks/blocks under the lock and invoke outside it. A cancelled task's own completion still re-enters `performIfCurrentDataTask:`, sees itself as stale, and is dropped — no double-failure path.
- DPoP-retry and replay branches now call `enqueueRequest:`/`replayRequest:` from inside the lock, widening the critical section — but everything that runs there (`SFNetwork.sendRequest:`, the refresh coordinator's kickoff) is non-blocking, so no deadlock risk.

**Escalation note:** per this repo's CLAUDE.md, any change to auth-replay/DPoP paths requires human review regardless of diff quality — flagging per that policy, though I found no credential-safety issue.

**Tests:** the 4 new barrier-forced race tests pause exactly at the real race window (between the early stale-check and the terminal claim) and assert single-delivery, not just "no crash." No test-correctness issues found.

**Test boundary already disclosed in the PR body:** the DPoP retry test is sequential, not barrier-forced, even though DPoP retry now shares the same ownership-claim helper as the raced paths. Worth a barrier-forced DPoP test as a follow-up? Non-blocking question.

**CI:** both red checks (`NativeLoginManagerTests.testPasswordValidation`, `LegacyLoginTests.testCAOpaque_DefaultScopes_WebServerFlow`) are unrelated — all 13 `SFRestAPIDataTaskRaceTests` passed in the same run, consistent with known dev-wide flakiness on those two suites.

**Static analysis:** the 2 Clang nullability warnings are on pre-existing lines this diff doesn't touch (confirmed via blame).

wmathurin:
@JohnsonEricAtSalesforce Thanks for calling out the remaining DPoP test boundary — addressed in c136f7bda.

The new `testDPoPNonceRetryClaimRacingAuthenticationReplayDoesNotInstallSecondSuccessor` pauses Task 1 after it recognizes `use_dpop_nonce` but immediately before its retry ownership claim. Authentication replay installs Task 2 during that pause; when Task 1 resumes, the test verifies that its stale claim cannot set `dpopNonceRetried` or install Task 3, and that Task 2 delivers the sole terminal callback.

I also updated the PR description and token-lifecycle guide to document this interleaving. Local validation after the follow-up: 15/15 focused race tests and 1,034/1,034 SalesforceSDKCore tests passed.

PR #4170, opened 2026-09-09
Prevent REST task publication after cleanup

wmathurin:
## Problem

W-24142318 reports the same checked-continuation double-resume crash addressed by #4168, but during a multi-user logout/login flow. The prior fix made retry and terminal completion atomic once a URL session task existed. It did not cover a request that had been admitted but had not published its initial task yet.

The remaining race was:

1. `send:` stores the completion blocks and adds the request to `activeRequests`.
2. `enqueueRequest:` begins preparing the URL request before creating its data task.
3. Logout cleanup runs in that gap, removes the request, and invokes its failure block with the logout error. This is the first continuation resume.
4. Request preparation, or a delayed `loginWithCompletion:` callback, finishes after cleanup.
5. The old code unconditionally creates and publishes a data task.
6. That task later fails and invokes the same failure block again, producing the fatal second continuation resume.

The multi-user flow makes this more likely because logout clears the affected users credentials before posting the logout notification. A concurrent request can therefore wait for login completion while cleanup drains its already-admitted request.

## Change

Immediately before selecting a network and creating/publishing the data task, `enqueueRequest:` now verifies under the existing `SFRestAPI` state lock that the request is still present in `activeRequests`.

If cleanup or refresh-failure handling already removed the request, enqueue returns without creating a task. Network selection remains inside the same critical section so a rejected custom-host enqueue cannot allocate an unused network instance.

This gives the race two safe outcomes:

- **Cleanup wins:** cleanup removes the request and delivers one failure. The delayed enqueue observes that the request is inactive and stops.
- **Enqueue wins:** the task is published while holding the lock. Cleanup then invalidates and cancels that task before delivering one failure.

The change is internal and does not modify public APIs, retry policy, or authentication behavior.

## Regression test

The new test deterministically pauses initial request preparation after admission but before task publication, then runs cleanup and releases preparation.

- **Before the fix:** the paused sender publishes a task after cleanup; delivering HTTP 500 calls the cleanup failure block a second time, and XCTest reports multiple fulfillment of the same expectation.
- **After the fix:** enqueue creates no task, the request remains retired, and cleanup is the only terminal callback.

The test comment also documents that delayed login completion reaches the same enqueue/membership gate.

## Validation

- New focused regression: **1 passed, 0 failed**
- `SFRestAPIDataTaskRaceTests`: **16 passed, 0 failed**
- CI `SalesforceSDKCore` test job: **passed in 12m45s**
- Static analysis: **passed**
- SalesforceSDKCore Codecov project and patch checks: **passed**

A local full-suite attempt reached 777 tests: 757 passed and 20 credential-dependent tests failed authentication setup because the local access/refresh tokens had expired. Those failures did not reproduce in SalesforceSDKCore CI. The AuthFlowTester UI check is excluded from readiness per work-item direction.


wmathurin:
Addressed feedback [2] in 48982e7db: the regression test now explicitly documents that a delayed `loginWithCompletion:` callback enters the same `enqueueRequest:` path after cleanup removes active membership. The focused regression was rerun and passed.

No production change for [3]; keeping network selection inside the lock intentionally prevents a rejected custom-host enqueue from allocating/leaking a unique network instance and keeps selection plus publication serialized.

sfdctaka (approved):
LGTM!

PR #4171, opened 2026-09-09
Fix RTR user agent after app restart

wmathurin:
## Summary
- Bind token refresh requests to the credential owner’s explicit per-user User-Agent instead of ambient current-user state.
- Add deterministic coverage for a different current user, explicit token-request header construction/fallback, and DPoP nonce-retry User-Agent preservation.
- Add an AuthFlowTester cold-restart regression that captures the actual `/services/oauth2/token` User-Agent and requires `RT`, `A2`, and `OT`.
- Document the first-post-restart refresh invariant.

## Why no timestamp reconciliation
No persisted-state repair is needed. Warm refreshes later in the original process already carry `RT`, and the post-restart regression passes once refresh resolves the credential owner explicitly. The failing request also omitted other per-user markers, confirming request-context loss rather than `RT` persistence loss.

## Validation
- Pre-fix focused regression failed as expected: the endpoint request had no explicit UA and omitted the owner’s RT/test markers.
- Focused SalesforceSDKCore tests: 10 passed, 0 failed, 0 skipped.
- DPoP nonce-retry User-Agent regression: 1 passed, 0 failed; full `SFSDKOAuthTokenEndpointResponseTests`: 5 passed, 0 failed, 0 skipped.
- AuthFlowTester build-for-testing passed.
- Focused live AuthFlowTester UI regression: 1 passed, 0 failed (103.817s).
- `git diff --check` passes.

## Workspace spec
- https://git.soma.salesforce.com/SalesforceMobileSDK/SalesforceMobileSDK-Workspace/pull/99

## Work item
- W-24141870


sfdctaka (approved):
LGTM!

PR #4172, opened 2026-09-11
Avoid REST lock during task creation

wmathurin:
## Context

PR #4170 fixed the cleanup/admission race by validating active membership at task publication. That publication gate is correct, but its implementation also moved `networkForRequest:` and `SFNetwork.sendRequest` under `@synchronized(self)`.

`SFNetwork.sendRequest` synchronously evaluates `SalesforceSDKManager.userAgentString`, which is public/application-overridable, and first-use SDK initialization can wait for main-thread WebKit work. This creates a possible lock inversion: a background REST send owns the per-`SFRestAPI` monitor while waiting for main/application work, while main-thread logout or screen-lock cleanup waits for that REST monitor.

Because Objective-C `@synchronized` is recursive, moving only the visible initial-send call was insufficient. The authentication replay and DPoP nonce retry paths could recursively enqueue a replacement while an outer ownership action still held the same monitor.

### 1. Split task creation from task start

- Add the non-public `SFNetwork+Internal.h` contract with `dataTaskForRequest:dataResponseBlock:` and `resumeDataTask:`.
- The first method creates and returns a suspended `NSURLSessionDataTask`; the second starts it.
- Public `sendRequest:dataResponseBlock:` preserves its existing create-and-resume behavior.
- Public callers may legally supply an immutable `NSURLRequest`, so `sendRequest:` mutable-copies it before adding the Mobile SDK user-agent header. `SFRestAPI` likewise mutable-copies the result of `prepareRequestForSend:` before header mutation.

### 2. Make enqueue a three-phase publication protocol

`enqueueRequest:shouldRetry:` now:

1. prepares the request, selects the network, evaluates the user agent, and creates a suspended task **outside** the REST monitor;
2. briefly acquires the monitor to verify `activeRequests` membership and publish `request.sessionDataTask`;
3. releases the monitor, then resumes the published task or cancels an unpublished task retired by cleanup.

The large indentation change in this method is the existing URL-session completion body moving out of the synchronized block along with task construction. Its success/error/timeout/status parsing and terminal callback behavior are not being redesigned.

### 3. Reserve retries under the monitor; build them outside it

- DPoP nonce retry atomically verifies current-task ownership, clears the old task, and marks the one retry used. Nonce harvesting and successor enqueue happen after unlocking.
- Authentication replay snapshots eligible active requests and clears each old task under the monitor. Old-task cancellation and successor enqueue happen after unlocking.
- Replay calls `enqueueRequest:` directly instead of `send:` so cleanup cannot be undone by re-admitting a retired request.
- On refresh success, replay reservation and clearing `refreshCycleActive` are one atomic state transition. A request admitted after the snapshot can therefore start a new refresh if it receives a 401 while replay tasks are being built.
- Stale-task logging, token refresh, logging, task cancellation, and task resume all run outside the REST monitor.

## Resulting ordering guarantees

- If cleanup wins before publication, the suspended task is discarded and never started.
- If publication wins first, cleanup clears ownership and cancels the task; a later cancellation callback is stale.
- Initial send, 401 replay, and DPoP replay cannot perform network/user-agent/application work while holding the REST monitor.
- A retry owns exactly one reserved successor while it is constructed outside the monitor.
- Client and delegate callbacks remain outside the monitor.
- No public API is added or changed.

## Log evidence

The supplied app log is consistent with this lock-order risk: login finishes, a background thread starts connected-app-settings retrieval, main begins the screen-lock flow, and logging stops at the screen-lock notification. It does **not** contain a crash exception, watchdog termination, hang sample, `SFRestAPI` stack frames, or lock ownership, so it supports the timing hypothesis but does not prove the deadlock.

## Tests

- 20/20 `SFRestAPIDataTaskRaceTests` passed, including deterministic initial-send, authentication-replay, and DPoP-replay lock-boundary tests plus refresh-cycle snapshot atomicity.
- 4/4 `SFNetworkTests` passed, including immutable-request user-agent coverage.
- 3/3 `SFRestAPIReplayRequestTests` passed.
- Final affected run: **27 passed, 0 failed** on iPhone 17 Pro / iOS 27.0 after rebasing onto current `upstream/dev`.
- A broad self-contained Core run executed 892 tests: 891 passed; one unrelated account-manager fixture exception passed on isolated rerun. The unfiltered local suite remains blocked by expired integration-test credentials, so fresh PR CI is required for the credential-backed classes.

## Documentation

Updates `docs/auth/token-lifecycle.md` with the suspended-task publication protocol, retry reservation rules, refresh-cycle transition, and lock-boundary invariants.

Follow-up to #4170 / W-24142318.

---

## Follow-up cleanup (dead code + docs)

After review, removed a production-dead path and documented the logout case:

- **Removed `replayRequest:response:`** from `SFRestAPI`. It had no production call site and duplicated the live retry path's `refreshCycleActive` election — but *without* the task-ownership recheck the real path performs, making it a divergent, misleading shim. The live retry path in `enqueueRequest:` calls `startAuthenticationRefreshForRequest:response:` directly.
- **Fixed the log selector** in `startAuthenticationRefreshForRequest:` to `NSStringFromSelector(_cmd)` so its log lines name the method that actually runs (previously hardcoded to the deleted selector). No behavior change.
- **Retargeted the 3 `SFRestAPIReplayRequestTests`** to invoke `startAuthenticationRefreshForRequest:response:` (the real error-block path they assert) and dropped the stale test-category declaration in `SFRestAPIDataTaskRaceTests`.
- **Added `token-lifecycle.md` §9.6 "Logout during an active refresh cycle"** plus invariant #12: cleanup drains `activeRequests` and clears `refreshCycleActive` under the lock, so a later refresh success (`resendActiveRequestsRequiringAuthentication`) or error (`flushPendingRequestQueue`) branch operates on an empty snapshot and delivers no second callback — and neither path can deadlock, since both take the lock only to read/drain and invoke client code after release.

Re-ran the affected suites after this change: **27 passed, 0 failed** (`SFRestAPIDataTaskRaceTests` 20/20, `SFRestAPIReplayRequestTests` 3/3, `SFNetworkTests` 4/4) on iPhone 16 Pro / iOS 18.6.

---

## Follow-up 2 — Residual multi-user logout/login deadlock (class monitor)

A client that cherry-picked this PR still reproduced a freeze: several users logged in, REST requests firing right after login, one user logged out while a login was completing. The app hung; a cold restart returned **already logged in** (no re-login). That "restart, no re-login" is diagnostic — `applyCredentials:` persists the refreshed credentials and current-user identity *before* any blocking, so login completed and the *running* app then deadlocked. This PR's per-instance `@synchronized(self)` hardening does not touch the lock involved, which is why the cherry-pick didn't help.

### Root cause — a distinct ABBA on the SFRestAPI *class* monitor

- **Edge 1 (class monitor → `_accountsLock`).** `+sharedInstanceWithUser:` / `+removeSharedInstanceWithUser:` resolved `currentUser` *inside* `@synchronized([SFRestAPI class])`. The `currentUser` getter takes `SFUserAccountManager`'s `_accountsLock` when `_currentUser` is nil — exactly the logout window, since `setCurrentUserInternal:nil` nils it.
- **Edge 2 (`_accountsLock` → class monitor).** `setCurrentUserInternal:` fires the `currentUser` KVO synchronously while holding `_accountsLock`; an app-registered `currentUser` observer that re-enters `SFRestAPI` (e.g. `+sharedInstance` / `RestClient.shared`) then wants the class monitor.

```text
Thread B (bg, request right after login)     Thread M (main, setCurrentUserInternal:)
  holds  @synchronized([SFRestAPI class])       holds  _accountsLock
  wants  _accountsLock  <──────────────────────► fires currentUser KVO
                                                   -> app observer re-enters SFRestAPI
                                                   wants @synchronized([SFRestAPI class])
```

The SDK ships both edges but the closing hop is **app-supplied** (no SDK-internal `currentUser` observer re-enters `SFRestAPI`, and `notifyUserChange` posts outside `_accountsLock`), which is why only some apps reproduce it.

### Precise change

- Resolve `currentUser` **before** entering `@synchronized([SFRestAPI class])` in `+sharedInstanceWithUser:` and `+removeSharedInstanceWithUser:`. The class monitor then never nests `_accountsLock`, so the cycle cannot close regardless of what an app `currentUser` observer does. The monitor still serializes the shared-instance registry; only the current-user read is hoisted out.
- No public API is added or changed. One new **internal-only** hook, `SFRestAPI.currentUserResolutionHookForTesting` in `SFRestAPI+Internal.h`, exists solely so the regression can deterministically reproduce the interleave; it is always nil in application code.

### Test

- New `SFRestAPIDataTaskRaceTests.test_givenLogoutWindow_whenCurrentUserKVOObserverReentersClassMonitor_thenSharedInstanceDoesNotDeadlock`: parks a thread at the class-monitor/`currentUser` seam and drives a `currentUser` KVO observer that re-enters the class monitor while `_accountsLock` is held. Passes with the fix (~0.23s); **deadlocks / times out** if the resolution is moved back inside the class monitor — verified in both directions.
- Focused suites: **`SFRestAPIDataTaskRaceTests` 21/21**, `SFNetworkTests` 4/4, `SFRestAPIReplayRequestTests` 3/3. Credential-backed account-manager suites require CI (expired local integration credentials, as in prior phases).

### Documentation

Adds `token-lifecycle.md` **§8 "Shared-Instance Acquisition and Lock Ordering"** and **invariant #13**. Trailing sections renumbered (Cleanup §9, Scenarios §10 incl. §10.6 "Logout during an active refresh cycle", Invariants §11).

Follow-up to #4170 / W-24142318.


sfdctaka (approved):
LGTM!

PR #4174, opened 2026-09-15
Fix safe collection write starvation under concurrent REST requests

wmathurin:
## Summary

- make every SFSDKSafeMutableSet and SFSDKSafeMutableDictionary mutation use a synchronous barrier so a write is published before its method returns
- preserve concurrent reader execution while serializing writes against those readers
- enumerate SFSDKSafeMutableSet from a stable snapshot and invoke the caller block outside the collection queue
- add deterministic blocking-hash publication tests and an enumeration-reentrancy test

## Why

AuthFlowTester batches of 20 async REST requests exposed two instances of the same cooperative-executor starvation shape.

### Trace 1: SFRestAPI active requests

1. Swift tasks enter RestClient.send(request:) through withCheckedThrowingContinuation and synchronously reach SFRestAPI request admission.
2. SFRestAPI calls addObject: on activeRequests. The prior implementation submitted a dispatch_barrier_async mutation and returned before the request was in the set.
3. Admission later calls containsObject: synchronously before assigning and resuming the URLSession task.
4. Under a burst, cooperative-executor threads can block in synchronous reads behind queued asynchronous barrier writes. If those barriers do not receive an execution thread, no request reaches resume and no continuation completes.

### Trace 2: SFNetwork shared instances

1. During the same enqueue path, networkForRequest: selects a shared ephemeral, background, or custom SFNetwork instance.
2. sharedInstanceWithIdentifier:sessionConfiguration: synchronously reads sharedInstances for the identifier.
3. On a miss, it creates the network and the prior subscript assignment only queued an asynchronous barrier before returning.
4. Concurrent request tasks can then synchronously read the registry behind that queued write, creating the same queued-write/read dependency.

This change does not make compound lookup-and-create sequences atomic. It makes each mutator synchronously publish its result and removes the return-before-write behavior that enabled both starvation cycles.

## Production call-site audit and blast radius

SFSDKSafeMutableSet has one production consumer:

- SFRestAPI.activeRequests: request admission, pre-resume membership checks, terminal removal, cancellation enumeration, and cleanup/flush snapshots.

SFSDKSafeMutableDictionary production consumers are:

- SFLogger.loggerList: low-cardinality per-component logger memoization, additionally guarded by synchronization on SFLogger.
- SalesforceSDKManager.snapshotViewControllers and nativeLoginViewControllers: per-scene UI controller registries used during scene and authentication lifecycle events.
- KeyValueEncryptedFileStore.globalStores and userStores, including per-user maps: caches shared store instances by store and user name; it is not the per-record data-write path.
- SFUserAccountManager.authSessions: per-scene login/auth-session registry.
- SFNetwork.sharedInstances: cache of ephemeral, background, and custom network/session instances.
- SFRestAPI.sfRestApiList: low-cardinality per-user/scope REST client registry; shared lookup/create is also guarded by synchronization on SFRestAPI.

The change intentionally covers all public set and dictionary mutators so their publication contract is consistent. No in-tree production calls were found for the set bulk mutators or dictionary addEntries:/setDictionary:. removeAllObjects is limited to cache reset and teardown paths. The audited consumers are low-cardinality registries, lifecycle state, or request tracking; no in-tree high-volume bulk-mutation hot path was identified. SFSDKSafeMutableArray is deliberately out of scope.

## API behavior and compatibility

There are no source or binary API signature changes. Mutations now provide the stronger conventional guarantee that the change is visible when the method returns. The tradeoff is that a caller waits for prior readers and the mutation itself instead of submitting fire-and-forget work. In particular, a mutation invoked on the main thread may now wait for readers already executing on the collection queue. The concurrent queue is retained so independent reads still execute in parallel.

Set enumeration uses a stable snapshot. The caller block runs outside the private synchronization queue, so it can mutate the set without self-deadlocking and arbitrary caller code does not hold the queue. Mutations during enumeration do not change that snapshot and may interleave with callback execution.

## Validation

Focused SalesforceSDKCommon tests on iPhone 17 Pro / iOS 26.1 simulator:

- SFSDKSafeMutableDictionaryTests: 2 passed
- SFSDKSafeMutableSetTests: 8 passed
- Total: 10 passed, 0 failed, 0 skipped

Result bundle: /private/tmp/fix-safe-collection-write-starvation-review-2.xcresult

The blocking-hash tests pause inside the backing collection mutation and prove addObject: and setObject:forKey: cannot return before publication. The enumeration test deterministically mutates the set from its callback and verifies completion without sleeps or deadlock.


sfdctaka (approved):
LGTM!

PR #4175, opened 2026-09-17
W-24224590: Select request authentication by path

wmathurin:
## Summary

- Let `SalesforceSDKManager` synchronously choose `ui_sid` Bearer authentication by request path when a DPoP access token is present.
- Default the policy to select no paths; apps opt specific paths into `ui_sid` Bearer, and all other requests keep DPoP.
- Centralize the decision in REST authentication handling and add coverage for policy, fallback, nonce, and race behavior.
- Update DPoP authentication documentation.

## Test plan

- Focused SalesforceSDKCore unit tests: 79 passed.
- SalesforceSDKCore build: passed.


sfdctaka (approved):
Reviewed the full diff plus surrounding source (`DPoPRequestDecorator`, `SFRestAPI`, `SFRestRequest`, credential model). This is tightly scoped and well-tested — approving.

Verified it holds up:
- **Gate is consistent end to end.** `isDPoPTokenType && uiSid nonempty && policy(path)`, and the credential model only ever populates `uiSid` for DPoP tokens and clears it on downgrade, so the invariant lines up.
- **Nonce-retry gate has no false negatives.** Genuine DPoP requests still carry both `Authorization: DPoP …` and the proof header, so the anchored/case-insensitive check keeps the retry; `Bearer <ui_sid>` attempts correctly skip it. `finalRequest` is the stamped request, so it's the right object to inspect.
- **Other call sites stay DPoP.** Identity and revoke/refresh remain on the 4-arg overload — only the REST client path opts into the policy. Good.
- Additive, backward-compatible public API with doc comments; `setValue(nil, …)` header removal is asserted by the LWR test.

Two non-blocking notes:
1. **Expired `ui_sid` → 401 replay.** A `Bearer <ui_sid>` `/lwr` request that 401s runs normal token refresh; replay works only if the refresh updates `uiSid` (it does), and degrades gracefully to DPoP if a refresh returns no `ui_sid`. Worth a quick sanity-check against a live LWR org.
2. **Subtle (practically inert) semantic change.** `SFRestRequest` moved from the 4-arg overload (strict token-type) to the credentials overload (`shouldAttachDPoP`, with the nil-token-type key-material fallback). Can't flip behavior in practice since `tokenType` is always persisted for logged-in users and no REST calls happen in the `/authorize→/token` window — just flagging that the gate changed.

